PR #6, opened 2014-07-09
[ADD] product_pack module v.7.0

tafaRU:
Moving: https://code.launchpad.net/~agilebg/openerp-product-attributes/adding_product_pack_7/+merge/195786


pedrobaeza:
Hi, Alex,

I have tried the module, and there are some functional remarks:
- There is the need to prevent possibility to add product pack to itself, otherwise it's easy to get infinite recursion when 'stock_depends' is marked.
- Product pack tree view can be editable to avoid to open a popup with the same few two fields in it.
- The logical order for the tree is product, quantity, not the reverse.
- On product view, Inventory tab, you must hide **update** button next to _Quantity on hand_ field when _Stock depends of components_ field is marked to avoid the creation of a fake inventory.
- On the same line, you must prevent on inventories to add a line of this type of products.
- When you add a product pack on a sale order, save, and then change quantity of this pack, pack lines are not properly updated with new quantities.
- And finally, if you make a delivery, product pack always get hanged on the stock.picking on state _Waiting availability_

Regards.

paulius-sladkevicius:
When duplicating a sale order or purchase order containing a product with a pack:
- In the new SO/PO, only a pack exist without components.
- In the initial SO/PO, the pack components are also duplicated

tafaRU:
thank you so much for https://github.com/tafaRU/product-attribute/pull/1! 

We're planning to analyze your contribs and @pedrobaeza  hints asap. Be patient ;-)

lmignon:
Hi Alex,

Thank you for sharing this module with the community.  I've tried the module since I've the same kind of requirements for a client. 
Basically, what we need is to find a solution to safely manage that, in a store, if a person wants to buy a can of Coke and it remains only packs of 6 cans she'll probably take it out of a pack.
At first, your module seems to provide the solution to this problem: Create a product for the can and a product pack for the pack of 6 cans with 'Stock depends of components' checked. 
- Pros:
  - The 2 products exists as it in the product list
  - Qties for product pack seems to be based on components (IOW, if we have 6 cans in stock, we have 1 pack of 6. If we sell 1 can, it remains 5 cans and 0 pack of 6)
  - A product pack can be composed of different products (looks like a BOM in mrp)
- Cons
  - By playing with the module, I found that stock movements were generated false  (with Stock depends of components checked). I'll try to explain why I think they are wrong with the following scenario: I'll purchase 1 pack of 6 cans and sell 1 can out of the pack.
    1 . Create a PO
    1. Add 1 pack of 6 cans. (As result you have 2 PO lines, 1 for the pack with qty = 1 and price = XX and an other one for the content (here 6 cans) with qty = 1 and price = 0. (why not)
    2. Receive the product. (you have to receive the 2 lines, the pack itself and the content of the pack...)
    3. Open the product form for the can, qty available is now 6.
    4. Open the product form for the 6 pack cans, qty available is now 1.
    5. If we look in the stock moves, we can see that both movements were generated. One for the pack and an other one for the components.
    6. Create a sale order 
    7. Add 1 can 
    8. If we look in the stock moves, we can see that 1 movement is generated for the can.
    9. Open the product form for the can, qty available is now 5.
    10. Open the product form for the 6 pack cans, qty available is now 0.
        The problem is that the module breaks the balance between inward and outward movements of the stock. ('X -> stock'  -  'stock -> Y' <> 0). Therefore, all processes based on this principle (inventories, reports, stock rules) are potentially broken.

For our client, we've decided to use 'product.packaging' to cover its needs. In our case, the purchase and sale of a pack is still considered as a purchase of the same product in a particular packaging. This packaging is then used to compute the real quantity in the base unit of the product. To facilitate the use of product.packaging in PO and SO, we will add the field packaging_qty on the line and calculate the qty field based on the rule qty = packaging_qty \* product_packaging.qty. In doing so, we should have nothing more to do because product.packaging is already supported everywhere (po.lines, so.line, stock.move). I don't  know if it's the right way to cover our needs but I've not found an existing module in the community to cover this functional need.

Personally :-1:  to include this PR in product-attribute.

Regards,

lmi

pedrobaeza:
@lmignon, although your concerns are correct, and need to be addressed, product_pack is for more than a packaging shortcut. You can pack two products of very different nature together. For example, a computer + keyboard + mouse. You can sell it apart, or join them in a pack. There's no possible way to do it with packaging. So the value of the module is still needed, but it has to be fine grained.

I think Alex has only ported this module as is, and now he is receiving punches from everywhere, but we can collaborate for refining the module and include it on OCA.

lmignon:
@pedrobaeza Sure, my scope is limited. The proposed functionalities are great. That's why I tried it because I saw more opportunities than my current approach. So I just made my comments in the spirit of cooperation and to be constructive, not to give bunches.

product_pack/pack.py
```diff
@@ -0,0 +1,484 @@
...
+#    You should have received a copy of the GNU Affero General Public License
+#    along with this program.  If not, see <http://www.gnu.org/licenses/>.
+#
+###############################################################################
+
+import math
+from openerp.osv import fields, orm
+
+
+class product_pack(orm.Model):
```
bwrsandman, line 27: Please name the class and python file after the model and have one model per file

product_pack/pack.py
```diff
@@ -0,0 +1,484 @@
...
+            ondelete='cascade', required=True
+        ),
+        'quantity': fields.float('Quantity', required=True),
+        'product_id': fields.many2one(
+            'product.product', 'Product', required=True
+        ),
+    }
+
+
+class product_product(orm.Model):
```
bwrsandman, line 42: Please name the class and python file after the model and have one model per file

product_pack/pack.py
```diff
@@ -0,0 +1,484 @@
...
+                        pack_stock = pack_stock_next
+
+                # result is the minimum stock of all subproducts
+                result[product.id] = pack_stock
+            else:
+                result[product.id] = stock[product.id]
+        return result
+
+
+class sale_order_line(orm.Model):
```
bwrsandman, line 138: Please name the class and python file after the model and have one model per file

product_pack/pack.py
```diff
@@ -0,0 +1,484 @@
...
+        ),
+        'pack_child_line_ids': fields.one2many(
+            'sale.order.line', 'pack_parent_line_id', 'Lines in pack'),
+    }
+    _defaults = {
+        'pack_depth': 0,
+    }
+
+
+class sale_order(orm.Model):
```
bwrsandman, line 157: Please name the class and python file after the model and have one model per file

product_pack/pack.py
```diff
@@ -0,0 +1,484 @@
...
+                        cr, uid, [id], {'sequence': sequence, }, context)
+
+        if updated_orders:
+            """ Try to expand again all those orders that had a pack in this
+            iteration. This way we support packs inside other packs. """
+            self.expand_packs(cr, uid, ids, context, depth + 1)
+        return
+
+
+class purchase_order_line(orm.Model):
```
bwrsandman, line 322: Please name the class and python file after the model and have one model per file

product_pack/pack.py
```diff
@@ -0,0 +1,484 @@
...
+        'pack_child_line_ids': fields.one2many(
+            'purchase.order.line', 'pack_parent_line_id', 'Lines in pack'
+        ),
+    }
+    _defaults = {
+        'pack_depth': 0,
+    }
+
+
+class purchase_order(orm.Model):
```
bwrsandman, line 347: Please name the class and python file after the model and have one model per file

product_pack/pack_view.xml
```diff
@@ -0,0 +1,48 @@
+<?xml version="1.0" encoding="utf-8"?>
+<openerp>
+	<data>
+
+		<!-- Product view modification -->
+		<record model="ir.ui.view" id="nan_product_pack_form">
+			<field name="name">product.product.pack.form</field>
+			<field name="model">product.product</field>
```
bwrsandman, line 8: Please name the class and python file after the model and have one model per file

product_pack/pack_view.xml
```diff
@@ -0,0 +1,48 @@
...
+						<field name="pack_line_ids" colspan="4" nolabel="1"/>
+					</page>
+				</notebook>
+			</field>
+		</record>
+
+		<!-- product.pack.line -->
+		<record model="ir.ui.view" id="product_pack_line_form">
+			<field name="name">product.pack.line.form</field>
+			<field name="model">product.pack.line</field>
```
bwrsandman, line 28: Please name the class and python file after the model and have one model per file

product_pack/pack_view.xml
```diff
@@ -0,0 +1,48 @@
...
+			<field name="arch" type="xml">
+				<form string="Pack Product Line">
+					<field name="quantity"/>
+					<field name="product_id"/>
+				</form>
+			</field>
+		</record>
+		<record model="ir.ui.view" id="product_pack_line_tree">
+			<field name="name">product.pack.line.tree</field>
+			<field name="model">product.pack.line</field>
```
bwrsandman, line 38: Please name the class and python file after the model and have one model per file

rafaelbn:
Hi @tafaRU , are you going to continue with this work? Let me know please, thanks!

cc @liebana please take a look here

tafaRU:
@rafaelbn, we don't think to continue with this work, so the PR can be closed.
Let me know, thanks.

rafaelbn:
@sergio-incaser please take this work, we will finish it. OK? Thanks

yajo:
Take notice this is v7 branch.

sergio-teruel:
Oks..

pedrobaeza:
This one is already migrated to v8 and improved: https://github.com/ingadhoc/odoo-addons/tree/8.0/product_pack

rafaelbn:
I opened an issue https://github.com/ingadhoc/odoo-addons/issues/244 asking to migrate already work done to OCA

PR #11, opened 2014-08-28
search first product.price.type instead of take id 1

laetitia-gangloff:
In some case (rare) product.price.type with id 1 doesn't exist, so it is not possible to set base 1.

To reproduce the problem, simply delete product price type with id 1 and set fixed price on pricelist item. Save doesn't work, because base is empty.


rvalyi:
:+1: LGTM; failing test is unrelated

pedrobaeza:
:+1: OK for me too.

Regards.

lmignon:
:+1:  less error prone

Regards

PR #22, opened 2014-10-23
product_gtin: port to v8

alexis-via:
Rename badly-named variable: ean13_len -> ean_len
Small code simplification


rvalyi:
:+1: LGTM trivial fixes

alexis-via:
@foutoucour 

I won't port it to the new API in this PR. But feel free to do it.

foutoucour:
:-1:

alexis-via:
@foutoucour 
The policy of OCA is not to force the use of the new API in all modules ported to v8. So you should not disapprove the PR only for that reason.
Odoo v8 has been released more than a month ago and the OCA product-attribute branch for v8 is still empty ; so we urgently need to get the modules out of **unported**, or OCA will loose its credibility. That's what I am trying to do with my contributions. A port to the new API can be done in another PR without problems.

sbidoul:
I fully agree with @alexis-via here. Port to new API is not mandatory for 8.0.

:+1:

lepistone:
I tried to migrate quite a few modules to the new API and I think that for a while we should not require it for oca modules. When it stabilzes and the core starts to use it more we can change our mind.

rvalyi:
Hello, I also agree, that migrating the API is a "nice to have" but should not be mandatory yet. So I keep with my approval, further enhancements will be welcome later.

pedrobaeza:
:+1:

I proceed with the merge, and anyone can port it to new API and make a PR if he/she wants.

PR #23, opened 2014-10-24
Add product_code_mandatory and product_code_unique

clonedagain:
This PR supersedes the merge proposal made on Launchpad some time ago by @lbellier: https://code.launchpad.net/~numerigraphe-team/openerp-product-attributes/7.0-add-bremskrel-modules/+merge/219858
"I propose these 2 new modules :
product_code_mandatory: all is in the module name.
product_code_unique: having a unique product code, without sequence.
These modules have come from BREMSKERL branch (they are OK to merge).
Loïc BELLIER"

### DIFF
--- a/product_code_mandatory/__init__.py
+++ b/product_code_mandatory/__init__.py
@@ -1,0 +1,1 @@
+import product
--- a/product_code_mandatory/__openerp__.py
+++ b/product_code_mandatory/__openerp__.py
@@ -1,0 +1,38 @@
+# -*- coding: utf-8 -*-
+##############################################################################
+#
+#    OpenERP Module
+#    
+#    Copyright (C) 2010-2011 BREMSKERL-REIBBELAGWERKE EMMERLING GmbH & Co. KG
+#    Author Marco Dieckhoff
+#
+#    This program is free software: you can redistribute it and/or modify
+#    it under the terms of the GNU Affero General Public License as
+#    published by the Free Software Foundation, either version 3 of the
+#    License, or (at your option) any later version.
+#
+#    This program is distributed in the hope that it will be useful,
+#    but WITHOUT ANY WARRANTY; without even the implied warranty of
+#    MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+#    GNU Affero General Public License for more details.
+#
+#    You should have received a copy of the GNU Affero General Public License
+#    along with this program.  If not, see <http://www.gnu.org/licenses/>.
+#
+##############################################################################
+{
+    "name": "Make product code mandatory",
+    "version": "1.01",
+    "depends": ["product"],
+    "author": "Marco Dieckhoff, BREMSKERL",
+    "category": "Inventory Control",
+    "description": """Makes product code (default_code) a mandatory field 
+    """,
+    'depends' : ['product'],
+    "init_xml": [],
+    'update_xml': [],
+    'demo_xml': [],
+    'installable': True,
+    'active': False,
+#    'certificate': '${certificate}',
+}
--- a/product_code_mandatory/product.py
+++ b/product_code_mandatory/product.py
@@ -1,0 +1,16 @@
+# -*- coding: utf-8 -*-
+
+from openerp.osv import osv, fields
+
+
+class product_code_mandatory_product(osv.osv):
+    _name = "product.product"
+    _inherit = "product.product"
+
+    _columns = {
+        'default_code':  fields.char('Reference', size=64, required=True),
+    }
+
+    _defaults ={
+        'default_code': lambda *a: 'NEW',
+    }
--- a/product_code_unique/__init__.py
+++ b/product_code_unique/__init__.py
@@ -1,0 +1,1 @@
+import product
--- a/product_code_unique/__openerp__.py
+++ b/product_code_unique/__openerp__.py
@@ -1,0 +1,38 @@
+# -*- coding: utf-8 -*-
+##############################################################################
+#
+#    OpenERP Module
+#    
+#    Copyright (C) 2010-2011 BREMSKERL-REIBBELAGWERKE EMMERLING GmbH & Co. KG
+#    Author Marco Dieckhoff
+#
+#    This program is free software: you can redistribute it and/or modify
+#    it under the terms of the GNU Affero General Public License as
+#    published by the Free Software Foundation, either version 3 of the
+#    License, or (at your option) any later version.
+#
+#    This program is distributed in the hope that it will be useful,
+#    but WITHOUT ANY WARRANTY; without even the implied warranty of
+#    MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+#    GNU Affero General Public License for more details.
+#
+#    You should have received a copy of the GNU Affero General Public License
+#    along with this program.  If not, see <http://www.gnu.org/licenses/>.
+#
+##############################################################################
+{
+    "name": "Make product code unique",
+    "version": "1.01",
+    "depends": ["product"],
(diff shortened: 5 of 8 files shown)


product_code_mandatory/__init__.py
```diff
@@ -0,0 +1 @@
+import product
```
pedrobaeza, line 1: `from . import product`

product_code_mandatory/__openerp__.py
```diff
@@ -0,0 +1,38 @@
...
+    "description": """Makes product code (default_code) a mandatory field 
+    """,
+    'depends' : ['product'],
+    "init_xml": [],
+    'update_xml': [],
+    'demo_xml': [],
+    'installable': True,
+    'active': False,
+#    'certificate': '${certificate}',
+}
```
pedrobaeza, line 38: Line at the end.

product_code_mandatory/__openerp__.py
```diff
@@ -0,0 +1,38 @@
...
+    "category": "Inventory Control",
+    "description": """Makes product code (default_code) a mandatory field 
+    """,
+    'depends' : ['product'],
+    "init_xml": [],
+    'update_xml': [],
+    'demo_xml': [],
+    'installable': True,
+    'active': False,
+#    'certificate': '${certificate}',
```
pedrobaeza, line 37: Remove this line

product_code_mandatory/__openerp__.py
```diff
@@ -0,0 +1,38 @@
...
+    "version": "1.01",
+    "depends": ["product"],
+    "author": "Marco Dieckhoff, BREMSKERL",
+    "category": "Inventory Control",
+    "description": """Makes product code (default_code) a mandatory field 
+    """,
+    'depends' : ['product'],
+    "init_xml": [],
+    'update_xml': [],
+    'demo_xml': [],
```
pedrobaeza, line 34: Remove this or put 'demo'.

product_code_mandatory/__openerp__.py
```diff
@@ -0,0 +1,38 @@
...
+    "name": "Make product code mandatory",
+    "version": "1.01",
+    "depends": ["product"],
+    "author": "Marco Dieckhoff, BREMSKERL",
+    "category": "Inventory Control",
+    "description": """Makes product code (default_code) a mandatory field 
+    """,
+    'depends' : ['product'],
+    "init_xml": [],
+    'update_xml': [],
```
pedrobaeza, line 33: Put 'data' or remove both lines (init_xml and update_xml).

product_code_mandatory/product.py
```diff
@@ -0,0 +1,16 @@
...
+class product_code_mandatory_product(osv.osv):
+    _name = "product.product"
+    _inherit = "product.product"
+
+    _columns = {
+        'default_code':  fields.char('Reference', size=64, required=True),
+    }
+
+    _defaults ={
+        'default_code': lambda *a: 'NEW',
```
pedrobaeza, line 15: This can be `'default_code': 'NEW',`

product_code_unique/__init__.py
```diff
@@ -0,0 +1 @@
+import product
```
pedrobaeza, line 1: `from . import product`

pedrobaeza, line 1: Line at the end.

product_code_unique/__openerp__.py
```diff
@@ -0,0 +1,38 @@
...
+    "description": """Makes product code (default_code) a unique field 
+    """,
+    'depends': ['product'],
+    "init_xml": [],
+    'update_xml': [],
+    'demo_xml': [],
+    'installable': True,
+    'active': False,
+#    'certificate': '${certificate}',
+}
```
pedrobaeza, line 38: Same remarks as the other module.

product_code_unique/product.py
```diff
@@ -0,0 +1,43 @@
...
+            # Result may contain the current product if it's active: we remove it here.
+            if product.id in res:
+                res.remove(product.id)
+            if len(res):
+                # If we have any results left
+                # we have duplicate entries
+                return False
+        return True
+
+    _constraints = [(_check_defaultcode_and_variants,
```
pedrobaeza, line 25: Why do you have two constrainst: one SQL and the other ORM?

product_code_mandatory/product.py
```diff
@@ -0,0 +1,16 @@
+# -*- coding: utf-8 -*-
+
+from openerp.osv import osv, fields
+
+
+class product_code_mandatory_product(osv.osv):
```
lmignon, line 6: Why do you name your class product_code_.... and not product_product? IMO the usage is to keep the class name from the inherited model.
osv.osv is deprecated. I would prefer orm.Model

pedrobaeza:
Apart from that comments, I see a potential problem: if you install both modules, and create a new product letting the default 'NEW' code, this will prevent the creation of any new record, because SQL constraint is triggered always while you don't change this code.

product_code_unique/product.py
```diff
@@ -0,0 +1,43 @@
+# -*- coding: utf-8 -*-
+
+from openerp.osv import osv
+from openerp.tools.translate import _
+
+
+class product_code_unique_product(osv.osv):
+    _name = "product.product"
```
lmignon, line 8: remove '_name=' since you inherit from product.product

product_code_mandatory/product.py
```diff
@@ -0,0 +1,16 @@
...
+    _name = "product.product"
```
lmignon, line 7: remove '_name=' since you inherit from product.product

clonedagain:
Yes, when both modules are installed, users are supposed to enforce an internal reference convention of one kind or another, so keeping "NEW" would be a mistake.
Would you suggest something else?

product_code_unique/product.py
```diff
@@ -0,0 +1,43 @@
+# -*- coding: utf-8 -*-
+
+from openerp.osv import osv
+from openerp.tools.translate import _
+
+
+class product_code_unique_product(osv.osv):
```
lmignon, line 7: Same remarks as for product_code_mandatory_product

clonedagain:
I'll redo the PEP8 things, sorry I ported it directly from LP.

pedrobaeza:
You can remove the default, so user is enforced to put it. In both cases you will have the problem that when you install mandatory module, all products will have 'NEW' in their codes, as this is defined as default. When you try to install second module (unique), it will fail to put SQL constraint, because there are already conditions that don't pass. Letting empty, the check is not done.

I'm also intringued about variants in this equation...

product_code_unique/product.py
```diff
@@ -0,0 +1,43 @@
...
+    _name = "product.product"
+    _inherit = "product.product"
+
+    def _check_defaultcode_and_variants(self, cr, uid, ids):
+        for product in self.browse(cr, uid, ids):
```
lmignon, line 12: I prefer a 'read' to avoid reading the full recordset of product.product

lmignon:
Thank you for sharing these modules. Apart from a few comments in the code, nothing more to say than what has already been said by @pedrobaeza

product_code_unique/product.py
clonedagain, line 12: Actually browse was proved to be more efficient in most cases (because of the prefetching if I remmeber correctly)

lmignon, line 12: @clonedagain You are right, it's the official answer from Odoo https://www.odoo.com/fr_FR/forum/help-1/question/solved-browse-vs-read-performance-67752. I am, however, quite skeptical and I often wonder if Odoo developers sometimes look at the number of SQL queries generated by the orm.

PR #24, opened 2014-11-14
[REF] auto-fix conventions

moylop260:


eLBati:
tests broken because of https://github.com/OCA/product-attribute/pull/25

PR #29, opened 2014-12-03
[ADD] product_manufacturer: add module removed in Odoo 8.0

acysos:
Hello,

Odoo has removed this module from 8.0. For my installations is useful, so I have migrated to the new API.

I think that can be useful for someone, so I request to add to the community repository.

Greetings,
Ignacio Ibeas

### DIFF
--- a/product_manufacturer/__openerp__.py
+++ b/product_manufacturer/__openerp__.py
@@ -1,0 +1,45 @@
+##############################################################################
+#
+#    OpenERP, Open Source Management Solution
+#    Copyright (C) 2004-2009 Tiny SPRL (<http://tiny.be>).
+#
+#    This program is free software: you can redistribute it and/or modify
+#    it under the terms of the GNU Affero General Public License as
+#    published by the Free Software Foundation, either version 3 of the
+#    License, or (at your option) any later version.
+#
+#    This program is distributed in the hope that it will be useful,
+#    but WITHOUT ANY WARRANTY; without even the implied warranty of
+#    MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+#    GNU Affero General Public License for more details.
+#
+#    You should have received a copy of the GNU Affero General Public License
+#    along with this program.  If not, see <http://www.gnu.org/licenses/>.
+#
+##############################################################################
+{
+    'name': 'Products Manufacturers',
+    'version': '1.0',
+    'author': 'OpenERP SA',
+    'contributors': ['Acysos SL <info@acysos.com>'],
+    'category': 'Purchase Management',
+    'depends': ['product'],
+    'demo': [],
+    'description': """
+A module that adds manufacturers and attributes on the product form.
+====================================================================
+
+You can now define the following for a product:
+-----------------------------------------------
+    * Manufacturer
+    * Manufacturer Product Name
+    * Manufacturer Product Code
+    * Product Attributes
+    """,
+    'data': [
+        'security/ir.model.access.csv',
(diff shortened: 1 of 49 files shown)


pedrobaeza:
Travis error is unrelated.

Thanks :+1:

pedrobaeza:
Going a bit deeper, Travis error is not unrelated at all. It happens in another module, but it is because of this new addition. The problem is the following: this module is adding another field called `attribute_ids` in the view, which turns into crazy the inheritance of the view of the _product_m2m_categories_ module. Can you please include in this PR the modification to this other module to have an xpath for the inheritance view (or another reference different from attribute_ids) to avoid this problem?

Meanwhile, I put this :-1: to avoid accidental merges.

acysos:
I don't find how to solve the travis fail, can some help me?

Greetings

pedrobaeza:
Ignacio, please rename _attribute_ids_ field for the module product_manufacturer to another name (for example _manufacturer_attribute_ids_, and try again. With that, I think you will get all solved.

acysos:
The travis continue with the same error. I have tested the two modules in the same database and work without any problem.

pedrobaeza:
One thing: why don't you just remove attributes, because they are redundant with new product attributes and doesn't add anything relevant? With that move, I think the problem will be gone.

acysos:
Yes, it can be removed, the function is different, because the new attributes add variants and this only add values, but can be redundant. I will go to update it.

acysos:
I have signed the CLA, this an error.

oihane:
@pedrobaeza, as you asked @acysos has deleted attributes part, I have been taking a look and it gives an unexpected travis error, I do not understand because as Ignacio says it does not have any problem installing
:+1:

acysos:
Yes, I think that is a problem in travis, because the two modules work without problems.

Greetings

product_manufacturer/__openerp__.py
```diff
@@ -0,0 +1,43 @@
...
+{
+    'name': 'Products Manufacturers',
+    'version': '1.0',
+    'author': 'OpenERP SA',
+    'contributors': ['Acysos SL <info@acysos.com>'],
+    'category': 'Purchase Management',
+    'depends': ['product'],
+    'demo': [],
+    'description': """
+A module that adds manufacturers and attributes on the product form.
```
pedrobaeza: Please extract this to a README.rst file and remove from here, because Odoo can parse these files and so it's compatible with both GitHub and Odoo.

pedrobaeza:
Good to see that you have finally found that the problem was with the ID of the view.

One last thing you have to do is to remove the attributes part from the translations.

product_manufacturer/__openerp__.py
acysos: I don't understand this?

oihane: You have to add a file `README.rst`  where you will write the same as it is in description.
In that way, you can see it in Odoo and also in Github.
When you have done this remove **description** from `__openerp__.py` file

pedrobaeza: In v8, you can put this text in a separated file called README.rst, and thus it can be parsed also by code view in GitHub. See an example here: https://github.com/odoomrp/odoomrp-wip/tree/8.0/stock_purchase_unit

pedrobaeza:
Thank you very much for all your efforts, Ignacio.

:+1:

acysos:
Not at all. Now to the next module.

rvalyi:
:+1:  LGTM

PR #30, opened 2014-12-05
8.0 product template header

foutoucour:
Simple module to create a header for the product.template form view.

It was used in the context to develop several action buttons for product.template but the architecture of the project asked that each buttons were developed in different module.

With this module, each button were easily added in a unique header without strange depencencies.


pedrobaeza:
A better option for this is to declare a common xml-id for all the modules, for example `base.product_template_header`, that all the modules that are going to use it declare. If there is one already present, it is overwritten without problem. If not, you get the header. You can see an example with these modules: https://github.com/odoomrp/odoomrp-wip/blob/8.0/product_last_mrp_info/views/product_view.xml#L4 and https://github.com/odoomrp/odoomrp-wip/blob/8.0/product_last_price_info/views/product_view.xml#L4

foutoucour:
This module has been removed after having follow @pedrobaeza adviced

PR #31, opened 2014-12-11
[ADD] product supplier check

vrenaville:
Add a constraint that check that the supplier enter is not your company
It's a improvement combining to this https://github.com/OCA/OCB/pull/94

### DIFF
--- a/product_supplier_check/product_supplier_check.py
+++ b/product_supplier_check/product_supplier_check.py
@@ -1,0 +1,50 @@
+# -*- coding: utf-8 -*-
+##############################################################################
+#
+#    Author Vincent Renaville. Copyright 2013 Camptocamp SA
+#
+#    This program is free software: you can redistribute it and/or modify
+#    it under the terms of the GNU Affero General Public License as
+#    published by the Free Software Foundation, either version 3 of the
+#    License, or (at your option) any later version.
+#
+#    This program is distributed in the hope that it will be useful,
+#    but WITHOUT ANY WARRANTY; without even the implied warranty of
+#    MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+#    GNU Affero General Public License for more details.
+#
+#    You should have received a copy of the GNU Affero General Public License
+#    along with this program.  If not, see <http://www.gnu.org/licenses/>.
+#
+##############################################################################
+from openerp.osv import orm
+
+
+class product_product(orm.Model):
+    _inherit = "product.product"
+
+    def _supplier_not_company(self, cr, uid, ids, context=None):
+        company_pool = self.pool['res.company']
+        # Get all company
+        all_company_ids = company_pool.search(cr, uid, [], context=context)
+        all_partner_company = company_pool.read(cr, uid,
+                                                all_company_ids,
+                                                ['partner_id'],
+                                                context=context)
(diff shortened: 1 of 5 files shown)


vrenaville:
Travis Failed but not on my module

product_supplier_check/product_supplier_check.py
```diff
@@ -0,0 +1,50 @@
...
+                                                all_company_ids,
+                                                ['partner_id'],
+                                                context=context)
+        company_partner_ids = [x['partner_id'][0] for x in all_partner_company]
+        # Get all partner attached to partner
+        current_product = self.browse(cr, uid, ids, context=context)[0]
+        product_supplier_ids = [x.name.id for x in current_product.seller_ids]
+        intersection_partner = set(company_partner_ids).intersection(
+            product_supplier_ids)
+        if intersection_partner:
```
pedrobaeza, line 40: Better `return not intersection_partner`

product_supplier_check/product_supplier_check.py
```diff
@@ -0,0 +1,50 @@
...
+            # some supplier refer to our company so refuse it
+            return False
+        else:
+            return True
+
+    _constraints = [
+        (_supplier_not_company, """You cannot defined
```
pedrobaeza, line 47: Don't put triple quota because it introduces line feeds. You can put:

```
(_supplier_not_company, "You cannot defined "
"your company has a product supplier",
...
```

pedrobaeza:
I'm seeing that this functionality is very weak for a module. Why don't you simply remove 'Supplier' check on company partner?

gdgellatly:
:-1: 

@vrenaville This might seem harsh, it isn't meant to be, I really do get what you are trying to do here.

I have a few reasons for this:
It is very old without update
A python check is being down on a one2many model, rather than the model itself
It isn't multicompany compatible
It takes no account of child records or the company_id of the seller record.
It will give different results depending on user access rights 
It could be done entirely as an SQL check constraint on product.supplierinfo, it has no reason to depend on product at all.  Moreover product.seller_ids is really related to product templates.
No tests.
It is targeted at 7.0 and I am not sure we want to target new modules with minimal functionality at an old branch.

I don't agree with @pedrobaeza that unticking supplier is the same, but realistically this kind of check should really be done not here but in procurement workflow.  From memory it does raise an error there, but it can be overridden to suppress that error and choose a different action.  In fact I imagine the kernel of the idea for this module was that error message causing scheduler failure.  I agree we need a way to solve this, just don't think this is the right way.

It is not seller_ids that is relevant, but rather seller_id or whatever record is selected by procurement workflow.  As an example, I do have this problem.  Multicompany 9 companies all buying off each other.  Commonly 1 will act as a central supplier to the other 8, so no company_id set on supplierinfo.  Then a lower sequence with company_id set for that 1 (or else raise a production order, as buy when seller is maker triggers MO in workflow). Then catch and suppress that stupid error but that was a quick fix.

I don't know but feels to me like rather than a seperate module in product-attribute it might be better as part of a module in a more procurement workflow related repo.

vrenaville:
@gdgellatly Thanks for your comment, I have a create this module for a wrong reason, We found that the solution of Pedro is more efficient, we juste cleaned wrong data by SQL.
We think that this problem is introduced by migrations from V5 -> V7

PR #33, opened 2014-12-17
add module product_unique_code and move sql_constraints from product_sequence to product_unique_code

chafique-delli:
.

PR #34, opened 2014-12-18
[ADD] product_unique_code.

LeartS:


product_unique_code/__openerp__.py
```diff
@@ -0,0 +1,43 @@
...
+##############################################################################
+
+
+{
+    'name': 'Product Unique Code',
+    'version': '1.0',
+    'category': 'Product',
+    'description': """
+Product Unique Codes
+===========================
```
pedrobaeza: Put same length of the title.

LeartS:
I'm actually surprised the build passes, considering there are products with duplicate default_code in the demo data and the module gives error during installation when that's the case, because it can't create the unique constraint.

pedrobaeza:
That error is bypassed by the Travis checks.

product_unique_code/product.py
```diff
@@ -0,0 +1,18 @@
...
+
+    _sql_constraints = [
+        ('unique_default_code',
+         'unique(default_code)',
+         'The code must be unique'),
+    ]
+
+    @api.one
+    def copy(self, default=None):
+        new_code = self.default_code + '-2'
```
gurneyalex: this will fail if you try to copy the same product twice as  you will end up with 2 products with dfefault_code = 'orig_code-2'

gurneyalex:
@Learts you need to patch the demo data in the demo data of your addon to fix the default code.

product_unique_code/product.py
LeartS: Yeah well, it will also fail if the user already has, for whatever reason, another product with default code `orig_code-2`.
Finding a pseudo-random "free" default_code is imo overkill, the user should edit the code anyway, `-2` is just to show a default value. The same way I don't think users leave `(Copy)` in the product name when they duplicate it.

Maybe it's better to add a specific error that tells them to change the code before duplicating the product again?

pedrobaeza: I will prefer to put field as not required, and with attribute copy=False to not propagate previous value. Constraint doesn't apply to duplicated null values.

gurneyalex: +1 for @pedrobaza's solution

LeartS: @pedrobaeza the unique constraint doesn't apply to null values at all, at least it seems from my tests.
But I've tested leaving `required=True` with `copy=False` (which I didn't know about, thanks!) and it seems like it works: it creates and opens in edit mode the duplicate, but doesn't let you save without specifying a code (but it lets you discard, in which case the duplicated product will have null code o,o)

So, `copy=False` works with both `required=True` or `required=False`. What's your opinion on this? Should unique code imply required code? Also keep in mind that with `required=True` the code field is highlighted, which might hint the user he should change the field.

sebastienbeau:
Hi I think it will be better to merge this previous merge here : https://github.com/OCA/product-attribute/pull/33
Which is a basic extraction of the unique_code feature in the product sequence module.
And than port this merge into 8 version and change the API.
Please also keep the copyright in the header ;)

LeartS:
The module is so small that I don't see any difference between writing it from scratch or "porting it" from v7, and surely faster than wait for another PR request to be reviewed and accepted.

Regarding the header, are you talking about the `product.py` file having no license/copyright notices at the start? Is there an OCA conventions on having those on every single .py file? I think having it on the __openerp__.py file is already [more than enough](http://hackerboss.com/get-rid-of-templates/)

yvaucher:
@LeartS I cannot remind of a talk about a convention on this

However, from what I can read on http://www.gnu.org/licenses/agpl-3.0.html

```
To do so, attach the following notices to the program. It is safest to attach them
to the start of each source file to most effectively state the exclusion of warranty;
and each file should have at least the "copyright" line and a pointer to where the full notice is found
```

You would at least something like that

```
# Copyright 2014 <author>
# <notice where to find full licence>
```

And this won't harm

```
# -*- coding: utf-8 -*-
```

Maybe it's time to talk about license on community mailing list?

LeartS:
> Maybe it's time to talk about license on community mailing list?

I think it's something that should be discussed. Can you start a thread about it?

clonedagain:
:-1: 
This is at least the third proposal for the same basic module here :
- https://github.com/OCA/product-attribute/pull/23
- https://github.com/OCA/product-attribute/pull/33

And the same feature exists also in the project "multi-company", only making it company-aware :
- https://github.com/OCA/multi-company/issues/8

Please stop, comment on existing proposals and let's converge.

yvaucher:
@LeartS https://lists.launchpad.net/openerp-community/msg07097.html talk launched here for license

PR #39, opened 2015-01-13
MIG product_sequence

angelmoya:


product_sequence/README.rst
```diff
@@ -0,0 +1,6 @@
+A module that adds sequence to the product.
+====================================================================
+
+This module allows to associate a sequence to the product reference.
+The reference (default code) is unique (SQL constraint) and required.
+Note: This module is incompatible with nan_product_sequence.
```
pedrobaeza: You can put a contributors section with your name here. There's a suggested template for OCA modules here: https://github.com/OCA/maintainers-tools/blob/master/template/module/README.rst

product_sequence/models/__init__.py
```diff
@@ -0,0 +1,23 @@
+# -*- encoding: utf-8 -*-
+##############################################################################
+#
+#    OpenERP, Open Source Management Solution
+#    Copyright (c) 2015 Domatix (http://domatix.com)
+#                       Angel Moua <angel.moya@domatix.com>
```
pedrobaeza: s/Moua/Moya (I think you want to put your name correctly :wink:)

pedrobaeza:
As in https://github.com/OCA/product-attribute/pull/33, I would remove here the constraint about code uniqueness.

product_sequence/README.rst
```diff
@@ -0,0 +1,27 @@
+A module that adds sequence to the product.
+====================================================================
```
pedrobaeza: This line should be of the same length of the title

product_sequence/README.rst
yvaucher: This is done

yvaucher:
:+1:

product_sequence/README.rst
```diff
@@ -0,0 +1,27 @@
...
+
+This module allows to associate a sequence to the product reference.
+The reference (default code) is unique (SQL constraint) and required.
+Note: This module is incompatible with nan_product_sequence.
```
pedrobaeza: This module is no longer on 8, so you can remove it.

pedrobaeza:
:+1:

pedrobaeza:
I merge, and we will resolve later the split of the module in 2.

PR #44, opened 2015-01-30
add check method for ean8, ean13 and upc

chafique-delli:
add check method for ean8, ean13 and upc

### DIFF
--- a/product_gtin/product_gtin.py
+++ b/product_gtin/product_gtin.py
@@ -124,7 +124,8 @@ class product_product(orm.Model):
             "http://en.wikipedia.org/wiki/Global_Trade_Item_Number"),
     }
 
-    _constraints = [(_check_ean_key, 'Error: Invalid EAN/GTIN code', ['ean13'])]
+    _constraints = [(_check_ean_key,
+                     'Error: Invalid EAN/GTIN code', ['ean13'])]
 
 
 class product_packaging(orm.Model):
--- a/product_gtin/product_gtin.py
+++ b/product_gtin/product_gtin.py
@@ -96,7 +96,7 @@ DICT_CHECK_EAN = {8: check_ean8,
 def check_ean(eancode):
     if not eancode:
         return True
-    if not len(eancode) in [8, 11, 12, 13, 14]:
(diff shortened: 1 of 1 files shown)


product_gtin/product_gtin.py
```diff
@@ -37,6 +37,53 @@ def check_ean(eancode):
...
+        check = check_ean8(eancode)
+    if len(eancode) == 11:
+        check = check_ean11(eancode)
+    if len(eancode) == 12:
+        check = check_upc(eancode)
+    if len(eancode) == 13:
+        check = check_ean13(eancode)
+    if len(eancode) == 14:
+        check = check_gtin14(eancode)
+    return check
```
rvalyi: Hello, I think at bare minimum you should use elif instead of just if. Or eventually some kind of dynamic dispatch such as http://stackoverflow.com/questions/11479816/what-is-the-python-equivalent-for-a-case-switch-statement as there is no case switch in Python.

pedrobaeza: Or better yet:

```
check_func = getattr(self, 'check_ean_%s' %len(eancode), None)
if check_func:
    return check_func(eancode)
return False
```

And you will have to make homogeneus all method names (check_ean or check_gtin, what you want).

bealdav:
:+1:

rvalyi:
in fact Pedro suggestion was better I believe https://github.com/OCA/product-attribute/pull/44#discussion_r23855973
But that's ok for me now :+1:

product_gtin/product_gtin.py
```diff
@@ -53,6 +77,34 @@ def check_ean(eancode):
...
+                  12: check_upc,
+                  13: check_ean13,
+                  14: check_gtin14,
+                  }
+
+
+def check_ean(eancode):
+    if not eancode:
+        return True
+    if not len(eancode) in [8, 11, 12, 13, 14]:
```
lmignon: in DICT_CHECK_EAN

bealdav:
seems to be ok for merge, thanks chafique :+1:

PR #57, opened 2015-03-12
[WIP] [MIG] product_multi_image: Module totally redone from old product_images

pedrobaeza:
This is still a work in progress.

Refactorized:
- Take into account templates and products
- Code for creating and updating
- Views
- All!

New planned features:
- Three possible sources: URL, file or DB.
- Possibility to share images between different products.
- New kanban view
- Others not yet thought.


product_multi_image/models/product_image.py
```diff
@@ -0,0 +1,168 @@
...
+        filters='*.png,*.jpg,*.gif')
+    comments = fields.Text(string='Comments')
+    product_id = fields.Many2one(
+        comodel_name='product.product', string='Product', required=True,
+        ondelete='cascade')
+
+    @api.one
+    def unlink(self):
+        if self.path:
+            os.path.isfile(self.path) and os.remove(self.path)
```
guewen: It is unlikely to happen but there is a small chance that 2 calls are done at the same time, one call will raise an OSError. I think the line could be replaced by the following because we do not mind if it does no longer exist.

```
try:
    os.remove(self.path)
except OSError:
    pass
```

guewen: Also, I think the file should be deleter _after_ the call to `super`, so if something `raise` in `super`, the file won't be deleted.

pedrobaeza: You're right, thanks.

product_multi_image/models/product_image.py
```diff
@@ -0,0 +1,168 @@
...
+        return super(ProductImage, self).unlink()
+
+    @api.constrains('type', 'url')
+    def _check_url(self):
+        if self.type == 'url' and not self.url:
+            raise exceptions.Warning(
```
guewen: According to the docs, a `ValidationError` should be raised in the `@api.constrains`

pedrobaeza: It also works with this, but I can be a little more specific

guewen: Of course it works because the raise blows up anyway. But if I write code and want to catch errors in `constrains`, I should expect to catch `ValidationError` errors.

bealdav:
@sebastienbeau, take a look, you have done job on this matter

flotho:
Hi,

I don't understand the problem with runbot. 
The issue seems to come from the dataset that has already 2 codes identical.
Any chance to be explained what is the process to change this?

LeartS:
I forced a rebuild, let's see what happens.

LeartS:
Well, seems to be an error due to the module `product_sequence`. The strange thing is that module has some code specifically to handle this case: https://github.com/OCA/product-attribute/blob/8.0/product_sequence/models/product_product.py#L26

Maybe it never gets called? I see the module [partner_firstname](https://github.com/OCA/partner-contact/blob/8.0/partner_firstname/models.py#L134) uses `_install_<module-name>` to do something similar.

flotho:
Hi @LeartS 

Thanks for your report.
Actually the branch from @pedrobaeza missed this part of code.
So I made this https://github.com/pedrobaeza/product-attribute/pull/1
Once it will be pulled we'll be able to rerun the test.

Thanks

LeartS:
@flotho I've just noticed that this PR is older than that code, @pedrobaeza should just rebase this branch and it should fix itself.

flotho:
@LeartS 
Ok sorry, I didn't checked well your comment, I did it for @pedrobaeza, we only have to wait now.

pedrobaeza:
Branch rebased. Let's see CIs' result.

flotho:
Hi,

I don't understand the reason of the runbot built failure. It talks about mrp, yet there is no mrp in this PR.

flotho:
Any chance to have some odoo core updates and rerun this one?

rafaelbn:
cc @yajo @sergio-incaser @carlos-incaser

product_multi_image/README.rst
```diff
@@ -0,0 +1,39 @@
+Multiple images in products
+===========================
+
+This module implements the possibility to have multiple images for a product,
+that it's to say, an image gallery.
+
```
yajo, line 6: This is missing some important parts such as usage instructions, try on runbot, report issues, etc.

product_multi_image/models/product_image.py
```diff
@@ -0,0 +1,164 @@
...
+        compute="_get_image_sizes", string="Small-sized image",
+        help="Small-sized image. It is automatically resized as a 64 x 64 px "
+             "image, with aspect ratio preserved. Use this field anywhere a "
+             "small image is required.")
+    comments = fields.Text(string='Comments')
+    product_id = fields.Many2one(
+        comodel_name='product.template', string='Product', required=True,
+        ondelete='cascade')
+
+    def _make_pretty(self, name):
```
yajo, line 120: `@api.model`.

product_multi_image/models/product_image.py
```diff
@@ -0,0 +1,164 @@
...
+        return name.replace('_', ' ').capitalize()
+
+    @api.onchange('url')
+    def onchange_url(self):
```
yajo, line 124: Please prefix with `_` following guidelines.

product_multi_image/models/product_image.py
```diff
@@ -0,0 +1,164 @@
...
+        if self.url:
+            filename = self.url.split('/')[-1]
+            self.name, self.extension = os.path.splitext(filename)
+            self.name = self._make_pretty(self.name)
+
+    @api.onchange('path')
+    def onchange_path(self):
```
yajo, line 131: Please prefix with `_` following guidelines.

product_multi_image/models/product_image.py
```diff
@@ -0,0 +1,164 @@
...
+        if self.path:
+            self.name, self.extension = os.path.splitext(os.path.basename(
+                self.path))
+            self.name = self._make_pretty(self.name)
+
+    @api.onchange('filename')
+    def onchange_filename(self):
```
yajo, line 138: Please prefix with `_` following guidelines.

product_multi_image/models/product_image.py
```diff
@@ -0,0 +1,164 @@
...
+        if self.filename:
+            self.name, self.extension = os.path.splitext(self.filename)
+            self.name = self._make_pretty(self.name)
+
+    @api.constrains('type', 'url')
+    def _check_url(self):
```
yajo, line 144: Should these checks be made multi-record aware?

product_multi_image/models/product_product.py
```diff
@@ -0,0 +1,84 @@
...
+import logging
+from openerp import models, fields, api, _
+
+_logger = logging.getLogger(__name__)
+
+
+class ProductProduct(models.Model):
+    _inherit = "product.template"
+
+    @api.one
```
yajo, line 31: `@api.multi` or nothing please.

product_multi_image/models/product_product.py
```diff
@@ -0,0 +1,84 @@
...
+                self.image_ids[0].write({'type': 'db',
+                                         'file_db_store': image})
+            else:
+                self.image_ids = [(0, 0, {'type': 'db',
+                                          'file_db_store': image,
+                                          'name': _('Main image')})]
+        elif self.image_ids:
+            self.image_ids[0].unlink()
+
+    @api.one
```
yajo, line 54: `@api.multi` or nothing please.

product_multi_image/models/product_product.py
```diff
@@ -0,0 +1,84 @@
...
+    def _set_main_image(self):
+        self._set_image(self.product_image)
+
+    @api.one
```
yajo, line 58: `@api.multi` or nothing please.

product_multi_image/models/product_product.py
```diff
@@ -0,0 +1,84 @@
...
+    def _set_main_image_medium(self):
+        self._set_image(self.image_medium)
+
+    @api.one
```
yajo, line 62: `@api.multi` or nothing please.

product_multi_image/models/product_image.py
```diff
@@ -0,0 +1,164 @@
...
+                    os.makedirs(dir_path)
+            except OSError, e:
+                raise exceptions.Warning(
+                    _('The image filestore can not be created, %s') % e)
+        return True
+
+    type = fields.Selection(
+        selection=[('url', 'URL'), ('file', 'OS file'), ('db', 'DB')],
+        default='db')
+    name = fields.Char(string='Image title', required=True)
```
yajo, line 91: Shouldn't this be translatable?

product_multi_image/models/product_image.py
```diff
@@ -0,0 +1,164 @@
...
+        help="Medium-sized image. It is automatically resized as a 128 x "
+             "128 px image, with aspect ratio preserved, only when the image "
+             "exceeds one of those sizes. Use this field in form views or "
+             "some kanban views.")
+    image_small = fields.Binary(
+        compute="_get_image_sizes", string="Small-sized image",
+        help="Small-sized image. It is automatically resized as a 64 x 64 px "
+             "image, with aspect ratio preserved. Use this field anywhere a "
+             "small image is required.")
+    comments = fields.Text(string='Comments')
```
yajo, line 115: Shouldn't this be translatable?

product_multi_image/models/product_product.py
```diff
@@ -0,0 +1,84 @@
...
+        comodel_name='product.image', inverse_name='product_id',
+        string='Product images', copy=True)
+    product_image = fields.Binary(
+        string="Main image", compute="_get_main_image", store=False,
+        inverse="_set_main_image")
+    image_medium = fields.Binary(
+        compute="_get_main_image", inverse="_set_main_image_medium",
+        store=False)
+    image_small = fields.Binary(
+        compute="_get_main_image", inverse="_set_main_image_small",
```
yajo, line 76: Why different inverse methods that all do the same?

product_multi_image/models/product_product.py
```diff
@@ -0,0 +1,84 @@
...
+        store=False)
+
+    @api.multi
+    def write(self, vals):
+        if 'image_medium' in vals and 'image_ids' in vals:
```
yajo, line 81: What if image_medium in vals and image_ids not? Wouldn't this fail? Also, why only image_medium and not image_small or product_image? Maybe I'm missing something...

product_multi_image/models/res_company.py
```diff
@@ -18,24 +18,21 @@
...
-            return self.browse(cr, uid, id, context=context).local_media_repository
-        user = self.pool.get('res.users').browse(cr, uid, uid, context=context)
-        return user.company_id.local_media_repository
+    local_media_repository = fields.Char(
+        string='Images Repository Path',
+        help='Local directory on the Odoo server where all images are '
+             'stored.')
+
+    @api.multi
+    def get_local_media_repository(self):
```
yajo, line 34: I see no use of this function in your new code.

yajo:
Full code review done for now, with some remarks as you see. You did _828 additions and 448 deletions_, some tests would be appreciated. Also please fix runbot to start functional review. Thanks!

pedrobaeza:
@Yajo, that was done a lot of time ago. My skills have improved since then :wink: I'll check your comments later

product_multi_image/models/product_image.py
```diff
@@ -0,0 +1,164 @@
...
+    def onchange_filename(self):
+        if self.filename:
+            self.name, self.extension = os.path.splitext(self.filename)
+            self.name = self._make_pretty(self.name)
+
+    @api.constrains('type', 'url')
+    def _check_url(self):
+        if self.type == 'url' and not self.url:
+            raise exceptions.ValidationError(
+                'You must provide a URL for the product image.')
```
elicoidal, line 147: a URL or an URL?

product_multi_image/models/product_image.py
```diff
@@ -0,0 +1,164 @@
...
+
+    @api.constrains('type', 'file_db_store')
+    def _check_store(self):
+        if self.type == 'db' and not self.file_db_store:
+            raise exceptions.ValidationError(
+                'You must provide an attached file for the product image.')
+
+    _sql_constraints = [
+        ('uniq_name_product_id', 'UNIQUE(product_id, name)',
+         _('A product can have only one image with the same name')),
```
elicoidal, line 163: add period

yajo:
I'm thinking of another approach, let's see what you think.

What about putting all the gallery-related stuff in an abstract model in a separate module, and then inherit it in products and add the views?

I'm sure with time more and more models will benefit of having a gallery (events, blog...). This way we'd reuse the abstract one and repeat less code.

pedrobaeza:
Yeah, it's a very good idea. We need them 2 modules: _base_multi_image_, with all the stuff that handles the possible sources, cache... and then inherit in _product_multi_image_, _event_multi_image_... Are you going to do it?

yajo:
Let's ask @rafaelbn.

rafaelbn:
I agree, let's go. Please, talking about images they must be stored **outside** data base. How to solve this?

pedrobaeza:
For now, there's 3 possibilities:
- Store in database
- Put a path of the file server. This is located outside of the DB, but the file handling is made outside Odoo.
- Put a URL. This is for avoiding making verbatim copies of images, if you already have them on another hosting (like in Magento, Prestashop...).

We can add a fourth option that is "stored as a attachment".

yajo:
I'm forking this to develop. Close for now if you wish.

pedrobaeza:
OK, eager to see the new approach

PR #58, opened 2015-03-17
[ADD] Add product_internal_reference_search module

foutoucour:
Simple module that allows to search a product by its internal reference


laetitia-gangloff:
Hello,

sorry, but I don't understand the interest of this module.
There is already a search on default_code :
https://github.com/odoo/odoo/blob/8.0/addons/product/product_view.xml#L13

bwrsandman:
This was patched in odoo a couple days after this PR was opened

PR #59, opened 2015-03-25
[IMP] Migration of base_custom_attributes

fevxie:
Initial migration base_custom_attribute,   Still work in progress for fully test

PR #62, opened 2015-03-27
[IMP]8.0 product custom attributes migr

fevxie:
Initial migration product_custom_attributes
still work in  progress

depends on #59


pedrobaeza:
What is purpose of this migration? Product attributes, variants and so on are already covered by standard.

fevxie:
@pedrobaeza Thanks for your reply,  the reason I migration this module is the product attributes and variants in standard is not satisfy what I wanted,  in the standard product module someone can add an attribute and values for a product,  if the attribute have multi values there will generate multi products for the attributes. if I work on e-commerce platform e.g magento, It have an attributes system for products one attribute is an field in the database every attribute have an attribute type, textarea, selected etc. and the standard products can not do this.  my question is the same as : https://www.odoo.com/forum/help-1/question/how-add-custom-attributes-to-product-v8-58693

I think the standard product attribute is exactly for product variants but not 100% for custom product attributes.

product_custom_attributes/product.py
```diff
@@ -18,48 +18,50 @@
...
-                     context=context)[0]['attribute_set_id']
-            if not set_id:
-                res[i] = []
-            else:
-                res[i] = self.pool.get('attribute.group').search(cr, uid,
-                      [('attribute_set_id', '=', set_id[0])])
-        return res
+    @api.one
+    def _attr_grp_ids(self):
+        import pdb
```
nhomar, line 39: a pdb unecesary

pedrobaeza:
@fevxie, thanks for the reply. I also think the standard solution is not 100% perfect, but replicating a full alternative is not the solution also, but extend the standard to fit your needs.

Under OdooMRP project, we have worked in this way, providing a module to avoid the automatic creation of product variant (https://github.com/odoomrp/odoomrp-wip/tree/8.0/product_variants_no_automatic_creation), and providing configurators for sales (https://github.com/odoomrp/odoomrp-wip/tree/8.0/sale_product_variants), purchases (https://github.com/odoomrp/odoomrp-wip/tree/8.0/purchase_product_variants) and mrp (https://github.com/odoomrp/odoomrp-wip/tree/8.0/mrp_product_variants) for leading with attributes. 

There is also an expansion for adding custom attributes (https://github.com/odoomrp/odoomrp-wip/tree/8.0/product_variants_types), and therir corresponding sale/purchase/mrp auto-install modules.

Please check them instead of making more effort in this migration.

fevxie:
@pedrobaeza Thanks, I will check the modules and close this PR.

gurneyalex:
note: @fevxie probably has a ECLA as he works for Elico Corp (cc @elicoidal)

PR #68, opened 2015-05-14
[ADD] modules to restrict the users that can view the product cost.

JordiBForgeFlow:
# Product Cost Security

This module adds a new user group "Display product cost". The module itself
will not apply the restriction, but needs to be complemented with a module
that actually implements the restriction in a view, such as:
- product_standard_price_visible
- stock_standard_price_visible
# Product Standard Price Visible

This module restricts the visibility of the field 'cost' associated to the
product only to users that belong to the group 'Display product cost'.
# Product stock standard price visible

This module restricts the visibility of the field 'cost' associated to the
product only to users that belong to the group 'Display product cost',
in stock-related views.

### DIFF
--- a/product_cost_security/__init__.py
+++ b/product_cost_security/__init__.py
@@ -1,0 +1,20 @@
+# -*- coding: utf-8 -*-
+##############################################################################
+#
+#    Copyright (C) 2015 Eficent (<http://www.eficent.com/>)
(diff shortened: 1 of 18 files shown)


ghost:
Please make 1 pull request per module and move your module descriptions to a file named README.rst

PR #71, opened 2015-05-19
[MIG] Base Custom Attributes

ghost:
@joaoalf


pedrobaeza:
I think these modules don't make sense with the new variants implementation in v8, doesn't it?

ghost:
@pedrobaeza The variant implementation in v8 will do the job for most companies. What you can acheive with these modules is way beyond what you can do with product variants.

pedrobaeza:
But you should put this functionality in top of the v8 base, not to create an incompatible set of modules with another data structure. I know that it's an annoyance to change the entire model, but we have treat with that and make very cute things with variants in OdooMRP (*_variants modules): https://github.com/odoomrp/odoomrp-wip, and if you need one of the missing features, it's not too complicated to add it to current model. For example, we have expanded attributes to admit custom values and ranges in https://github.com/odoomrp/odoomrp-wip/tree/8.0/mrp_product_variants_types, and we are also developing a sale configurator guided through restrictions in consequent attributes.

For me, it's a nonsense to diverge in the standard model.

florian-dacosta:
@pedrobaeza I am not so sure it is incompatible with standard odoo model.
I mean, the base custom attribute module allow to add attribute to any model not just product.template

Also I think we can need the standard odoo attributes and these attribute.attribute together.

For instance, with prestashop, I think we would need both.
The odoo attributes would match what is called the combinations and combination values in prestashop.
The attributes.attributes would match the products feature in prestashop. The goal of these product features in prestashop is not to create variant, but just to add any custom field to the product.
In prestashop, combinations and features are 2 thing different, it would be the same in odoo.

Well, I am not sure we would maintain a module if it is used only by prestashop connector, but I think it can have many uses...
At least the base custom attributes which, once again, can be interesting for any model.
What do you think?

pedrobaeza:
Hi, Florian, you are right. Although product attributes has no sense now in v8, having this base module for any general purpose can make sense.

But then, the target for the module shouldn't be here, because this repository is for product related thing. The problem is that there's no a clear candidate. Maybe server-tools can be one.

ghost:
@pedrobaeza Should I move this PR to server-tools ?

guewen:
@pedrobaeza 

> But you should put this functionality in top of the v8 base, not to create an incompatible set of modules with another data structure.

I don't see how the v8 core datastructure could be extended to allow such things as attribute values are a  float, a selection or a many2one.
I think they made a too simplistic datamodel for attributes which is not extensible for more advanced use cases so this module is still required when we reach the limits of the core implementation. Still, I hope that someone can prove me I'm wrong.
Do you have propositions?

pedrobaeza:
Well, I was always referring to the product part, but you're right that the other part is not covered.

guewen:
I refer to the product part too when I say it is too limited. The core product attributes can only contain char values.
When you say:

> Although product attributes has no sense now in v8

I disagree for the aforementioned reasons

pedrobaeza:
But are you going to maintain two set of attributes (regular ones and these)? In OdooMRP, we have expanded the regular ones adding some fields to product.attribute and product.attribute.value. You can see for example https://github.com/odoomrp/odoomrp-wip/tree/8.0/product_attribute_types and https://github.com/odoomrp/odoomrp-wip/tree/8.0/product_packaging_through_attributes

ghost:
@pedrobaeza Will odoomrp-wip modules be merged into OCA someday or is it meant to stay a seperate project ? In the second case, why maintain 2 different projects ?

pedrobaeza:
Yeah, in fact we already have started to move some modules, and in Odoo experience there was a commitment in the community to help migrating all the modules to OCA.

We worked in this branch to make a fast-development cycle avoiding burocracy (look at the number of commits/PRs), but it's only a temporary state.

guewen:
> But are you going to maintain two set of attributes (regular ones and these)? In OdooMRP, we have expanded the regular ones adding some fields to product.attribute and product.attribute.value. You can see for example https://github.com/odoomrp/odoomrp-wip/tree/8.0/product_attribute_types and https://github.com/odoomrp/odoomrp-wip/tree/8.0/product_packaging_through_attributes

This is indeed interesting and seems well done. I'm maybe too negative on what is achievable only by extending the existing model. Anyway, it surely needs more thinking to see if really it can match to what was done in product_custom_attributes.

pedrobaeza:
For now, I think we can continue with this module migration as a foundation for other features (in product or in another model), and let's see requirements for product attribute extensions to see if they fit in current model or require this one.

What do you think about the repo for this one?

ghost:
My opinion is that this module belongs to the server-tools repo. Do you have any other repo in mind ?

guewen:
@dufresnedavid @pedrobaeza I agree with server-tools

ehdem:
See #93

PR #72, opened 2015-05-19
[MIG] Product Custom Attributes

ghost:
@joaoalf


pedrobaeza:
The same as https://github.com/OCA/product-attribute/pull/71#issuecomment-103692932

florian-dacosta:
Hello,

I am testing this module but I have a bug.
The 'open_attributes' button does not work when we are on a product form.
It only works when we are on a template form.

If you click this button from a product.product, you will get this error : 
AttributeError: 'product.product' object has no attribute 'open_attributes'

florian-dacosta:
Well, I see it is a general problem with buttons on product.product.
But we definitly need to be able to open the attributes from the products, and not only from the templates.

I am not sure what is the good solution.
Maybe simply  define the method "open_attributes" on product.product class and call the method on template, this way : 

```
    _inherit = "product.product"

    @api.multi
    def open_attributes(self):
        self.ensure_one()
        return self.product_tmpl_id.open_attributes()
```

What do you think?

ghost:
@florian-dacosta sorry, I am not working on this module right now.
@ehdem, I think you had the same problem ? Did you fix it in another branch ? if so, you may push your changes in this branch.

ehdem:
@florian-dacosta : use my branch: https://github.com/OCA/product-attribute/pull/93

florian-dacosta:
Ok, thanks for the feedback, I will look the other branch!

PR #74, opened 2015-06-01
migrate product_attribute_types addon from odoomrp-wip

rdeheele:
@pedrobaeza @max3903 @anajuaristi
This PR proposes to add the addon created here https://github.com/odoomrp/odoomrp-wip

Description
"This module extends product attributes, adds new type range, also the option of
defining if it is required and a default value for the attribute."

It uses the migration method explained here https://github.com/odoomrp/odoomrp-wip/wiki/Migration-to-OCA to keep commits history.

PR #75, opened 2015-06-02
Add product_attribute_types

rdeheele:
Recreates https://github.com/OCA/product-attribute/pull/74

Proposes product_attribute_types created here https://github.com/odoomrp/odoomrp-wip

"This module extends product attributes, adds new type range, also the option of
defining if it is required and a default value for the attribute."

The addon name can be discussed.


pedrobaeza:
Please take into account my previous comments in the PR https://github.com/OCA/product-attribute/pull/74

product_attribute_types/models/product.py
```diff
@@ -0,0 +1,47 @@
...
+    attr_type = fields.Selection(string='Type', store=False,
+                                 related='attribute_id.attr_type')
+
+
+class ProductAttributeValue(models.Model):
+    _inherit = "product.attribute.value"
+
+    attr_type = fields.Selection(string='Type',
+                                 related='attribute_id.attr_type')
+    numeric_value = fields.Float('Numeric Value', digits=(12, 6))
```
StefanRijnhart: How about a decimal precision instead of hardcoding?

rdeheele: @StefanRijnhart good idea, I'll add it.

pedrobaeza: The problem is that depending on the use of this field, the precision can be different. For now, we use it for at least three things:
- the quantity of the product associated in another field (module _mrp_bom_through_attributes_).
- the conversion factor between packages (module _product_packaging_through_attributes_).
- another one I didn't remember just now :wink: 

so I let it with enough resolution for each use case.

Anyways, you can create a new decimal precision for this, and let the user decides, but don't put any of the existing ones.

rdeheele: I see, I propose to create a new one named "Product Attribute".

product_attribute_types/README.rst
```diff
@@ -0,0 +1,34 @@
+.. image:: https://img.shields.io/badge/licence-AGPL--3-blue.svg
+    :alt: License: AGPL-3
+
+Product attribute types
+=======================
+
+This module extends product attributes, adds a new type range, and also the
+option of defining if it is required and a default value for the attribute.
+
```
StefanRijnhart: I would state here that this module provides basic functionality that is extended upon by more functional modules.

product_attribute_types/data/product_attribute_data.xml
```diff
@@ -0,0 +1,11 @@
+<?xml version="1.0" encoding="utf-8"?>
+<openerp>
+    <data noupdate="1">
+
+        <record forcecreate="True" id="decimal_product_attribute" model="decimal.precision">
```
StefanRijnhart: forcecreate's default is True already.

StefanRijnhart:
Oh, I think you need to regenerate the pot to make the name of the precision translatable

product_attribute_types/models/product.py
```diff
@@ -0,0 +1,58 @@
...
+    numeric_value = fields.Float('Numeric Value',
+                                 digits=dp.get_precision('Product Attribute'))
+    min_range = fields.Float('Min',
+                             digits=dp.get_precision('Product Attribute'))
+    max_range = fields.Float('Max',
+                             digits=dp.get_precision('Product Attribute'))
+
+    _sql_constraints = [
+        ('range_check',
+         'CHECK (min_range < max_range)',
```
StefanRijnhart: This constraint cannot be added on a database with demo data, and if it could be added, it would break adding values for attributes of other types than range because in that case min_range == max_range == 0.0.

pedrobaeza: This should be a python constraint that only applies when the attribute type is range.

rdeheele:
it seems that it's the product_sequence addon that fails on runbot build.

product_attribute_types/__init__.py
```diff
@@ -0,0 +1,20 @@
...
+#    but WITHOUT ANY WARRANTY; without even the implied warranty of
+#    MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+#    GNU General Public License for more details.
+#
+#    You should have received a copy of the GNU General Public License
+#    along with this program.  If not, see http://www.gnu.org/licenses/.
+#
+##############################################################################
+
+from . import models
```
bealdav: no code here => no license required. Let the cloud lighter ;-)

product_attribute_types/README.rst
```diff
@@ -0,0 +1,35 @@
+.. image:: https://img.shields.io/badge/licence-AGPL--3-blue.svg
+    :alt: License: AGPL-3
+
+Product attribute types
+=======================
+
```
bealdav: I can't find OCA guidelines, but as far I know, plural form is deprecated for module names

product_attribute_types/models/__init__.py
```diff
@@ -0,0 +1,20 @@
...
+#    but WITHOUT ANY WARRANTY; without even the implied warranty of
+#    MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+#    GNU General Public License for more details.
+#
+#    You should have received a copy of the GNU General Public License
+#    along with this program.  If not, see http://www.gnu.org/licenses/.
+#
+##############################################################################
+
+from . import product
```
bealdav: no code here => no license required

bealdav:
Thanks to all contributors for this nice job, only change module name and then :+1:

pedrobaeza:
Yeah, this can be rename to _product_attribute_multi_type_ for keeping the meaning and respect the singular form

rdeheele:
I've just changed it, thanks

pedrobaeza:
:+1:

bealdav:
:+1:

StefanRijnhart:
Thanks for the changes! :+1:

gdgellatly:
:-1: @rdeheele @pedrobaeza @StefanRijnhart @bealdav 

sorry I know this has been merged, but I have the same complaint as the latest pull request.  The required part of this module simply does not behave as expected.  (I would expect it to behave exactly like the mandatory boolean of v7 product_variant_multi) and I don't understand quite what the behaviour should be for other types other than select. (Although I have a great use case for it)

I get that this is designed for odoomrp, but it still needs to work standalone.

If I select not required on an attribute line, I would expect it to do the following.
Line 1: Green, Blue Not required
Line 2: Large, Small Required

Variants
Large
Small
Large, Green
Large Blue
Small Green
Small Blue

However at the moment it has no effect whatsoever.  See screenshots
screenshot from 2015-07-03 15 55 17
screenshot from 2015-07-03 15 55 05

Whats more what is the expected behaviour of the new types when variants are created?
I added a range for Length of 1-50 called Std
and 2 numerics for Height 1 and 2.

This is the result
screenshot from 2015-07-03 16 09 20
screenshot from 2015-07-03 16 09 06

Honestly for this repo to work I think we are going to need a module called lets say product_variant_base which refactors and overwrites the Odoo variant creation which is just plain wrong and for all the attribute line modules to inherit from it with good hooks to produce desired behavior.

I also think that a module like this doesn't belong here, at least without the modules that required its creation because it is just very confusing.

As another note, just for fun, I'd like to see product_sequence taken out until it plays well with variants.  All the products in that screen shot had proper part codes before I changed the attribute lines.

product_attribute_multi_type/README.rst
```diff
@@ -0,0 +1,35 @@
+.. image:: https://img.shields.io/badge/licence-AGPL--3-blue.svg
+    :alt: License: AGPL-3
+
+Product attribute multi type
+============================
+
+This module extends product attributes, adds 2 new types: range & numeric, and
+also the option of defining if it is required and a default value for the attribute.
+It brings the basic functionnality that can be used then by more fonctional modules.
```
gdgellatly, line 9: spelling -> functionality and functional

pedrobaeza:
@gdgellatly, I don't know what you expect to find with 'required' property, but this is something that works in combination with another modules like https://github.com/odoomrp/odoomrp-wip/tree/8.0/sale_product_variants or https://github.com/odoomrp/odoomrp-wip/tree/8.0/purchase_product_variants, but if you can exercise a test that fails about what you expect, we can work on it.

product_attribute_multi_type/README.rst
pedrobaeza, line 9: Fix directly in https://github.com/OCA/product-attribute/commit/9c855dee31524535946b31ea21fa47022280e50a

gdgellatly:
Hi @pedrobaeza ,  You kind of make my point.  This addon only currently works with the module 
product_variants_no_automatic_creation installed.  But basically would you agree that the behaviour, if it is to be a standalone module not depending on that module and not of type Hidden, should be identical with or without product_variants_no_automatic_creation installed.

In other words right now we have a scenario where without product_variants_no_automatic_creation the required field does absolutely nothing, but with it installed it behaves as expected.  Specifically it allows the creation of a variant that does not include the optional attribute.

My concern is that none of this is documented anywhere and the dependencies do not make this clear at all.

But in any case, I can't look at this for the next couple of weeks, but what I would propose is this.

The purpose of attribute lines is primarily to define and create variants.  Any module that utilises this model must be mindful of that.
Therefore functionality attributed to attribute lines logically should affect their creation in a way the user would expect.
Having a required field on attribute line is desirable.

Obviously you are coming at this from your quite tightly coupled odoomrp project.  I am coming at this from replicating and enhancing the capabilities of product variant multi and related modules from v7 and having a generic as possible OCA repo.  So we should be able to get a good outcome. 

What has become clear is that before any of this can be fixed (also see the other PR for attribute_line_sequence) is that to make any of these modules work we need to refactor create_variant_ids of core to make it more modular.  Its current implementation is programatically efficient but doesn't allow anything other than Odoo's idea of variants which is simplistic and wrong.  It appears the entire function has been written for the sole purpose of being able to match existing variants with a change in attribute lines and takes serious shortcuts to do so.

My question is then, do we propose this in the core module, or do we create our own (or both)?

But we have a lot of crossover and think we can really make things work.

Take for example something like product.supplierinfo.  I have seen your module for having this at variant level, however I think I have a better solution, I just can't get it to work.  The main difference is is that you overwrite seller ids with a one2many on product_id, but I do it using _compute.  It works beautifully for listing, displaying and getting supplierinfo as it can list and sequence both product specific and template information, but I haven't worked out the inverse yet and can't get new records saving.  There is also a question of usability (it affects both of us), in that a user sometimes want to set at template level for a specific product, and sometimes at product level update for template.  I've got a couple of things to try.

product_variant_cost - again we have taken 2 different approaches.  My approach for the last 5 years has been to override standard_price field (in fact what you did with seller_ids) and while somewhat meaningless to keep inventory reports correct is keep the template average as the weighted average of all variants.  Nowhere in odoo is template standard_price written directly so pretty easy to make work.  The odoomrp approach is to create a new field and override functions that write to it.  I must admit, when I first decided to do this, that was my approach, but I found myself chasing bugs forever and being restricted to what combinations of cost methods etc I could use and constantly having to make small changes to community modules.  Now I don't have much experience with the new WMS so maybe it is all OK now, but again another area where we can collaborate.

product_variant_default_code - akretion is working on this.  I see some good things in the odoomrp version, and I like the simplicity of the akretion one.  The akretion one fits 95% of my needs, but the odoomrp one meets 100% but is more complicated.  However it is simple to see how they could be combined.  Both of them are superior to v7 product_variant_multi, although maybe using {} or ${} rather than [] for token is more natural.

sale_product_variants - is an interesting one.  I am in the middle of porting some similar code (well once I have product_variant creation working properly).  The odoomrp approach is to specify these on the sale order line and use templates.  My own approach to date is to specify the common ones on the sale order and restrict search domain, expanding out as no matches are found.  I am really interested to see how we can combine these. (My use case is that generally an order is only for one colour and one metal type, but often accessories have neither).  I like the idea of using templates - faster search etc, but the idea of specifying attributes on each line would slow down order entry and crowd out the line too much.  I get complaints if they even have to use a mouse.

Anyway there is more, sorry for the long comment, but we should work together to make this repo awesome for anybody needing advanced product features.

PR #76, opened 2015-06-09
customer context in product view port to v8

gdgellatly:
Remove non working code
Change module name to better reflect new scope
Update button to smart button
Reorganize to new coding guidelines
Add README.RST
bump minor version number


gdgellatly:
I think you will find I have signed the CLA

gurneyalex:
CLA ok :smile_cat:

customer_pricing_in_product_view/README.rst
```diff
@@ -0,0 +1,82 @@
...
+'Product Pricing' button to the partner view which when clicked will open the product list
+with the pricelist set to the customers pricelist.
+
+Installation
+============
+
+To install this module, you need to:
+
+* do nothing special
+
```
gurneyalex, line 17: I think this paragraph should be removed

customer_pricing_in_product_view/README.rst
```diff
@@ -0,0 +1,82 @@
...
+Configuration
+=============
+
+To configure this module, you need to:
+
+* do nothing
+
```
gurneyalex, line 24: I think this paragraph should be removed

gurneyalex:
runbot status related to #77 

2 non blocking remarks. 

:+1:

customer_pricing_in_product_view/README.rst
gdgellatly, line 17: @gurneyalex I wrote it as a joke initially, however the more I think about it I think it should stay in or we should have some standard wording of the same effect and add it to quality tools.  Telling a user what they don't need to do is equally valuable as telling them what they do need to do, and if we put it everywhere and make consistent it will prompt module developers when it changes.

Maybe a simple line like, This Module has no specific installation requirements, and for the one below, This module requires no post installation configuration

rousseldenis:
:+1: tested on runbot

gdgellatly:
@pedrobaeza can we make a call on whether we merge this or not.  I'm long since over v8 so happy to close as well and I've done the v10 port work in #246 .

pedrobaeza:
OK, let's merge it to have it for those using v8.

Issue #77, opened 2015-06-09
product_sequence causes tests to fail


gdgellatly:
@AngelMoya-Domatix 

Product Sequence is causing tests on this repo to fail test.  It sets a static default and then a unique constraint.

Moreover I don't understand quite why it is done this way.  Could we not just use copy=False, default=<get sequence> and remove all the methods entirely?  This would number all existing products automatically and ensure duplicates were given a new number.

angelmoya:
Hi, I'm not author of this code, but:
- with copy = false, you lost original code, maybe you want to maintain part of original code.
- with method write, setting code="/" you can set new code of sequence, this is useful setting codes after module installation.

The test should not fail, because create method update this value. This is the same way sales sequence do it.

pedrobaeza:
This is already fixed, so I close.

PR #78, opened 2015-06-09
fix constraint on product_sequence

gdgellatly:
Fix constraint to allow NULLS at initial install
Update demo data to set default codes

This PR should resolve failing tests on this branch as well as making the module more robust for existing installations

### DIFF
--- a/product_sequence/README.rst
+++ b/product_sequence/README.rst
@@ -4,6 +4,15 @@ A module that adds sequence to the product.
 This module allows to associate a sequence to the product reference.
 The reference (default code) is unique (SQL constraint) and required.
 
+Installation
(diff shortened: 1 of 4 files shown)


gdgellatly:
closes issue #77

gdgellatly:
Need some help here, the unique constraint combined with making the field required makes the initial run fail, even though afterwards the ORM call to write actually updates those defaults set to '/' to the sequence.  It seems the demo data is loaded after the module install so that doesn't help either.  It is causing the tests to fail and holding up all the PR's on this branch.

gdgellatly:
Replacing with a new PR, messed up git rebase

PR #81, opened 2015-06-19
[ADD] New module <product_last_price_info>

oihane:


pedrobaeza:
I'm seeing that we can use the trick of declaring this view with the same ID in both modules that use it and avoid this dummy module.

oihane:
Yes, that's true, but then the modules won't be installed automatically, user will be forced to install the required ones by hand

pedrobaeza:
Well, I prefer that option, because this way you need to setup multiple repositories.

oihane:
Ok, I close this, and change the others

PR #84, opened 2015-06-29
[ADD] add product_attribute_line_sequence addon

rdeheele:
Allows to order attributes on the product form.

### DIFF
--- a/product_attribute_line_sequence/README.rst
+++ b/product_attribute_line_sequence/README.rst
@@ -1,0 +1,31 @@
+.. image:: https://img.shields.io/badge/licence-AGPL--3-blue.svg
+    :alt: License: AGPL-3
+
+Product attribute line sequence
+===============================
+
+This module adds a sequence field on product attribute lines
+to order attributes in the product form.
+
+
+Credits
+=======
+
+Contributors
+------------
+* Romain Deheele <romaindeheele@gmail.com>
+
+Maintainer
+----------
+
+.. image:: https://odoo-community.org/logo.png
+   :alt: Odoo Community Association
+   :target: https://odoo-community.org
+
+This module is maintained by the OCA.
+
+OCA, or the Odoo Community Association, is a nonprofit organization whose
+mission is to support the collaborative development of Odoo features and
+promote its widespread use.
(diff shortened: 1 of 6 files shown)


pedrobaeza:
Thanks :+1:

gdgellatly:
:-1: for now for the same reasons as my original comments in the OCB proposal, except now tested.  I really want the functionality, but it needs to work as expected.

See screen shots
These 2 are existing Ipad Retina with attributes reordered to put colour first

screenshot from 2015-07-03 09 18 23
screenshot from 2015-07-03 09 18 47

These 2 are a duplicated Ipad Retina which deletes the attribute lines, readded in the order of Color, Memory, Wifi, then reordered to Wifi, Memory Color.
screenshot from 2015-07-03 09 22 37
screenshot from 2015-07-03 09 22 12

As you can see it just doesn't do what a user would expect, which is reorder the attributes.

The interesting thing about this test, is that it also didn't order them in the order of the attribute lines originally, but rather in the order of the attribute ids.  Looking at this it is because _rec_name is set to attribute_id.  So the good news is I expect this can be solved by just adding an _order to product_attribute_line of _order = 'sequence, id' (hopefully id works, it seems necessary because of the rec_name). 

Getting this behaviour is really important, especially for the PR's for generating product codes etc.

gdgellatly:
Hmmm I've also noticed that it is reordering attribute values not in the order entered but by id.  That might affect my product_attribute_group modules.

gdgellatly:
Crap it is worse than that.  Man I hate the way Odoo writes some things, look at this nightmare screenshot.  In create_variant_ids, a method so horrible to start with it uses int(id) twice in fact with sorted to order things regardless of what you do.

screenshot from 2015-07-03 09 48 20

Looks like its going to need an OCB proposal to fix.

pedrobaeza:
I remove my positive review until Graeme's comments are achieved, because indeed it's a blocking issue

gdgellatly:
@pedrobaeza @rdeheele If you could help support https://github.com/odoo/odoo/issues/7814 we might be able to get this working.

PR #85, opened 2015-07-08
8.0 product variant inactive

ahilali:


gdgellatly:
@ahilali I think you have mixed up your branches or something, there are 3 modules in this PR.  the image ones are causing travis and runbot to fail.

product_variant_inactive/README.rst
```diff
@@ -0,0 +1,45 @@
+.. image:: https://img.shields.io/badge/licence-AGPL--3-blue.svg
+ :alt: License: AGPL-3
+
+Module name
+===========
+
+This module was written to extend the functionality of Product Variants
+and allow you to make activate or deactivate products using the two buttons
```
rvalyi, line 8: s/make activate or deactivate products using the two buttons/enable or disable product variants with two buttons/

rvalyi:
same remark as @gdgellatly. Seems cool otherwise, will test better once you clean the branch as we also need that here.

PR #86, opened 2015-07-09
[IMP] Generate a new sequence on copy

joaoalf:
Improve copy behavior.

### DIFF
--- a/product_sequence/product_product.py
+++ b/product_sequence/product_product.py
@@ -30,7 +30,7 @@ class product_product(orm.Model):
         'default_code': fields.char('Reference',
                                     size=64,
                                     select=True,
-                                    readonly=True),
+                                    required=True),
         }
 
     _sql_constraints = [
--- a/product_sequence/product_product.py
+++ b/product_sequence/product_product.py
@@ -30,7 +30,7 @@ class product_product(orm.Model):
         'default_code': fields.char('Reference',
                                     size=64,
                                     select=True,
-                                    required=True),
+                                    readonly=True),
         }
 
     _sql_constraints = [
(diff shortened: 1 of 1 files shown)


pedrobaeza:
Making this field readonly is not a good idea, because there are people that wants to modify manually the code.

joaoalf:
@pedrobaeza I beg to differ. IMHO the whole point of having a sequence field is to have it filled without the user interference. Since it has a unicity constraint, changing it manually will only create the possibility of conflicts later.

pedrobaeza:
I see this as a default sequence when no value is specified, but letting the user to customize when he wants. There's no possibility of conflict, because the constraint will raise when trying to save a duplicated value, and you can always access to the sequence to modify it.

joaoalf:
@pedrobaeza What about this case?

The actual sequence is PR/00099 but among the product records there's a product code that was manually set to PR/00100 by one user. Then anothe user tries to create a new product and the system automatically generates a new sequence PR/00100 but fails in the create() because the unicity constraint. The poor user will be confronted with 2 options: To change the sequence itself or to change the code of the PR/00100 product to avoid further conflicts. In both cases the user will be probably forced to abandon the creation of the product to fix the problem causing frustration and time loss. 

Letting the field default_code open is valid as long as we can insure that manually create codes will be outside the sequence range and this is not the case right now.

pedrobaeza:
That improbable case can be easily fixed calling the system administrator for changing the sequence next number to 101. You are only thinking on people putting manual product codes that matches the sequence, but most of the times the manual overriding of the code will be for setting a totally different code, for example, the supplier code (for instance, X34545AB).

joaoalf:
@pedrobaeza I think you misundertood me. It's not about me thinking only about this or that case. It's about me knowing that at the current version the module allows this kind of confict to happen and this renders it useless for some enviromnents.

That said. Since there's still a conflict with the base tests because of the read-only on the field I'm willing to put the read-only part into another PR.

pedrobaeza:
This has CIs statuses red and no further devolpment, so I close. If there's still interest, please open a new PR fixing CIs.

PR #87, opened 2015-07-09
[IMP] Make the field default_code read-only

joaoalf:
Make the field default_code read-only.

### DIFF
--- a/product_sequence/__openerp__.py
+++ b/product_sequence/__openerp__.py
@@ -37,6 +37,7 @@ Note: This module is incompatible with nan_product_sequence.
     ],
     "data" : [
         'product_sequence.xml',
+        'product_sequence_view.xml'
     ],
     'installable': True,
     'active': False,
--- a/product_sequence/__openerp__.py
+++ b/product_sequence/__openerp__.py
@@ -36,6 +36,7 @@ Note: This module is incompatible with nan_product_sequence.
         'product',
     ],
     "data" : [
+        'security/product_sequence.xml',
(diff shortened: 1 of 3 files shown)


joaoalf:
@pedrobaeza This is a meet-in-the-middle solution to complement PR #86

gdgellatly:
This feels like it starts to get really messy.  I've got nothing against this PR in isolation, but looking down the list we have targeted at 7.0 already

A module to split out uniqueness from sequence.
Another one for product_unique_code
Another one for product_code_unique and product_code_mandatory
This one to make default_code readonly

Then in 8.0 I did quite a bit of work changing product_sequence to make the branch green (merged) plus add some tests (not merged).  But the module is still causing issues in travis for other modules if they create products.
Another one to generate the sequence based off the product_category which is quite weird.

Honestly I think what we need to do here is make some decisions and have one or two PR's that close most of these.  I think we forget about the category one for now.

My personal feeling on this is that firstly current changes to 8.0 should be backported to 7.0 (plus tests) so we can at least have runbot and travis green as a starting point.

Then we should split out product_sequence into two modules.
1: Contains the functionality of product_code_mandatory, product_unique in one module (maybe2, but preferable 1).
2: Contains product_sequence plus this PR depending on modules in 1.

My gut feel is that 1 has wider appeal than 2.  But then we can kill 5 or 6 PR's.

pedrobaeza:
This has CIs statuses red and no further development, so I close. If there's still interest, please open a new PR fixing CIs.

PR #92, opened 2015-08-18
[port][8.0] product_with_supplier_unit_price

legalsylvain:
- port product_with_supplier_unit_price;
- use new api;
- use OCA convention;
- add demo data and give correct access right to admin user to test correctly; 
- Improve display if there are many prices for a product and a supplier;

product-attribute

Thanks for your review. (you can skip the review of the first commit that is just a move from "unported" folder) 
Kind regards.


product_with_supplier_unit_price/__openerp__.py
```diff
@@ -15,30 +15,27 @@
...
-product_supplierinfo()
-
-# vim:expandtab:smartindent:tabstop=4:softtabstop=4:shiftwidth=4:
+{
+    'name': 'Supplier Unit Price',
+    'summary': 'Supplier Unit Price on Product Form',
+    'version': '8.0.1.0',
+    'author': 'Savoir-faire Linux,GRAP,Odoo Community Association (OCA)',
+    'website': 'https://odoo-community.org/',
+    'license': 'GPL-3',
```
pedrobaeza: Why GPL?

pedrobaeza:
I think a better name can be _product_supplierinfo_tree_price_info_.

pedrobaeza:
I see that the module only considers two price sections, but not any arbitrary number of price scales. I also see that the string format is a bit cryptic and long, so I propose to change it to this one:

```
1-49: 100.0; >=50: 95.0
>=1: 100.0
```

What do you think?

product_with_supplier_unit_price/__openerp__.py
legalsylvain: It's an error from V7 I guess. licence header is AGPL. I'll change it.

legalsylvain:
> I think a better name can be product_supplierinfo_tree_price_info.

Sound's good.

```
1-49: 100.0; >=50: 95.0
>=1: 100.0
```

Excellent suggestion, Thanks ! I was looking for something less cryptic but I had no idea.

legalsylvain:
@pedrobaeza : Done. (cf screenshot). Thanks for your suggestion !

product_supplierinfo_tree_price_info/models/product_supplierinfo.py
```diff
@@ -0,0 +1,59 @@
...
+
+from openerp import fields, api
+from openerp.models import Model
+from openerp.tools.translate import _
+
+
+class ProductSupplierinfo(Model):
+    _inherit = 'product.supplierinfo'
+
+    @api.multi
```
laetitia-gangloff, line 31: Why there is no api.depends on pricelist_ids ?

product_supplierinfo_tree_price_info/models/product_supplierinfo.py
```diff
@@ -0,0 +1,59 @@
...
+                        supplierinfo.pricelist_ids[i + 1].min_quantity,
+                        supplierinfo.pricelist_ids[i].price)
+                txt += '>=%s : %s' % (
+                    supplierinfo.pricelist_ids[size - 1].min_quantity,
+                    supplierinfo.pricelist_ids[size - 1].price)
+                supplierinfo.unit_price = supplierinfo.pricelist_ids[0].price
+                supplierinfo.unit_price_note = txt
+
+    unit_price_note = fields.Char(
+        compute=_get_unit_price, multi='unit_price', string='Unit Price')
```
pedrobaeza: Put compute method between quotes to allow easy inheritance

product_supplierinfo_tree_price_info/models/product_supplierinfo.py
```diff
@@ -0,0 +1,59 @@
...
+
+    unit_price = fields.Float(
+        compute=_get_unit_price, multi='unit_price',
```
pedrobaeza: The same

product_supplierinfo_tree_price_info/models/product_supplierinfo.py
pedrobaeza, line 31: It is not needed when store=False.

product_supplierinfo_tree_price_info/models/product_supplierinfo.py
```diff
@@ -0,0 +1,59 @@
...
+from openerp import fields, api
+from openerp.models import Model
+from openerp.tools.translate import _
+
+
+class ProductSupplierinfo(Model):
+    _inherit = 'product.supplierinfo'
+
+    @api.multi
+    def _get_unit_price(self):
```
LeartS: function field method should start with `_compute` (Odoo and OCA guideline)

legalsylvain: Thanks, I didn't know !

product_supplierinfo_tree_price_info/models/product_supplierinfo.py
```diff
@@ -0,0 +1,59 @@
...
+        for supplierinfo in self:
+            if len(supplierinfo.pricelist_ids) == 0:
+                supplierinfo.unit_price = 0
+                supplierinfo.unit_price_note = _('-')
```
LeartS: Why a translateable `-` where in lines 42 and 46 the symbols are not?

LeartS:
I would rename `views/view.xml` in `views/product_supplierinfo.xml` https://github.com/OCA/maintainer-tools/blob/master/CONTRIBUTING.md#file-naming

LeartS:
Apart minor remarks, :+1:

legalsylvain:
@LeartS : Thanks for your review. Changes done. Regards.

pedrobaeza:
Please check CIs errors

legalsylvain:
@pedrobaeza : done.

pedrobaeza:
Thanks for all the changes, @legalsylvain. Trying the module on runbot, I have found one inconsistence: if I define 1 and 5 as quantities, I get this:

```
1.0-5.0: 100
>= 5.0: 95
```

In reality, it should be `1.0-4.99: 100`, because 5.0 already gets the other price. To achieve this you should decrease to the next interval the following formula:

```
1 / 1 * (get_precision('Product Unit of Measure') + 1)
```

legalsylvain:
@pedrobaeza : I didn't understood your formula, but I applied your idea.
Regards.

pedrobaeza:
I like the way you have done it more than mine, @legalsylvain. A little PEP8 problem, and we can go.

legalsylvain:
Hi @pedrobaeza : is it now ok for you ? 
Kind regards.

product_supplierinfo_tree_price_info/demo/res_groups.yml
```diff
@@ -0,0 +1,37 @@
...
+#    but WITHOUT ANY WARRANTY; without even the implied warranty of
+#    MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+#    GNU Affero General Public License for more details.
+#
+#    You should have received a copy of the GNU Affero General Public License
+#    along with this program.  If not, see <http://www.gnu.org/licenses/>.
+#
+##############################################################################
+
+- !record {model: res.groups, id: base.group_no_one}:
```
pedrobaeza, line 23: Demo data should be oriented for filling some data relevant for the module, not for preparing admin user for the manual test of it, don't you think?

pedrobaeza, line 23: You can include here some product.supplierinfo that helps to put in place an example for the price intervals.

product_supplierinfo_tree_price_info/models/product_supplierinfo.py
```diff
@@ -0,0 +1,60 @@
...
+                txt += '>=%s : %s' % (
+                    supplierinfo.pricelist_ids[size - 1].min_quantity,
+                    supplierinfo.pricelist_ids[size - 1].price)
+                supplierinfo.unit_price = supplierinfo.pricelist_ids[0].price
+                supplierinfo.unit_price_note = txt
+
+    unit_price_note = fields.Char(
+        compute='_compute_unit_price', multi='unit_price', string='Unit Price')
+
+    unit_price = fields.Float(
```
pedrobaeza, line 56: Maybe you can call this max_unit_price and make a max of the mapped prices. 99.9% of times it will be the first price (because you define intervals putting first the lowest quantities/higher prices), but it can be that you change this order or a higher quantity is more expensive (very weird indeed).

product_supplierinfo_tree_price_info/demo/res_groups.yml
legalsylvain, line 23: @pedrobaeza : Thanks for your review again ! You're on every front !

1/ the objective of demo data (**for me**) is to provide full testable environment: 
- for users. (developper or functional). The aim : "One click and you can test on runbot". So, for that, I allways set admin into all the needed groups to see correctly how the module works without any extra settings;
- for computer, providing data that will be available for automatic tests ;

2/ I put it in the other file ./product_supplierinfo.yml. Or maybe, you're talking about other things...

pedrobaeza, line 23: About point 1, both positions are reasonable, so let's stay as is.

For point 2, I didn't see the rest of the files, because I got lost a bit with the fragmentation. I would put all together because they all belong to a same user view (the product template). It's the same as purchase order and its lines, that usually come together in the same file for clarity.

product_supplierinfo_tree_price_info/models/product_supplierinfo.py
legalsylvain, line 56: I'm porting the module and the name is unit price. not 'max price'.  
In the previous version of this module, the price returned was the price associated to the row (qty=1, if exists). 
https://github.com/OCA/product-attribute/pull/92/files#diff-b88e9df436c587a50b78fe87cff35f29L39
I realized a little changes, to return the price associated to the min qty, (because by default, qty=0, so the unit price doesn't return nothing). And it was no very understandable for end users.
But, return another price (the max price) is for me another thing, and so another field. Don't you think ?

pedrobaeza, line 56: Well, it's more or less the same, because unitary price is the almost the same as maximum supplier price, but you are right that is not exactly the same. We can leave it as is

pedrobaeza:
Although there are some diverging opinions, they can both consider valid, so my :+1:

legalsylvain:
2 :+1:. Can be this PR be merged ?

pedrobaeza:
Yeah, of course!

PR #96, opened 2015-09-11
[ADD] Dimensions fields in product.template

nicomacr:


flotho:
Any possible review for this one?

pedrobaeza:
If you add the fields at product.template level, you don't need to add it again at product.product level. What is the motivation anyway for this change?

flotho:
Hi @pedrobaeza, 
My point of view is that it should work as the EAN.
Dimensions an of course weight could be defined by default on template but could be different for specific variant as the attributes could impact weight.
Is it a weird point of view?

regards

pedrobaeza:
OK, seems reasonable. Please make the suggested changes.

product_dimension/product.py
```diff
@@ -50,3 +50,16 @@ def convert_to_meters(self, measure, dimensional_uom):
...
         domain="[('category_id.name', '=', 'Length / Distance')]",
         help='UoM for length, height, width')
+
+
+class Product_template(models.Model):
+
+    _inherit = 'product.template'
+
+    length = fields.Float(related='product_variant_ids.length')
+    height = fields.Float(related='product_variant_ids.height', oldname='high')
```
gurneyalex: oldname: not needed as this is a new field

gurneyalex:
why no volume on product_template? 

Please add an automated test showing that the dimensions and volume can be different in the product.template and a related product.product.

nicomacr:
Thank you for the suggestion, I'm not very familiar with the subject of the test , there are some tutorial?

nicomacr:
There is no need to add volume on product.template because it is there by default. We have add the on change on product.template so it also computes from there when you have 1 product.template --> 1 product.product relationship. I really don't know how to build that test. Can you please accept this pr the same way?

jjscarafia:
This PR is now ok, can you please review it?
@gurneyalex we add a test to get weight from product template.
We didn't work on your suggestion because weight is a native field of odoo defined on "product_template" and not redefined on "product_product", so, it exists only on tempaltes.

The aim of this PR is to allow you to configure dimensions from templater or product views.

Thanks!

jjscarafia:
Hello! this PR is ready and fully tested. Modules is unusable with this fix.

jjscarafia:
Hi @pedrobaeza can you help us with this PR? This module is unusable without this fix.
How can we have PR get reviewed more quickly? We are working to send all our modules to OCA but with more than 16 days to approve a PR it discourage that task. Please let us know if we should do something different.
Thanks,
Juan

pedrobaeza:
:+1: 

As the module has tests and everything is green, I'm going to merge although there is no a second review because it has been a long time.

jjscarafia:
@pedrobaeza  Thanks!

PR #99, opened 2015-10-13
check_upc() fails when check-digit is 0

pjmcdermott:
I discovered that the check_upc() routine was not working when the check-digit was zero, as it was missing the:

```
if check == 10:
 check = 10
```

that the other `check_x()` routines have.

Once I started poking around the rules for GTIN checksums, I discovered that the algorithm is identical for EAN-8, EAN-13, UPC, EAN-14---you just have to start at the right hand side (ignoring the check digit).

So this has led to some major re-factoring as there was considerable duplication of the GTIN checking algorithm.

I also added some unit tests, as it was a gap that this module did not have any real-world EANs or UPCs to verify the `check_x()` routines with.

Finally, once all of the identifiers are being checked by the same function, the section to route the identifier to the appropriate check function was redundant, so I zapped that too!

### DIFF
--- a/product_gtin/product_gtin.py
+++ b/product_gtin/product_gtin.py
@@ -83,21 +83,21 @@ def check_gtinx(code):
 VALID_GTIN_LENGTHS = [8, 10, 12, 13, 14]
 
 
-def check_gtin(eancode):
-    if not eancode:
+def check_gtin(code):
+    if not code:
         return True
-    if len(eancode) not in VALID_GTIN_LENGTHS:
+    if len(code) not in VALID_GTIN_LENGTHS:
         return False
-    if len(eancode) == 10:
+    if len(code) == 10:
         # Should be an ISBN-10
-        return check_isbn(eancode)
+        return check_isbn(code)
     else:
         # SHould be GTIN-x
         try:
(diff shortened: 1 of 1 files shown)


hugosantosred:
I can confirm this bug with UPC Codes when check digit is 0. For example this code 810354020810 is valid but Odoo says that its not.

Functionally this PR is OK for me :+1:

I'll try also to make the technical review.

Thanks! @pjmcdermott

product_gtin/product_gtin.py
```diff
@@ -191,14 +139,72 @@ class res_partner(orm.Model):
...
+            'EAN', size = max(VALID_GTIN_LENGTHS),
             help="Code for %s" % HELP_MESSAGE),
         }
 
     _constraints = [(_check_ean_key, CONSTRAINT_MESSAGE, ['ean13'])]
+    
+    
+if __name__ == '__main__':
+    
+    import unittest
```
pedrobaeza, line 157: Make tests in the tests folder and with Odoo structure.

pedrobaeza:
Please squash commits to not have too much for a simple patch.

PR #104, opened 2015-10-28
Add new field 'default_code' on product_brand

lmignon:


product_brand/product_brand.py
```diff
@@ -49,12 +49,18 @@ class ProductBrand(models.Model):
...
     )
     products_count = fields.Integer(
         string='Number of products',
         compute='_get_products_count',
     )
 
+    code = fields.Char(
+        string='Reference',
+        index=True
+        )
```
LeartS: Could you please put the closing parenthesis where the other fields put it, for consistency?

LeartS:
Code review :+1: 

I do not know if such a feature deserves its own module in an OCA repo, it's trivial and I think not much used.
At least, I would put the unique part of https://github.com/OCA/product-attribute/pull/105 here, and leave only the sequence part there. Allowing for duplicates code is a very rare requirement, I think.

lmignon:
@LeartS Thank you for the review. Changes done.

laetitia-gangloff:
:+1: (code and functional test)

PR #107, opened 2015-11-02
[FIX] Added size=-1 for compatibility with integer indexes in Selecti…

gfcapalbo:
…on field.

### DIFF
--- a/product_pricelist_fixed_price/model/product_pricelist_item.py
+++ b/product_pricelist_fixed_price/model/product_pricelist_item.py
@@ -33,6 +33,7 @@ class ProductPricelistItem(models.Model):
 
     base_ext = fields.Selection(selection='_price_field_get_ext',
                                 string='Based on',
+                                size=-1,
                                 required=True,
                                 default=lambda self:
                                     self.default_get(


pedrobaeza:
What is the rationale behind this?

gfcapalbo:
Important FIx for allowing base_ext to work correctly. Selection fields in Odoo will usually have strings as indexes for every option, if the index is an INTEGER as in base and base_ext , it is mandatory to add size = -1 to the field attributes in order to allow the ORM to correctly evaluate the field and not return an UNKNOWN option.

pedrobaeza:
Does this also happens on 7.0?

gfcapalbo:
yes , untested , but the ORM code is unchanged. it will check for size=-1 to make integers work as indexes. As a matter of fact the 7.0 version had size=-1.
This is an undocumented Feature of selection field. It should be in the documentation of selection Fields I believe.  No fix needed for 7.0, it is already OK.

pedrobaeza:
Yeah, indeed it should be documented. I didn't know about it.

gfcapalbo:
Will this be approved? any thumbs up? @pedrobaeza  how can i advertise this?
I really don't want the defective version to stay in the main branch.
Here is the code in models.py that explains why we need size=-1 for selection fields with integer indexes

https://github.com/odoo/odoo/blob/8.0/openerp/models.py#L195-L196

pedrobaeza:
OK, as this is a critical bug, I fast-track it with the corresponding acknowledge by my part.

gfcapalbo:
thanks.

PR #108, opened 2015-11-09
[MIG]Product brand

flotho:


product_brand/product_brand_view.xml
```diff
@@ -129,18 +129,19 @@
...
             <field name="name">product.search.form</field>
             <field name="model">product.product</field>
             <field name="inherit_id" ref="product.product_search_form_view"/>
             <field name="arch" type="xml">
-                <field name="categ_id" position="after">
+                <field name="name" position="after">
                     <field name="product_brand_id"/>
                 </field>
-                <group string='Group by...' position="inside">
+<!--                <group string='Group by...' position="inside">
```
pedrobaeza: Use `<filter name="template_id" posititon="after">` instead

product_brand/product_brand_view.xml
```diff
@@ -152,10 +153,10 @@
                 <field name="name" position="after">
                     <field name="product_brand_id"/>
                 </field>
-                <group string='Group by...' position="inside">
+<!--                <group string='Group by...' position="inside">
```
pedrobaeza: Use `<group posititon="inside">` instead. It will find the only group in the search

pedrobaeza:
Bump version number

LeartS:
Why does runbot report as successful when the build stopped with an (unrelated) error?

lasley:
What is the status of this PR?

rvalyi:
Hello repo PSC, I think this PR can be closed. Indeed product_brand is available in 9.0
https://github.com/OCA/product-attribute/tree/9.0/product_brand
cc @pedrobaeza

PR #112, opened 2015-11-19
[8.0][base_multi_image][product_multi_image] Multi images for products split in abstract and concrete models

yajo:
According to https://github.com/OdooCommunityWidgets/website_multi_image/issues/34#issuecomment-157990715, I'm pushing this PR that continues job from https://github.com/OCA/product-attribute/pull/57 to discuss about design, but it is a WIP right now.


pedrobaeza:
The name should be _base_multi_image_ and I'm not sure if the repo target should be this or server-tools, and let _product_multi_image_ for here. Anway, this is a WIP and we can let here the discussion.

multi_image_base/README.rst
```diff
@@ -0,0 +1,92 @@
...
+Usage
+=====
+
+To use this module, you need to:
+
+* Follow instructions on the module based on this one that you install.
+
+To develop a module based on this one:
+
+* See module ``multi_image_product`` as an example.
```
pedrobaeza: _product_multi_image_: always first the name of the "area".

multi_image_base/README.rst
```diff
@@ -0,0 +1,92 @@
+.. image:: https://img.shields.io/badge/licence-AGPL--3-blue.svg
+   :target: http://www.gnu.org/licenses/agpl-3.0-standalone.html
+   :alt: License: AGPL-3
+
+====================
+Multiple images base
+====================
+
+This module was written to extend the functionality of any model to support
```
elicoidal: This modules extends the...

multi_image_base/README.rst
```diff
@@ -0,0 +1,92 @@
...
+having multiple images (a gallery) attached to it and allow you to manage them.
```
elicoidal: remove 'having'

multi_image_base/README.rst
elicoidal: product_image_multi?

multi_image_base/models/image.py
```diff
@@ -0,0 +1,169 @@
...
+        help="Image path")
+    url = fields.Char(
+        'Image remote URL')
+    image_main = fields.Binary(
+        "Full-sized image",
+        compute="_get_image")
+    image_medium = fields.Binary(
+        "Medium-sized image",
+        compute="_get_image_sizes",
+        help="Medium-sized image. It is automatically resized as a 128 x "
```
elicoidal: Carriage Return after 'as a' to keep the size words together

multi_image_base/models/image.py
```diff
@@ -0,0 +1,169 @@
...
+             "128 px image, with aspect ratio preserved, only when the image "
+             "exceeds one of those sizes. Use this field in form views or "
+             "some kanban views.")
```
elicoidal: remove 'some'

product_multi_image/README.rst
```diff
@@ -0,0 +1,41 @@
+Multiple images in products
```
elicoidal, line 6: I miss the license badge

product_multi_image/README.rst
```diff
@@ -0,0 +1,41 @@
...
+
+
+Contributors
+------------
+
+* Pedro M. Baeza <pedro.baeza@serviciosbaeza.com>
+* Rafael Blasco <rafabn@antiun.com>
+* Jairo Llopis <yajo.sk8@gmail.com>
+
+Maintainer
```
elicoidal, line 57: I miss the bug tracker section

yajo:
I refactored most code to make it easier to inherit, and moving as much as possible to the base module, following the method learned with https://github.com/OCA/server-tools/pull/313. It fixed some bugs by the way, but there is one remaining that I'd thank help:
- Create a product.
- Go to images tab.
- Add one.
- That image gets set as the product's image.

Now the reverse way does not work:
- Create a product.
- Set it an image.
- No multi_image record is created, but the product gets the image!

yajo:
Finally it works (I hope). Ready to review!

rafaelbn:
Travis fails

yajo:
I had to write 4a43237be6c6db9cefe1430103340bb836bd32c3 to avoid conflicts caused by https://github.com/odoo/odoo/issues/10799. Let's see if it gets :green_heart: now.

rafaelbn:
Thanks! :green_apple: please @pedrobaeza could you change label WIP by need review!

dreispt:
The `multi_image_base` should be renamed as Pedro suggested, and split into it's own PR on OCA/server-tools.
Also, this is a backend feature, so I'm puzzled at the `website` dependency.

yajo:
The name is alright because odoo names for modules are `<section>_<functionality>`. In this case `base` indicates the functionality of the `multi_image` section, not the other way around.

Also I kept both modules in 1 PR because one does nothing without the another. I understand the base one should go to server-tools, but please review this and when I have enough :+1: I will split it before merging, so you can test now in runbot and so.

pedrobaeza:
@Yajo, you can say what you want, but the name is not correct. All is prefixed with `base`, not suffixed. If not, I challenge you to show me one 8.0 standard module with the `base` word at the end. Please stop debating all the reviews you receive and accept humbly the suggestions made by the reviewers.

yajo:
Hmm you are right, I did not know that exception. About the website dependency it's about the usage of `assets_frontend` view.

product_multi_image/README.rst
yajo, line 6: This is now fixed.

product_multi_image/README.rst
yajo, line 57: This is now fixed too.

dreispt:
The multi_image module is a generic feature, not specifically linked to products.
So I insist it should go in OCA/server-tools.

dreispt:
The `website` dependency is not needed then: you can add the assets to web instead of website.

yajo:
@dreispt 

> The multi_image module is a generic feature, not specifically linked to products.
> So I insist it should go in OCA/server-tools.

I agree. But could you (and others) please test and _conditionally approve_ this before I split? It's just because one module without the other don't make too much sense and will break bots.

> The website dependency is not needed then: you can add the assets to web instead of website.

I'm afraid the [`assets_frontend` view is defined in `website`](https://github.com/odoo/odoo/blob/8.0/addons/website/views/website_templates.xml#L14). Anyway, it's not too common to want to have an image gallery for something you don't want to show up in your website, don't you think? This is another case of perfect vs practical. I think a module that just adds that template would be an overkill.

pedrobaeza:
I don't agree about the website part, @Yajo. You can have multiple images because you synchronize Prestashop/Magento products, and you only want to maintain them. For that, you don't need website.

yajo:
Do you want 2 modules then?

dreispt:
@Yajo you mean 3?

yajo:
yes, sorry.

dreispt:
Can't you just replace the `website` dependency by a `web` one?

base_multi_image/views/assets.xml
```diff
@@ -0,0 +1,16 @@
+<?xml version="1.0" encoding="utf-8"?>
+<!-- © 2016 Antiun Ingeniería S.L. - Jairo Llopis
+     License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html). -->
+
+<openerp>
+<data>
+
+<template id="assets_frontend" inherit_id="website.assets_frontend">
```
yajo, line 8: @dreispt Here's why I need website (or split module).

dreispt:
@Yajo check https://github.com/OCA/web/blob/9.0/help_popup/views/popup_help_view.xml

base_multi_image/views/owner_templates.xml
```diff
@@ -0,0 +1,54 @@
+<?xml version="1.0" encoding="utf-8"?>
+<!-- © 2016 Antiun Ingeniería S.L. - Jairo Llopis
+     License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html). -->
+
+<openerp>
+<data>
+
+<!-- Call this template in your owner's website page to get a carousel.
+     It needs a `multi_image_owner` variable set. -->
```
yajo, line 9: @dreispt These templates are to reuse in frontend website, not backend, so assets must go in frontend.

yajo, line 9: See usage example in https://github.com/OCA/event/pull/33/files#diff-068a0c4a89bc9582a1c4cf77d01be469R32

dreispt:
I don't understand we we need frontend features in a backend module. Confusing...
Front end features should go in their own modules and use a prefix `website_`

base_multi_image/views/owner_templates.xml
yajo, line 9: Since this is a base module, it provides tools to reuse in submodules, like these templates. These templates are useless without the attached CSS, and you cannot attach it without the website module. I hope you understand the dependency is **required**, so either I split the module or it stays there.

rafaelbn:
Hi @Yajo , please which is the state of this PR, please could you give a resume? Thanks

@dreispt we don't need here front-end features, **this module is done to provide base functionalities to future modules as product_multi_image, event_multi_image, ....** always in back-end. Of course when we will have such modules as event_multi_image (which depend on this one) we will do website_event_multi_image to show this multi images in front-end.

Note: like this module OCA has more base_\* modules like https://github.com/OCA/partner-contact/tree/8.0/base_contact, .... I think idea is clear.

yajo:
Well after confusions have been cleared, let me split this into 3 modules and PR. Thanks everyone!

yajo:
Thanks everyone! :wink: This is now properly split. Please review: https://github.com/OCA/server-tools/pull/374, https://github.com/OCA/website/pull/171, https://github.com/OCA/product-attribute/pull/135, https://github.com/OCA/event/pull/31, https://github.com/OCA/event/pull/33.

dreispt:
@Yajo Shoudn't there be a PR for OCA/product-attribute also?

yajo:
True, I forgot to say it: https://github.com/OCA/product-attribute/pull/135

PR #118, opened 2015-12-09
[8.0][product_pack] New module to have product packs

yajo:
This fixes https://github.com/ingadhoc/odoo-addons/issues/244. Related to https://github.com/OCA/product-attribute/pull/6.

This is probably a WIP, but I'd thank reviews from original authors.

CC @rafaelbn @ingadhoc @jjscarafia.


rafaelbn:
As commented in https://github.com/OCA/sale-workflow/pull/221#issuecomment-157310333, please @liebana @hugosantosred @ismaelcj could you make a review in this PR technically and functionally please? cc @pedrobaeza

product_pack/README.rst
```diff
@@ -0,0 +1,129 @@
...
+
+Usage
+=====
+
+To create a product pack, you need to:
+
+* Go to *Sales > Products > Products*.
+* Create a product.
+* Name it.
+* Enable the *Pack* check box.
```
rafaelbn, line 74: How must be set "product type"? Consumable? Service? Stockable?

product_pack/README.rst
```diff
@@ -0,0 +1,129 @@
...
+
+Assuming *Pack* is the product pack and *subproducts* are those set to form
+the *pack*, these are your pricing options:
+
+Detailed Component Prices
+-------------------------
+
+This will show 1 line per *subproduct* and 1 line for the *pack*.
+
+All of them will use the price set in the product.
```
rafaelbn: Change "set in the product" by "set in the pack" or "set in the subproducts", It's not clear. By the way in takes the sum of all of them.

product_pack/README.rst
```diff
@@ -0,0 +1,129 @@
...
+
+The *pack* line will cost the sum of the *subproducts*' prices.
+
+Not Detailed Assisted Price
+---------------------------
+
+This will show 1 line for the *pack*.
+
+The *pack* line will cost the sum of the *subproducts*' prices, and will have a
+button for editing its details.
```
rafaelbn: This must be improved. If we don't want detailed info then we have only 2 options:
1. Price of the sum of the subproducts -> "Not detailed totalized price"
2. Price of the pack

**assisted price** is an addtional option in both cases (1) and (2)

rafaelbn:
We must improve help:



IMHO We must change label "Pack?" for "Product pack" and change help "TODO"

rafaelbn:
- Please add a menu "Product pack" in:
  - Sales - Products
  - (if purchases) it would be nice to have also Puchases - Products
  - (if stock) it would be nice to have also Warehouse - Products


- If is a product pack, variants are forbidden


- Please add a filter for pack in **all** product.template views  (this will help for example when adding a pack in a SO)

rafaelbn:
- [BUG] If product_pack installed stock is always 0 if you uninstall you will see real stock


- (cc @pedrobaeza) When creating an invoice from a stock.picking the _order_ from sale order line is not the same in the invoice

rafaelbn:
I think README it's not clear and "Product types name" are not intuitive. Please change like this:
- About Detail, this means what is shown in sale order:
  - Detailed: pack+sub-products
  - Not detailed: only pack
- About Price, this means which price will be taken:
  - Full price: pack+sub-products
  - Pack price: pack only
  - Sub-products price: sub-products

Resume of options:
1. The price is sub-product's price + pack's price
   - Pack and sub-products are shown: "**Detailed full price**"
2. The price is sub-product's price
   - Pack and sub-products are shown: "**Detailed sub-products price**"
   - Only pack is shown: "**Not detailed sub-products price**"
3. The price is pack's price
   - Pack and sub-products are shown: "**Detailed pack price**"
   - Only pack is shown: "**Not detailed pack price**"

Option in all cases: "Assisted": button for editing its details

Thanks

product_pack/README.rst
yajo, line 74: Does not matter.

jjscarafia:
Hi, are you still interested on mergin this module to OCA? I can help with reviews and development.
Thanks!

rafaelbn:
Hi @jjscarafia we are interested in merging this to OCA, of course. After my last comment I talk with @Yajo and he told me that we must change lot of things to make it work as I commented in https://github.com/OCA/product-attribute/pull/118#issuecomment-163555677.

Pleas @Yajo write down here what we must do (also discussed with @pedrobaeza ) and let know to @jjscarafia .

In this moment the functionality of this module is not what it is expected for general purpose.

Thanks

jjscarafia:
Thanks for the response @rafaelbn. @Yajo can you please let me knwo what we should change? If you want contact me by skype (jjscarafia) or hangout.
Thanks!

yajo:
It was some time ago and I don't remember exactly, but I think the main problem comes from having a single field for `product.template.pack_type`. There should be 2 fields as stated in https://github.com/OCA/product-attribute/pull/118#issuecomment-163555677. One to choose how many lines should be displayed in the SO, and another to choose the price calculation method.

jjscarafia:
Thanks @Yajo , I Like the suggestions. I will work on them and comment it here.
Thanks again!

ovnicraft:
@rafaelbn i am migrating to 10.0 https://github.com/ingadhoc/product/pull/88 your comments are welcome !

PR #120, opened 2015-12-17
[8.0][base_custom_info][product_custom_info] New modules to have custom information associated to any model (concretely to products right now)

yajo:
Here go **2 modules** needed to have custom information in products.
1. `base_custom_info` defines an abstract model to have custom attributes in any other model in 1 LOC.
2. `product_custom_info` implements that logic for products.

I decided to merge both in the same PR because the base one does not do anything without an implementation, and the product one does not work without the base one. I hope you understand.

@sergio-incaser @carlos-incaser @rafaelbn.


rafaelbn:
runbot fails

cc @pedrobaeza

pedrobaeza:
Please squash a bit the commits.

yajo:
closing to send base to server-tools

PR #121, opened 2015-12-17
[8.0][product_custom_info] New module to have custom properties in products

yajo:
Continuing from #120, this is only the product part, now that the base part is in https://github.com/OCA/server-tools/pull/313.

It allows to add custom attributes to products.

Obviously, this will fail until https://github.com/OCA/server-tools/pull/313 is merged.

@rafaelbn @pedrobaeza @carlos-incaser.


oca_dependencies.txt
```diff
@@ -0,0 +1 @@
+server-tools
```
pedrobaeza: Put here
```
server-tools https://github.com/Yajo/server-tools custom-info
```
and you'll be able to test this PR while the other PR is without being merged.

product_custom_info/__init__.py
```diff
@@ -0,0 +1,6 @@
+# -*- coding: utf-8 -*-
+# © 2015 Antiun Ingeniería S.L. - Sergio Teruel
+# © 2015 Antiun Ingeniería S.L. - Carlos Dauden
+# License AGPL-3 - See http://www.gnu.org/licenses/agpl-3.0.html
+
+from . import models
```
pedrobaeza: Line at the end

product_custom_info/README.rst
```diff
@@ -0,0 +1,61 @@
...
+    :target: http://www.gnu.org/licenses/agpl-3.0-standalone.html
+    :alt: License: AGPL-3
+
+===================
+Product Custom Info
+===================
+
+This module extends products to allow having custom properties on them.
+
+Usage
```
pedrobaeza, line 23: Add installation section indicating the dependency on OCA/server-tools

product_custom_info/models/product_custom_info.py
```diff
@@ -0,0 +1,11 @@
...
+# © 2015 Antiun Ingeniería S.L. - Sergio Teruel
+# © 2015 Antiun Ingeniería S.L. - Carlos Dauden
+# License AGPL-3 - See http://www.gnu.org/licenses/agpl-3.0.html
+
+from openerp import models
+
+
+class ProductTemplate(models.Model):
+    _name = "product.template"
+    _inherit = ["product.template", "custom.info"]
```
pedrobaeza: You should also inherit properties creating a new table, or this won't work when having multiple models that uses these tables.

Yajo: Not. See https://github.com/OCA/server-tools/pull/313#discussion_r48032780.

Yajo:
Well, that line I added in https://github.com/Yajo/product-attribute/commit/fac5eb89e85d6c721187f5ce9dee9548f2e59091 seems ignored in runbot, so not quite useful... Is that expected?

yajo:
WIP no more, https://github.com/OCA/server-tools/pull/313 is merged.

I rebased and removed fac5eb89e85d6c721187f5ce9dee9548f2e59091. Let's see if it gets :green_heart:

product_custom_info/README.rst
```diff
@@ -0,0 +1,66 @@
...
+===================
+
+This module extends products to allow having custom properties on them.
+
+Installation
+============
+
+* Install `OCA/server-tools <https://github.com/OCA/server-tools>`_.
+
+Usage
```
pedrobaeza, line 23: It lacks section Configuration to indicate where you create the Custom Info Templates.

product_custom_info/README.rst
```diff
@@ -0,0 +1,66 @@
...
+===================
+Product Custom Info
+===================
+
+This module extends products to allow having custom properties on them.
+
+Installation
+============
+
+* Install `OCA/server-tools <https://github.com/OCA/server-tools>`_.
```
pedrobaeza: Indicate the exact module it depends.

yajo:
This one should be ready.

pedrobaeza:
I see that I cannot create by hand templates to test this, because model fields are readonly. The problem should be in base_custom_info.

yajo:
Well, it's the expected behavior. It's readonly to avoid users changing model of template and breaking things. You are expected to manage templates from the product form. Please follow usage instructions and you'll see it works fine.

yajo:
... and in any case, if you consider that a bug, it's a discussion for `base_custom_info` as you pointed.

Issue #101, opened 2015-10-21
product_dimension multi language problem


levkar:
domain filter doesn't work in other languages than English.

https://github.com/OCA/product-attribute/blob/9.0/product_dimension/product.py#L51

maybe a boolean field in product.category and changing domain would be the solution...

I'm very busy to fix, test and make a PR but  something like this might work:

class ProductUomCateg(models.Model):
    _inherit = 'product.uom.categ'
is_lenght = fields.Boolean()

```
    <record id="product.product product_uom_categ_vol" model="product.uom.categ">
        <field name="is_length" eval="True"/>
    </record>

dimensional_uom_id = fields.Many2one(
    'product.uom',
    'Dimensional UoM',
    domain="[('category_id.is_lenght ', '=', True)]",
    help='UoM for length, height, width')
```

PR #124, opened 2016-01-05
[FIX] product_dimension:

hugosantosred:
Now domain on dimensional_uom_id is working also with other languages than english.

Fixes issue #101


pedrobaeza:
:+1: 

Fast-tracking as this is trivial and prevent to work the module.

PR #125, opened 2016-01-07
[ADD] product_dimension_volumetric_weight

hugosantosred:
Adds the volumetric weight to products. This volumetric weight is automatically
calculated when volume or dimension unit for length, width and height is
changed.

This value is calculated using the following formula:

Volume \* UoM Ratio for volumetric weight (kg/m3) = Kg

### DIFF
--- a/product_dimension_volumetric_weight/models/product.py
+++ b/product_dimension_volumetric_weight/models/product.py
@@ -1,0 +1,30 @@
+# -*- coding: utf-8 -*-
+# © 2015 FactorLibre - Hugo Santos <hugo.santos@factorlibre.com>
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+from openerp import models, fields, api
+
+
+class ProductUom(models.Model):
+    _inherit = 'product.uom'
+
+    weight_volumetric_ratio = fields.Float(
+        'Volumetric Weight Ratio (Kg/m3)',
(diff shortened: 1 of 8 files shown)


product_dimension_volumetric_weight/models/product.py
```diff
@@ -0,0 +1,30 @@
...
+    @api.depends('dimensional_uom_id', 'volume')
+    def _get_volumetric_weight(self):
+        for product in self:
+            product.weight_volumetric = (
+                product.volume *
+                product.dimensional_uom_id.weight_volumetric_ratio)
+
+    weight_volumetric = fields.Float(
+        string='Volumetric weight', compute="_get_volumetric_weight",
+        store=True, readonly=True)
```
tremlin, line 30: The `readonly` should not be necessary because the field has no inverse function.

product_dimension_volumetric_weight/models/product.py
```diff
@@ -0,0 +1,30 @@
...
+
+
+class ProductUom(models.Model):
+    _inherit = 'product.uom'
+
+    weight_volumetric_ratio = fields.Float(
+        'Volumetric Weight Ratio (Kg/m3)',
+        help="Used as variable for the volumetric weight calculation "
+        "using the formula. (Volume * Ratio) = Kg",
+        default=100)
```
tremlin, line 14: Is there a reason for this default value? Can't we leave it empty?

PR #131, opened 2016-02-16
[FIX] m2o field use in product_profile

bealdav:
Python change validity is confirmed by data in profile_product_example module

Thanks for review @sebastienbeau @pedrobaeza


hparfr:
:+1:

mourad-ehm:
:+1:

PR #134, opened 2016-02-23
Add module product_webcam

misern2:
Taking the hr_webcam module as a reference, I just created a new module called product_webcam in order to take pictures and store them into product table.


misern2:
I need some help with coveralls check. Where I can get more info ?

pedrobaeza:
Clicking on details you get the code lines that are not tested. As this module is not difficult to test (how do you automatically test a webcam?), it's not mandatory. I personally don't like this one, because it uses a Flash component, but I'm not going to block it.

bealdav:
Thanks for this contribution, but when you say 

> Taking the hr_webcam module as a reference

you mean there is duplicate code ? If yes, there is an opportunity to make a base_ module, don't you think ?

product_webcam/product_webcam_data.xml
```diff
@@ -0,0 +1,12 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<openerp>
+    <data>
+        
+        <record id="action_client_employees_menu" model="ir.actions.client">
+            <field name="name">Open Employees Menu</field>
```
tremlin, line 6: There are some copy&paste artifacts from `hr_webcam`.

A proper base module in `web_webcam`(?) for all the common webcam-related web components would probably be the best solution.

misern2:
You can close this PR if you want. I think this module https://github.com/OCA/web/tree/10.0/web_widget_image_webcam is a better solution.

PR #135, opened 2016-02-24
[8.0][product_multi_image] Multi images for products

yajo:
This will be WIP until https://github.com/OCA/server-tools/pull/374 gets merged.

Implements image gallery for products. Most logic is handled by `base_multi_image`.

@rafaelbn


carlosdauden:
When I install the module product images are "losed"...
They should migrate to the new model?

carlosdauden:
Functional test in local :+1:

product_multi_image/README.rst
```diff
@@ -0,0 +1,70 @@
...
+   :alt: License: AGPL-3
+
+===========================
+Multiple Images in Products
+===========================
+
+This module implements the possibility to have multiple images for a product,
+that it's to say, an image gallery.
+
+Usage
```
pedrobaeza, line 20: Add Installation section talking about the dependency to the base module and in which repo it is.

product_multi_image/hooks.py
```diff
@@ -0,0 +1,9 @@
+# -*- coding: utf-8 -*-
+# © 2016 Antiun Ingeniería S.L. - Jairo Llopis
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+
+from openerp.addons.base_multi_image.hooks import post_init_hook_for_submodules
+
+
+def post_init_hook(cr, registry):
+    post_init_hook_for_submodules(cr, registry, "product.template", "image")
```
pedrobaeza: Shouldn't this be `product.product` instead?

pedrobaeza:
For the icon, maybe you can add a symbol/object to represent the product for distinguishing from the base module.

andhit-r:
@Yajo i could not force rebuild runbot to test

yajo:
All of the above fixed and/or documented. But this will :x: until https://github.com/OCA/server-tools/issues/385 is merged. It's a little patch, go there before please :blush:

yajo:
https://github.com/OCA/server-tools/pull/385 merged. Let's see if you lose no more images, @carlos-incaser. :wink:

atchuthan:
:+1: code reviewed

pedrobaeza:
:+1:

lukebranch:
@Yajo ,

Thanks for your work with this module. I'd like to check what the OCA's policy is on including OCA modules as dependencies in paid themes.

The website_multi_image module seems like it would be a vital component of most themes currently available for Odoo. Many have their own versions of website_multi_image included within the theme itself, however I think @Yajo's module handles this far better, and I believe it will quickly become the de-facto standard (for 8.0/9.0/Master).

I'm not interested in starting another endless licensing debate, i'm just curious to know what the OCA's policy is given the situation i've just described so that I can plan ahead and decide whether to include the multi-image module as a direct dependency or if I should declare it as 'compatible with' Yajo's modules.

My apologies for diverting from the topic of this issue thread. If there is a more appropriate place to ask this question I will remove this comment and ask wherever you suggest.

pedrobaeza:
@lukebranch, indeed this is not the right place for this discussion, but in sort, you can't. If you derive from an AGPL module, you have to be AGPL too.

lukebranch:
@pedrobaeza ,

Thanks for the confirmation.

yajo:
Hi @lukebranch, of course you can as long as your license is [compatible with AGPL](http://www.gnu.org/licenses/license-list.html#GPLCompatibleLicenses). And [you are allowed to ask for money for your themes](http://www.gnu.org/licenses/gpl-faq.html#DoesTheGPLAllowDownloadFee).

lukebranch:
@Yajo ,

Thanks for the confirmation, I look forward to testing the module in production soon and a few themes i'm working on.

By the way, for website_multi_image i've just come across a new GPU-accelerated slider that looks quite promising and is much more actively maintained than OwlCarousel here:

https://github.com/nolimits4web/Swiper

I haven't looked into it in detail yet, however it looks like it could work quite well for this purpose.

PR #137, opened 2016-03-04
[FIX] avoid attrs readonly/invisible for default profile fields

bealdav:
fields view modify fields to make invisible or readonly according to groups.

product fields with matching profile fields with prefix 'profile_default_' should stay visible and editing.

This PR avoid that.

Thanks to evaluate this PR

### DIFF
--- a/product_profile/README.rst
+++ b/product_profile/README.rst
@@ -76,7 +76,7 @@ class ProductProfile(models.Model):
 
    In this case 'categ_id' field (from product.template) is populated
    with 'profile_default_categ_id' value but can be updated manually by the user.
-   Carefull: each time you change profile, the default value is also populated
+   Careful: each time you change profile, the default value is also populated
    whatever the previous value. Custom value is only keep if don't change the profile.
 
 
--- a/product_profile/README.rst
+++ b/product_profile/README.rst
@@ -39,6 +39,7 @@ Configuration
    Example of fields declaration in your own module:
 
 ```python
+
 class ProductProfile(models.Model):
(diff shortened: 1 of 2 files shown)


product_profile/product.py
```diff
@@ -48,7 +52,6 @@ def _get_types(self):
              "(not synchronized with product.template fields)")
     explanation = fields.Text(
         required=True,
-        oldname='description',
```
legalsylvain, line 51: Hi @bealdav,
You should not remove this line, because old version of product_profile 8.0 define description field.
(ref : https://github.com/OCA/product-attribute/commit/4f866da826e717cf952ee6e6da8f31611352fb4c#diff-6564fbea77f6e37ef19b6a92abeb4a40R52)
if you remove oldname attribute you will break upgrade process for other OCA users, using non up to date 'product_profile' module. you can consider remove it only for 9.0 ++ series.

product_profile/README.rst
```diff
@@ -63,15 +65,19 @@ class ProductProfile(models.Model):
...
     ...
     profile_default_categ_id = fields.Many2one(
         'product.category',
         string='Default category')
+
 ```
 
    In this case 'categ_id' field (from product.template) is populated
    with 'profile_default_categ_id' value but can be updated manually by the user.
+   Carefull: each time you change profile, the default value is also populated
```
hparfr: careful

hparfr:
:+1:

chafique-delli:
:+1:

PR #140, opened 2016-03-18
[REF] make behavior homogeneous for product profile

bealdav:
- behavior now is the same between onchange and import data on profile_default_ fields
- refactoring removed 10 lines, and add comments
- tested in a better way with example data and 5 test units
- code coverage up to 90 % for product_profile, +5 % for the whole branch

READY FOR MERGE


product_profile_example/profile.py
```diff
@@ -37,7 +20,7 @@ def _get_types(self):
     purchase_ok = fields.Boolean(
         string='Can be Purchased')
     available_in_pos = fields.Boolean()
-    route_ids = fields.Many2many(
+    profile_default_route_ids = fields.Many2many(
```
lmignon, line 23: By renaming a field, you breaks the API and potentially breaks modules based on this API

product_profile/product.py
```diff
@@ -79,52 +79,51 @@ def _get_profile_fields(self):
                 if field not in fields_to_exclude]
 
     @api.model
-    def _get_profile_data(self, profile_id, filled_fields=None):
+    def _get_vals_from_profile(self, profile_id):
```
lmignon: renaming a method breaks the API and should only occurs when upgrading the addons ( for ex: 8.0 -> 90)

lmignon:
@bealdav Thank you for the changes but I'm a little bit worried to see that these changes break the API and introduce potential incompatibilities if others modules use this API.

product_profile_example/profile.py
bealdav, line 23: Here is just in module example. It's not a module that can be used in production. 50% demo, 50% documentation
No risk here.

product_profile/product.py
bealdav: I propose to maintain the old method which call the new method and write instruction for future refactoring in v9. Then no api changed but future v9 migration becomes clearly defined

What do you think ?

lmignon: Why not.

bealdav:
@lmignon I returned to old api, thanks to notice me. Can you confirm it's ok now ?

bealdav:
@pedrobaeza @lmignon @gurneyalex please, could you put in 'needs review. thanks

lmignon:
@bealdav Thank you for this refactoring. LGTM :+1: (Code review + basic functional tests)

hparfr:
:+1:

product_profile/product.py
```diff
@@ -80,56 +80,48 @@ def _get_profile_fields(self):
...
-                if PROF_DEFAULT_STR == field[:LEN_DEF_STR]:
-                    self[field[LEN_DEF_STR:]] = (
-                        self.profile_id[field])
-                else:
-                    self[field] = self.profile_id[field]
-            except ValueError as e:
-                raise UserError(format_except_message(e, field, self))
-            except Exception as e:
-                raise UserError("%s" % e)
+        if self.profile_id:
```
pedrobaeza, line 145: This is `@api.multi` by default, so you should iterate over self for each record:

```
for record in self:
    if record.profile_id:
...
```

product_profile/product.py
```diff
@@ -80,56 +80,48 @@ def _get_profile_fields(self):
...
-            vals.update(
-                self._get_profile_data(vals['profile_id'], vals.keys()))
-        return super(ProductMixinProfile, self).write(
-            {k: v for k, v in vals.items() if PROF_DEFAULT_STR not in k})
-
-    @api.model
-    def _get_profiles_to_filter(self):
-        """ Inherit if you want that some profiles doesn't have a filter """
-        return [(x.id, x.name) for x in self.env['product.profile'].search([])]
+            vals.update(self._get_profile_data(vals['profile_id']))
```
pedrobaeza, line 162: With this, you're making that the product is updated each time you save it to the latest profile specs. I think this can provoke a lot of side effects and changes not warned to the user.

pedrobaeza:
Why the example is not put as demo data, so we don't need two modules for the same feature?

product_profile/product.py
lmignon, line 145: @pedrobaeza Are you sure? In the Odoo core they never iterate over `self` in a `api.onchange` method.

pedrobaeza, line 145: Yes, and they're limiting the practical application of this over a multi-recordset. They are only thinking this as a UI helper, but this function, making it multi-record aware, can be used in any automation. Not making that, you are forcing user to call the method record by record. The "only" problem is if you have to return a domain or a warning, that is not appropriated in a multi-record environment.

lmignon, line 145: IMO we should keep the same practical application as Odoo to avoid to introduce confusion. The `onchange` method is to notify one model instance that a value for a specific field has changed.

bealdav:
@pedro, I can't use standard demo data because, demo data haven't the same dependencies than product_profile
product_profile depends on prodiuct
product_profile depends on sale, pos, mrp
I don't know any alternative way. And you @pedrobaeza ?

product_profile/product.py
bealdav, line 162: @pedrobaeza Ok you prefer than I update all products wrote with this profile when I update profile model ?

EDIT: this commit fix the case with a test [FIX] ensure updated profile impact products

Thanks to review

product_profile/product.py
bealdav, line 145: summary: 2 ways to do the same thing. My code is like Odoo here. No real example of best practice in guidelines on this matter.
@pedrobaeza take care to not only think to UI. It's exactly the purpose of the method `_get_profile_data()` to think UI and import/external access. Onchange is made shorter for that. 
Here your argument can block this PR which improve module and increase code coverage, isn't it ?
Thanks for shared your view.

sebastienbeau:
:+1: Ok for me, indeed in this case we need a demo/test module

product_profile/product.py
pedrobaeza, line 162: What I prefer is a button that allows to update all products of that profile, but not make it automatically.

product_profile/product.py
pedrobaeza, line 145: OK to let the method as single recordset, but please put at least a `self.ensure_one()` instruction to get the error of the proper place if using this in a multi-recordset.

pedrobaeza:
I don't see why do you need that dependencies in the example module. Are they for the tests? The tests should check the functionality of this module, not other features.

bealdav:
I need these dependencies to have a more realistic example with data in common models, not only product. 
If you read README.rst you can see

> Main use case: a lot of modules are installed (mrp, purchase, sale, pos) and products configuration becomes harder for end users: too many fields to take care of.

Only product is not representative that lead complexity to the use case.

I don't know why it's a problem now and why it was not when you merged it ? You could complain when there is no much documentation/test/demo, but please not there. Thanks

product_profile/product.py
bealdav, line 162: As you said:

> With this, you're making that the product is updated each time you save it to the latest profile specs. I think this can provoke a lot of side effects

You're right, but if I put a button and somebody forget to click the button, then you have side effects: with the same profile applied you get records with different settings on fields which must be control by profile.

Instead of button, I propose to have a red text on product.profile form which explain the impact of the update of the profile. What do you think ?

product_profile/product.py
bealdav, line 145: OK for that

product_profile/product.py
bealdav, line 162: Hi @pedrobaeza , what do you think about this message to end user ?

prof_warning

pedrobaeza:
OK then, merging although the difference about the dependency question. Previously, I didn't see that question about the dependencies (take into account that I review a lot of PRs, and sometimes some things pass unnoticed for me).

bealdav:
thanks @pedrobaeza

PR #141, opened 2016-03-18
[8.0][fix] onchange method was incompletely migrated and produced a stack trace

Olivier-LAURENT:
- When changing the value of `base_ext` to `-3` (i.e. fixed price), a stack trace is produced.
- when changing this field to another value, the `base` field was not updated leaving visible and required (but most of cases irrelvant) the field related to another pricelist


Olivier-LAURENT:
wrong upstream branch sorry

PR #142, opened 2016-03-18
[8.0][fix] onchange method was incompletely migrated and produced a stack trace

Olivier-LAURENT:
- when changing the value of `base_ext` to `-3` (i.e. fixed price), a stack trace is produced.
- when changing this field to another value, the `base` field was not updated leaving visible and required (but most of cases irrelvant) the field related to another pricelist

### DIFF
--- a/product_pricelist_fixed_price/tests/test_product_pricelist_fixed_price.py
+++ b/product_pricelist_fixed_price/tests/test_product_pricelist_fixed_price.py
@@ -1,0 +1,64 @@
+# -*- coding: utf-8 -*-
+# © 2016  Olivier Laurent, Acsone SA/NV (http://www.acsone.eu)
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+
+import openerp.tests.common as common
+
+
+class TestProductPricelistFixedPrice(common.TransactionCase):
+
+    def test_onchange_base_ext(self):
+        """ Check for the value of 'base' field when changing the value of
+            'base_ext' field:
+            * for sale price list
+            * for purchase price list
+        """
+        item_obj = self.env['product.pricelist.item']
+        plv = self.env.ref('product.ver0')
+        base_sale = item_obj._get_default_base({'type': 'sale'})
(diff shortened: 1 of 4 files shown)


product_pricelist_fixed_price/tests/test_product_pricelist_fixed_price.py
```diff
@@ -0,0 +1,64 @@
...
+        vals['base_ext'] = -3
+
+        # check again, it must be the same as default 'base' value
+        item = item_obj.new(values=vals)
+        item.change_base_ext()
+        self.assertNotEqual(item.base, item.base_ext)
+        self.assertEqual(item.base, base_pur)
+        self.assertEqual(item.price_discount, -1.0)
+
+        pass
```
pedrobaeza, line 64: Remove this

pedrobaeza:
Minor remark, but :+1:

rven:
I was running against the same problem, but noticed it was already been fixed by this pull request.

:+1: :+1: :+1:

pedrobaeza:
OK, I merge removing the `pass` in the merge operation.

rven:
The following part is not working as it should be:

```
    @api.onchange('base_ext')
    def change_base_ext(self):
        base = self.base_ext
        if base == -3:
            base = self._get_default_base(
                {'type': self.price_version_id.pricelist_id.type})
            self.price_discount = -1
        self.base = base
```

The problem is that the following is not working in an onchange event:
`self.price_version_id.pricelist_id.type`

Odoo returns NewId objects in the onchange methods and handles only one level of many2one.

In this case `self.price_version_id.pricelist_id.type` is empty and the function `_get_default_base` returns -1 instead of 1.
This way the field base_pricelist_id is shown and required.

Is there a solution for the NewId problem in Odoo?

pedrobaeza:
The statement `self.price_version_id.pricelist_id.type` works perfectly even on new records, so the problem should be other.

rven:
The problem does not occur anymore, I don't know the reason of the earlier failure.

PR #144, opened 2016-03-31
8.0 fixed pricelist fixed

NL66278:
Restore full functionality of fixed price - pricelist on 8.0 branch.


product_pricelist_fixed_price/__openerp__.py
```diff
@@ -1,42 +1,20 @@
...
     "website": "www.serviciosbaeza.com",
     "license": "AGPL-3",
     "depends": [
         "product",
     ],
-    "demo": [],
     "data": [
         'view/product_pricelist_item_view.xml',
     ],
-    "init_hook": "init_hook",
```
Olivier-LAURENT, line 40: It was however the right way to upgrade (only once) the database but like this:
`"post_init_hook": "post_init_hook",`

product_pricelist_fixed_price/model/product_pricelist_item.py
```diff
@@ -1,50 +1,124 @@
...
+        Values for price_discount and price_round will not be checked,
+        because create and write will automagically set appropiate
+        values.
+        """
+        if self.base_ext == FIXED_PRICE_TYPE and not self.product_id:
+            raise ValidationError(
+                _("Product required for fixed price item.")
+            )
+
+    def _auto_end(self, cr, context=None):
```
Olivier-LAURENT: If the `post_init_hook()` is really executed (see comment above), this method is (normally) useless

product_pricelist_fixed_price/init_hook.py
```diff
@@ -1,24 +0,0 @@
...
-#    MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
-#    GNU Affero General Public License for more details.
-#
-#    You should have received a copy of the GNU Affero General Public License
-#    along with this program.  If not, see <http://www.gnu.org/licenses/>.
-#
-##############################################################################
-
-
-def post_init_hook(cr):
```
Olivier-LAURENT, line 23: If recovered, signature must be:
`post_init_hook(cr, registry)`

product_pricelist_fixed_price/model/product_pricelist_item.py
```diff
@@ -1,50 +1,124 @@
...
+        if self.base_ext == FIXED_PRICE_TYPE:
+            self.base = self._get_fixed_price_base()
             self.price_discount = -1
-        self.base = base
+            self.price_round = 0.0
+            self.price_min_margin = 0.0
+            self.price_max_margin = 0.0
+        else:
+            self.base = self.base_ext
+            if self.price_discount == -1.0:
```
Olivier-LAURENT, line 110: Would it not be better to let the user to decide in this case ? Maybe his change does not come from the value `-3` keeping the formula valid without forcing `discount=0`... Maybe his change is coming from the value `-3` making the formula highly suspect by forcing `discount=0` if the user does not touch also initial `surcharge`

product_pricelist_fixed_price/model/product_pricelist_item.py
```diff
@@ -1,50 +1,124 @@
...
+        base = self.env['product.price.type'].search([], limit=1, order='id')
+        assert base, _("No record found in product.price.type")
+        return base[0].id
+
+    @api.model
+    def _modify_vals(self, vals):
+        """Ensure consistent values for fixed pricelist items.
+
+        The passed vals parameter is used for both input and output.
+        base should be default if base-ext = -1, in all other cases base and
```
Olivier-LAURENT: -3 ?

product_pricelist_fixed_price/__openerp__.py
NL66278, line 40: I decided to re-use the method (overriding _auto_end) that was also used in the 7.0 branch. Also because it contained a guard against unneeded or even unwanted changes: only changing base_ext if not already set.

product_pricelist_fixed_price/model/product_pricelist_item.py
NL66278: You are right, will change this.

product_pricelist_fixed_price/model/product_pricelist_item.py
NL66278, line 110: My reasoning was that the change that the user would accidentally leave the -1 discount, was much higher than the user actually wanting a -1 discount, when basing a price on another pricelist (effectively ignoring the price from that other pricelist).

product_pricelist_fixed_price/model/product_pricelist_item.py
```diff
@@ -1,50 +1,124 @@
...
+    @api.constrains('base_ext')
+    def _check_fixed_price(self):
+        """Ensure fixed prices always refer to a specific product.
+
+        Values for price_discount and price_round will not be checked,
+        because create and write will automagically set appropiate
+        values.
+        """
+        if self.base_ext == FIXED_PRICE_TYPE and not self.product_id:
+            raise ValidationError(
```
Olivier-LAURENT, line 41: An import is missing:
`global name 'ValidationError' is not defined`

product_pricelist_fixed_price/README.rst
```diff
@@ -2,7 +2,6 @@
    :target: http://www.gnu.org/licenses/agpl-3.0-standalone.html
    :alt: License: AGPL-3
 
-=============================
```
pedrobaeza, line 5: Why remove this line?

product_pricelist_fixed_price/__openerp__.py
pedrobaeza, line 40: That guard can be put on the post_init_hook, but the current 8.0 and later way is with these hooks.

product_pricelist_fixed_price/model/product_pricelist_item.py
pedrobaeza: I agree with @Olivier-LAURENT. Make this on post_init_hook

product_pricelist_fixed_price/__openerp__.py
NL66278, line 40: @pedrobaeza @Olivier-LAURENT I put all the logic to set initial value of base_ext on post_init_hook now.

product_pricelist_fixed_price/model/product_pricelist_item.py
NL66278, line 41: @Olivier-LAURENT It is imported now.

product_pricelist_fixed_price/__openerp__.py
```diff
@@ -1,42 +1,21 @@
...
 {
     "name": "Fixed price in pricelists",
-    "version": "8.0.1.0",
-    "author":  "Serv. Tecnol. Avanzados - Pedro M. Baeza,"
-               "Therp B.V.,"
-               "Odoo Community Association (OCA),"
-               "Giovanni Francesco Capalbo",
+    "version": "8.0.1.0.1",
+    "author": "Serv. Tecnol. Avanzados - Pedro M. Baeza,"
+              "Odoo Community Association (OCA)",
```
pedrobaeza, line 10: Put Therp here

pedrobaeza:
Little comment, but :+1:

product_pricelist_fixed_price/__openerp__.py
NL66278, line 10: Done

product_pricelist_fixed_price/__openerp__.py
```diff
@@ -1,42 +1,22 @@
...
 {
     "name": "Fixed price in pricelists",
-    "version": "8.0.1.0",
-    "author":  "Serv. Tecnol. Avanzados - Pedro M. Baeza,"
-               "Therp B.V.,"
-               "Odoo Community Association (OCA),"
-               "Giovanni Francesco Capalbo",
+    "version": "8.0.1.0.1",
+    "author": "Serv. Tecnol. Avanzados - Pedro M. Baeza,"
+              "Therp BV",
```
pedrobaeza: Remove the `,` at the end of this line (you have to put it inside the quotes).

Olivier-LAURENT:
Hi @NL66278,
The test I proposed in the previous PR has no longer sense now with this one.
You have to delete it or modify it.
Then :+1: and thx for this work

pedrobaeza:
@NL66278, @Olivier-LAURENT, I didn't see the Travis error. Can you please propose a PR with the fix?

PR #146, opened 2016-04-10
[FIX] allow to use extra attribute price with pricelist_per_product

bealdav:
Allow to use extra attribute price with pricelist_per_product module


pedrobaeza:
:+1:

bealdav:
travis test: Module product_pricelist_fixed_price: 1 failures, 0 errors
but test on pricelist_per_product is ok everywhere

PR #150, opened 2016-04-18
[ADD] module pricelist_item_generator

bealdav:
This module allows to create pricelist items in a bulk way, by separate price elements and products informations. It avoids a lot of manual entries.

 REQUEST REVIEW

Here is entry of data
In this object, we separate price / quantity from products / product category, to be able to reuse price quantity definition for several products or category.
one
# 

---

Here is impact on pricelist version
We now have 2 one2many, but both are on the same object (items), 'auto' field allow to separate them. Then native behavior is kept in the second o2m. The first one is readonly.
two
# 

---

Generator Configuration

To go on to simplify entry of data, we can copy the one2many if these boolean are checked.
three


bealdav:
tests fail comes from 'product_pricelist_fixed_price' module. We should study the interaction of these modules. Unit tests on my module are ok

pedrobaeza:
The problem is also on the main branch, so it should be fixed there.

bealdav:
any comment on this PR, @pedrobaeza ? thanks

pedrobaeza:
Sorry, I can't review this one now.

bealdav:
As defined by runbot, all unit tests in my module are OK, can I have a review, thanks.

bealdav:
After a first wave of review, i'll add images to PR

bealdav:
unit test coverage of this module has been raised to 93 %.

pricelist_item_generator/models/pricelist.py
```diff
@@ -0,0 +1,415 @@
...
+from openerp import fields, models, api, _
+from openerp.exceptions import Warning as UserError
+import openerp.addons.decimal_precision as dp
+from collections import defaultdict
+
+
+TODO_SELECTION = [('update', 'Update'), ('create', 'Create')]
+
+
+def update_item_tpl_or_condition(vals, self):
```
lasley: IMO it would make more sense if `self` is the first arg in order to match standard Python instance methods.

Also the `api.multi` decorator should be added to this method in order to be more compatible.

pricelist_item_generator/models/pricelist.py
```diff
@@ -0,0 +1,415 @@
...
+    _description = "Pricelist Items Generator"
+
+    sequence = fields.Integer(
+        string='Sequence', default=5,
+        help="Used to generate the sequence of the price item")
+    name = fields.Char(
+        string='Name', required=True,
+        help="Copied towards pricelist price item")
+    active = fields.Boolean(
+        string='Active', default=False, copy=False,
```
lasley: Booleans already default to False, no need to define it.

pricelist_item_generator/models/pricelist.py
```diff
@@ -0,0 +1,415 @@
...
+                tmpl_vals = gen._get_vals_from_objs([tmpl])
+                for vals in vals_conditions:
+                    values = gen._synchronize_items(tmpl_vals[0], vals)
+                    prd_list_item_m.create(values)
+            self._unset_todo_create(tmpl_objs, condition_objs)
+            # Then we update all values
+            price_items = prd_list_item_m.search(
+                [('price_generator_id', '=', gen.id)])
+            map_items = {'%s-%s' % (
+                         str(x.product_condition_id.id),
```
lasley: No need to explicitly convert to string here- Python interpolation is already taking care of the typing w/ your use of `%s`

pricelist_item_generator/models/pricelist.py
```diff
@@ -0,0 +1,415 @@
...
+                for condition in condition_vals:
+                    values = gen._synchronize_items(tmpl, condition)
+                    p_list_item_id = map_items['%s-%s' % (
+                        values.get('product_condition_id'),
+                        values.get('item_template_id'))]
+                    prd_list_item_m.browse(p_list_item_id).write(values)
+            # unset todo
+            gen.product_condition_ids.write({'todo': False})
+            gen.item_template_ids.write({'todo': False})
+            gen.write({'to_update': False})
```
lasley: There's a lot of consecutive writes here. Maybe something like:

```
gen.write({
    'to_update': False,
    'product_condition_ids': [
        (1, r.id, {'todo': False}) for r in gen.product_condition_ids
    ],
    'item_template_ids': [
        (1, r.id, {'todo': False}) for r in gen.item_template_ids
    ],
})
```

pricelist_item_generator/models/pricelist.py
```diff
@@ -0,0 +1,415 @@
...
+                         for x in price_items}
+            tmpl_vals = gen._get_vals_from_objs(gen.item_template_ids)
+            condition_vals = gen._get_vals_from_objs(gen.product_condition_ids)
+            for tmpl in tmpl_vals:
+                for condition in condition_vals:
+                    values = gen._synchronize_items(tmpl, condition)
+                    p_list_item_id = map_items['%s-%s' % (
+                        values.get('product_condition_id'),
+                        values.get('item_template_id'))]
+                    prd_list_item_m.browse(p_list_item_id).write(values)
```
lasley: It would be nice if this model had a M2M relation w/ `product.pricelist.item`. You could then utilize that col and the Odoo ORM to group all of these writes in with the one below, which would be significantly more efficient.

pricelist_item_generator/models/pricelist.py
```diff
@@ -0,0 +1,415 @@
...
+                    [('price_generator_id', '=', generator.id)])
+                if items_tpl:
+                    items2unlink = self.env['product.pricelist.item'].search([
+                        ('item_template_id', 'in', [x.id for x in items_tpl])])
+                    if items2unlink:
+                        items2unlink.unlink()
+                # 'todo' field to initial value
+                generator.item_template_ids.write({'todo': 'create'})
+                generator.product_condition_ids.write({'todo': 'create'})
+            generator.write(vals)
```
lasley: Please group these writes into one using the Odoo ORM and the trick mentioned above. You would just want to make an array of `(1, record.id, {'values': 'dict'})` as you loop instead of directly writing.

pricelist_item_generator/models/pricelist.py
```diff
@@ -0,0 +1,415 @@
...
+        # Some records shouldn't be written with the same values
+        todo_update = update_item_tpl_or_condition(vals, self)
+        super(PricelistProductCondition, self).write(vals)
+        if todo_update:
+            super(PricelistProductCondition, todo_update).write(
+                {'todo': 'update'})
+        return True
+
+
+class AbstractPriceListItemGenerator(models.AbstractModel):
```
lasley: Class names should be the UpperCamelCaseForm of the model name, so in this instance it should be `AbstractPricelistItemGenerator`

pricelist_item_generator/models/pricelist.py
```diff
@@ -0,0 +1,415 @@
...
+    _name = "abstract.pricelist.item.generator"
+    _description = "Abstract Pricelist Items Generator"
+
+    price_generator_id = fields.Many2one(
+        comodel_name='pricelist.item.generator',
+        string='Generator', readonly=True)
+    min_quantity = fields.Integer(default=1)
+
+
+class PricelistItemTemplateBase(models.Model):
```
lasley: Class names should be the UpperCamelCaseForm of the model name, so in this instance it should be `PricelistItemTemplate`

pricelist_item_generator/models/pricelist.py
```diff
@@ -0,0 +1,415 @@
...
+        help='Specify the minimum amount of margin over the base price.')
+    price_max_margin = fields.Float(
+        string='Max. Price Margin',
+        digits_compute=dp.get_precision('Product Price'),
+        help='Specify the maximum amount of margin over the base price.')
+    todo = fields.Selection(
+        selection=TODO_SELECTION, readonly=True, string='Next Action',
+        help="Required action on Pricelist Items")
+
+    @api.one
```
lasley: Please don't use `api.one` in new code - it is deprecated.

pricelist_item_generator/models/pricelist.py
```diff
@@ -0,0 +1,415 @@
...
+    @api.constrains('price_min_margin', 'price_max_margin')
+    def _check_margin(self):
+        if self.price_max_margin and self.price_min_margin and (
+                self.price_min_margin > self.price_max_margin):
+            raise UserError(_("Error! The minimum margin should be lower "
+                              "than the maximum margin."))
+
+    @api.one
```
lasley: Please don't use `api.one` in new code - it is deprecated.

pricelist_item_generator/models/pricelist.py
```diff
@@ -0,0 +1,415 @@
...
+    @api.constrains('price_discount', 'price_surcharge')
+    def _check_price_elements(self):
+        # check price validity
+        if self.price_discount == 0 and self.price_surcharge == 0:
+            raise UserError(
+                _("'Discount' or 'Surcharge' must be different of 0."))
```
lasley: s/of/from

pricelist_item_generator/models/pricelist.py
```diff
@@ -0,0 +1,415 @@
...
+        result = []
+        result.extend([(1, _('Public Price')), (-1, _('Other Pricelist'))])
+        return result
+
+
+class PricelistItemTemplate(models.Model):
+    """ This model is called by a second class
+    """
+    _inherit = ['abstract.pricelist.item.generator', 'pricelist.item.template']
+    _name = 'pricelist.item.template'
```
lasley: Please add a `_description` to all new models

pricelist_item_generator/models/pricelist.py
```diff
@@ -0,0 +1,415 @@
...
+
+    _order = 'sequence ASC, id ASC'
+
+    price_generator_id = fields.Many2one(
+        comodel_name='pricelist.item.generator', required=True)
+
+    @api.model
+    def create(self, vals):
+        self.env['pricelist.item.generator'].search(
```
lasley: You can use browse here instead of search - `...].browse(vals['price_generator_id'])...`

pricelist_item_generator/models/pricelist.py
```diff
@@ -0,0 +1,415 @@
...
+        super(PricelistItemTemplate, self).write(vals)
+        if todo_update:
+            super(PricelistItemTemplate, todo_update).write(
+                {'todo': 'update'})
+        return True
+
+
+class ProductPricelistItem(models.Model):
+    _inherit = ['abstract.pricelist.item.generator', 'product.pricelist.item']
+    _name = 'product.pricelist.item'
```
lasley, line 10: Please add a `_description` to all new models.

pricelist_item_generator/models/pricelist.py
```diff
@@ -0,0 +1,415 @@
...
+
+    auto = fields.Boolean(
+        string='Auto', default=False,
```
lasley: False is already the default value for Boolean, no need to redeclare

lasley:
Thanks for the submission @bealdav - useful module here. I left some comments inline.

Also, please break your models & views up one per file - naming them the camel_case_form of the model name (with _view suffix for views). This is per the [OCA guidelines](https://github.com/OCA/maintainer-tools/blob/master/CONTRIBUTING.md), and really helps things from a maintainability standpoint when dealing with large sets of projects.

pricelist_item_generator/models/pricelist.py
bealdav: This function is not inside a class, then that intentionnaly that I put self arg  in secondth position to avoid confusion with a method. api multi is for method in class.

pricelist_item_generator/models/pricelist.py
bealdav: it's not a new model, there is `_inherit`, description is in original model declaration

pricelist_item_generator/models/pricelist.py
bealdav, line 10: it's not a new model, there is `_inherit`

pricelist_item_generator/models/pricelist.py
bealdav: Not sure it's deprecated https://github.com/odoo/odoo/issues/12408#issuecomment-226421501

Documentation here contains api.one https://www.odoo.com/documentation/9.0/reference/orm.html#openerp.api.constrains

pricelist_item_generator/models/pricelist.py
bealdav: I have 2 class on 'pricelist.item.template': PricelistItemTemplateBase and PricelistItemTemplate.
Python prevent 2 class with the same name.

pricelist_item_generator/models/pricelist.py
bealdav: Done

bealdav: Done

pricelist_item_generator/models/pricelist.py
bealdav: Not deprecated for constraint, see comment below

bealdav:
@lasley thanks a lot for your complete review. I hope you'll find answers to your questions

lasley:
Oops sorry I missed this! I won't block for the api.one, that convo is pretty deep and this is 8.0 code anyways. Thanks again for the awesome module 👍

pricelist_item_generator/README.rst
```diff
@@ -0,0 +1,73 @@
...
+
+This module allows to create pricelist items in a bulk way, by separate price elements and products informations. It avoids a lot of manual entries.
+
+
+Configuration
+=============
+
+To configure this module (for non admin user), you need to:
+
+#. Go to Settings > Configuration > Sales > Quotation and Sales > Customer Features: check 'Use pricelists'
```
pedrobaeza: Split this list element in 2 at "check..." level

pricelist_item_generator/README.rst
```diff
@@ -0,0 +1,73 @@
...
+
+Usage
+=====
+
+To use this module, you need to:
+
+#. Go to Sales > Configuration > Pricelists > Price Items Generator
```
pedrobaeza: Explain the possibilities you have in this menu option.

pricelist_item_generator/README.rst
```diff
@@ -0,0 +1,73 @@
...
+.. image:: https://odoo-community.org/website/image/ir.attachment/5784_f2813bd/datas
+   :alt: Try me on Runbot
+   :target: https://runbot.odoo-community.org/runbot/135/8.0
+
+
+Known issues / Roadmap
+======================
+
+* Only support sales pricelist: others possible
+* For V9 version we could add dates to item templates to avoid manual entries
```
pedrobaeza: I don't understand this point

pricelist_item_generator/__openerp__.py
```diff
@@ -0,0 +1,29 @@
+# coding: utf-8
+# © 2016 David BEAL @ Akretion
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+
+{
+    'name': 'Pricelist Item Generator',
+    'version': '8.0.0.0.1',
```
pedrobaeza: 8.0.1.0.0

pricelist_item_generator/__openerp__.py
```diff
@@ -0,0 +1,29 @@
...
+    'category': '',
```
pedrobaeza: Put same category as module product

pricelist_item_generator/__openerp__.py
```diff
@@ -0,0 +1,29 @@
...
+    'summary': "Create/Update Pricelist Items in a massive way",
+    'author': 'Akretion,Odoo Community Association (OCA)',
+    'website': 'http://www.akretion.com',
+    'license': 'AGPL-3',
+    'depends': [
+        'sale',
+        ],
```
pedrobaeza: Align the closing bracket the same as the others

pricelist_item_generator/__openerp__.py
```diff
@@ -0,0 +1,29 @@
...
+        'views/setting_view.xml',
+        'security/ir.model.access.csv',
+    ],
+    'demo': [
+        'demo/pricelist.item.generator.csv',
+        'demo/pricelist.product.condition.csv',
+        'demo/pricelist.item.template.csv',
+    ],
+    'installable': True,
+    'images': [
```
pedrobaeza: Remove empty keys

pricelist_item_generator/i18n/pricelist_item_generator.pot
```diff
@@ -0,0 +1,518 @@
+# Translation of Odoo Server.
```
pedrobaeza: Remove pot file. It's useless and confuses users.

pricelist_item_generator/models/pricelist.py
```diff
@@ -0,0 +1,426 @@
...
+        default = dict(default or {})
+        default['name'] = '%s (Copy)' % self.name
+        # Conditional One2many copy
+        if self.copy_item_template:
+            default['item_template_ids'] = self._get_o2m_values(
+                self.item_template_ids, 'item_template_id')
+        if self.copy_product_condition:
+            default['product_condition_ids'] = self._get_o2m_values(
+                self.product_condition_ids, 'product_condition_id')
+        return super(PricelistItemGenerator, self).copy(default)
```
pedrobaeza: It's a good practice to continue keeping keyword arguments as is, so please put `...copy(default=default)`

pricelist_item_generator/models/pricelist.py
```diff
@@ -0,0 +1,426 @@
...
+            tmpl_objs = defaultdict(list)
+            condition_objs = {}
+            # Firstly we are creating missing records
+            for item in gen.item_template_ids:
+                if item.todo == 'create':
+                    tmpl_objs[item] = [x for x in gen.product_condition_ids]
+            for condition in gen.product_condition_ids:
+                if condition.todo == 'create':
+                    condition_objs[condition] = [
+                        x for x in gen.item_template_ids
```
pedrobaeza: Use here `filtered` and you won't lose the recordset concept:

```
condition_objs[condition] = gen.item_template_ids.filtered(
    lambda x: x.todo != 'create')
```

pricelist_item_generator/README.rst
```diff
@@ -0,0 +1,73 @@
...
+
+Credits
+=======
+
+Images
+------
+
+* Icon images comes from http://icons8.com/
+
+* Odoo Community Association: `Icon <https://github.com/OCA/maintainer-tools/blob/master/template/module/static/description/icon.svg>`_.
```
pedrobaeza: You can remove this line as you're not using the general icon.

pricelist_item_generator/README.rst
```diff
@@ -0,0 +1,73 @@
...
+check there if your issue has already been reported. If you spotted it first,
+help us smashing it by providing a detailed and welcomed feedback.
+
+Credits
+=======
+
+Images
+------
+
+* Icon images comes from http://icons8.com/
```
pedrobaeza: Can you specify the exact URL?

bealdav:
And yes I have to split files, I prepare it: pricelist.py, generator.py and abstract.py

bealdav:
Thanks @lasley and @pedrobaeza for your complete review.

EBII:
bien

PR #151, opened 2016-04-26
[IMP] product_multi_image: Handle images in variants

pedrobaeza:
With this PR, you can:
- Add images only in product templates, so no need of duplicating bytes.
- Select in which variants the images are available.
- Handle from the variants the addition or removal of images.


pedrobaeza:
This one is ready to be reviewed.

cc @carlosdauden @sergio-teruel @Yajo @rafaelbn

pedrobaeza:
The error in Travis is unrelated (due to product_pricelist_fixed_price).

carlosdauden:
:+1:

sergio-teruel:
:+1: Tested and ok.

product_multi_image/models/image.py
```diff
@@ -0,0 +1,22 @@
...
+
+
+class Image(models.Model):
+    _inherit = "base_multi_image.image"
+
+    product_variant_ids = fields.Many2many(
+        comodel_name="product.product", string="Visible in these variants",
+        help="If you leave it empty, all variants will show this image. "
+             "Selecting one or several of the available variants, you "
+             "restrict the availability of the image to that variants.")
```
yajo: "those variants"

product_multi_image/models/product_product.py
```diff
@@ -0,0 +1,116 @@
...
+            # Remember the list of images that were before changes
+            previous_images = product.product_tmpl_id.image_ids.filtered(
+                lambda x: (not x.product_variant_ids or
+                           product.id in x.product_variant_ids.ids))
+            for image in product.image_ids:
+                if isinstance(image.id, models.NewId):
+                    # Image added
+                    image.owner_id = product.product_tmpl_id.id
+                    image.owner_model = "product.template"
+                    image.product_variant_ids = [(6, 0, product.ids)]
```
yajo, line 70: I'd prefer new api here, it's easier to understand. Just `image.product_variant_ids = product` should do the trick.

pedrobaeza, line 70: This is not a question of new/old API, it's the syntax to get the records linked (instead of created, duplicated or something similar). Other ambiguous syntax like the one you say can be interpreted different through the evolution of ORM.

product_multi_image/models/product_product.py
```diff
@@ -0,0 +1,116 @@
...
+                 'product_tmpl_id.image_ids.product_variant_ids')
+    def _get_multi_image(self):
+        """Needed for changing dependencies in this class."""
+        super(ProductProduct, self)._get_multi_image()
+
+
+class ProductProductOld(orm.Model):
+    """It is needed to use v7 api here because core model fields use the
+    ``multi`` attribute, that has no equivalent in v8, and it needs to be
+    disabled or bad things will happen.
```
yajo: I think that a link to https://github.com/odoo/odoo/issues/10799 would fit here.

yajo:
Tested in runbot, I have an issue.
1. Set an image to a product template.
2. create the variants.
3. Image gets lost.

Besides, runbot page displays some errors that seem related:

```
2016-04-27 01:02:29     ERROR   server  openerp.sql_db:241 execute

bad query:  SELECT "base_multi_image_image"."id" FROM "base_multi_image_image"
                        WHERE "base_multi_image_image".id IN ('one2many_v_id_169')  ORDER BY "base_multi_image_image"."sequence" ,"base_multi_image_image"."owner_model" ,"base_multi_image_image"."owner_id" ,"base_multi_image_image"."id"  

  File "/srv/openerp/instances/openerp-oca-runbot/parts/odoo-extra/runbot/static/build/3146184-151-5bd61f/openerp/sql_db.py", line 234, in execute
    res = self._obj.execute(query, params)
DataError: invalid input syntax for integer: "one2many_v_id_169"
LINE 2: ...            WHERE "base_multi_image_image".id IN ('one2many_...
                                                             ^
```

Code seems good except for some little remarks.

product_multi_image/models/product_product.py
yajo, line 70: In new api AFAIK, you do not need all that. If it's a NewId instance it gets created, otherwise it gets linked. All is automatic.

pedrobaeza:
I'll check your use case. Probably there will be an assignation image=False in the variant creation. We should inhibit the inverse function in that case.

For the error in the log, it seems a bug in the ORM that I have already seen commented by others (I think @hbrunn), so for now I don't think it's important, because it doesn't prevent the normal behaviour, although it's a bit ugly.

yajo:
In such case an entry to known issues would be enough.

product_multi_image/models/product_product.py
pedrobaeza, line 70: But this is a special case in a calculated one2many, so this is not covered by the scope of the new API (indeed, Raphael Collet told me directly that there shouldn't be computed one2many), only many2many for on the fly created records, but linking with the proper syntax shouldn't be a problem. I had to it this way for not losing the kanban view, because many2many fields don't have it.

Anyway, I can try, but my fear is that in v9/v10, we might need to rewrite the expression.

yajo, line 70: This new api keeps getting worse :confused:, leave it as it is if you fear that then, thanks for the info.

pedrobaeza:
@Yajo, I have fixed the user case that was incorrect, and include 2 new tests for testing this case. Please review before you leave tomorrow.

yajo:
I have this problem now (I don't know if it is expected, but for what I read in the README seems like it's a bug):
- Create a product.template.
- Set its main image through the classic mechanism.
- Create 2 variants.
- Go to first variant.
- Change its main image through the classic mechanism.
- The new image is applied to all variants and the template.

pedrobaeza:
For now, that's the expected behavior:
- The variant main image is assigned to the first multi-image found.
- As this image is the first image of the template, then the template also switches this image
- As this image is available for all variants, the main image of the rest of the variants is switched to this new one.

There are use cases that recommends one strategy or the other... What do you think?

yajo:
I think it's good to keep the behavior as close to core as possible. AFAIK core Odoo sets the template image into the variant unless the variant has its own. That should be the same with this module.

However I understand the complexity of this given the current implementation, so if you prefer to add a notice in the known issues and leave this for the future, it's OK for me.

rafaelbn:
:+1:

pedrobaeza:
cc @Tecnativa

atchuthan:
:+1:

atchuthan:
@pedrobaeza For OCA/e-commerce runbot(**version 8.0**), seems to be failing due to product_multi_image.

Please check version 8.0 of any session in runbot link below:
https://runbot.odoo-community.org/runbot/repo/github-com-oca-e-commerce-113

Not sure why this module is affected in OCA/e-commerce.

I noticed this problem when I created a PR(https://github.com/OCA/e-commerce/pull/115) at OCA/e-commerce where runbot failed.  More info on log: https://runbot.odoo-community.org/runbot/build/3150078

pedrobaeza:
Solved in https://github.com/OCA/product-attribute/commit/c26675cf89146e00c7a707a1d41912ea105efc46

Launch rebuilds of the corresponding builds.

atchuthan:
@pedrobaeza thanks for the quick fix.

PR #155, opened 2016-05-12
[NEW] Add new module: product_attribute_priority

mileo:
captura de tela de 2016-05-12 16 35 19
captura de tela de 2016-05-12 16 35 31
## 

captura de tela de 2016-05-12 16 35 50
captura de tela de 2016-05-12 16 35 59


moylop260:
Could you check the following conventions?

``` bash
************* Module product_attribute_priority.__openerp__
product_attribute_priority/__openerp__.py:1: [C8201(no-utf8-coding-comment), ] No UTF-8 coding comment found: Use `# coding: utf-8` or `# -*- coding: utf-8 -*-`
product_attribute_priority/__openerp__.py:19: [C8102(manifest-required-key), ] Missing required key "license" in manifest file
************* Module product_attribute_priority.models.product
product_attribute_priority/models/product.py:1: [C8201(no-utf8-coding-comment), ] No UTF-8 coding comment found: Use `# coding: utf-8` or `# -*- coding: utf-8 -*-`
```

More info here:
https://travis-ci.org/OCA/product-attribute/jobs/129856262#L419

moylop260:
👍

product_attribute_priority/models/product.py
```diff
@@ -0,0 +1,58 @@
...
+from openerp import models, fields
+
+
+class ProductAttribute(models.Model):
+    _inherit = 'product.attribute'
+    _order = 'priority, name'
+
+    priority = fields.Integer(
+        string='Priority',
+        help='Define priority of exhibition in reports',
```
guewen: `help` needs an update

product_attribute_priority/models/product.py
```diff
@@ -0,0 +1,58 @@
...
+
+    _inherit = 'product.attribute.value'
+    _order = 'priority, name'
+
+    priority = fields.Integer(
+        string='Priority',
+        readonly=True,
+        store=True,
+        related='attribute_id.priority',
+        help='Define priority of exhibition in reports'
```
guewen: `help` needs an update

product_attribute_priority/models/product.py
```diff
@@ -0,0 +1,58 @@
...
+
+    _inherit = 'product.attribute.line'
+    _order = 'priority'
+
+    priority = fields.Integer(
+        string='Priority',
+        readonly=True,
+        store=True,
+        related='value_ids.priority',
+        help='Define priority of exhibition in reports'
```
guewen: help needs an update

product_attribute_priority/models/product.py
mileo: Any suggestion? Can be:

> "Set the display order of the variants in visions and reports"

Thanks in advance

mileo:
@moylop260 and @guewen Issues fixeds

moylop260:
Thanks!
I reconfirm my +1

product_attribute_priority/__init__.py
```diff
@@ -0,0 +1,2 @@
+# -*- encoding: utf-8 -*-
```
bealdav: s/encoding/coding or remove line because ascii in fact, the default coding.

product_attribute_priority/__openerp__.py
```diff
@@ -0,0 +1,37 @@
...
+#    This program is distributed in the hope that it will be useful,
+#    but WITHOUT ANY WARRANTY; without even the implied warranty of
+#    MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+#    GNU General Public License for more details.
+#
+#    You should have received a copy of the GNU General Public License
+#    along with this program.  If not, see http://www.gnu.org/licenses/.
+#
+##############################################################################
+
```
bealdav, line 3: short version

```
# coding: utf-8
# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
```

product_attribute_priority/models/product.py
bealdav: Suggestion

> Set the display order of the attributes in backoffice and reports (in website too if installed)

product_attribute_priority/views/product_view.xml
```diff
@@ -0,0 +1,18 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<openerp>
+    <data>
+
+        <record model="ir.ui.view"
+            id="product_attribute_priority_tree_view">
```
bealdav: You may reuse the same id than the original if you want. Whatever module name prefix the id string finally

bealdav:
I like your module: efficient and short. Maybe `priority` field could be renamed `sequence` ?

product_attribute_priority/models/product.py
```diff
@@ -0,0 +1,58 @@
...
+        string='Priority',
+        help='Define priority of exhibition in reports',
+        default=10,
+    )
+
+
+class ProductAttributeValue(models.Model):
+
+    _inherit = 'product.attribute.value'
+    _order = 'priority, name'
```
bealdav: I think it'll interesting if we could add another field (`value_sequence` or any). 
Yes because we have an issue on attribute values on website but also in some other views: 

If I have dimensions: `1 cm, 3 cm, 6 cm, 10 cm`  

> 10  is in second position after 1 cm, whereas it should be last

Then we have `_order = 'priority, value_sequence, name'`.
With this settings, order of the values could be defined for those particular values.

What do you think ? Do you accept a PR on your PR ?

Thanks

pedrobaeza: Why not simply change the priority in that case instead of having one more field?

bealdav: priority is applied to product.attribute, and new field to product.attribute.value.
other priority fields are only related (readonly) to first one.
If I change the priority, it has no effect on attribute value, don't you think ?

pedrobaeza: OK, understood :+1:

mileo: :+1:

mileo:
@bealdav Can you make another pr with your modifications? Or make a PR to my fork?

bealdav:
:+1:

hparfr:
👍  usefull

mourad-ehm:
:+1:

PR #157, opened 2016-05-18
[9.0] [MIG] product_dimension

ghost:


product_dimension/product.py
```diff
@@ -21,7 +21,7 @@
 
 
 class Product(models.Model):
-    _inherit = 'product.product'
+    _inherit = 'product.template'
```
pedrobaeza, line 24: You shouldn't change this. It's on purpose at variant level.

ghost, line 24: ah, then nothing I did should be kept...I don't need it there so I'll just keep mine for now.

leemannd, line 24: Hello, In v8.0 there are both "product.template" & "product.product". I need the product dimension in V9.0.
Should I make a PR from the actual v8.0 version of this module.
@pedrobaeza Why is it kept on purpose at variant level?

pedrobaeza, line 24: Yes, follow this migration guide: https://github.com/OCA/maintainer-tools/wiki/Migration-to-version-9.0

leemannd, line 24: PR in WIP state https://github.com/OCA/product-attribute/pull/159

guewen:
Will be replaced by https://github.com/OCA/product-attribute/pull/159
@idncom you might be interested in #159 as it enables the dimension both on the template and on the variants (as on 8.0).

PR #158, opened 2016-05-20
Module for add default code in product template

gonxi:


product_template_deault_code/models/product_template.py
```diff
@@ -0,0 +1,43 @@
...
+                name = '[%s] %s' % (code,name)
+            return (d['id'], name)
+        result = []
+        for product in self:
+            mydict = {
+                'id': product.id,
+                'name': product.name,
+                'default_code':product.default_code,
+            }
+            result.append(_name_get(mydict))
```
bealdav, line 42: I think, you should call `super()` to override and not erase other behavior

PR #159, opened 2016-05-26
[9.0] [MIG] product dimension

leemannd:
- Added commits from 8.0
- Port from 8.0 to 9.0


product_dimension/models/product.py
```diff
@@ -0,0 +1,79 @@
...
+    height = fields.Float()
+    width = fields.Float()
+    dimensional_uom_id = fields.Many2one(
+        'product.uom',
+        'Dimensional UoM',
+        domain=_get_dimension_uom_domain,
+        help='UoM for length, height, width')
+
+
+class Product_template(models.Model):
```
mdietrichc2c: class ProductTemplate(models.Model):

product_dimension/models/product.py
```diff
@@ -0,0 +1,79 @@
...
+
+    def convert_to_meters(self, measure, dimensional_uom):
+        uom_meters = self.env['product.uom'].search([('name', '=', 'm')])
+
+        return self.env['product.uom']._compute_qty_obj(
+            from_unit=dimensional_uom,
+            qty=measure,
+            to_unit=uom_meters)
+
+    length = fields.Float(related='product_variant_ids.length')
```
mdietrichc2c, line 72: Why are they defined twice for template and product? Also, the functions are never called for templates as they are related fields (on many2one).

What should be done is set the functions and fields on product.template only ; the fields will be accessible from product.product.

mdietrichc2c, line 72: After discussing it and reading the original PR : I didn't understand that it was the goal to set it on the template, and edit it on all products at the same time. So it's not an issue.

guewen, line 72: More than that, the main benefit is that if you don't work with variants and have 1 variant on templates, then you'll have the dimension at the template level too.

leemannd:
@pedrobaeza If it's ok for you.

leemannd:
@idncom Here is a proposition of port that should fit your needs

flotho:
Hi @pedrobaeza, @guewen , any chance to be merged?

product_dimension/models/product.py
```diff
@@ -0,0 +1,78 @@
...
+
+    @api.model
+    def _get_dimension_uom_domain(self):
+        return [
+            ('category_id', '=', self.env.ref('product.uom_categ_length').id)
+        ]
+
+    length = fields.Float()
+    height = fields.Float()
+    width = fields.Float()
```
gurneyalex, line 41: I'm wondering if we want to have a decimal precision here (this is really a question).

gurneyalex:
one small interrogation, but :+1: 

thanks for porting.

gurneyalex:
@flotho I suggest you make a review of the module and state it in this discussion: it will speed up the process...

product_dimension/models/product.py
leemannd, line 41: @gurneyalex I see no harm in putting decimal precision. For the moment all the only I know that has dimensional fields using float without decimal precision (not ported: https://github.com/OCA/product-attribute/blob/9.0/product_size/product_size.py#L57) .

gurneyalex, line 41: @leemannd let's keep it that way for now. Can be added later or in a separate module.

mdietrichc2c:
:+1:

pedrobaeza:
Please squash together all the commits that come from the same person. We have for example 3 commits with the exact same title. Other from @leemannd with only patches from comments. When done, I'll merge.

leemannd:
@pedrobaeza Hello, the commit history should be better now.

panos-ch:
The module works OK with products with no attributes. Please consider making it work with products that include attributes and also be able to add dimensions to the different variants of the product.

leemannd:
@panos-ch Hello, I'm seeing your feedback now. Can you be more specific? With which kind of attributes? With another module installed?

panos-ch:
@leemannd Hi, no other modules are installed. The dimension fields are not displayed for product templates that have variants. Also, it would be nice to expand the module to work at a product variant level. I can add some screenshots if it's not clear

leemannd:
@panos-ch For now, I haven't used variants. For the moment I don't have the time to expand it. Maybe you could do a PR or an Issue with detailed informations?

PR #160, opened 2016-06-06
[FIX] profile when default in vals

bealdav:


sebastienbeau:
code review and tested

chafique-delli:
:+1:

PR #164, opened 2016-06-22
product_unique_default_code Module

rami-wafaie:
Make default_code field as a mandatory field in product.product + default_code should be unique

### DIFF
--- a/product_unique_default_code/product.py
+++ b/product_unique_default_code/product.py
@@ -13,5 +13,6 @@ class Product(models.Model):
     @api.constrains('default_code')
     def _check_unique(self):
         for obj in self:
-            if 1 < self.search_count([('default_code', '=', obj.default_code)]):
+            if 1 < self.search_count(
+                    [('default_code', '=', obj.default_code)]):
                 raise UserError(_("Internal Reference must be unique!"))
--- a/product_unique_default_code/product.py
+++ b/product_unique_default_code/product.py
@@ -1,0 +1,17 @@
+# -*- coding: utf-8 -*-
+# © 2016 initOS Gmb
+# License GPL-3.0 or later (http://www.gnu.org/licenses/gpl.html).
+
+from openerp import models, api, _
+from openerp.exceptions import Warning
+
+
+class Product(models.Model):
+    _inherit = 'product.product'
+
+    @api.multi
+    @api.constrains('default_code')
+    def _check_unique(self):
+        for obj in self:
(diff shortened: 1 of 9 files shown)


andhit-r:
@rami-wafaie i haven't test this but what would happen when user try to install this module on database with existing non-unique product?

Tested functionality on runbot :+1:

andhit-r:
@rami-wafaie it's better if you add unittest. Thanks.

rami-wafaie:
@andhit-r  I tested it on database which has existing non-unique products, no thing happened, the old products remained the same,  I will consider to add some checks regarding this point,

Added tests

product_unique_default_code/tests/test_product_unique_default_code.py
```diff
@@ -0,0 +1,35 @@
...
+from openerp.tests.common import TransactionCase
+from openerp.exceptions import Warning as UserError
+
+
+class TestProductUniqueDefaultCode(TransactionCase):
+
+    def setUp(self):
+        super(TestProductUniqueDefaultCode, self).setUp()
+
+        self.model = self.env['product.product']
```
gdgellatly: should this not be part of setup

product_unique_default_code/product.py
```diff
@@ -0,0 +1,20 @@
...
+
+    default_code = fields.Char('Internal Reference', copy=False)
+
+    @api.multi
+    @api.constrains('default_code')
+    def _check_unique(self):
+        for obj in self:
+            if 1 < self.search_count(
+                    [('default_code', '=', obj.default_code)]):
+                raise UserError(_("Internal Reference must be unique!"))
```
gdgellatly, line 20: Hi,

As a semantics issue I feel the conditional should read if count > 1.  A unique index on default_code does the same thing although I suppose if duplicates already exist the sql method will not install plus you get a nicer error message here.  

While I didn't test I think there are also possible issues with this constraint.  For example User A creates a product in Company A call PROD.  User B creates a product in Company B called PROD.  Depending on user B's access and record rule setup the constraint either passes or fails?

Also I think it will make variants impossible to use by default as they don't have a default_code at time of creation as standard.  Probably needs a test for that scenario.

gdgellatly: Actually there is another issue here.  The search won't take into account inactive products, maybe thats a desired behaviour but worth mentioning.

rousseldenis:
@rami-wafaie Do you plan to finish this ? As it is very old, it will be closed soon.

rousseldenis:

PR #165, opened 2016-06-23
[FIX] Fix field base_pricelist_id visibility when base is fixed price

hurrinico:
If you set fixed price on a pricelist item save and go back on the item you can see base_pricelist_id field.
This PR fix the problem

### DIFF
--- a/product_pricelist_fixed_price/view/product_pricelist_item_view.xml
+++ b/product_pricelist_fixed_price/view/product_pricelist_item_view.xml
@@ -57,6 +57,9 @@
                     <attribute name="attrs"
                         >{'invisible': [('base_ext', '=', -3)]}</attribute>
                 </field>
+                <field name="base_pricelist_id" position="attributes">
+                    <attribute name="attrs">{'invisible': [('base_ext', '=', -3)]}</attribute>
+                </field>
             </field>
         </record>
 


tafaRU:
@pedrobaeza, trying the current PR, I found another issue:

the first time https://github.com/OCA/product-attribute/blob/8.0/product_pricelist_fixed_price/model/product_pricelist_item.py#L61 is called, self is not a record yet, so type is equal to False.
As it's been already done in https://github.com/OCA/product-attribute/blob/8.0/product_pricelist_fixed_price/model/product_pricelist_item.py#L100 we could put 'sale' as fallback value, but in this way we wouldn't be able to use the 'purchase' pricelist type.

What's your opinion?

Thank you in advance.

eLBati:
@tafaRU @pedrobaeza see https://github.com/OCA/product-attribute/pull/175

Consider `product_pricelist_fixed_price` would work with both `sale` and `purchase` default type. What we need to avoid is to have `base` = -1, because we would fall into this
https://github.com/odoo/odoo/blob/8.0/addons/product/pricelist.py#L298, `rule.base_pricelist_id` would be empty and `price` would be `False`

eLBati:
This PR is :+1: for me

tafaRU:
:+1:

archetipo:
:+1:

PR #166, opened 2016-06-23
[MIG]9.0 product dimension

flotho:


pedrobaeza:
There's already https://github.com/OCA/product-attribute/pull/159 for migrating product_dimension

PR #168, opened 2016-06-24
[8.0][product_custom_info] New module to have custom properties in products

yajo:
Relaunch #121 from right repo. Sorry for that.

@Tecnativa @pedrobaeza @carlosdauden @rafaelbn


rafaelbn:
Simply nice, thanks :+1:

product_custom_info/__openerp__.py
```diff
@@ -0,0 +1,27 @@
...
+        'security/ir.model.access.csv',
+        'views/product_view.xml',
+    ],
+    'author': 'Antiun Ingeniería S.L., '
+              'Incaser Informatica S.L., '
+              "Tecnativa, "
+              "Odoo Community Association (OCA)",
+    'website': 'http://www.antiun.com',
+    'license': 'AGPL-3',
+    'installable': True,
```
lasley, line 26: I believe the `application` key is required here for some automation. @pedrobaeza - am I remembering correctly?

lasley:
Code (one comment inline, not totally certain on) + functional 👍

product_custom_info/__openerp__.py
pedrobaeza, line 26: `application` is used for marking a module as a "full-featured" module that allows to use a new area. This can't be considered like that. Some examples of modules with application=True can be the `medical` main module or `donation` module.

lasley, line 26: We don't have to explicitly define false? I remembered you saying something about this one time, but maybe it was just in reference to the installable key.

pedrobaeza, line 26: The default value for application is False. The comment was about `'installable': True` to be explicitly put (although the default value is that) because a fault in the branch migration script that can make this fails (https://github.com/OCA/maintainer-tools/blob/ca0941c654a4c319686aff9cd82b66ab0fb9a06b/tools/migrate_branch.py#L167-L172)

lasley, line 26: Ahhh cool thanks for the elaboration (as always) @pedrobaeza - I'll make sure to commit it to memory fully this time 😄

rafaelbn:
This PR is ready to be merged. @OCA/product-maintainers could you take a look please? Thanks!

pedrobaeza:
:+1:

PR #175, opened 2016-07-21
FIX product_pricelist_fixed_price: creation of new fixed price list

eLBati:
FIX the following use case:

Install product_pricelist_fixed_price and sale
Create a new product P1
Create a new price list 'fixed' with a version and a rule linked to product P1, based on 'fixed price' and set 'new price' = 100
Save
Create a new sale order, select 'fixed' price list, add a line, select P1 product, get

No valid pricelist line found ! :Cannot find a pricelist line matching this product and quantity.
You have to change either the product, the quantity or the pricelist

### DIFF
--- a/product_pricelist_fixed_price/model/product_pricelist_item.py
+++ b/product_pricelist_fixed_price/model/product_pricelist_item.py
@@ -57,8 +57,9 @@ class ProductPricelistItem(models.Model):
             if base_ext != FIXED_PRICE_TYPE:
                 base = base_ext
             else:
-                base = self._get_default_base(
-                    {'type': self.price_version_id.pricelist_id.type})
+                base = self._get_default_base({
+                    'type': self.price_version_id.pricelist_id.type or 'sale'
+                })
         else:
             # getting here we are sure base is in vals
             base = vals['base']


archetipo:
@eLBati  thanks Travis is red please fix flake8

pedrobaeza:
Have you tried with last versions of Odoo/OCB? This is not a fault in the module, but a bug in onchange mechanism, but with this, you're fixing sale behaviour, but not for purchase,

eLBati:
@pedrobaeza just tried with last revision of OCB 8.0 + product-attribute 8.0 and got the same error.

Also, for purchase case would work anyway, because, to make `product_pricelist_fixed_price` correctly work , the `base` field can assume any possible value of the selection that is > 0 (that is, it must not be `(-1, _('Other Pricelist'))` nor `(-2, _('Supplier Prices on the product form'))`)

eLBati:
@archetipo thanks, fixed

pedrobaeza:
OK, thanks for the explanation

:+1:

archetipo:
:+1:

tafaRU:
:+1:

PR #169, opened 2016-06-24
 [8.0][product_pack] New module to have product packs

yajo:
Relaunching #118 from right repo, sorry for that.

This fixes adhoc-dev/odoo-addons#244. Related to #6.

This is probably a WIP, but I'd thank reviews from original authors.

@Tecnativa @rafaelbn @ingadhoc @jjscarafia.

PR #180, opened 2016-09-19
[9.0][product_bundle] New module to have product bundles.

yajo:
This is a continuation from https://github.com/OCA/product-attribute/pull/169, dramatically simplified.

@Tecnativa


product_bundle/models/product.py
```diff
@@ -0,0 +1,159 @@
...
+# License LGPL-3.0 or later (http://www.gnu.org/licenses/lgpl).
+from openerp import _, api, fields, models
+from openerp.exceptions import ValidationError
+from openerp.tools.float_utils import float_round
+
+
+class ProductTemplate(models.Model):
+    _inherit = "product.template"
+
+    bundle_ok = fields.Boolean(
```
carlosdauden, line 12: The name of this field could be "is_bundle"?

carlosdauden (commented):
When you are creating a new bundle (product.template), if you don't save before than add Bundled products, Bundled products don't save.

product_bundle/models/product.py
yajo: The `_ok` suffix seems to be the standard...

rafaelbn:
Hi @Yajo , which is the status of this PR? Why is WIP @pedrobaeza ? Thanks!

cc @jjscarafia here you are ;-D

yajo:
This is to be reviewed.

product_bundle/README.rst
```diff
@@ -0,0 +1,87 @@
...
+
+==============
+Product Bundle
+==============
+
+This module extends the functionality of products to support having a bundle of
+products in a single one.
+
+Internally, bundles will always be consumable products, but you will be able to
+see their stock, which is extracted from the maximum amount of stockable
```
rafaelbn, line 13: s/maximum/minimum

rafaelbn (approved):
tested in runbot

rafaelbn:
@pedrobaeza this is not a WIP is NR , please change tag, thanks. For me is 👍

pedrobaeza:
Stock part is not yet covered.

product_bundle/i18n/es.po
```diff
@@ -0,0 +1,208 @@
...
+# Translators:
+# Albert Cervera i Areny <albert@nan-tic.com>, 2009
+msgid ""
+msgstr ""
+"Project-Id-Version: odoo-addons (8.0)\n"
+"Report-Msgid-Bugs-To: \n"
+"POT-Creation-Date: 2015-09-09 23:57+0000\n"
+"PO-Revision-Date: 2015-09-10 00:00+0000\n"
+"Last-Translator: Juan Jose Scarafia <scarafia.juanjose@gmail.com>\n"
+"Language-Team: Spanish (http://www.transifex.com/adhoc/ingadhoc-odoo-addons-8-0/language/es/)\n"
```
jjscarafia, line 14: I believe that this module came, in some way, from product_pack (https://github.com/ingadhoc/product/tree/9.0/product_pack). What about authors?
Thanks!

pedrobaeza: No, that's only something not updated. We created from the scratch, but @Yajo reuse the folder and didn't remove these files. You only need to check code to see that nothing is equal...

jjscarafia: ok, just wondering, no need to check code. Thanks!

pedrobaeza:
Closing for now as we don't have any project requiring this and there's still work to do. Anyone wanting to work on this can retake the work.

rafaelbn:
It would be better to merge it, as it was working. Just a known issues about stock. But any way... here it is

rafaelbn:
Working in https://github.com/OCA/product-attribute/pull/498

PR #182, opened 2016-10-04
[FIX] Extended copy method for model product.product

espo-tony:
Currently this module only extends the copy method for the model product.template, in order to provide a non-duplicated value for the field prefix_code.
Unfortunately, when a user tries to duplicate a product (model product.product), the copy method of the product.template model is not called (only the copy method of the product.product model is called and it doesn't recall the copy method of the product.template model, only its create method). 
The result is that the system tries to duplicate the field prefix_code on the new product.template record and this raises an error, as the field prefix_code is declared to be unique.

The solution that I propose is to simply extend the copy method in the product.product model in order to prevent the system from duplicating a unique field.

### DIFF
--- a/product_code_builder_sequence/models/product_product.py
+++ b/product_code_builder_sequence/models/product_product.py
@@ -46,3 +46,18 @@ class ProductTemplate(models.Model):
                     'prefix_code': self.prefix_code + _('-copy'),
                 })
         return super(ProductTemplate, self).copy(default)
+
+
+class ProductProduct(models.Model):
+    _inherit = 'product.product'
+
+    @api.multi
+    def copy(self, default=None):
+        for product in self:
+            if default is None:
+                default = {}
+            if product.prefix_code:
+                default.update({
+                    'prefix_code': self.prefix_code + _('-copy'),
+                })
+        return super(ProductProduct, self).copy(default)
--- a/product_code_builder_sequence/models/product_product.py
+++ b/product_code_builder_sequence/models/product_product.py
@@ -38,13 +38,13 @@ class ProductTemplate(models.Model):
 
     @api.multi
     def copy(self, default=None):
-        for product in self:
-            if default is None:
-                default = {}
-            if product.prefix_code:
-                default.update({
-                    'prefix_code': self.prefix_code + _('-copy'),
-                })
+        self.ensure_one()
+        if default is None:
+            default = {}
+        if 'prefix_code' not in default and self.prefix_code:
+            default.update({
+                'prefix_code': self.prefix_code + _('-copy'),
+            })
         return super(ProductTemplate, self).copy(default)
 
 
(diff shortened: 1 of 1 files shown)


product_code_builder_sequence/models/product_product.py
```diff
@@ -46,3 +46,18 @@ def copy(self, default=None):
...
+
+class ProductProduct(models.Model):
+    _inherit = 'product.product'
+
+    @api.multi
+    def copy(self, default=None):
+        for product in self:
+            if default is None:
+                default = {}
+            if product.prefix_code:
```
StefanRijnhart: Plese check if prefix_code is not already in 'default'

product_code_builder_sequence/models/product_product.py
```diff
@@ -46,3 +46,18 @@ def copy(self, default=None):
                     'prefix_code': self.prefix_code + _('-copy'),
                 })
         return super(ProductTemplate, self).copy(default)
+
+
+class ProductProduct(models.Model):
```
StefanRijnhart, line 51: Can you document the reason of the code duplication in the doc string?

StefanRijnhart (changes requested):

product_code_builder_sequence/models/product_product.py
gdgellatly: Note @StefanRijnhart this is a direct copy of the existing product.template code, so the existing module has the same error.  

@espo-tony In addition the existing module also has the error of product.prefix_code morphing into self.prefix_code on the following line.

StefanRijnhart: @gdgellatly Yes, this was a long standing issue since https://bugs.launchpad.net/openobject-addons/+bug/942699 but my fix was finally merged in June this year! https://github.com/odoo/odoo/pull/12397

StefanRijnhart: Or do you mean the same code for product.template, a couple of lines above? https://github.com/OCA/product-attribute/pull/182/files#diff-570b3649efcb95036f9fa9884a839eb2L44

espo-tony: @StefanRijnhart , Implemented the check for an already existing prefix_code in default, in both copy methods. Thanks for the hint.

@gdgellatly , You're right: product.prefix_code was morphing in self.prefix_code. This never caused troubles as copy methods are supposed to run on a single record, therefore self and product are, in most cases, the same object. 
I've fixed this mistake in the latest commit. Thanks for your review.

product_code_builder_sequence/models/product_product.py
espo-tony: This is not a code duplication. I'm just reproducing the existing feature of the method product.template.copy() into the method product.product.copy().

This is needed because when you duplicate a product (model product.product) the method product.template.copy() is, correctly, not invoked. Therefore the method tries to duplicate the field prefix_code on the new product, which is not allowed as it's a violation of the SQL Unique constraint on the field prefix_code.

product_code_builder_sequence/models/product_product.py
```diff
@@ -53,11 +53,11 @@ class ProductProduct(models.Model):
...
     @api.multi
     def copy(self, default=None):
-        for product in self:
-            if default is None:
-                default = {}
-            if product.prefix_code:
-                default.update({
-                    'prefix_code': self.prefix_code + _('-copy'),
-                })
+        self.ensure_one()
```
hurrinico, line 56: i think it's better to use a for cicle on self instead of ensure one.

hurrinico (commented):
Just the ensure_one comment for me, the rest is :+1:

product_code_builder_sequence/models/product_product.py
hurrinico: i'm not thinking about the return, so maybe in this case it's ok to use ensure_one so it's :+1:  for me

astirpe (approved):

PR #188, opened 2016-10-15
[MIG][10.0] product_manufacturer: tested with EE and CE

bistaray:
Note: The view this hooked to is in the stock module, hence the additional dependency.  Also, I don't know the work needed to port the MIGRATION folder (is this code only used during database migration?), so I have not made any changes to it - if that is still needed someone needs to port that part.  I tested on a DB with no records.

**Community:**

**Enterprise:**

### DIFF
--- a/product_manufacturer/__manifest__.py
+++ b/product_manufacturer/__manifest__.py
@@ -1,4 +1,4 @@
-# -*- encoding: utf-8 -*-
+# coding: utf-8
 ##############################################################################
 #
 #    OpenERP, Open Source Management Solution
--- a/product_manufacturer/__manifest__.py
+++ b/product_manufacturer/__manifest__.py
@@ -18,17 +18,17 @@
 #
 ##############################################################################
 {
-    'name': 'Products Manufacturers',
-    'version': '9.0.1.0.0',
-    'author': "OpenERP SA,Odoo Community Association (OCA)",
+    'name': 'Product Manufacturers',
+    'version': '10.0.1.0.0',
+    'author': "OpenERP SA, Odoo Community Association (OCA)",
     'license': 'AGPL-3',
     'contributors': ['Acysos SL <info@acysos.com>'],
     'category': 'Purchase Management',
-    'depends': ['product'],
+    'depends': ['product','stock'],
     'demo': [],
     'data': [
         'views/product_manufacturer_view.xml'
     ],
     'auto_install': False,
-    'installable': False,
+    'installable': True,
 }
--- a/product_manufacturer/__manifest__.py
+++ b/product_manufacturer/__manifest__.py
@@ -25,7 +25,7 @@
     'license': 'AGPL-3',
     'contributors': ['Acysos SL <info@acysos.com>'],
     'category': 'Purchase Management',
-    'depends': ['product','stock'],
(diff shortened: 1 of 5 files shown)


raycarnes:
Attached is a digitally signed copy of the ICLA.

Ray Carnes.

pedrobaeza:
I have change the title for easing the review of pending PRs. Please take not for other occasions.

pedrobaeza:
CLA checked and added.

Please check the Travis status.

product_manufacturer/__manifest__.py
```diff
@@ -18,17 +18,17 @@
...
-    'version': '9.0.1.0.0',
-    'author': "OpenERP SA,Odoo Community Association (OCA)",
+    'name': 'Product Manufacturers',
+    'version': '10.0.1.0.0',
+    'author': "OpenERP SA, Odoo Community Association (OCA)",
     'license': 'AGPL-3',
     'contributors': ['Acysos SL <info@acysos.com>'],
     'category': 'Purchase Management',
-    'depends': ['product'],
+    'depends': ['product','stock'],
```
pedrobaeza: You shouldn't depend on this module, but add the group in other place.

bistaray: 'stock' does not add the Inventory page in the form view.  The Inventory page was where these fields were anchored in the previous versions.

This is what the Product Template form view looks like with 'stock' not installed:



Choosing another group to put it (other than the Inventory page supplied by the stock module) would require chosing either the 'General Information' page or creating a new page.  What do you advise?

pedrobaeza:
General information looks good to me, as this is not "Inventory" related info, don't you think?

bistaray:
Hmm - I prefer a new page "Manufacturer" then I can also shorten the names of all the field labels so they fit nicer in the UI.

pedrobaeza:
With this, we have the risk of increasing a lot the number of pages if each module adds a new one. I don't see it very ugly. Maybe we can shorten field labels instead.

pedrobaeza:
You can use "Manuf. product code", "Manuf. product name" and so on.

bistaray:
I have abbreviated the field names and placed them on the General Information tab:

pedrobaeza (approved):
Please squash and merge to whom that merges the PR

bistaray:
[product_manufacturer.zip](https://github.com/OCA/product-attribute/files/531980/product_manufacturer.zip)

(Adding ZIP version of the complete module to allow testing by non-technical users)

pedrobaeza:
That's not needed, as we have runbot for that purpose. https://odoo-community.org/page/review

bistaray:
I don't see a way for a non technical user to install this module locally to test without a zip.

pedrobaeza:
The question is that you don't to install the module. Just go to the runbot link, and try it in a instance with the module installed. This is for even less technical people, because they don't need to know how to install a module.

bistaray:
I agree with you that it is possible to test via runbot and review the code via git.  I am handling the case where a user WANTS to test locally.

pedrobaeza:
OK, another choice for the user before merging :wink:

gurneyalex (approved):

PR #189, opened 2016-10-19
[MIG][9.0] Upgrade pricelist_per_product to 9.0

woodbrettm:
## Overview
- As most of the pricelist_per_product concepts exist in the core, I removed the logic.
- Overrides made to views as follows depending on the pricelist sale settings
  - **Single sale price per product**
    - Defaults kept (pricelist items invisible in product form)
  - **Different prices per customer segment**
    - Defaults kept (pricelist items in product form inline editable)
  - **Advanced pricing based on formula**
    - pricelist items in product form now visible
    - Editing a pricelist_item in the product form opens up the pricelist_item form instead of being edited inline
- pricelist_id added to pricelist form
## Issues
- Can't seem to figure out how to autofill the product_tmpl_id field when creating a new pricelist_item. This isn't relevant when the tree is inline-editable and when editing an existing pricelist_item (the product_tmpl_id is autofilled there as shown in image). @lasley 
### Different prices per customer segment

product_form_inline
### Advanced pricing based on formula

product_form


lasley:
You default things with context:

```
<field name="item_ids" 
       context="{'default_product_tmpl_id': id,
                 'default_other_field': field_name,
                 }" />
```

lasley:
@bmw95 - Hit me up in chat when you get some time; would like to talk about commit messages

woodbrettm:
@lasley thanks for the tips. I updated the context and that works a lot better now. Ready for review.

lasley:
@bmw95 I'm confused about all the changes you made. Why change the icon to default OCA, why are so many files missing, why remove things from the Roadmap that you didn't solve, why remove Configuration instructions from the Readme?

woodbrettm:
@lasley apologies for the slight delay. I reviewed the original code and have a few points:
- Icon was removed because the view it's showing is out-of-date. However if you suggest, I could update with a similar icon just with the new context
- Ah yeah I don't think mass updating does exist in the core does it : |. I'll add that back to the readme
- Instead of removing the config instructions I'll update with new context. Will update the index.html file instead of deleting as well
- In terms of the models, translations, and views, those were removed as from what I can see the concepts exist in the core. Just some adjustments were needed to override some things in the core views.

Interested to hear your thoughts though perhaps I missed some things.

woodbrettm:
@lasley ping

pricelist_per_product/__openerp__.py
```diff
@@ -0,0 +1,24 @@
...
+# Copyright 2015 Akretion
+# Copyright 2016 LasLabs Inc.
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+
+{
+    "name": "Pricelist Per Product",
+    "summary": "Display pricelist items in products",
+    "version": "9.0.1.0.0",
+    "category": "Sales",
+    "website": "https://laslabs.com",
```
lasley: Why did you change this to our website?

lasley (changes requested):
@bmw95 - Regarding the icon, please add the old one back or quickly create a new one. It's not really a big deal either way, but switching something customized for something generic is not good.

index.html file is not used. We use only the ReadMe

pricelist_per_product/__openerp__.py
```diff
@@ -0,0 +1,24 @@
...
+    "author": "Akretion, Laslabs, Odoo Community Association (OCA)",
```
lasley: I think author is a bit of an overreach here too. From what I can tell, you didn't do much here. My general rule is a 30-40% logic addition/refactor deserves authorship. Version migrations, bugfixes, etc. for the most part you're just a contributor.

lasley:
@bealdav - would you mind taking a look at this? I'm having a bit of trouble understanding the ReadMe & all features that exist in the v8 version of this, mainly in regards to the price grid. Is it correct that all the things being removed here are actually in core now?

bealdav:
I'll review it in few days. Thanks to noticed me.
Interesting vision:

> My general rule is a 30-40% logic addition/refactor deserves authorship. Version migrations, bugfixes, etc. for the most part you're just a contributor.

I have no strong opinion on this, but interesting to have guidelines.

lasley:
> I have no strong opinion on this, but interesting to have guidelines.

Yeah I'm not sure how to word this, but I've been thinking about it for a while now. There is definitely a line, but I'm still somewhat figuring it out by coaching my internal peeps. Once that line in the sand is established internally, I think it'll be easier to write it all out. 

This is a fairly new thing for me as I've been watching v10 migrations add authors for `awk` usage. I also noticed that there is a max char limit in the author col, which also contributes to this statement.

bealdav:
@bmw95, Here is quick review:
> As most of the pricelist_per_product concepts exist in the core, I removed the logic 
- Yes you removed all the logic, even logic which are not in core. :-(
- Between v8 and v9, `version` were dropped: OK view is updated
- How to access to all products with the same pricelist. Dropped and should be reintroduced: module name is pricelist_per_product, then this feature stick exactly with the topic

pl_smart

The button allow to go to products screen. **Then you can filter products of this domain make batch operation on it**. Please make it available on the view of pricelist instead of version.
pl_prd
- In v8: o2m of pricelist.version in sales tab, allowing to add product.pricelist.item with tree view directly. The purpose here was to improve ergonomy and productivity in encoding data. The drawback was only fixed price was supported. Maybe you could reintroduce this logic, don't you think ? I you don't want more productive UI, just use the standard UI, why a module for that ? If you could fix this, I could help you by supporting other method that only fixed price.
- Native product pricelist item form is complex: in your case, it seems default values are not set: when I try, I can add, from a product.template, a product variant of this template or of any template. Here is another reason to not reuse the native behavior if usability is worst.

Then if you're ok to fix these items, I could help you to support other methods than only fixed price. In v9 or in V10. Let me know. Thanks a lot for your contribution.

bealdav:
@lasley concerning 'price grid'. This concept was introduced to limit to managed fixed price in the pricelist.

lasley:
> concerning 'price grid'. This concept was introduced to limit to managed fixed price in the pricelist.

Oh sweet so literally a grid of prices. This totally makes sense in hindsight!

Thanks for the review @bealdav, we will loop back around and fix this up in next sprint.

tedsalmon:
Hey @bealdav!

Few questions for you based on your review:

>How to access to all products with the same pricelist. Dropped and should be reintroduced: module name is pricelist_per_product, then this feature stick exactly with the topic

It seems this already exists; when you click into a `Pricelist`, the "Pricelist Items" grid contains a list of all products in that pricelist. Are there any data points that's missing that we should re-introduce?

>In v8: o2m of pricelist.version in sales tab, allowing to add product.pricelist.item with tree view directly. The purpose here was to improve ergonomy and productivity in encoding data. The drawback was only fixed price was supported. Maybe you could reintroduce this logic, don't you think ? I you don't want more productive UI, just use the standard UI, why a module for that ? If you could fix this, I could help you by supporting other method that only fixed price.

Since `pricelist.version` is gone, we've added the pricelist tree under the product.template `Sales` tab. From there, we can create new pricelists for the item which we're looking at. During my testing, I didn't find any issue adding fixed or complex prices for the item which I was editing. Am I missing something? :)

>Native product pricelist item form is complex: in your case, it seems default values are not set: when I try, I can add, from a product.template, a product variant of this template or of any template. Here is another reason to not reuse the native behavior if usability is worst.

So, do we want to overwrite core behavior and only allow the addition of pricelist item exclusively for the product we're viewing? That way we can't add variants or pricelist items for any other products.

Thanks!
-Ted

bealdav:
@tedsalmon part 1/
> the "Pricelist Items" grid contains a list of all products in that pricelist

You're right you have the list of the products. But it's just a list from o2m without any features on it. This is very different from a tree view list with a lot capabilities: no checkbox to select items, no 'more button' then no action allowed by other modules (mass edit or any), no search capabilities on products of this domain, no thing.

bealdav:
part 2/
> I didn't find any issue adding fixed or complex prices for the item

On my side, I found it: default values are not set as said previously

bealdav:
part 3/
> only allow the addition of pricelist item exclusively for the product we're viewing

Yes only the product.template or any product.product with same template: it was the goal of the module (restrict pricelist for a product.template domain).

If you're not sure to understand what I mean (sorry for my bad english) you may prefer than I take in charge this PR on my side ? Let me know. Thanks

tedsalmon:
@bealdav,

Thanks for the clarification! I'll take care of it though; I need to get more familiar with Pricelists anyways 😆 . Could you take a look at 4890ec880d913c0677c15647e48862d351d8c6d5 and make sure this addresses part 1?


Thanks again!
-Ted Salmon

bealdav:
Ok it's a step in the right direction. I hope you can go on with product.pricelist.item ...
Thanks

tedsalmon:
Hey @bealdav,

I believe I have attended to your requirements. Please let me know if I missed anything :)

Thanks!
-Ted

bealdav:
Please @tedsalmon just test https://github.com/OCA/product-attribute/tree/8.0/pricelist_per_product 'try on runbot', you'll see all differences. But you can let it as for v9, I'll improve it in v10 to fill it directly in tree view. Thanks

tedsalmon:
@bealdav,

Interesting; I definitely ran them side by side and at least under Products and Pricelists, the functionality seemed congruent, specifically with the inline-fillable Tree views. Regardless, we can move forward if you'd like :)

Thanks!
-Ted

bealdav:
:+1:

sylvainc:
:+1:

lasley (approved):

bealdav:
@tedsalmon It seems you missed #233 in your code

tedsalmon:
Whoops, fixed @bealdav! I also added explicit tests, updated the verbiage and functionality to account for the missing now defunct `product.pricelist.version` :)

pricelist_per_product/models/product_pricelist.py
```diff
@@ -0,0 +1,40 @@
...
+        total = [pt_id for pt_id in self.item_ids.mapped('product_tmpl_id')]
+        self.product_template_count = len(total)
+
+    @api.multi
+    def button_template_in_pricelist(self):
+        """ Return a tree form of ``product.template`` for the pricelist """
+        self.ensure_one()
+        pids = []
+        for rec in self.item_ids.mapped('product_tmpl_id'):
+            pids.append(rec.id)
```
Garamotte: `pids = self.item_ids.mapped('product_tmpl_id').ids`

Garamotte (commented):
Just a very minor change :)

pricelist_per_product/models/product_pricelist.py
tedsalmon: That's pretty - I like it! Thanks!

tedsalmon:
@sylvain-garancher,

Updated - Thanks!

pricelist_per_product/views/product_template_view.xml
```diff
@@ -0,0 +1,40 @@
...
+    </record>
+
+    <record id="product_template_form_view" model="ir.ui.view">
+        <field name="name">product.template.common.form</field>
+        <field name="model">product.template</field>
+        <field name="inherit_id" ref="product.product_template_form_view" />
+        <field name="arch" type="xml">
+            <xpath expr="//div[@name='pricelist_item']" position="attributes">
+                <attribute name="groups">product.group_sale_pricelist</attribute>
+            </xpath>
```
bealdav: Hi ted, I don't understand why you inherit `ref="product.product_template_form_view"` twice ? 
It could be in the same record.

tedsalmon: No idea -- must have been an oversight but it's fixed now :)

lasley (approved):
LGTM thanks @bmw95 & @tedsalmon

bealdav:
ping @sebastienbeau could you merge ?

Issue #153, opened 2016-05-03
[9.0C] Product Brand - Logo Is not resized in Sales >> Product Brand - displays image in original full-width and height (screenshot attached)


lukebranch:
I've just come across an issue with Product Brand in Odoo 9 (CE).

Steps to reproduce:

1) upload a large picture as a brand image
2) view kanban view in Sales >> Product  Brand

Expected Behaviour:
Logo is resized to the width of the kanban container as a thumbnail or small image.

<img src="https://cloud.githubusercontent.com/assets/2337666/14979991/6dc7880a-1158-11e6-96c0-0dbf0eb2cd1b.png" />

pedrobaeza:
@JayVora-SerpentCS I assigned you as you were the original author of the module, isn't it?

JayVora-SerpentCS:
We aren't the author,but we will contribute

JayVora-SerpentCS:
Thanks, you are so quick!

lukebranch:
@pedrobaeza @JayVora-SerpentCS ,

Thanks for looking into this! I'll check again with the latest code.

EDIT:
Tested and it works great now, thanks! 

FEATURE REQUEST:
The image is still the full-size image, however it has been resized with CSS to 64x64px which works just fine in our case. I can imagine if there were hundreds of brands in there it might be worthwhile utilising the image, image_medium, and image_small features to avoid performance issues, and maintain the aspect ratio. This would ensure that for databases with perhaps even thousands of brands the performance and loads times for the page would be optimised.

PR #190, opened 2016-10-20
Coreected #153

JayVora-SerpentCS:
#153  - Display small size image on kanban as usual.


pedrobaeza (approved):
Tried on runbot. Thanks for the patch. Fast-tracking!

PR #193, opened 2016-10-24
[FIX] prevent unseful writing on products in product_profile

bealdav:
The objective is to fix performance issue.
False changes in product.profile can bring to a lot of write on product.product without this patch.

ping @sebastienbeau @chafique-delli

### DIFF
--- a/product_profile/product.py
+++ b/product_profile/product.py
@@ -76,14 +76,15 @@ class ProductProfile(models.Model):
         """ Profile update can impact products: we take care
             to propagate ad hoc changes """
         new_vals = vals.copy()
+        to_exclude_profile_fields = get_profile_fields_to_exclude()
         for key in vals:
             if (key[:LEN_DEF_STR] == PROF_DEFAULT_STR or
-                    key in get_profile_fields_to_exclude() or
-                    self.check_unuseful_key_in_vals(new_vals, key)):
+                    key in to_exclude_profile_fields or
+                    self.check_useless_key_in_vals(new_vals, key)):
                 new_vals.pop(key)
-        # super call must be after check_unuseful_key_in_vals() call
+        # super call must be after check_useless_key_in_vals() call
         # because we compare value before and after write
-        res = super(ProductProfile, self).write(vals)
+        res = super(ProductProfile, self).write(new_vals)
         if new_vals:
             for rec in self:
                 products = self.env['product.product'].search(
@@ -96,7 +97,7 @@ class ProductProfile(models.Model):
         return res
 
     @api.model
-    def check_unuseful_key_in_vals(self, vals, key):
+    def check_useless_key_in_vals(self, vals, key):
         """ If replacing values are the same than in db, we remove them.
             Use cases:
             1/ if in edition mode you switch a field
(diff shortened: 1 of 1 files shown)


chafique-delli:
Thank you @bealdav  for this fix.
 :+1:

product_profile/product.py
```diff
@@ -72,23 +75,47 @@ def _get_types(self):
...
-                keys2remove.append(key)
-        for key in keys2remove:
-            # we remove keys which have no matching in product template
-            vals.pop(key)
-        if vals:
-            products = self.env['product.product'].search(
-                [('profile_id', '=', self.id)])
-            if products:
-                products.write({'profile_id': self.id})
+                    key in get_profile_fields_to_exclude() or
```
hparfr: get_profile_fields_to_exclude() should be put outside the loop

product_profile/product.py
```diff
@@ -72,23 +75,47 @@ def _get_types(self):
...
+                    self.check_unuseful_key_in_vals(new_vals, key)):
+                new_vals.pop(key)
+        # super call must be after check_unuseful_key_in_vals() call
+        # because we compare value before and after write
+        res = super(ProductProfile, self).write(vals)
```
hparfr: new_vals instead of vals ?

product_profile/product.py
```diff
@@ -72,23 +75,47 @@ def _get_types(self):
...
+                    [('profile_id', '=', rec.id)])
+                if products:
+                    _logger.info(
+                        " >>> %s Products updating after updated '%s' "
+                        "product profile" % (len(products), rec.name))
+                    products.write({'profile_id': rec.id})
         return res
 
     @api.model
+    def check_unuseful_key_in_vals(self, vals, key):
```
hparfr: 'useless' instead of 'unuseful' ?

product_profile/product.py
```diff
@@ -72,23 +75,47 @@ def _get_types(self):
...
+            2/ if profile data are in csv file there are processed
+               each time module containing csv is loaded
+            we remove field from vals to minimize impact on products
+        """
+        comparison_value = self[key]
+        if self._fields[key].type == 'many2one':
+            comparison_value = self[key].id
+        elif self._fields[key].type == 'many2many':
+            comparison_value = [(6, False, self[key].ids), ]
+        if vals[key] == comparison_value:
```
hparfr: `return vals[key] == comparison_value`

product_profile/product.py
```diff
@@ -72,23 +75,47 @@ def _get_types(self):
...
+        """ If replacing values are the same than in db, we remove them.
+            Use cases:
+            1/ if in edition mode you switch a field
+               from value A to value B and then go back to value A
+               then save form, field is in vals whereas it shouldn't.
+            2/ if profile data are in csv file there are processed
+               each time module containing csv is loaded
+            we remove field from vals to minimize impact on products
+        """
+        comparison_value = self[key]
```
hparfr, line 110: use `else` in the following `if`

hparfr (commented):
👍  minor remarks

product_profile/product.py
bealdav: why else instead of elif ?

PR #195, opened 2016-10-24
[MIG][10.0] product_dimension

gurneyalex:


product_dimension/models/product.py
```diff
@@ -8,7 +8,7 @@
 from openerp import api
```
leemannd: s/openerp/odoo

gurneyalex: fixed, thanks

jcoux (approved):
:+1: Tested on runbot : it works !

leemannd (approved):
:+1:  Looks good to me.

product_dimension/views/product_view.xml
```diff
@@ -4,32 +4,28 @@
...
         </xpath>
       </field>
     </record>
 
     <record model="ir.ui.view" id="product_template_only_form_view">
-      <field name="name">product_template_form_view</field>
+      <field name="name">product.template with properties and dimensions</field>
       <field name="model">product.template</field>
-      <field name="inherit_id" ref="product.product_template_only_form_view"/>
+      <field name="inherit_id" ref="stock.view_template_property_form"/>
```
pedrobaeza, line 21: Changing dependency for only a convenient view might be too much. Can't you place this in another place?

susport:
👍 Good

product_dimension/views/product_view.xml
LeartS: `product` is not an app (in the odoo manifest sense), so IMO is not a big deal having modules depend on "real apps" (like `sale`, `purchase`, or `stock`). If you install only the `product` module, do you even have a menuitem to access them? I'm not sure!

pedrobaeza: Yeah, I agree, but depending only on product allows you to decide which app you want to install. You can install sale without stock and need to manage dimensions in product, but with this, you are forced to have the stock module installed. The same if you decide to depend on sale and you only need purchase. So the best option is to rely only on a view in product module (and it's reasonable to do it I think).

cliffkujala:
What is the status of this migration?

gdgellatly:
@gurneyalex please note this module has a nasty little bug.  Unfortunately underscore.js does not allow the use of the field name 'length' - see #210 I expect if we close and reopen this PR it will fail travis tests.

gurneyalex:
@gdgellatly  rebased and pushed.

gdgellatly:
@gurneyalex Guess travis wasn't updated as indicated.  This bug drove me mental for literally years until by some fluke someone solved it for me.  When you give a field the name length, and use it an embedded one2many list view, bad things happen.  This is because underscore.js uses the length attribute of the 'object' and if the odoo record has an attribute called length it takes precedence.  

Guess its more an FYI than blocking seeing this module never embeds it anywhere.  Certainly my modules now have lngth fields when talking about a products dimensions.

gurneyalex:
@gdgellatly do you think it is worth a field rename ?  I don't really understand the issue since other standard addons in odoo have a field called 'length' already (e.g. delivery). So surely it cannot be *that* bad?

@pedrobaeza you said in #210 that pylint should flag this, yet the rebased PR is green. Should we worry about this?

pedrobaeza:
It's flagged, but as beta warning for now: https://travis-ci.org/OCA/product-attribute/jobs/233559765#L516

You should rename it for avoiding problems if you embed the view. Odoo core is not always the best place to look at for extensibility problems.

gdgellatly:
@gurneyalex this was the test addon I used for reporting to Odoo https://github.com/gdgellatly/odoo_bug_demos/tree/master/testo2m .  The symptoms are whenever an embedded one2many displays a field named length, no text displays, no records are updated, new lines are blank etc.  But saving does update them.

Now on one hand it is extremely unlikely this module would ever suffer from it as its more form view based, on the other if it ever happens because someone really needs to edit in an embedded o2m, noone knows why it doesn't work.

Its related to a change made to form.js in v7 circa December 2015, strangely enough to solve another _.js problem related to many2many tags.

product_dimension/models/product.py
```diff
@@ -24,11 +23,7 @@ def onchange_calculate_volume(self):
 
     def convert_to_meters(self, measure, dimensional_uom):
         uom_meters = self.env['product.uom'].search([('name', '=', 'm')])
```
sylvainvh, line 25: Why not use the XML ID ? 
uom_meters = self.env.ref('product.product_uom_meter')

sylvainvh: I you use Odoo in an another language it doens't work

sylvainvh: @gurneyalex I created a PR with these two modification. Do what you want with this PR :p

jcdrubay (approved):
I think it is good to merge.

Yet, I think that both comments could be taken into account. Possibly in a next version, it could be added to the roadmap in the README if no-one has time to take into account now:
 - use xml_id in convert_to_meters
 - depend only on product

sylvainvh:
Do you think that we can merge (if @gurneyalex accept my PR) this PR rapidly ?

pedrobaeza:
There's still the problem on depending on stock.

flotho:
Hi everyone. 
FMI, who could solve those conflicting files ?

regards

flotho:
Hi @gurneyalex , @pedrobaeza ,

Who is responsible of resolving such conflicts?

leemannd:
@flotho Normally the author of the PR is the responsible for fixing his PR, but you can help.
You can propose a PR on @gurneyalex 's branch and it will be present in this PR once merged.

pedrobaeza:
Superseeded by #491

PR #198, opened 2016-11-02
[FIX] product_profile in case new value empty

bealdav:
ping @sebastienbeau

### DIFF
--- a/product_profile/product.py
+++ b/product_profile/product.py
@@ -76,10 +76,10 @@ class ProductProfile(models.Model):
         """ Profile update can impact products: we take care
             to propagate ad hoc changes """
         new_vals = vals.copy()
-        to_exclude_profile_fields = get_profile_fields_to_exclude()
+        excludable_fields = get_profile_fields_to_exclude()
         for key in vals:
-            if (key[:LEN_DEF_STR] == PROF_DEFAULT_STR or
-                    key in to_exclude_profile_fields or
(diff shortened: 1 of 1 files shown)


legalsylvain:
Hi,
Thanks for this PR.
I find it technically ok, but maybe not in this module.
In 99,9% of the case, it will be ok, but maybe in some cases, not calling super will have an unexpected side effect. (because sometimes write function not only write, (but does a log, for exemple).
I mean, avoiding realizing a write if vals are empty can be an interesting feature but maybe in a generic module like 'model_avoid_emty_write', with inheritance in 'model.base'.
What do you think ? Any other PoV ? I personnaly don't use this module for the time being, so my PoV is limited.
regards.

bealdav:
@legalsylvain Ok, do you accept my PR if I had a comment in thi code to schedule me to do the module you ask me in next month.
It's a really good idea but it's an important fix and I have no time for now to do it.
cc @sebastienbeau @hparfr 
thanks for your feedback

bealdav:
@legalsylvain Ok I scrap previous commit and transform in cosmetic commit to avoid change write behavior, as you said me.
Now PR is safe

legalsylvain:
:+1: thanks !

bealdav:
ping @hparfr

hparfr:
👍 
startswith is definitely more readable  :triumph:

PR #202, opened 2016-11-13
[MIG] 9.0 product_attribute_priority

lmignon:
migration from 8.0


bealdav:
:+1:

sylvainc:
:+1:

product_attribute_priority/views/product_view.xml
```diff
@@ -0,0 +1,25 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<odoo>
+
+    <record model="ir.ui.view" id="attribute_tree_view">
+        <field name="model">product.attribute</field>
+        <field name="inherit_id" ref="product.attribute_tree_view"/>
+        <field name="arch" type="xml">
+            <field name="name" position="before">
+                <field name="sequence" widget="handle"/>
```
bealdav: Please remove this line, because odoo implements it, now
https://github.com/odoo/odoo/blob/9.0/addons/product/product_view.xml#L532

product_attribute_priority/models/product_attribute.py
```diff
@@ -0,0 +1,17 @@
...
+SEQUENCE_ATTR_HELP = SEQUENCE_HELP + " (defined in Product Attribute)"
+
+
+class ProductAttribute(models.Model):
+    _inherit = 'product.attribute'
+    _order = 'sequence, name'
+
+    sequence = fields.Integer(
+        string='Sequence', oldname='priority', default=10,
+        help=SEQUENCE_HELP)
```
bealdav, line 17: Just a remark, nothing to do: Field 'sequence' is defined in Odoo v9, but please let it here for extra args, like oldname, help

product_attribute_priority/views/product_view.xml
bealdav: @lmignon it seems it's the only thing which prevent to merge. Thanks to fix

bealdav:
:+1:

PR #203, opened 2016-11-13
[MIG] 9.0 product_code_builder_sequence 

lmignon:
migration from 8.0
depends of #204

PR #204, opened 2016-11-13
[MIG] 9.0 product_code_builder

lmignon:
migration from 8.0
depends of #203


product_code_builder/__openerp__.py
```diff
@@ -0,0 +1,28 @@
...
+{
+    'name': "Product Code Builder",
+    'author': "Akretion, Odoo Community Association (OCA)",
+    'summary': "Create product references according to attributes",
+    'website': "https://www.akretion.com",
+    'license': 'AGPL-3',
+    'category': 'Product',
+    'version': '9.0.1.0.0',
+    'depends': [
+        'product',
```
bealdav, line 14: only one 1 dependency sufficient ? the last, I think

bealdav:
Thanks for your work @lmignon. 
You have make product_attribute_priority and product_code_builder compatible. 
Ca c'est fait, merci
@EBII please review and test in v9 before to go to v10, thanks

EBII:
seems ok and doing the job

product_code_builder/__openerp__.py
bealdav: Any news @lmignon ? Do you want a PR, there ?

lmignon: @bealdav Sorry for the late reply... Both are required since I extends the views defined in product and therefore I have a direct dependency on product. In case of direct dependency it's always better to avoid to rely on the fact that this dependency is transitively acquired by an other one.

bealdav: ok

product_code_builder/models/product_attribute_value.py
```diff
@@ -0,0 +1,18 @@
+# coding: utf-8
+# © 2015 Benoît GUILLOT <benoit.guillot@akretion.com>
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+
+from openerp import fields, models
+
+
+class ProductAttributeValue(models.Model):
+    _inherit = "product.attribute.value"
+    _code = "code"
```
gdgellatly, line 10: What does this line do? I'm trying to port to 10 but can find no usage of it anywhere in odoo

bealdav: This is a mistake https://github.com/OCA/product-attribute/commit/726c16ec05b0c65d542322fe529e551edf057ab5
it must be _order \= 'code' . Sorry for that

bealdav:
@gdgellatly  Note that product_attribute_priority (dependent module) is ported to v9

gdgellatly:
@bealdav OK I've given it my best shot.  Note in v10 most of the features in product_attribute_priority already exist in product module - see #244

pedrobaeza:
I think this is not needed anymore with the revamp product_variant_default_code in OCA/product-variant. Isn't it?

gdgellatly:
I thought that revamp was v10?

pedrobaeza:
Uhm, right. OK, but I'm seeing this in a place that maybe is not the best, and it requires 3 modules to work. I think it's better to backport the other one.

gdgellatly:
I've long said that these modules don't belong here so no argument from me.
And If someone wants to backport the other one again no argument. It is
more feature complete.

rousseldenis:

PR #212, opened 2016-11-29
Allow the write of default_code if set in vals dict

grindtildeath:
To be consistent with the create method, we shall check if default_code is not defined in the vals dictionary, so that when a default_code is passed to this method, this default_code won't be overwritten with a new sequence number.

### DIFF
--- a/product_sequence/models/product_product.py
+++ b/product_sequence/models/product_product.py
@@ -32,7 +32,9 @@ class ProductProduct(models.Model):
     @api.multi
     def write(self, vals):
         for product in self:
-            if product.default_code in [False, '/']:
+            if (product.default_code in [False, '/'] and
+                    ('default_code' not in vals or
+                     vals['default_code'] == '/')):
                 sequence = self.env.ref('product_sequence.seq_product_auto')
                 vals['default_code'] = sequence.next_by_id()
             super(ProductProduct, product).write(vals)
--- a/product_sequence/models/product_product.py
+++ b/product_sequence/models/product_product.py
@@ -32,8 +32,8 @@ class ProductProduct(models.Model):
     @api.multi
     def write(self, vals):
         for product in self:
-            if (product.default_code in [False, '/'] and
-                        vals.get('default_code', '/') == '/'):
+            if (product.default_code in [False, '/'] and vals.get(
+                    'default_code', '/') == '/'):
                 sequence = self.env.ref('product_sequence.seq_product_auto')
                 vals['default_code'] = sequence.next_by_id()
             super(ProductProduct, product).write(vals)
--- a/product_sequence/models/product_product.py
+++ b/product_sequence/models/product_product.py
@@ -33,8 +33,7 @@ class ProductProduct(models.Model):
     def write(self, vals):
         for product in self:
             if (product.default_code in [False, '/'] and
-                    ('default_code' not in vals or
-                     vals['default_code'] == '/')):
+                        vals.get('default_code', '/') == '/'):
                 sequence = self.env.ref('product_sequence.seq_product_auto')
                 vals['default_code'] = sequence.next_by_id()
             super(ProductProduct, product).write(vals)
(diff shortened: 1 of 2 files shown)


yvaucher (approved):

product_sequence/models/product_product.py
```diff
@@ -32,7 +32,9 @@ def create(self, vals):
     @api.multi
     def write(self, vals):
         for product in self:
-            if product.default_code in [False, '/']:
+            if (product.default_code in [False, '/'] and
+                    ('default_code' not in vals or
```
pedrobaeza: You can reduce this with `vals.get('default_code', '/') != '/'`

grindtildeath: @pedrobaeza 
I rewrote the condition but your example should have "equals to" instead of "not equals to" :

`
>>> def check_default_code(dict):
...     return ('default_code' not in vals or vals['default_code'] == '/')
... 
>>> vals = {}
>>> check_default_code(vals)
True
>>> vals['default_code'] = '/'
>>> check_default_code(vals)
True
>>> vals['default_code'] = '.'
>>> check_default_code(vals)
False
>>> 
>>> def check_revision_default_code(dict):
...     return vals.get('default_code', '/') != '/'
... 
>>> vals = {}
>>> check_revision_default_code(vals)
False
>>> vals['default_code'] = '/'
>>> check_revision_default_code(vals)
False
>>> vals['default_code'] = '.'
>>> check_revision_default_code(vals)
True
>>> 
>>> def check_revision_v2_default_code(dict):
...     return vals.get('default_code', '/') == '/'
... 
>>> vals = {}
>>> check_revision_v2_default_code(vals)
True
>>> vals['default_code'] = '/'
>>> check_revision_v2_default_code(vals)
True
>>> vals['default_code'] = '.'
>>> check_revision_v2_default_code(vals)
False
`

product_sequence/models/product_product.py
```diff
@@ -32,7 +32,8 @@ def create(self, vals):
     @api.multi
     def write(self, vals):
         for product in self:
-            if product.default_code in [False, '/']:
+            if (product.default_code in [False, '/'] and vals.get(
+                    'default_code', '/') == '/'):
```
gdgellatly, line 36: This seems wrong to me.  It seems very unlikely both those conditions would be met except for the very specific behaviour the commit is for.  For example if a default_code already existed and it was removed, this test evaluates False yet in this situation we would expect a pass.  Also if a default_code did not exist, and we were writing something else, say uom_id this test also fails where it would be expected to pass. 

A much simpler way would be to call super first, then run the conditional if with a simple assignment write. e.g.
```python
result = super()
for p in self:
   if p.default_code in {False, '/'):
        <exec code>
        p.default_code = seq.next_by_id()
return result
```
there is an issue with performance, however given it is only ever going to execute that code block once ever on a product in normal usage and usually then one at a time it is acceptable IMO

gdgellatly (changes requested):

product_sequence/models/product_product.py
grindtildeath: @gdgellatly Thanks for your review.

However I disagree with your suggestion :

>  It seems very unlikely both those conditions would be met except for the very specific behaviour the commit is for

The second condition I added is always True if 'default_code' key isn't in vals dict, and as the first condition doesn't change, I don't understand why it seems so unlikely...

>  For example if a default_code already existed and it was removed, this test evaluates False yet in this situation we would expect a pass.

This is already the case actually with this module, my PR doesn't pretend to address this behaviour.

>  Also if a default_code did not exist, and we were writing something else, say uom_id this test also fails where it would be expected to pass.

If we want to write on uom_id it will be done anyway : 
- Should the product already have a default_code, this default_code won't change, and the new uom_id will be written as is the case before the PR. 
- Should the product be without a default_code and the default_code isn't in vals dict, a new one will be written from the sequence as was previously the case with this module too. 

So I don't think it's a great idea to call super first and make another update afterwards. Moreover should you write False on default_code, it would be done through super() but then will be overwritten by the sequence, so it doesn't improve anything.

gdgellatly:
@grindtildeath oops you are right. Having the in statement and get in same
conditional for same field confused me. I'll update when I'm neat a pc.

gdgellatly:
Except the false case isn't covered in vals.get.

gdgellatly (approved):

pedrobaeza (approved):

Issue #215, opened 2016-12-09
RFC: Product features


rubencabrera:
From a client's request, I'm working on a module to add product features in a particular way:

+ Features and products have a M2M relation.
+ If a template has a feature, all products inherit with exactly the same value. 
+ A feature has an image and an alt text. 

I've seen a similar development somewhere but can't find it now, so I'm starting from scratch, defining a new model. 

Is this the right repo to make a PR when I'm done? Should I use some existing model? Product attributes don't fit the purpose, because all variants have the same features as the template. Product categories could do the deed, but they are not meant for this purpose and using them seems confusing.

pedrobaeza:
Why don't you use product_multi_image?

lukebranch:
@rubencabrera ,

Maybe i'm mistaken, but I think the module you're referring to is product_custom_info:

https://github.com/OCA/product-attribute/tree/8.0/product_custom_info

I'm not sure if it covers your exact requirements, however it might be a good start.

I've done a quick and dirty port to version 9.0 and 10.0 for that module and base_custom_info, and can share the code (mostly just view changes and some minor code tweaks to make it compatible with the new API) if it's helpful.

Regarding having an image connected to the feature, I can see how this might be useful (eg. icon representing the feature for display in your website layout), however perhaps product_custom_info could be extended to match your requirements.

pedrobaeza:
product_custom_info is already migrated to 9.0: https://github.com/OCA/product-attribute/pull/177

rubencabrera:
@pedrobaeza I'm already using product_multi_image, but images don't add information about the extra feature.

On the other hand, as @lukebranch said, product_custom_info seems to be on the right track. The only problem is that the user has to manually set a value for the custom property on each product template. Maybe this feature is way too simple for the functionality this module provides. 

I'm using a very similar module previously developed and adapting it to a more general use case hoping it's useful for someone else.

lasley:
@rubencabrera Will the existing product attributes not work? Seems like we're just talking about a Product Template <=> Variant relationship, right? Or are you proposing to apply these attributes at the variant level?

rubencabrera:
Hi @lasley. Features are set at template level and they are common to every variant. Attributes generate combinations where each variant has only one set of values and the rest are excluded. You can check hte PR #216 to see it.

pedrobaeza:
Closing as there's a PR.

PR #216, opened 2016-12-09
ADD product_feature

rubencabrera:
ADD a module to add product features with an image.

Product features are added to the template and common to all variants. The relationship is a M2M and the name of the feature must be unique. 

Related to #215


pedrobaeza:
Looking at the code, this overlaps a lot with _product_custom_info_ module, that only needs to add an image field to their templating structure, so :-1: for this. It also lacks all the OCA guidelines conventions on the code.

rubencabrera:
All the guidelines conventions? Seriously? Give me a hint so I can solve it, because I used another module from the repo as template.

product_custom_info_module is very complete and I tried to demo it to the client so we could just add an image and make it work, but it needs too many steps for such a simple task that the customer asked for a custom development, the result is what I shared, hoping someone would find it useful.

pedrobaeza:
product_custom_info is not complicated. Check OCA guidelines here: https://github.com/OCA/maintainer-tools/blob/master/CONTRIBUTING.md

lasley:
@rubencabrera - a lot of the 8.0 modules don't comply with current guidelines. In the same repo Pedro linked, there is a template module that has some excellent samples in it of essentially anything you need: https://github.com/OCA/maintainer-tools/tree/master/template/module

We keep the samples updated to the guides, so it is much easier than using pre-existing modules that may have aged.

Regarding this module - I dunno if it's necessarily a duplication because we are missing the image feature in the `product_custom_info`. I think though that this might lead to a dangerous data path.

I feel like this data/images are going to be far from Odoo core, so it's always going to be an outlier that has to be designed around & causing kind of an N+1 module proliferation. An example would be integration of this into something like `website_sale`, showing the features in an invoice, etc.

Do you mind explaining a bit of use case for context here?

rubencabrera:
Thanks for your thorough answer, @lasley 

I'll use the templates to check the modules and change wherever I see differences. 

The use case for this module is oriented towards web_sale and mobile apps connecting to odoo. Besides product images, which show different relevant images about each particular product or product variant, some use cases require more general attributes. 

Things like "100% cotton" if we sell clothes, "Water resistant" for watches or "handcrafted" for something else. The image attached to the feature is intended to be shown to customers and maybe some reports, as you suggest. 

The thing about product_custom_info is that we set values per each product, which is not needed for this more straight forward feature idea. Also, when defining a new template for custom info, we can add properties with the same name as existing one(s). I don't think product_custom_info is overcomplicated _per se_, it's just that this use case is way more simple than what it covers. 

Also, I found where I saw this module: at 2 different clients. They already have similar developments from different partners, so I still think someone will find this simple solution suitable for already existing needs.

PR #218, opened 2017-01-02
[9.0][NEW] product_list_price_tax_include: Write price in product template with taxes include

sergio-teruel:
Moved from sale-workflow: https://github.com/OCA/sale-workflow/pull/386

With this module you can write list price with taxes include.
You can write list price and price with tax is calculated or you can write list price with taxes and list price is also calculated.

cc@Tecnativa


rafaelbn (approved):
Tested in runbot

rafaelbn:
@Tecnativa

luismontalba (approved):
Functionality tested on runbot

pedrobaeza (changes requested):
Only one change: name should be product_pricelist_tax_included

cubells (approved):
cc @Tecnativa

renamed module.

code review and tested.

lasley:
Looks like @pedrobaeza's rename comment has been attended to - merging.

PR #221, opened 2017-01-06
[FIX] force rewrite of dimentional_uom_id on created new product

vrenaville:
When you create a product, the dimentional_uom_id must be write on the product

### DIFF
--- a/product_dimension/models/product.py
+++ b/product_dimension/models/product.py
@@ -72,7 +72,8 @@ class ProductTemplate(models.Model):
     @api.model
     def create(self, vals):
         product_template_id = super(ProductTemplate, self).create(vals)
-        # TODO: this is needed to set given values to first variant after creation
+        # TODO: this is needed to set given values to first variant
+        # after creation
         # these fields should be moved to product as lead to confusion
         if vals.get('dimensional_uom_id'):
             related_vals = {}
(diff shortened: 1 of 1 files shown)


leemannd (approved):
Tests could be improved.

pedrobaeza:
And the TODO seems to be a very big issue...

leemannd:
@pedrobaeza The comment and technique used by @vrenaville is from odoo's core https://github.com/odoo/odoo/blob/9.0/addons/product/product.py#L679

pedrobaeza:
Yeah, but that doesn't mean that is OK IMO. You know that Odoo doesn't do several things the right way, and the TODO is because of something. You can avoid it in this case I think.

leemannd:
You're right, sayin it's in core isn't a suitable answer. Currently, this PR does the trick and without introducing something wrong. For the moment, I don't have the time to investigate further. :(

guewen:
@pedrobaeza do you have a suggestion?

pedrobaeza:
There are several:

- Remove field from template.
- Launch the write on the variants, not the rewrite of the same field. Anyway, I see no code for rewriting the field on the variants, so is this really working?

rousseldenis:

PR #222, opened 2017-01-12
[9.0][ADD] product_pricelist_simulation: Show price in product for all pricelists

sergio-teruel:
cc @Tecnativa


product_pricelist_simulation/models/product.py
```diff
@@ -0,0 +1,38 @@
...
+        <div class="col-md-%s">
+            <div class="col-md-12"><strong>%s</strong></div>
+            <div class="col-md-12">%s %s</div>
+        </div>
+        """ % (columns, pricelist.name, pricelist.currency_id.symbol, price)
+        return render
+
+    @api.multi
+    def _compute_pricelist_simulate(self):
+        self.ensure_one()
```
pedrobaeza: Make it multi record aware

pedrobaeza (changes requested):

bealdav:
I like the idea of your module, nice, but I think it could be better (for performance with really big db) to have access these infos with 'more options' button than directly in the form.
Maybe it could be a choice settings ?

Thanks a lot to Tecnativa one more time

sergio-teruel:
@bealdav Thanks for yor feedback, I have on mind this improvement... I have to finish tests, and I will do more changes...
OK?? :smile:

pedrobaeza:
Closing as no more interest in this

PR #224, opened 2017-01-17
[ADD][10.0] Product Pricelist Tier

woodbrettm:
Hi there,

**This module extends product pricelist items with a new, tiered type**. 

As seen below, the user can set a specified cost ($400) when a customer purchases the minimum quantity (20) of products. The price discount field (20%) is then automatically calculated (though can still be edited if desired), and will be applicable on orders of minimum quantity items or more.

Features from the formula option were made visible as well.

Thanks!
Brett

capture

@lasley 
@tedsalmon


lasley:
@bmw95 - the price discount does not update on runbot

product_pricelist_tier/models/product_pricelist_item.py
```diff
@@ -0,0 +1,69 @@
...
+        string='Tiered Price',
+        digits=dp.get_precision('Product Price'),
+        help='The desired price at minimum quantity you specify. '
+        'e.g. You want 10 bowls to cost a total of $20. '
+        'Set this field to 20, and set minimum quantity to 10. The '
+        'price discount will automatically be calculated for you.',
+        store=True,
+    )
+
+    @api.onchange('compute_price')
```
lasley: This should also be onchange `tiered_price` - you are using it in the calculations

woodbrettm:
@lasley fixed. I added a post init hook to trigger the initial calculations.

lasley:
Why is a hook needed? Initial calculations for what? This module adds its own assets and needs to modify nothing pre-existing.

woodbrettm:
@lasley apologies, was a bit vague. When I checked the previous runbot build before updating, the price discount was automatically being updated as expected when I made changes to the minimum quantity or tiered price, however it was not being initially calculated because the demo data was not triggering the onchange. I added a post init hook to fix that.

lasley:
@bmw95 - this still doesn't work & it doesn't have anything to do with the demo data. Create a pricelist item for any product variant, set the tiered price to anything, and the discount remains unchanged.

woodbrettm:
@lasley fixed. The dev ticket requested a constraint be added such that the pricelist item can only be applied on a product, however the constraint would only run when you save the pricelist item, and then click save on the pricelist form (found that pretty odd). Adjusted so you're notified in the pricelist item form directly.

Fixed and squashed, pending any other comments.

lasley:
Instead of an error, why don't you just hide "Tiered Price" when it's not applicable? Seems more intuitive IMO. The constraint is simply to make sure that we don't save bad data, and should be kept.

```
attrs="{'invisible': [domain_for_invisibility]}"
```

woodbrettm:
@lasley I think I know what you're getting at. I made the Applied On fields invisible when pricing is tiered (and set applied_on automatically to product). The only validation error that will pop up now is if the user sets a min quantity of 0 or None. Is that what you're looking for?
capture

lasley:
As mentioned in my comment, the constraint should have been kept. 

You also did the invisibility backwards. If anything other than "Product" is selected, "Tiered Price" should not be a visible option. You shouldn't remove the "Apply On" when "Tiered Price" is selected.

woodbrettm:
Got it, that makes more sense now. Will update with a fix in a sec.

lasley:
@bmw95 - Ok so yeah it's not possible to make part of the select invisible, sorry for suggesting that avenue - completely spaced that detail 😆 

Ok so the new strategy should be to add an onchange method for the "Apply On" & "Tiered Price" col. Basically the same logic you have in the constraint - if Apply On isn't Product & "Tiered Price" has something in it, return a warning.

You could probably cheat by moving the verification logic out into its own method, then calling from the constraint and the onchange.

Onchange can return a warning key to do what you want - https://www.odoo.com/forum/how-to/developers-13/what-should-onchange-methods-do-and-return-57760

product_pricelist_tier/views/product_pricelist_item_view.xml
```diff
@@ -0,0 +1,35 @@
...
+<odoo>
+
+    <record id="product_pricelist_item_form_view" model="ir.ui.view">
+        <field name="name">product.pricelist.form</field>
+        <field name="model">product.pricelist.item</field>
+        <field name="inherit_id" ref="product.product_pricelist_item_form_view" />
+        <field name="arch" type="xml">
+            <xpath expr="//field[@name='applied_on']" position="attributes">
+                <attribute name="attrs">
+                    {'invisible': [('compute_price', '==', 'tiered')]}
```
tedsalmon: `s/==/=`

product_pricelist_tier/views/product_pricelist_item_view.xml
```diff
@@ -0,0 +1,35 @@
...
+                </div>
+            </xpath>
+            <xpath expr="//group[field[@name='base']]" position="attributes">
+                <attribute name="attrs">
+                    {'invisible': [('compute_price', '!=', 'tiered'), ('compute_price', '!=', 'formula')]}
+                </attribute>
+            </xpath>
+            <xpath expr="//field[@name='base']" position="attributes">
+                <attribute name="attrs">
+                    {'invisible': [('compute_price', '==', 'tiered')]}
```
tedsalmon: `s/==/=`

tedsalmon (commented):
I'll wait for an explicit approval once you've made the changes @lasley requested. For now, here's my review.

woodbrettm:
@lasley @tedsalmon Updated

woodbrettm:
Should this module depend on product instead of sale? I noticed other modules had put sale, so I did that as well, however I'm pretty sure product meets the requirements for this module.

lasley:
Product contains everything we need as you noted. We should definitely try to minimize dependencies as much as possible.

woodbrettm:
Updated.

tedsalmon (approved):
Changes LGTM and are functional 👍 

One thing though, If it isn't more difficult to add, could you set the `Price` column on create/write when setting the tiered price? The column doesn't update until the price list is saved, which threw me off and made me think I was doing something wrong.

screen shot 2017-01-30 at 15 54 26

Thanks!

lasley:
I would bet that there's an onchange that needs to be called manually.

woodbrettm:
@tedsalmon @lasley This issue happens with all of the pricelist item options, however as mentioned, only the Price column doesn't update. My best guess would be to add an `@api.depends('item_ids.price')` or something like that to the pricelist model.

However since this is not restricted to the scope of pricelist_tiers, just wondering whether you want me to go ahead and solve this here, or whether you have another solution in mind?

bealdav:
Hi @bmw95 any news on this interesting PR ?

woodbrettm:
Hi @bealdav the module is finished, however there were issues mentioned when saving the pricelist item and the one2many tree 'price' column in the pricelist form view not updating until you click save in the pricelist form. Like so:
1. Click pricelists (advanced formula selected in settings view)
2. Click 'Public Pricelist'
3. Click 'Edit'
4. Click 'All Products'
5. Ensure 'Compute Price' set to 'Formula'
6. Change 'Price Discount' field to '20'
7. 'Save'
8. Notice the 'Price' column in the one2many tree view does not update until you click 'Save' in the pricelist form view.

This issue affects all of the compute price choices, so I felt fixing that issue here is not particularly ideal. As per my previous comment, pending insight from @lasley as to whether I should still go ahead and fix the issue here or not.

lasley:
@bmw95 - don't worry about it if the built-ins are functioning the same way.

woodbrettm:
This is good to merge unless any comments.

lasley (approved):

lasley:
Would a PSC mind reviewing please?

PR #233, opened 2017-02-20
[8.0] fix sequence for pricelist items in pricelist_per_product module

chafique-delli:
In sale order line, we want to use as a priority the product pricelist item associated with the product variant.But in Odoo, it's the product pricelist item which has the smallest sequence that is used.
That's why, I put a smaller sequence to the product pricelist item of the product variant and a larger sequence to the product pricelist item of the product template.

### DIFF
--- a/pricelist_per_product/pricelist.py
+++ b/pricelist_per_product/pricelist.py
@@ -93,7 +93,7 @@ class ProductPricelistItem(models.Model):
         }
 
     @api.model
-    def update_sequence_price_grid(self, vals):
+    def _get_sequence_price_grid(self, vals):
         """
             In sale order line, we want to use as a priority
             the product pricelist item associated with the product variant.
@@ -105,18 +105,20 @@ class ProductPricelistItem(models.Model):
         """
         product_id = vals.get('product_id', self.product_id)
         product_tmpl_id = vals.get('product_tmpl_id', self.product_tmpl_id)
+        related_sequence = 15
         if product_id:
-            vals['related_sequence'] = 5
+            related_sequence = 5
         elif product_tmpl_id:
-            vals['related_sequence'] = 10
+            related_sequence = 10
+        return related_sequence
 
     @api.model
     def create(self, vals):
         if self.env['product.pricelist.version'].browse(
                 vals['price_version_id']).price_grid:
-            self.update_sequence_price_grid(vals)
             vals.update({
                 'price_discount': -1,
+                'related_sequence': self._get_sequence_price_grid(vals),
                 'base': 1,
             })
         return super(ProductPricelistItem, self).create(vals)
@@ -125,6 +127,7 @@ class ProductPricelistItem(models.Model):
     def write(self, vals):
         for item in self:
             if 'product_id' in vals:
-                item.update_sequence_price_grid(vals)
+                vals['related_sequence'] = item._get_sequence_price_grid(
+                    vals)
             super(ProductPricelistItem, item).write(vals)
         return True
--- a/pricelist_per_product/pricelist.py
+++ b/pricelist_per_product/pricelist.py
@@ -72,7 +72,9 @@ class ProductPricelistItem(models.Model):
         readonly=True)
     related_sequence = fields.Integer(
         String='Sequence',
-        related="sequence")
+        related="sequence",
+        help="Allows to modify the sequence manually because "
+             "the sequence field is difficult to modify because 'handle'.")
 
     @api.multi
     def button_product(self):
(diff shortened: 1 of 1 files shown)


pricelist_per_product/pricelist.py
```diff
@@ -91,12 +93,39 @@ def button_product(self):
...
+        elif (vals.get('product_tmpl_id') and vals.get('related_sequence') and
+              vals['related_sequence'] != 10):
+            vals['related_sequence'] = 10
+        return vals
+
+    @api.model
     def create(self, vals):
         if self.env['product.pricelist.version'].browse(
                 vals['price_version_id']).price_grid:
+            vals = self.update_sequence(vals)
```
bealdav: left member vals useless

pricelist_per_product/pricelist.py
```diff
@@ -91,12 +93,39 @@ def button_product(self):
         }
 
     @api.model
+    def update_sequence(self, vals):
+        if (vals.get('product_id') and vals.get('related_sequence') and
+                vals['related_sequence'] != 5):
```
bealdav: s / vals.get('related_sequence') and vals['related_sequence'] != 5 / vals.get('related_sequence')!= 5

pricelist_per_product/pricelist.py
```diff
@@ -91,12 +93,39 @@ def button_product(self):
...
+            vals['related_sequence'] = 5
+        elif (vals.get('product_tmpl_id') and vals.get('related_sequence') and
+              vals['related_sequence'] != 10):
```
bealdav: idem upper

pricelist_per_product/pricelist.py
```diff
@@ -91,12 +93,39 @@ def button_product(self):
...
-                'sequence': 1,
                 'base': 1,
             })
         return super(ProductPricelistItem, self).create(vals)
+
+    @api.multi
+    def write(self, vals):
+        for item in self:
+            values = {}
+            if (not vals.get('product_id') and item.product_id and
```
bealdav: it could be easier if you write: `product_id = vals.get('product_id', item.product_id)` instead to evaluate `if (not vals.get('product_id') and item.product_id ...`. Idem for product_tmpl_id and sequence
Then you can compare product_id and product_tmpl_id directly in their future values

pricelist_per_product/pricelist.py
```diff
@@ -91,12 +93,39 @@ def button_product(self):
         }
 
     @api.model
+    def update_sequence(self, vals):
```
bealdav: 1/ You should explain why this method with a docstring
2/ method name could be in collision with some method in other module: what do you think of `update_sequence_according_to_product_or_product_tmpl` or something like that

bealdav (changes requested):

pricelist_per_product/pricelist.py
```diff
@@ -72,7 +72,9 @@ class ProductPricelistItem(models.Model):
         readonly=True)
     related_sequence = fields.Integer(
         String='Sequence',
-        related="sequence")
+        related="sequence",
+        help="Allows to modify the sequence manually because "
+             "the sequence field is difficult to modify because 'handle'.")
```
bealdav: Why 2 commits for these 2 things
You could squash ?

bealdav:
Tested with UI, it works. :+1:

bealdav:
ping @lasley could you review please ?

pricelist_per_product/pricelist.py
```diff
@@ -91,12 +93,38 @@ def button_product(self):
         }
 
     @api.model
+    def update_sequence_price_grid(self, vals):
```
lasley: Hmmm I'm not so sure how I feel about this method mutating the `vals`. Can we instead return a dict `{related_sequence: #}`, then use the return of this method to update the vals dict? eg
```
vals.update(
    self.update_sequence_price_grid(vals)
)
```

I know it seems backwards sort of, but I had to read both code block a few times in order to identify what was happening, which is a sign of too much complexity & coupling. This method doesn't actually own `vals`, so mutating it isn't necessarily the best course of action IMO.

chafique-delli:
@lasley, thank you for the review.
I fixed my code after your comment.

chafique-delli:
you can squash, when merge, thanks

lasley (approved):

lasley:
@tedsalmon can you please make sure to include this in #189

bealdav:
:+1:

andhit-r (approved):
:+1:

Issue #184, opened 2016-10-06
Migration to version 10.0


pedrobaeza:
# Todo

https://github.com/OCA/maintainer-tools/wiki/Migration-to-version-10.0
# Modules to migrate
- [x] ~~base_custom_attributes~~ : deprecated in 9.0 branch in favor of product_custom_info
- [ ] customer_context_in_product_view
- [ ] partner_custom_attributes 
- [ ] product_attribute_multi_type
- [x] product_brand - by @loxamir #191
- [ ] product_catalog_report
- [ ] product_categ_attributes
- [ ] product_custom_attributes
- [x] product_custom_info - By @fanha99 
- [X] ~~product_customer_code - By @pcatinean - #236~~ Deprecated for `product_supplierinfo_for_customer`
- [x] product_dimension - ~~By @gurneyalex #195~~ By @thomaspaulb - #491 
- [ ] product_electronic
- [ ] product_gift
- [ ] product_gtin
- [ ] product_icecat
- [ ] product_index
- [ ] product_listprice_upgrade
- [ ] product_lot_foundry
- [x] product_manufacturer - By @bistaray - #188 
- [x] product_multi_category - By @angelmoya - #242 
- [ ] product_multi_price
- [ ] product_pricelist_fixed_price
- [ ] product_prices_on_variant
- [ ] product_qt
- [ ] product_quick_stock_rule
- [ ] product_search_reference
- [x] product_sequence - by @sodexisteam #194 
- [ ] product_series
- [ ] product_size
- [X] ~~product_supplierinfo_tree_price_info~~ Not needed anymore as the structure has changed.
- [ ] product_supplierinfo_for_customer - @aheficent - #292
- [ ] product_supplierinfo_for_customer_sale - @aheficent https://github.com/OCA/product-attribute/pull/280#issuecomment-316748306
- [x] product_uom - By @lasley - #213 
- [ ] product_weight
- [ ] production_lot_custom_attributes

yajo:
@pedrobaeza I think you got confused:

> [x]  product_custom_info - By @Yajo - #177

#177 is for v9

pedrobaeza:
Right, totally. Fixed.

ysantiago:
@pedrobaeza What about module product_supplierinfo_for_customer?

pedrobaeza:
The module was not in the repository before. When decided what to do, it can be added here.

AaronHForgeFlow:
[...]

PR #234, opened 2017-02-27
Update product_dimension

andreampiovesana:


lasley:
Thanks for the submission @andreampiovesana. Looks like this module isn't installable, so I don't think the migration is complete. Please set the installable key to True in the manifest.

I was going to reference you to the Migration docs in the wiki, but it looks like that specific step isn't there. I created https://github.com/OCA/maintainer-tools/issues/273 for that & will review this once it's installable.

gdgellatly:
should be closed in favour of #195

lasley:
Indeed this is a dup, closing. 

@andreampiovesana - if ever in question as to whether something has been upgraded and is just pending review, I recommend you take a look at the issues section of the repo for one titled "Migration to version X" - where X is the version you want (v10 for this repo is #184)

PR #235, opened 2017-02-27
Update product_brand.py

andreampiovesana:


lasley (approved):
Simple enough, thanks @andreampiovesana 

I'm going to just merge this because it's been sitting so long & is definitely not going to cause an issue

PR #122, opened 2015-12-23
Porting 8.0 product_customer_code

hurrinico:
I've made a fork of https://github.com/OCA/product-attribute/pull/38 and rebase it

PR #236, opened 2017-02-27
[MGR] Migrated product_customer_code to version 10

PCatinean:


pedrobaeza:
Please check Travis status

PCatinean:
Fixes to pass all tests added

cvinh:
fix issue with product search (https://github.com/PCatinean/product-attribute/pull/1)

PCatinean:
@invitu thanks for the fix. I also thought about what was the intention, showing only the code, only the product, or a mix of both (I would have wanted to go this route).

But it's up for debate I guess

PCatinean:
I think it needs product_code.mapped('product').name_get()?

cvinh:
BTW, I think it would be useful to show both product and customer_code in display_name (in sale_order for example)
something like [default_code] Product (customer_code)

PCatinean:
@invitu I thought of the exact same pattern. If others agree I will make it like that np

lukebranch:
I believe something similar exists (unported from 7 or 8):

https://github.com/OCA/sale-workflow/blob/10.0/product_customer_code_sale/__manifest__.py

Not exactly the same as you described but quite close.

cvinh:
I think that we don't need product_customer_code_sale anymore if we append customer_code by overwriting ProductProduct.name_get()
@PCatinean do you agree ?

pedrobaeza:
Well, you still need to get somehow (with the context) the customer involved, so a minimum change in sale is needed.

cvinh:
@pedrobaeza self._context.get('partner_id') in ProductProduct does the job doesn't it ?

PCatinean:
@invitu it does not, that implies the partner_id has been sent via the context which does not happen by default.

One semi-solution would be to browse the active_id (whichever model it is) and scan for relational fields towards res.partner and use that (if there is just one or such). But this approach is messy and can have bad consequences

cvinh:
I just saw that another PR is pending here https://github.com/OCA/product-attribute/pull/122

@pedrobaeza suggested to use product_supplierinfo_for_customer https://github.com/OCA/product-attribute/pull/122#issuecomment-172921499 in order not to create a new model and it never happened

Another PR for product_customer_code_sale is here https://github.com/OCA/sale-workflow/pull/345

cvinh:
@PCatinean I think that we only need customer code in a sale_order context so partner_id can be retrieved. In other case, we don't need customer_code, so it's not a problem if we don't have it
I can find the same code in your PR (line 30 in product.py):
`    @api.model`
`     def name_search(self, name='', args=None, operator='ilike', limit=100):`
`         res = super(ProductProduct, self).name_search(`
`             name=name, args=args, operator=operator, limit=limit)`
`         if not res:`
`             partner_id = self._context.get('partner_id')`

cvinh:
@PCatinean could you check https://github.com/PCatinean/product-attribute/pull/4 ? It seems to work for me...

LoisRForgeFlow:
Runbot is not working even after rebuilt  :confused:

eLBati:
@PCatinean I would rename the fields as defined in https://github.com/OCA/product-attribute/pull/122, for instance
product_customer_codes --> product_customer_code_ids
product --> product_id

We can make a PR

PCatinean:
@eLBati I followed the contributing guideline from OCA: https://github.com/OCA/maintainer-tools/blob/master/CONTRIBUTING.md#variable-names

If this is required/needed by OCA maintainers then I will rename of course.

What else is needed to merge this PR?

eLBati:
@PCatinean variable names are different from field names, see
https://github.com/OCA/maintainer-tools/blob/master/CONTRIBUTING.md#fields

PCatinean:
@eLBati right you are, my bad I mixed thing up.

Updating now

product_customer_code/models/product_customer_code.py
```diff
@@ -0,0 +1,58 @@
...
+
+    company_id = fields.Many2one(
+        comodel_name='res.company',
+        string='Company',
+        required=False,
+        default=lambda self: self.env['res.company']._company_default_get(),
+    )
+
+    _sql_constraints = [
+        ('unique_code', 'unique(product_code,company,partner)',
```
LoisRForgeFlow: Check the runbot warning related to this: 
```
WARNING openerp_test odoo.models.schema: Table 'product_customer_code': unable to add constraint 'unique(product_code,company,partner)'!
If you want to have it, you should update the records and execute manually:
```

product_customer_code/__manifest__.py
```diff
@@ -8,53 +8,40 @@
 ############################################################################
 #    Coded by: el_rodo_1 (rodo@vauxoo.com)
 ############################################################################
```
lasley: Please use [this header](https://github.com/OCA/maintainer-tools/blob/master/template/module/__openerp__.py#L3)

product_customer_code/__manifest__.py
```diff
@@ -8,53 +8,40 @@
...
+    'version': '10.0.1.0.0',
     "author": "Vauxoo,Odoo Community Association (OCA)",
     "website": "http://www.vauxoo.com/",
     "license": "AGPL-3",
     "category": "Generic Modules/Product",
     "summary": "Add many Customers' Codes in product",
     "depends": [
-            "base",
-            "product",
+        "base",
```
lasley: No need for `base`

product_customer_code/__manifest__.py
```diff
@@ -8,53 +8,40 @@
...
+        "product",
     ],
     "description": """
 Customer' codes in product
```
lasley: This should be in a [ReadMe](https://github.com/OCA/maintainer-tools/blob/master/template/module/README.rst)

product_customer_code/models/product.py
```diff
@@ -0,0 +1,39 @@
+# -*- coding: utf-8 -*-
+
+###########################################################################
```
lasley: Short header

product_customer_code/models/product_customer_code.py
```diff
@@ -0,0 +1,58 @@
+# -*- coding: utf-8 -*-
+###########################################################################
```
lasley: Short header

product_customer_code/models/product.py
```diff
@@ -0,0 +1,39 @@
...
+#    Module Writen to OpenERP, Open Source Management Solution
+#
+#    Copyright (c) 2012 Vauxoo - http://www.vauxoo.com
+#    All Rights Reserved.
+#    info@vauxoo.com
+############################################################################
+#    Coded by: Rodo (rodo@vauxoo.com),Moy (moylop260@vauxoo.com)
+############################################################################
+
+from odoo import models, fields, api
```
lasley: Alphabetize imports

product_customer_code/models/product_customer_code.py
```diff
@@ -0,0 +1,58 @@
...
+#    Module Writen to OpenERP, Open Source Management Solution
+#
+#    Copyright (c) 2012 Vauxoo - http://www.vauxoo.com
+#    All Rights Reserved.
+#    info@vauxoo.com
+############################################################################
+#    Coded by: Rodo (rodo@vauxoo.com),Moy (moylop260@vauxoo.com)
+############################################################################
+
+from odoo import models, fields
```
lasley: Alphabetize imports

product_customer_code/models/product_customer_code.py
```diff
@@ -0,0 +1,58 @@
...
+class ProductCustomerCode(models.Model):
+    _name = "product.customer.code"
+    _description = "Add multiple product customer codes"
+
+    _rec_name = 'product_code'
+
+    product_code = fields.Char(
+        string='Customer Product Code',
+        required=True,
+        help="""This customer's product code will be used when searching into
```
lasley: This is adding a bunch of whitespace into the string. Please start and end the lines as so:
```
    help="This customer's product code will be used when searching into"
         "a request for quotation."
```

Hint in a lot of IDEs, you can simply hit "Return" while in a string to do this automagically

product_customer_code/models/product_customer_code.py
```diff
@@ -0,0 +1,58 @@
...
+        required=False,
+        default=lambda self: self.env['res.company']._company_default_get(),
+    )
+
+    _sql_constraints = [
+        ('unique_code', 'unique(product_code,company_id,partner_id)',
+         'Product customer code must be unique'),
+    ]
+
+    # TODO: Add index to product_code, partner
```
lasley: Please move this to roadmap in readme

product_customer_code/models/res_partner.py
```diff
@@ -0,0 +1,13 @@
+# -*- coding: utf-8 -*-
+
+from odoo import models, fields
```
lasley: Alpha

product_customer_code/models/res_partner.py
```diff
@@ -0,0 +1,13 @@
...
+
+
+class ResPartner(models.Model):
+    _inherit = 'res.partner'
+
+    product_code_ids = fields.One2many(
+        comodel_name='product.customer.code',
+        inverse_name='partner_id',
+        string='Products'
```
lasley: Please add a trailing comma

product_customer_code/security/product_customer_code_security.xml
```diff
@@ -1,14 +1,14 @@
 <?xml version='1.0' encoding='UTF-8'?>
-<openerp>
+<odoo>
```
lasley: `<odoo noupdate="1">` & can remove the data tag

lasley (changes requested):
Thanks @PCatinean - commentsinline

hurrinico:
@PCatinean functionally looks good, 
Once you fix @lasley request is 👍 for me

product_customer_code/models/product_customer_code.py
gdgellatly: Or just add the index.  That said an index on product_code is not required as a btree index on something that is usually ilike text searched is pointless.  The unique constraint has already created a composite index anyway.  Personally I'd just add index=True to partner_id and delete the TODO.

product_customer_code/views/product_customer_code_view.xml
```diff
@@ -0,0 +1,59 @@
...
+                    <filter string="Company" icon="terp-partner" domain="[]" context="{'group_by':'company'}"/>
+                    <filter string="Product" icon="terp-accessories-archiver" domain="[]" context="{'group_by':'product'}"/>
+                    <filter string="Partner" icon="terp-partner" domain="[]" context="{'group_by':'partner'}"/>
+                    <separator orientation="vertical"/>
+                </group>
+            </search>
+        </field>
+    </record>
+
+     <!-- <menuitem action="action_product_customer_code_tree" id="menu_product_customer_code" sequence="50" parent="product.menu_root" groups="group_product_customer_code_manager"/> -->
```
gdgellatly, line 57: Any reason this wasn't ported?

product_customer_code/models/product.py
```diff
@@ -0,0 +1,39 @@
...
+    @api.model
+    def name_search(self, name='', args=None, operator='ilike', limit=100):
+        res = super(ProductProduct, self).name_search(
+            name=name, args=args, operator=operator, limit=limit)
+        if not res:
+            partner_id = self._context.get('partner_id')
+            if partner_id:
+                product_customer_code_obj = self.env['product.customer.code']
+                product_codes = product_customer_code_obj.search([
+                    ('product_code', '=', name),
```
gdgellatly: This is the part I hate. Second guessing rationale especially when its just a port but here goes.  Anything you agree with I'd add to roadmap rather than amend seeing it is a port.

Do we really want an '=' operator here, why not just operator?  I guess if the use case is importing customer orders which have complete codes it makes sense.
Why do we only care if not res?  Why not len(res) < limit then a comp in return to remove dupes. It seems very feasible that at some point an indeterminate number of customer codes will overlap a normal code (especially if not using '=') 
Why partner_id specifically? Why not parents, children of partner?

Also wouldn't it have slightly odd results if for example an order line (be it sale, purchase, invoice, mrp) happened to have the context key partner_id as product name_get returns [default_code] name (from memory), and this returns customer code.

As I say, just questions, not criticisms just trying to functionally understand.

I've read through the views to understand better and I think I get it.  This module is never intended to be extended into the sale realm or used during a standard business document workflow, rather just display master data on demand given a customers own code.  In which case I guess all of the above comments would form a TODO for future use in sale orders etc.

gdgellatly (changes requested):
Hi Paul,

I've also just run this through a functional review/manual test.

Review setup - 
add security groups to admin
add a product customer code to some CPU - ABCDE for agrolait. (I also did ABCDEF for an ipad)
create a sale order - type ABCDE - the entry is there - it doesn't look right and should follow the format of product name get but wait for it.  It returns the Server consulting product and on save descriptions all convert to it.
Next Test. Order for Agrolait, Michael Fletcher - type ABCDE - nothing.
Next Test, Vendor Bills - behave identically as above, now should product customer codes be in vendor bills?

Something is not right with this module.  http://3292092-236-117ed7.runbot1.odoo-community.org/web#view_type=form&model=sale.order&menu_id=175&action=240

Also the product view options are only available if using variants.  There is no way to set on templates if using basic functionality.  But functionally there are a few issues here I think, and it comes down to solely relying on that name_get function and partner_id context key to do the heavy lifting.  It is mangling up the ids in the return and also the name_get is not consistent with product.

I think it really needs to search the associated table and return the product_ids and code and then reformat to essentially [p.customer_code] p.name or if it isn't meant to be used in sales or purchases then it needs a more unque context key.

product_customer_code/views/product_customer_code_view.xml
PCatinean: The module depends only on product. There seems to be no standard menu for products in V10 (that I could find with a very quick look) except in Sales which the module does not hold a dependency over.

product_customer_code/models/product.py
PCatinean: I do agree that showing results based on both product name / default_code and product_code would be more flexible. Otherwise the scenario of "search by code only when by name / default_code fails" is in effect.

If by some chance the name/default_code of a product is the same with a product_code you cannot select it.

Searching by child_of the commercial_partner_id is not a bad idea.

Maybe someone else can pitch in to this and offer an opinion

lasley: IMO these are the types of questions that need to be asked during migration, because this is the point where we have the most flexibility in the way that a module operates. Once we release stable, changing things becomes a lot more difficult.

I totally agree with using `operator` and that we should eval for equal length of res instead of just an either or. For the little bit of effort this would take, this would increase the functionality of this module quite a bit outside of its intended - which is good IMO as long as we don't sacrifice anything in the process.

I think that we may be getting a bit liberal with it by searching children though, because then the module's functionality would be changed drastically without the ability to revert. 

Consider a giant conglomerate with multiple business entities under it. These entities could all have separate product codes, but we would get an aggregate if searching the children without a way to filter away from these results. 

On the flip-side, current operation we just have to search more codes if we wanted some sort of aggregate report. I could also be misunderstanding this side of the proposal though, so please let me know if I was just talking stupid 😉

gdgellatly: @lasley we have the identical use case, except mine is real.  We deal with many national entities - think McDonalds, where there are literally hundreds of sub-entities that all use the exact same product code but are different entities.  However I think the entire argument can be alleviated to cover both use cases simply by
```
'|', ('partner_id', '=' partner.id),('partner_id','=', partner.commercial_id.id)
```
Then update README with expected behaviour.  That actually doesn't get my use case, because we also bill these sub-entities but we wouldn't use this module either.

I would need (pseudocode)
```
('partner_id', 'in' [partner.id, partner.commercial_partner_id.id, partner.parent_id.id,partner.commercial_partner_id.parent_id.id])
```
which I think is also reasonable.  In practice, its unlikely for any given customer entity that they would mix behaviours and this covers where product codes are set at a higher level.  We get individual, entity, parent entity in even reasonably complex setups.

I agree searching children is wrong, was just typing out loud at the time.

hurrinico:
@PCatinean @gdgellatly 
I've made [this](https://github.com/PCatinean/product-attribute/pull/5) in order to fix name search return and logics

PCatinean:
@hurrinico thanks for the PR, i was getting lost in too many tasks and got distracted from this one.

hurrinico:
About the view, IMO the best way is to move everything in product_customer_code_sale because module product doesn't introduce any menu.
We are going to port https://github.com/OCA/sale-workflow/tree/8.0/product_customer_code_sale in a few days and our intent is to move all the views in that module if it's ok

product_customer_code/models/product.py
```diff
@@ -0,0 +1,31 @@
...
+            name=name, args=args, operator=operator, limit=limit)
+        if (len(res) < limit):
+            partner_id = self._context.get('partner_id')
+            if partner_id:
+                product_customer_code_obj = self.env['product.customer.code']
+                product_codes = product_customer_code_obj.search([
+                    ('product_code', operator, name),
+                    ('partner_id', '=', partner_id)
+                ], limit=limit)
+                res.extend(product_codes.mapped('product_id').name_get())
```
gdgellatly, line 30: Is this what we want?  The behaviour before this commit was the customer code.  Now I assume it is the [product_code] product name if the display_default_code context key is set.  

Would something like 
```python
product_name_get = product_codes.mapped('product_id').with_context(display_default_code=False).name_get()
if self._context.get('display_default_code'):
    product_name_get = [(pk, '[%s] %s' % (product_codes.filtered(lambda r: r.product_id = pk).product_code, name) for pk, name in product_name_get]
res.extend(product_name_get)
```
or adjusting name_get be better.  Its not great code but hopefully it gives an idea what I mean.  Its just from a user perspective I think this will be confusing.  Type a product code and get a different one back.  Note I've not looked at product_customer_code_sale so maybe its handled there better.

gdgellatly:
@hurrinico @PCatinean @lasley also see #271 . IMHO this is why we need PSC input.  We can review at a 'tactical' level but miss the overall fit.  If we have product_customer_code, product_customer_code_sale and product_customer_supplierinfo_code(!?) all doing roughly the same thing, really we need to just say this one is it, it lives in this repo, we take these features from this one etc etc and stub the others.  Sure such rationalisation introduces some migration pain for existing users, but ultimately everyone gets a clearer idea of which module is the one to use, work on.

Also @hurrinico I made a usability comment on your name_get change you may wish to review.

product_customer_code/models/product.py
lasley: @gdgellatly - This code looks functionally equivalent to the previously merged version IMO, just using the new API. For ease, the previous version:

``` python
               id_prod_code = \		
                   product_customer_code_obj.search(cr, user,		
                                                    [('product_code',		
                                                      '=', name),		
                                                     ('partner_id', '=',		
                                                      partner_id)],		
                                                    limit=limit,		
                                                    context=context)		
               # TODO: Search for product customer name		
               id_prod = id_prod_code and product_customer_code_obj.browse(		
                   cr, user, id_prod_code, context=context) or []		
               for ppu in id_prod:		
                   ids.append(ppu.product_id.id)		
           if ids:		
               res = self.name_get(cr, user, ids, context)
```

product_customer_code/views/product_customer_code_view.xml
lasley: Please remove the commented action then. We can redeclare the menu when needed, as is customary in the v10 (sale, purchase, inventory, etc. all define product menus)

lasley:
I'm not familiar with `product_customer_supplierinfo_code`, but `product_customer_code` and `product_customer_code_sale` should remain independent modules IMO. Aren't the _x modules just binding on the appropriate context?

That said, it does need a rename - `sale_product_customer_code`

gdgellatly:
@lasley no the context is already bound.  In fact in my functional review I noted that the context is potentially problematic as it is set in views other than sale related ones.  partner_id is a very common context key.  All customer_code_sale does is add the column to sale order line, which I guess was kind of my point about the name_get return.  I don't really care, but it seems pointless to have a module which stores customer partcodes, then another to display them.  What other module would make use of such a base module?  Why would you not just have as a security option like uom or discount?  But in any case the main thrust was the fact that we have PR's for 2 modules doing the same thing more or less being the supplierinfo one and this.

eLBati:
@gdgellatly @lasley I don't know about ```product_customer_supplierinfo_code```, but ```product_customer_code_sale``` extends ```product_customer_code``` only adding and displaying ```product_customer_code``` on ```sale.order.line``` (we could also add that field to sale order report)

Recap on the main functionalities
 - ```product_customer_code``` allows to specify product codes depending on customers: every customer can have their product code, if needed
 - ```product_customer_code_sale``` extends ```product_customer_code``` to display product customer code on sale order line

About menu
```
<menuitem action="action_product_customer_code_tree" id="menu_product_customer_code" sequence="50" parent="product.menu_root" groups="group_product_customer_code_manager"/>
```
we would move it to ```product_customer_code_sale``` because ```product_customer_code``` depends on ```product``` only and ```product``` does not introduce menu items

> That said, it does need a rename - sale_product_customer_code

Agree

eLBati:
@gdgellatly I can't see the problem with ```partner_id``` in context: if the key is there, we assume that the current partner is partner_id, thus we search products  by customer code also; if it's not, standard search is performed.
Note that ```partner_id``` is only involved by ```name_search``` method

gdgellatly:
@eLBati @lasley I wasn't overly clear.  The 2 modules do roughly the same as the supplierinfo_customer modules.  As I say I don't  care whether you keep these as one or two, but its hard to converse if no one even wants to look at the other module to see if they are functionally similar or the same.  See https://github.com/OCA/product-attribute/tree/8.0/product_supplierinfo_for_customer and #271 

The potential problem with partner_id in context is every other model that isn't customer related.  For example, purchase orders.  In the context of purchases, is the partner a partner, a customer or a supplier?  Should customer codes be searched on purchases?  What if supplierinfo is also set? If functionally the answer is yes, no problem, it was just to me when I saw customer_code I guess I expected it used only in customer related contexts.  What about manufacturing, pickings etc?

lasley:
@gdgellatly - I see your point. Maybe what we need to do instead is abstract the terminology into more of a partner code instead. 

In the context of sales, it's a customer code. In the context of purchases, it's a vendor code. In the context of manufacturing, maybe a supplier code or something. 

Basically we're trying to maintain a code for Product A against Partners A, B, and C regardless of context. Then whatever glue modules simply add the context.

Am I oversimplifying this?

gdgellatly:
@lasley It wasn't what I was thinking but yes.  Ultimately there is a need to maintain partners codes and they wouldn't differ depending on the users context.  That said, I do wonder how big the case is that a vendor buys its own products from its customers, and those codes happen to overlap with internal codes, and those codes would be stored.  Only time I really see it is when vendor runs out of stock and needs to supply so rounds it up from their customers.  Maybe it doesn't matter, not every possible corner case or strange implementation is able to be mitigated.  My point was more was that using a very common context key for a specific purpose in a very common ORM method may have unintended side effects.  More a warning, than a recommendation.

However, I still urge you to look at the supplierinfo_for_customer module.  Having looked at both in detail now, functionally I think they serve almost identical use cases (in fact their readme's are nearly the same), but supplierinfo is very DRY in its approach and would additionally support pricing, qty breaks, descriptions.  I guess its full featured, enhancing the existing model, whereas this adds customer codes as a seperate model to emulate some of supplierinfo model but not all.  Not saying either approach is better or worse, just we potentially have two functionally very similar modules with very different technical approaches in the same repo.

eLBati:
@gdgellatly I looked at ```product_supplierinfo_for_customer```. It provides the same basic functionality of ```product_customer_code``` (defining product codes depending on customers) plus it allows to define prices and pricelists based on them.
Also, it is already available for 8.0, while, for ```product_customer_code```, we have an [open PR](https://github.com/OCA/product-attribute/pull/122).

I think we should drop ```product_customer_code``` in favour of ```product_supplierinfo_for_customer``` and add extensions of it where needed (sale/purchase orders, reports...)

AaronHForgeFlow:
Hi @PCatinean! Thank you for this proposal. In my opinion it makes sense to keep both modules, this one and the proposed here: https://github.com/OCA/product-attribute/pull/122 That one can be an extension of the  product_customer_code. What do you think @eLBati ?

eLBati:
Hi @aheficent, the module proposed in this PR and the one proposed in PR https://github.com/OCA/product-attribute/pull/122 are the same module.

Instead, ```product_supplierinfo_for_customer``` is at https://github.com/OCA/product-attribute/tree/8.0/product_supplierinfo_for_customer

AaronHForgeFlow:
@eLBati Yes, it's the same module name, but as you said, it allows to define prices and pricelists based on them. So it adds some extra features that wasn't there before. If you think that the approach in your PR is correct, then the PR in v10 should also incorporate those additional features.

BTW, I'm planning to add this to v9.

eLBati:
> Yes, it's the same module name, but as you said, it allows to define prices and pricelists based on them

No, ```product_customer_code``` does not.
```product_supplierinfo_for_customer``` does

AaronHForgeFlow:
@eLBati so I guess both modules cannot coexist, right?

eLBati:
@aheficent they overlap providing the same functionality "defining product codes depending on customers". This is why I propose to drop ```product_customer_code```

AaronHForgeFlow:
@eLBati All right, as product_supplierinfo_for_customer is already in OCA/8.0  I agree to keep using that module

AaronHForgeFlow:
We need to decide which module is going to be ported either this or product_supplierinfo_for_customer. @eLBati @PCatinean

PCatinean:
@aheficent I was not sure what to say, there a quite a few differences of opinion. Can't say I have a definitive one myself. If further changes are needed to the module I'll be glad to carry them out.

Regarding organisation of modules in OCA I don't have such a broad perspective as others to decide

[ EDIT ]
Maybe @lasley can have say as he is PSC?

lasley:
Take this for a grain of salt, because I don't use either of these. I looked at the two and agree that one should be deprecated.

I am thinking that `product_supplierinfo_for_customer` should be the preferred here due to more features + it already being in OCA/8. Looks like there are some other  product_supplierinfo* modules too, so it wins out in the realm of more integrated as well.

gdgellatly:
@lasley @PCatinean @aheficent 

Sorry, I didn't really mean to kick all this off, thought it would take a couple of days to sort through, not months. Just saw 2 modules with largely the same functionality and thought 1 would be better that combines best of both. I'm not invested either way as like @lasley I don't use either, but if I did it would be the product_supplierinfo_for_customer module.

PCatinean:
Well then it's settled, the other module will be used and this will be dropped.

lasley:
@gdgellatly - no need to be sorry, these types of discussions 100% have to happen. Sometimes it takes a while because we're all volunteers here, but it's all for the greater goal of having the ultra-supreme modules as the ones we support. Nobody said ultra-supreme was easy 😉

PR #240, opened 2017-03-15
[Mig 10] migration product profile 

EBII:
from 8.0 to V10 with test module
with oca and author


legalsylvain:
Hi @EBII. Could you make a PR cherry-picking V8 module ? It will be more easy to review.
kind regards.
Thanks for having ported this module.

PR #241, opened 2017-03-15
[MIG] 10 Product profile et example

EBII:


product_profile/models/product.py
```diff
@@ -0,0 +1,286 @@
+# coding: utf-8
+# © 2015 David BEAL @ Akretion
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+
+import logging
+from odoo import models, fields, api, _
+from odoo.osv import orm
+from odoo.exceptions import Warning as UserError
```
Garamotte: `from odoo.exceptions import UserError`

product_profile/models/product.py
```diff
@@ -0,0 +1,286 @@
...
+             "(not synchronized with product.template fields)")
+    sequence = fields.Integer(
+        help="Defines the order of the entries of profile_id field\n"
+             "(not synchronized with product.template fields)")
+    explanation = fields.Text(
+        required=True,
+        oldname='description',
+        help="An explanation on the selected profile\n"
+             "(not synchronized with product.template fields)")
+    type = fields.Selection(
```
Garamotte, line 59: Please don't use python's reserved names as field names.

product_profile/models/product.py
```diff
@@ -0,0 +1,286 @@
...
+        selection='_get_types',
```
Garamotte: New API provides a `selection_add` attribute, so you don't need to use a method to allow other modules to extend the selection list. You can directly put `selection=[('consu', 'Consumable'), ('service', 'Service')]` here instead.

product_profile/models/product.py
```diff
@@ -0,0 +1,286 @@
...
+        res = super(ProductProfile, self).write(new_vals)
+        if new_vals:
+            for rec in self:
+                products = self.env['product.product'].search(
+                    [('profile_id', '=', rec.id)])
+                if products:
+                    _logger.info(
+                        " >>> %s Products updating after updated '%s' "
+                        "product profile" % (len(products), rec.name))
+                    products.write({'profile_id': rec.id})
```
Garamotte: I don't understand this.
You write `profile_id = rec.id` only on records that already have this value ? Or I miss something ?

product_profile/models/product.py
```diff
@@ -0,0 +1,286 @@
...
+                    elm = etree.Element(
+                        'filter', **self._customize_profile_filters(my_filter))
+                    node[0].addprevious(elm)
+                res['arch'] = etree.tostring(doc, pretty_print=True)
+        return res
+
+    @api.model
+    def _get_profiles_to_filter(self):
+        """ Inherit if you want that some profiles doesn't have a filter """
+        return [(x.id, x.name) for x in self.env['product.profile'].search([])]
```
Garamotte, line 231: This is exactly `return self.name_get()`, isn't it ?

product_profile/security/group.xml
```diff
@@ -0,0 +1,11 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<openerp>
+    <data>
```
Garamotte: Replace `<openerp>` by `<odoo>` and remove `<data>`.

product_profile/views/config_view.xml
```diff
@@ -0,0 +1,18 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<openerp>
+    <data>
```
Garamotte: Replace `<openerp>` by `<odoo>` and remove `<data>`.

product_profile/views/product_view.xml
```diff
@@ -0,0 +1,53 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<openerp>
+    <data>
```
Garamotte: Replace `<openerp>` by `<odoo>` and remove `<data>`.

product_profile_example/__openerp__.py
```diff
@@ -0,0 +1,27 @@
+# coding: utf-8
```
Garamotte, line 1: Remove this file.

product_profile_example/product_product_demo.xml
```diff
@@ -0,0 +1,24 @@
+<?xml version="1.0"?>
+<odoo>
+    <data noupdate="1">
```
Garamotte: `<odoo noupdate="1">`

product_profile_example/profile.py
```diff
@@ -0,0 +1,31 @@
...
+# © 2015 David BEAL @ Akretion
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+
+from odoo import models, fields
+
+
+class ProductProfile(models.Model):
+    _inherit = 'product.profile'
+
+    def _get_types(self):
```
Garamotte: Override the field using `selection_add=[('product', 'Stockable Product')]` when you remove this method in the other module.

Garamotte (changes requested):
Could you add some unit tests on the `product_profile` module ?

product_profile/models/product.py
EBII: hello, 
the unit test exist in product_profile_example to test product_profile
please use it

product_profile/models/product.py
EBII: yes sure but is the odoo fields name : product.template.type so we use it.

product_profile/models/product.py
EBII: it is to force product update after  profile change
see next commit for update code

product_profile/models/product.py
EBII: but not the same model

Garamotte: Oh sorry, I knew I missed something.
So : `return self.env['product.profile'].search([]).name_get()`.

product_profile/models/product.py
EBII: yes sure but is the odoo fields name : product.template.type so we use it.

product_profile/models/product.py
EBII: i remove it : not necessary

product_profile/models/product.py
```diff
@@ -0,0 +1,269 @@
...
+                new_vals.pop(key)
+        # super call must be after check_useless_key_in_vals() call
+        # because we compare value before and after write
+        res = super(ProductProfile, self).write(new_vals)
+        if new_vals:
+            for rec in self:
+                products = self.env['product.product'].search(
+                    [('profile_id', '=', rec.id)])
+                data = products._get_profile_data({'profile_id': rec.id})
+                products.write(data)
```
bealdav, line 83: missing logger before write https://github.com/OCA/product-attribute/commit/56995f899884de7b1d41ffe68140beca1bd38c75#diff-6564fbea77f6e37ef19b6a92abeb4a40R92

EBII:
fail rebase remove PR to new one

PR #242, opened 2017-03-21
MIG 10.0 product_multi_category

angelmoya:


product_multi_category/README.rst
```diff
@@ -38,7 +31,7 @@ help us smashing it by providing a detailed and welcomed `feedback
 <https://github.com/OCA/
 product-attribute/issues/new?body=module:%20
 product_multi_category%0Aversion:%20
-9.0%0A%0A**Steps%20to%20reproduce**%0A-%20...%0A%0A**Current%20behavior**%0A%0A**Expected%20behavior**>`_.
+10.0%0A%0A**Steps%20to%20reproduce**%0A-%20...%0A%0A**Current%20behavior**%0A%0A**Expected%20behavior**>`_.
```
pedrobaeza: Use short version

pedrobaeza (approved):
Minor comment

product_multi_category/README.rst
angelmoya: What do you mean with shot version? 10 instead of 10.0 ??

pedrobaeza: Check current README template: https://raw.githubusercontent.com/OCA/maintainer-tools/master/template/module/README.rst. I mean the short link.

angelmoya:
@pedrobaeza fixed and squashed commits

florian-dacosta (approved):

Cedric-Pigeon (approved):

Cedric-Pigeon:
@pedrobaeza I think this one could be merged ;-)

Cedric-Pigeon:
@pedrobaeza Thanks but I just noticed  that setup was missing so I created a new PR https://github.com/OCA/product-attribute/pull/260

dreambits:
I tried using this one on nightly build of odoo 10 but it failed to find **categ_id** field.
I realized that categ_id field is added by stock module and so lower priority of view in here is causing it to not find the field.
So I changed it to 17 for now but what should be the recommended priority for this view to avoid side effects with other customization ?

P.s. Tried running this module in runbot and saw that **product_template_form_view** view itself has categ_id field so maybe this has been added after recent nighly build. Need to dig deeper.

P.p.s So I found this and it may have been merged after nightly: https://github.com/odoo/odoo/commit/8e5684250bbab7e18d46bf13fb512003602a9791

Issue #227, opened 2017-01-27
[RFC] product_default_image: Default images by category


lasley:
This is a fairly simple module that will provide a default image for products, similar to the partner avatars. This image will be configurable on both a global & a product category level.

A view will be added in the Product Configuration menu, which will allow for the configuration of the images and the categories.

The create method of `product.template` will be overridden. If the `image` key is not present, the default image will be generated based on category or global.

Anyone have any thoughts or pre-existing strategies/modules that we can base from?

pedrobaeza:
Two notes:

* Default image for categories should be set on each category.
* Bring default image on default_get for the product.template model, not on write/create.

lasley:
> Default image for categories should be set on each category.

Such as in the category form view, instead of a separate view? Yeah that's probably a better direction.

> Bring default image on default_get for the product.template model, not on write/create.

TIL `default_get`. Fully agree.

pedrobaeza:
> Such as in the category form view, instead of a separate view? Yeah that's probably a better direction.

That's my idea, yes.

lasley:
Closing to track in PR #243

PR #243, opened 2017-03-27
[10.0][ADD] product_default_image

woodbrettm:
Hi there,

This module is in response to #227.


product_default_image/models/product_template.py
```diff
@@ -0,0 +1,78 @@
...
+
+    _inherit = 'product.template'
+
+    @api.model
+    def default_get(self, vals):
+        vals = super(ProductTemplate, self).default_get(vals)
+
+        if vals.get('categ_id'):
+
+            company = self.env.user.company_id
```
woodbrettm: So I functionally tested the module in runbot and it works. However with the tests, I'm having an issue with this line here.

If I change self.env.user.company_id in the tests, logs show in the test environment the company is correctly changed (logged in the failure message of a test), however this company here is still left unchanged. 

I'm trying to get it so this company instead of being 'MyCompany', is now my own test data company.

Was wondering if someone could point out how I could properly do that when running tests.

Thanks!

product_default_image/views/product_category_view.xml
```diff
@@ -0,0 +1,21 @@
...
+
+<odoo>
+
+    <record id="product_category_form_view" model="ir.ui.view">
+        <field name="name">product.category.form</field>
+        <field name="model">product.category</field>
+        <field name="inherit_id" ref="product.product_category_form_view" />
+        <field name="arch" type="xml">
+            <xpath expr="//div[contains(@class, 'oe_title')]" position="before">
+		        <field name="image_medium"
```
tedsalmon: Using tabs instead of spaces

product_default_image/views/stock_config_settings_view.xml
```diff
@@ -0,0 +1,29 @@
...
+        <field name="name">stock settings</field>
+        <field name="model">stock.config.settings</field>
+        <field name="inherit_id" ref="stock.view_stock_config_settings" />
+        <field name="arch" type="xml">
+            <xpath expr="//header" position="inside">
+                <field name="company_id" invisible="1" />
+            </xpath>
+            <xpath expr="//field[@name='group_stock_packaging']" position="after">
+                <field name="product_image_target" widget="radio" />
+		        <field name="product_default_global_image"
```
tedsalmon: Tabs instead of four spaces

product_default_image/models/product_template.py
tedsalmon: So you're trying to return a specific company Record from `self.env.user.company_id`?

In your tests, you could do the following:
```
self.env.user.company_id = self.env['res.company'].browse(your_other_company_id)
```
It should be temporary, so you shouldn't have to reset the company_id before the next test fires off.

product_default_image/tests/test_setup.py
```diff
@@ -0,0 +1,67 @@
...
+            'name': 'Test Category',
+        })
+        self.category_2 = self.category_mod.create({
+            'name': 'Test Category',
+        })
+        self.company_1 = self.company_mod.create({
+            'name': 'Test Company',
+            'currency': self.currency_usd.id,
+        })
+        # self.user_1 = self.user_mod.create({
```
tedsalmon: Remove or use this code ;)

tedsalmon (changes requested):
Your build is failing and you have numerous warnings about invalid fields being used while creating certain objects. Please address these issues and I'll update my review :)

product_default_image/models/product_template.py
woodbrettm: Perhaps the reason why there's issues is that I'm trying to match the user with test data and not demo data. 
I was doing `self.env.user.company_id = self.company_1` where self.company_1 is a test company i created in the test_setup file.

product_default_image/views/stock_config_settings_view.xml
woodbrettm: interesting. My code editor is showing spaces with those two fields properly aligned. Looks off here though.

product_default_image/models/product_template.py
tedsalmon: As long as `self.company_1` is a Record of type `res.company`, it should work. *Should*

lasley: `self.env.user.company_id = self.company_1.id`

woodbrettm: Dang. Adding `.id` to it didn't work. And adding a demo company did not work either.

product_default_image/tests/test_setup.py
```diff
@@ -0,0 +1,67 @@
...
+
+        self.category_1 = self.category_mod.create({
+            'name': 'Test Category',
+        })
+        self.category_2 = self.category_mod.create({
+            'name': 'Test Category',
+        })
+        self.company_1 = self.company_mod.create({
+            'name': 'Test Company',
+            'currency': self.currency_usd.id,
```
woodbrettm: currency_id (causing the travis warnings)

product_default_image/models/product_template.py
woodbrettm: Ah figured it out. Companies have a `user_ids` field that only accepts users in that field.

product_default_image/hooks.py
```diff
@@ -0,0 +1,14 @@
...
+
+from odoo import api, SUPERUSER_ID
+
+
+def _update_prevent_categ_change_image(cr, registry):
+    with cr.savepoint():
+        env = api.Environment(cr, SUPERUSER_ID, {})
+        templ_mod = env['product.template']
+        templates = templ_mod.search([])
+        for template in templates:
```
lasley: `templates.write({`

product_default_image/models/product_template.py
```diff
@@ -0,0 +1,79 @@
...
+
+from odoo import api, fields, models, tools
+
+
+class ProductTemplate(models.Model):
+
+    _inherit = 'product.template'
+
+    prevent_categ_change_image = fields.Boolean(
+        help='Used to prevent automatic changes to image fields when categ_id '
```
lasley: I'm unclear on this. Why would we ever want the image fields to change, outside of the default_get?

product_default_image/models/product_template.py
```diff
@@ -0,0 +1,79 @@
...
+    @api.model
+    def create(self, vals):
+        vals['prevent_categ_change_image'] = True
+        return super(ProductTemplate, self).create(vals)
+
+    @api.model
+    def default_get(self, vals):
+        vals = super(ProductTemplate, self).default_get(vals)
+
+        if vals.get('categ_id'):
```
lasley: invert this logic to reduce nesting.

```
if not vals.get('categ_id'):
    return vals
```

product_default_image/models/product_template.py
```diff
@@ -0,0 +1,79 @@
...
+                vals['image_small'] = categ.image_small
+
+            elif target == 'global':
+                vals['image'] = company.product_default_global_image
+                tools.image_resize_images(vals)
+
+        return vals
+
+    @api.onchange('categ_id')
+    def _change_image_onchange_categ_id(self):
```
lasley: It's implicit, but prefer the explicit `@api.multi` please.

lasley: s/_change_image_onchange_categ_id/_onchange_categ_id

This also should be above the CRUD overrides per contribution docs

product_default_image/models/product_template.py
```diff
@@ -0,0 +1,79 @@
...
+        """ Checks if the user has uploaded a custom image or not. """
+
+        company = self.env.user.company_id
+        target = company.product_image_target
+
+        if target == 'category' and not self.prevent_categ_change_image:
```
lasley:
```
if target != 'category' or self.prevent_categ_change_image: 
    return
```

product_default_image/models/product_template.py
```diff
@@ -0,0 +1,79 @@
...
+                vals['image'] = company.product_default_global_image
+                tools.image_resize_images(vals)
+
+        return vals
+
+    @api.onchange('categ_id')
+    def _change_image_onchange_categ_id(self):
+        """ Checks if the user has uploaded a custom image or not. """
+
+        company = self.env.user.company_id
```
lasley: This isn't used until after condition thus is unnecessary assignment in some instances

product_default_image/models/product_template.py
```diff
@@ -0,0 +1,79 @@
...
+            image_fields = {
+                'image': self.image,
+                'image_medium': self.image_medium,
+                'image_small': self.image_small,
+            }
+
+            matched_categs = []
+            empty_image_field = []
+
+            for key, value in image_fields.iteritems():
```
lasley: What is the purpose of this (L66-74)?

product_default_image/models/res_company.py
```diff
@@ -0,0 +1,26 @@
...
+
+class ResCompany(models.Model):
+
+    _inherit = 'res.company'
+
+    product_image_target = fields.Selection(
+        string='Default Product Image',
+        selection=[
+            ('global', 'Global'),
+            ('category', "Category's Image"),
```
lasley: If category is chosen and the category doesn't have an image, does the global get used?

lasley (changes requested):

product_default_image/models/product_template.py
woodbrettm: Well for example, if default_get is only used, the default category on products is 'all', so the image for 'all' will load when going to create a product. However, what if I want the category to be something else and not the image for 'all' every time?

In the form I would simply change the category, but the image would not change with it. IMO that's a very poor experience, so I implemented the onchange method below.

This field prevents the image onchange from happening on products that have been created, as I found it problematic on existing products. So the image will only change with the category when creating a new product (unless the user has uploaded a custom image).

It would probably be nice to have on existing products, but I felt that was too far out of scope atm and too risky to work on.

It looks like atm the onchange is still happening on existing products in runbot, so that will need to be fixed unless you have something else in mind. Kind of weird this field is True in the db but False in the onchange.

product_default_image/models/res_company.py
woodbrettm: not atm no. Perhaps we should add a `both` or `global_category` selection?

product_default_image/models/product_template.py
woodbrettm: hmm. I felt doing this was cleaner however, as in the if statement i'd have to do:
`if self.env.user.company_id.product_image_target != 'category' or record.prevent_categ_change_image`

instead of `if target != 'category' or record.prevent_categ_change_image:`

I guess the extra length isn't too bad on the if statement. Would you like me to still remove the company assignment?

bmw95: It does clutter things up a fair amount though

woodbrettm:
@lasley this module works as expected now. Ready for next round of review.

tedsalmon (approved):
Much better - LGTM! 👍

product_default_image/models/stock_config_settings.py
```diff
@@ -0,0 +1,22 @@
...
+    _inherit = 'stock.config.settings'
+
+    product_image_target = fields.Selection(
+        string='Default Product Image',
+        related='company_id.product_image_target',
+        required=True,
+    )
+    product_default_global_image = fields.Binary(
+        string='Default Global Image',
+        required=True,
```
woodbrettm: Remove

product_default_image/models/product_template.py
```diff
@@ -0,0 +1,100 @@
...
+            * image_medium matches any category's image_medium field
+            * image_small matches any category's image_small field
+
+        image, image_medium, and image_small fields are all compared against
+        categories because, for example, in certain cases the image field
+        is None, but the image_medium field has data.
+        """
+        for record in self:
+
+            if not self.env.user.company_id.product_image_target == 'category':
```
woodbrettm: instead of `not`, probs better to do `!=` here, as it's easier to read.

lasley: This isn't using the record, so should be evaluated before the loop. It would result in the loop simply iterating through itself if the condition is true.

product_default_image/models/product_template.py
```diff
@@ -0,0 +1,100 @@
...
+            * image, image_medium, and image_small are all empty
+            * image matches any category's image field
+            * image_medium matches any category's image_medium field
+            * image_small matches any category's image_small field
+
+        image, image_medium, and image_small fields are all compared against
+        categories because, for example, in certain cases the image field
+        is None, but the image_medium field has data.
+        """
+        for record in self:
```
lasley: `self.filtered(lambda r: r.categ_change_image)`

product_default_image/models/product_template.py
```diff
@@ -0,0 +1,100 @@
...
+
+            if not self.env.user.company_id.product_image_target == 'category':
+                continue
+
+            if not record.categ_change_image:
```
lasley: Can remove this after above suggestion

product_default_image/models/product_template.py
```diff
@@ -0,0 +1,100 @@
...
+                continue
+
+            categ_mod = self.env['product.category']
+            categories = categ_mod.search([])
```
lasley: Searching for everything like this does not scale.

product_default_image/models/product_template.py
lasley: The line numbers have changed, but this question still stands. What exactly is happening here?

product_default_image/models/product_template.py
```diff
@@ -0,0 +1,100 @@
...
+                )
+
+            if matched_categs or len(empty_image_field) == 3:
+                record.image = record.categ_id.image
+                record.image_medium = record.categ_id.image_medium
+                record.image_small = record.categ_id.image_small
+
+    @api.model
+    def create(self, vals):
+        vals['categ_change_image'] = False
```
lasley: Why not just a default?

lasley (changes requested):

product_default_image/models/product_template.py
woodbrettm: oh nice, great idea.

product_default_image/models/product_template.py
woodbrettm: This block is matching up any of the image fields of the product with any category image fields.

If there's a match, it is assumed the user has not uploaded a custom image, and when they change the category, the image will change to the new category's image.

This block also detects if the product's image, image_medium, and image_small fields are None, in which case the image will also change when switching categories.

product_default_image/models/product_template.py
woodbrettm: This still should be False, as it's set to True in default_get (although I should probs just set `default=True` right on the field.

This just prevents the product's image from being changed when the user changes the categ_id after the product has been created.

product_default_image/models/product_template.py
woodbrettm: As per your comment below, searching all the categories and then comparing every time during onchange is not particularly scalable, so perhaps there's a better way to do this.

woodbrettm:
@lasley @tedsalmon This module has been rewritten and is ready for review.

On a side note, when you look through product.template you'll notice the following block around line 245:
```python
for img_bg, records in write_map.iteritems():

    for record in records:
        record.write({
            'image': img_bg,
            'img_type': to_type,
        })
    # records.write({
    #     'image': img_bg,
    #     'img_type': to_type,
    # })
```
Interestingly, there's an issue here where if I just write on the recordset, I verified that `img_type` (img_type might be like `no_image` or `default_category` image) is properly applied on all the images, however the correct `image` is sometimes only applied to a few of the records.

Even weirder, looping through the records and calling write on each record works. All the records have the correct image and to_type value when I do that. Could be something in `product` or in `stock`. Haven't quite checked yet but curious if either of you have experienced that issue before.

Other than this performance issue, the module is ready. Also, atm, `img_type` is still visible in the product template form views for debugging purposes. When approved I'll move that up into the form header.

Thanks!
Brett

product_default_image/models/product_category.py
```diff
@@ -0,0 +1,77 @@
...
+    def write(self, vals):
+
+        tools.image_resize_images(vals)
+        target = self.env.user.company_id.product_image_target
+
+        if target not in ['category', 'global_category']:
+            return super(ProductCategory, self).write(vals)
+
+        keys = ['image', 'image_medium', 'image_small']
+        imgs_present = [vals[key] for key in keys if key in vals]
```
lasley: Using an array is pretty dirty here IMO. Better is a dict:
```
imgs_present = {key: vals[key] for key in keys if key in vals}
```
Which would make the currently obfuscated `imgs_present[0]` into a readable `imgs_present['image']`

product_default_image/models/product_category.py
```diff
@@ -0,0 +1,77 @@
...
+
+        if not imgs_present:
+            return super(ProductCategory, self).write(vals)
```
lasley: Untested

product_default_image/models/product_category.py
```diff
@@ -0,0 +1,77 @@
...
+        }
+
+        if target == 'category' and not imgs_present[0]:
+            img_args.update({
+                'to_type': 'no_image',
+                'to_img_bg': None,
+            })
+
+        if target == 'global_category':
+            img_args['from_types'] += ['default_global']
```
lasley: `img_args['from_types'].append('default_global')`

product_default_image/models/product_template.py
```diff
@@ -0,0 +1,299 @@
...
+
+    _inherit = 'product.template'
+
+    auto_change_image = fields.Boolean(
+        string='Auto Change Image',
+        default=True,
+        help='Allow/disallow automatic changes to product image. '
+             'Uncheck to stop being changed to a default image.',
+    )
+    img_type = fields.Selection(
```
lasley: Do not abbreviate. Use the extra two bytes and spell `image` fully everywhere

product_default_image/models/product_template.py
```diff
@@ -0,0 +1,299 @@
...
+        target = company.product_image_target
+
+        if target in ['category', 'global_category'] and vals.get('categ_id'):
+            categ = self.env['product.category'].browse(vals['categ_id'])
+            vals.update({
+                'image': categ.image,
+                'img_type': 'default_category',
+            })
+
+        if target in ['global', 'global_category'] and not vals.get('image'):
```
lasley: You're repeating the `if target in` logic.

product_default_image/models/product_template.py
```diff
@@ -0,0 +1,299 @@
...
+            vals.update({
+                'image': company.product_image,
+                'img_type': 'default_global',
+            })
+
+        tools.image_resize_images(vals)
+        return vals
+
+    @api.model
+    def search_change_images(self, from_types, to_type, to_img_bg=None,
```
lasley: This method naming is unclear and does not clearly indicate record mutation. 

You should also add a doc block.

product_default_image/models/product_template.py
```diff
@@ -0,0 +1,299 @@
...
+            img_args['to_img_bg'] = to_img_bg
+
+        return templates.change_template_image(**img_args)
+
+    @api.model
+    def search_by_img_types(self, from_types, add_domain=None):
+        """ Constructs domain and searches using from_types, add_domain
+
+            Args:
+                from_types:
```
lasley: `Args:
    from_types (list): Selections from...
    add_domain (bool): `True` to add domain...
...
Example:
.... The note about what the example explicitly returns should be in the example ...
Returns:
    ProductTemplate: Recordset using a domain constructed of....
```

File: product_default_image/models/product_template.py [hunk 27]
```diff
@@ -0,0 +1,299 @@
...
+
+        if add_domain and not isinstance(add_domain, list):
+            raise TypeError('add_domain argument must be a list')
+
+        domain = [
+            ('company_id', '=', self.env.user.company_id.id),
+            ('auto_change_image', '=', True),
+            ('img_type', 'in', from_types),
+        ]
+        domain += add_domain
```
lasley: The domain part should be generated in a separate method in order to allow for modularity. On this note, the `add_domain` would likely become irrelevant - which I think you added in order to achieve the desired modularity?

product_default_image/models/product_template.py
```diff
@@ -0,0 +1,299 @@
...
+        return self.search(domain)
+
+    @api.multi
+    @api.onchange('categ_id')
+    def _onchange_categ_id(self):
+
+        company = self.env.user.company_id
+        target = company.product_image_target
+
+        if target not in ['category', 'global_category']:
```
lasley: `['category', 'global_category']` should be a class constant, or this check centralized.

product_default_image/models/product_template.py
```diff
@@ -0,0 +1,299 @@
...
+                continue
+
+            categ = record.categ_id
+
+            img_args = {
+                'to_type': 'default_category',
+                'in_cache': True,
+            }
+            if not categ.image and target == 'global_category':
+                img_args['to_type'] = 'default_global'
```
lasley: All these strings are incredibly important from what I can tell, but are being redeclared all over (`global_category`, `default_global`, etc). They should be constants in order to avoid mistyping as well as to allow quick future modification.

product_default_image/models/product_template.py
```diff
@@ -0,0 +1,299 @@
...
+                    in cache (helpful when calling from an onchange method
+                    in product.template)
+
+            Returns:
+                recordset if tmpl images have been changed.
+
+                None/False if any of below conditions are True.
+
+        """
+        if any([
```
lasley: `self.ensure_one`

product_default_image/models/product_template.py
```diff
@@ -0,0 +1,299 @@
...
+
+            record.change_template_image(**img_args)
+
+    @api.multi
+    def change_template_image(self, to_type, to_img_bg=None, in_cache=False):
+        """ Change product template image to specified to_type's image
+            or to_img_bg.
+
+            Args:
+                to_type:
```
lasley: https://sphinxcontrib-napoleon.readthedocs.io/en/latest/example_google.html

product_default_image/models/product_template.py
```diff
@@ -0,0 +1,299 @@
...
+        if any([
+                to_type not in [
+                    'default_global', 'default_category',
+                    'no_image', 'custom'],
+                to_type == 'custom' and not to_img_bg,
+                to_type == 'no_image' and to_img_bg,
+                not isinstance(in_cache, bool)]):
+            return
+
+        # {img_bg: records}
```
lasley: ?

product_default_image/models/product_template.py
```diff
@@ -0,0 +1,299 @@
...
+        write_map = defaultdict(lambda: self.env['product.template'].browse())
```
lasley: Lambda is superfluous

```
defaultdict(self.env['product.template'].browse)
```

product_default_image/models/product_template.py
```diff
@@ -0,0 +1,299 @@
...
+
+        if to_img_bg:
```
lasley: Hmmm I don't like this method at all, but I don't want to make you rewrite it either because I think this may be somewhat of a logic bomb.

@tedsalmon can you please make this more elegant?

product_default_image/models/product_template.py
```diff
@@ -0,0 +1,299 @@
...
+            'category': 'default_category',
+            'global_category': 'default_category',
+        }
+
+        if target not in to_keys:
+            return
+
+        if target == 'global_category':
+            categ = self.categ_id
+            if not any([categ.image, categ.image_medium, categ.image_small]):
```
lasley: Should use tuples instead of arrays in these `any`s. Tuples are significantly lower overhead because they're immutable. We don't want mutable, so that's fine here.

product_default_image/models/product_template.py
```diff
@@ -0,0 +1,299 @@
...
+
+        self.change_template_image(
+            to_type=to_keys[target],
+        )
+
+    @api.multi
+    def write(self, vals):
+
+        keys = ['image', 'image_medium', 'image_small']
+        imgs_present = [vals[key] for key in keys if key in vals]
```
lasley: This should probably be a helper method to generate

product_default_image/models/res_company.py
```diff
@@ -0,0 +1,131 @@
...
+                'product_image': some_image,
+                'product_image_target': 'global',
+            })
+
+            # 4
+            company.product_image = some_image
+            company.product_image_target = 'global'
+
+        """
+        target = vals.get('product_image_target')
```
lasley: @tedsalmon simplify this one while you're here too please

product_default_image/README.rst
```diff
@@ -0,0 +1,118 @@
...
+help us smash it by providing detailed and welcomed feedback.
+
+Credits
+=======
+
+Images
+------
+
+* Odoo Community Association: `Icon <https://github.com/OCA/maintainer-tools/blob/master/template/module/static/description/icon.svg>`_.
+
```
lasley: What about the glob_prod_img attribution?

lasley (changes requested):
Thanks @bmw95 - comments inline.

Also, please prefix the majority of your methods with `_`. There is not a reason for most of these to be exposed via public interface IMO

product_default_image/models/product_template.py
woodbrettm: This method is for a recordset so `self.ensure_one()` should not apply.

product_default_image/models/product_template.py
woodbrettm: Do you want this removed? I just put the comment as a reminder of what the structure of the dict is.

product_default_image/models/product_template.py
woodbrettm: Oh interesting.

product_default_image/models/product_template.py
woodbrettm: When should lambda be used? I noticed a lot of times it is used and sometimes it isn't.

product_default_image/models/product_template.py
woodbrettm: I'm good with either. Is it the amount of elifs? From my original code I reduced the amount by a huge margin, but perhaps for the next refactor a dict map could be made? Or is it setting the image as the key to the dict that's the issue?

product_default_image/models/product_template.py
woodbrettm: hmm yeah this code is used I think 2 or 3 times across 2 classes, so should I create a separate file with this method so I can import where needed?

product_default_image/models/res_company.py
```diff
@@ -0,0 +1,131 @@
...
+            'from_types': [
+                'no_image', 'default_global', 'default_category'
+            ],
+            'to_type': target_map[target],
+            'to_img_bg': None,
+            'in_cache': False,
+            'add_domain': None,
+        }
+
+        if 'product_image' in vals:
```
woodbrettm: `if 'product_image' in vals and target not in ['category', 'none']:`

product_default_image/models/res_company.py
woodbrettm: Just as a note, often `stock.config.settings` will call this `write()` method twice. It sends in `product_image` first, and then sends in `product_image_target` second. So the logic needs to account for that :|

product_default_image/models/product_template.py
lasley: This question is too generic - I don't know how to answer it. You don't need a lambda here because you don't need to pass args to the browse method

product_default_image/README.rst
woodbrettm: The image is public domain from clipart with some modifications I made. Would you still like it in?
https://openclipart.org/detail/98491/open-box

product_default_image/models/product_category.py
woodbrettm: Well that's cool. Didn't know you could do that.

product_default_image/models/product_category.py
woodbrettm: Is one more efficient than the other or is that just preference?

product_default_image/models/product_template.py
woodbrettm: Makes sense. I was debating on that. Will change.

product_default_image/models/product_template.py
woodbrettm: For the doc string, for the args could I point to the methods below that this method calls? (those ones have docstrings.). Or should I restate all that stuff here?

product_default_image/models/product_template.py
woodbrettm: Perhaps a helper method yeah.

product_default_image/models/product_template.py
woodbrettm: Also for the name, would something like `search_and_change_product_images` be better?

product_default_image/README.rst
lasley: Yeah we should always provide attribution, regardless of whether the license forces us to.

product_default_image/models/product_template.py
woodbrettm: The idea here was to ensure that
 ```
+        domain = [
 +            ('company_id', '=', self.env.user.company_id.id),
 +            ('auto_change_image', '=', True),
 +            ('img_type', 'in', from_types),
 +        ]
```

is always in the domain. This cannot and should not be changed imo, and then I added the `add_domain` argument so people can add additional domain entries to that.

File: product_default_image/models/product_template.py (hunk 29 above)
Comment (woodbrettm) [reply product_default_image/models/product_template.py]: okie dokes. I'm unclear of what you're gearing towards as to what that constant would look like. Would it be a dict to map certain things? I'm unsure of what you're thinking of in that aspect.

File: product_default_image/models/product_template.py (hunk 31 above)
Comment (woodbrettm) [reply product_default_image/models/product_template.py]: Ah woops I missed adding the arg type `to_type (str)`. Also seems inline description as well.

File: product_default_image/models/product_template.py (hunk 33 above)
Comment (woodbrettm) [reply product_default_image/models/product_template.py]: ah makes sense now

File: product_default_image/models/product_category.py (hunk 22 above)
Comment (woodbrettm) [reply product_default_image/models/product_category.py]: Ah because of the list creation = less fast

Comment (lasley) [reply product_default_image/models/product_category.py]: Primarily because of the array. A tuple would make it only about a +10% variance

Comment (woodbrettm) [reply product_default_image/models/product_category.py]: aha yeah just figured that and edited my comment as you were writing xD

**Comment by lasley** [comment]
@bmw95 - please attend to the review comments. We won't worry about rewriting the one area just yet

**Comment by woodbrettm** [comment]
Sounds good. For the helper methods I was thinking of creating an abstract class to be inherited. Would that be the right approach?

Also you mentioned to put the string values (e.g. 'default_category') in a constant variable (maybe a dict). Should I put that  in a separate file and then import the file where needed?

**Comment by lasley** [comment]
> For the helper methods I was thinking of creating an abstract class to be inherited. Would that be the right approach?

Yup - if they need to be used by multiple models, you just create an abstract then use multiple inheritance to glue the logic in.

> Also you mentioned to put the string values (e.g. 'default_category') in a constant variable (maybe a dict). Should I put that in a separate file and then import the file where needed?

Class constants in a model somewhere instead maybe? Maybe in the abstract model?

**Comment by woodbrettm** [comment]
Ah yeah that's a good idea to put in the abstract model. I'll do that then.

**Comment by woodbrettm** [comment]
@lasley updated! I opted to move the constants into their own file instead.

File: product_default_image/models/abstract_product_image.py [hunk 40]
```diff
@@ -0,0 +1,55 @@
...
+
+
+class AbstractProductImage(models.AbstractModel):
+
+    _name = 'abstract.product.image'
+    _description = 'Abstract Product Image'
+
+    @api.model
+    def _vals_get_images(self, vals,
+                         img_keys=('image', 'image_medium', 'image_small')):
```
lasley: IMO the default image keys should be a constant

product_default_image/models/abstract_product_image.py
```diff
@@ -0,0 +1,55 @@
...
+
+    @api.model
+    def _val_match_any(self, val, source_dict, source_dict_keys):
+        source_dict_keys = self._ensure_iterable_arg(
+            source_dict_keys
+        )
+        for key in source_dict_keys:
+            if val == source_dict.get(key, False):
+                return True
+
```
lasley: It would be best to explicitly return `False` IMO. While this is technically evaluating to `False` on the fail case due to the `None` value, `None` is indicative of no or a bad result typically vs. `False` is the result of a boolean method returning a fail

product_default_image/models/product_category.py
```diff
@@ -33,6 +37,7 @@ class ProductCategory(models.Model):
     )
 
     @api.model
+    @api.returns('self', lambda value: value.id)
```
lasley: Why was this required?

product_default_image/image_constants.py
```diff
@@ -0,0 +1,20 @@
+# -*- coding: utf-8 -*-
+# Copyright 2017 LasLabs Inc.
+# License LGPL-3.0 or later (http://www.gnu.org/licenses/lgpl.html).
+
+# From product.template
+# image_type field
+TYPES = {
```
lasley: Why are you using numeric keys? The point of a dict is so that the key describes the data - you might as well just use lists here.

Don't use a list though - I'm having an incredibly hard time remembering these type definitions. Any time you have to keep referencing another file to remember something, you are totally doing it wrong. 

I think the better solution would actual constants, not dicts.

```
TYPE_GLOBAL = 'default_global'
TYPE_CATEGORY = 'default_category'

```

product_default_image/models/product_category.py
```diff
@@ -0,0 +1,82 @@
...
+    image_small = fields.Binary(
+        string='Small-Sized Image',
+        attachment=True,
+        help='Small-sized image of the category. It is automatically '
+             'resized as a 64x64px image, with aspect ratio preserved. '
+             'Use this field anywhere a small image is required.',
+    )
+
+    @api.model
+    @api.returns('self', lambda value: value.id)
```
lasley: Why is this required?

product_default_image/models/product_template.py
```diff
@@ -0,0 +1,319 @@
...
+    auto_change_image = fields.Boolean(
+        string='Auto Change Image',
+        default=True,
+        help='Allow/disallow automatic changes to product image. '
+             'Uncheck to stop being changed to a default image.',
+    )
+    image_type = fields.Selection(
+        string='Image Type',
+        selection=[
+            ('default_global', 'Global'),
```
lasley: Should use the constants here

product_default_image/models/product_template.py
```diff
@@ -0,0 +1,319 @@
...
+                        ('auto_change_image', '=', True),
+                        ('image_type', 'in', from_types arg),
+                    ]
+
+                    You can add additional domain items using the add_domain
+                    arg, which will be appended to the above constructed
+                    domain.
+
+        """
+        if not add_domain:
```
lasley: `if add_domain is None`

product_default_image/models/product_template.py
```diff
@@ -0,0 +1,319 @@
...
+            add_domain = []
+
+        if not isinstance(from_types, list):
+            raise TypeError('from_types argument must be a list')
+
+        if add_domain and not isinstance(add_domain, list):
```
lasley: `if add_domain` is superfluous due to above

product_default_image/models/res_company.py
```diff
@@ -0,0 +1,136 @@
...
+             'Limited to 1024x1024.',
+    )
+
+    @api.model
+    def _default_product_image(self):
+        image_path = get_module_resource(
+            'product_default_image', 'static/src/img', 'glob_prod_img.png'
+        )
+        return tools.image_resize_image_big(
+            open(image_path, 'rb').read().encode('base64')
```
lasley: Bad - This file handler is left open.
```
with open(image_path, 'rb') as handler:
    image_data = handler.read()
return tools...
```

You could technically return from inside the with statement, but the handler wouldn't be closed until the end of the session.

product_default_image/models/res_company.py
```diff
@@ -0,0 +1,136 @@
...
+            company.product_image = some_image
+            company.product_image_target = 'global'
+
+        """
+        target = vals.get('product_image_target')
+
+        if 'product_image' not in vals and not target:
+            return super(ResCompany, self).write(vals)
+
+        if 'product_image' in vals:
```
lasley, line 71: You're duplicating the logic of checking whether the key is in the dict. Unify please

lasley (changes requested):

lasley:
@bmw95 - We're almost there! 😆  this is a difficult one

product_default_image/image_constants.py
woodbrettm: ah okay. I saw in the medical constants, integers were being used so that's why I did that here. Makes sense though, I was having to refer back to the file a lot.

lasley: There's a reason we're deprecating the medical constants 😉

product_default_image/models/res_company.py
woodbrettm: The first check ensures if neither `'product_image'` and `target` are present, to exit. After that, the logic accounts for if one of each is present, or both.

woodbrettm:
@lasley updated

lasley (approved):

product_default_image/models/abstract_product_image.py
```diff
@@ -0,0 +1,49 @@
...
+            img_keys (list, tuple, str): Contains string values of image
+                fields. If specifying multiple, use a list or tuple.
+
+        Returns:
+            list: Returns image values of the img_keys in the vals arg
+                if the keys are in the vals dict.
+
+        """
+        if not isinstance(img_keys, (list, tuple)):
+            img_keys = [img_keys]
```
obulkin: Minor comment but maybe use a tuple here instead since you're not mutating the collection.

product_default_image/models/product_template.py
```diff
@@ -0,0 +1,317 @@
...
+                None: if any of the conditions below are True.
+
+        """
+        if any([
+                to_type not in (NONE, GLOBAL, CATEGORY, CUSTOM),
+                to_type == GLOBAL_CATEGORY,
+                to_type == CUSTOM and not to_img_bg,
+                to_type == NONE and to_img_bg,
+                not isinstance(in_cache, bool)]):
+            return
```
obulkin: In at least some of these cases, it seems like you would want to raise an exception with a helpful error message, rather than just doing nothing and silently returning.

product_default_image/models/product_template.py
```diff
@@ -0,0 +1,317 @@
...
+                        * Deletes the image
+
+                    CUSTOM
+                        * Set to the image you specify using to_img_bg.
+                        Note that targeting to custom will prevent the
+                        image from automatically being changed in the
+                        future until deleted or manually re-mapped to a
+                        default image.
+
+                to_img (base64): Can be used as an override if the template
```
obulkin: This should be ``to_img_bg``, right?

product_default_image/models/product_template.py
```diff
@@ -0,0 +1,317 @@
...
+                    method in product.template)
+
+            Returns:
+                ProductTemplate: Recordset if tmpl images have been changed.
+                None: if any of the conditions below are True.
+
+        """
+        if any([
+                to_type not in (NONE, GLOBAL, CATEGORY, CUSTOM),
+                to_type == GLOBAL_CATEGORY,
```
obulkin: This line is redundant.

product_default_image/models/res_company.py
```diff
@@ -0,0 +1,135 @@
...
+            'from_types': [
+                NONE, GLOBAL, CATEGORY,
+            ],
+            'to_type': target,
+            'to_img_bg': None,
+            'in_cache': False,
+            'add_domain': None,
+        }
+
+        if target == GLOBAL_CATEGORY:
```
obulkin: This seems like some leftover code that needs to be dropped now that the logic at L126 is present.

obulkin (changes requested):
The only blocking comment is the one in ``res_company.py``

product_default_image/models/abstract_product_image.py
woodbrettm: The issue here is that, say if I pass in `'image_medium'` as my img_keys argument, tuples must be at least length 2, so it will be left as just a string even if I try to convert to a tuple. The extra logic to account for that is unnecessary imo.

product_default_image/models/product_template.py
woodbrettm: Nice catch :)

product_default_image/models/res_company.py
woodbrettm: Ah thanks for catching that.

product_default_image/models/product_template.py
woodbrettm: Yeah I was debating about that for awhile. I think perhaps it may be best to separate these into separate checks and raise errors accordingly.

product_default_image/models/abstract_product_image.py
obulkin: It's possible to have a tuple of length 1. The literal syntax for this involves adding a comma after the element. In this particular case, that would be ``(img_keys,)``.

woodbrettm:
Thanks for the review @obulkin. 

I remember now why I didn't raise the errors and returned instead, as there were some quirks with .write() in res.company. Those have been fixed now as well.

This is ready for further review, despite Travis down. When it's up I'll make sure to repush.

lasley:
Travis issue - https://github.com/OCA/maintainer-quality-tools/issues/472

obulkin (approved):
No problem. It looks like all my comments have been addressed so this should be good to go assuming that the tests pass.

Issue #206, opened 2016-11-16
RFC: Product Service Duration


lasley:
This module will add the concept of a service user_id and duration on `product.template`. 

The intent is to allow for the creation of services that are (optionally) performed by a certain user & have a standard duration in minutes.

The specific need on my end is to use as part of Physician scheduling functionality, pairing with the calendar module to allow selection of the service that is going to be performed (office visit, etc) in order to default a time block size and optional user. 

This seems useful in a lot of other cases though; basically any form of job scheduling where X units takes Y minutes of time.

Assuming I'm not missing some strategy that already exists for this, am I also in the right repo?

pedrobaeza:
For this, you already have as a base https://github.com/OCA/sale-workflow/tree/10.0/sale_start_end_dates, and sale_rental has a similar concept, but it doesn't include yet the scheduling.

lasley:
Thanks for the suggestion @pedrobaeza - I thought about `sale_start_end_dates` at first, but most of this logic is incompatible with my needs unfortunately. It is focused around dates on sales and invoices, while I need just a float_time on products dictating how long that service generally takes to perform (and possibly a person that performs it).

An example is a Doctor's Office visit, which will typically takes say 30 minutes. The scheduling assistant can select the associated product for an office visit during scheduling, which will enforce a 30 minutes time block for the appointment. The product can then be used for subsequent billing when the appointment is actually handled.

Another example in a different context with no need for billing the appointment, would be a restaurant taking reservations. They know that the average reservation for a party is 2 hours, so they use this service product to reserve the time block in the calendar. Maybe there is a cancellation fee associated with this product that is used for billing, but for the most part the billing is on the food itself.

In both examples these workflows are taking place from a calendar, and don't need the creation of a sale order during scheduling. The amount of time taken for the appointment isn't relevant to anyone past the scheduling, so doesn't really flow into the billing processes like the existing modules.

As I typed this out, I'm thinking maybe `event` is a better repo. While I did not originally intend for this base module to depend on event, or provide that logic, I'm having troubles thinking of what else it would be good for.

pedrobaeza:
I think you should depend on `resource` module and base all your work to think in a load/capacity planner. We have this in our R&D projects, but to be plugged to projects and manufacturing, but the problem is the same: you have several resources (a table, a doctor, an engineer, a workcenter...), and you need it for a time. If it's free, you can lock it for a time slot.

[...]

PR #251, opened 2017-03-31
[10.0][ADD] product service duration

woodbrettm:
Hi there,

Opened in response to #206.

### Description

Extension of products (services only) that allows them to be added to calendars.

As a note, if you add services with a `Minimum Time` (minimum time it
takes to complete that service) to an event, the meeting/event duration cannot
be less than the combined minimum time of those services.

For example, say your meeting involves two services, technical and design services.
If the technical service requires 2 hours and design requires 3 hours, then your
meeting must be at least 5 hours.

The above constraint does not apply on events that have `All Day` selected.

When changing the `Minimum Time` on services, this will not affect events that have
already ended.

Thanks!
Brett


product_service_duration/demo/calendar_event_demo.xml
```diff
@@ -0,0 +1,17 @@
...
+<odoo>
+
+    <record id="calendar_event_1" model="calendar.event">
+        <field name="name">Client Consultation</field>
+        <field name="start" eval="time.strftime('%Y-%m-10 12:00:00')" />
+        <field name="stop" eval="time.strftime('%Y-%m-10 17:00:00')" />
+        <field name="duration" eval="5.00" />
+        <field name="template_ids"
+               eval="[(4, ref('product_service_duration.product_template_1')),
+                      (4, ref('product_service_duration.product_template_2'))]" />
```
woodbrettm: Pretty sure I can do [(6, 0 [ids])] right?

product_service_duration/models/calendar_event.py
```diff
@@ -0,0 +1,45 @@
...
+                )
+                stop = fields.Datetime.from_string(
+                    record.stop
+                )
+
+                diff_secs = (stop - start).total_seconds()
+                event_hours = diff_secs / 3600
+
+            min_event_duration = sum(
+                [t.minimum_service_time for t in record.template_ids]
```
woodbrettm: Perhaps I should add `min_event_duration` as a computed field. Would be more useful for other devs that way.

product_service_duration/views/product_template_view.xml
```diff
@@ -0,0 +1,44 @@
...
+
+    <record id="product_template_form_view" model="ir.ui.view">
+        <field name="name">product.template.common.form</field>
+        <field name="model">product.template</field>
+        <field name="inherit_id" ref="product.product_template_form_view" />
+        <field name="arch" type="xml">
+            <xpath expr="//group[@name='group_standard_price']" position="after">
+                <group name="group_service">
+                    <field name="service_user_id" />
+                    <field name="minimum_service_time" />
```
woodbrettm: Should probably default these to invisible if the product is not a service.

product_service_duration/models/calendar_event.py
```diff
@@ -0,0 +1,45 @@
...
+            if not event_hours:
+                start = fields.Datetime.from_string(
+                    record.start
+                )
+                stop = fields.Datetime.from_string(
+                    record.stop
+                )
+
+                diff_secs = (stop - start).total_seconds()
+                event_hours = diff_secs / 3600
```
woodbrettm: Should I add any rounding to this? Not sure if that may cause issues when comparing event_hours against the duration of the meeting, since they could potentially have a different set of decimal points.

woodbrettm:
Travis is failing from unused imports at the moment. Will fix after reviews.

product_service_duration/demo/calendar_event_demo.xml
lasley: Yup, that's actually what you want to do. Subsequent updates of this record will actually continue to insert the same records into the m2m, because the old ones aren't being removed. `6` is just a `5` followed by a `4` for each `id` in `ids`

product_service_duration/models/calendar_event.py
lasley: You should never round without an explicit reason, and definitely not before saving data. Rounding should always be a display thing

product_service_duration/models/calendar_event.py
lasley: That's fine, yeah. Just make sure to store it.

product_service_duration/models/calendar_event.py
```diff
@@ -0,0 +1,45 @@
...
+
+            min_event_duration = sum(
+                [t.minimum_service_time for t in record.template_ids]
+            )
+            if min_event_duration > event_hours:
+                raise ValidationError(_(
+                    'The services involved requires a %s hour '
```
lasley: s/requires/require

product_service_duration/models/calendar_event.py
```diff
@@ -0,0 +1,45 @@
...
+                    'long meeting.' % min_event_duration,
```
lasley: This doesn't work. You're interpolating on the second string, not the first. 

Regardless, interpolation should take place outside of the translation. Otherwise you need a new translation for any variable combination.
```
_('this is data %s') % data
```

product_service_duration/models/product_template.py
```diff
@@ -0,0 +1,28 @@
...
+import odoo.addons.decimal_precision as dp
+
+
+class ProductTemplate(models.Model):
+
+    _inherit = 'product.template'
+
+    service_user_id = fields.Many2one(
+        string='Service User',
+        comodel_name='res.users',
```
lasley: `domain="[('company_id', 'child_of', company_id)]"`

product_service_duration/models/product_template.py
```diff
@@ -0,0 +1,28 @@
...
+        help='Optionally set a user on performing this service',
+    )
+    minimum_service_time = fields.Float(
+        string='Minimum Service Hours',
+        digits_compute=dp.get_precision('Product UoM'),
```
lasley: [s/digits_compute/digits](https://github.com/odoo/odoo/search?utf8=%E2%9C%93&q=%22dp.get_precision%28%22&type=)

product_service_duration/tests/test_calendar_event.py
```diff
@@ -0,0 +1,86 @@
...
+
+from odoo import fields
+
+from odoo.tests.common import TransactionCase
+from odoo.exceptions import ValidationError
+
+import time
+
+
+class TesteventEvent(TransactionCase):
```
lasley: Bad class name

lasley (changes requested):
Thanks @bmw95

product_service_duration/models/product_template.py
woodbrettm: What's the reason for this domain?

lasley: This is not multi company friendly as it is. A company would only be selling services for their users.

woodbrettm: oh got it. Makes sense now.

woodbrettm:
@lasley updated.

product_service_duration/views/calendar_event_view.xml
```diff
@@ -0,0 +1,58 @@
...
+
+    <record id="view_calendar_event_form_popup" model="ir.ui.view">
+        <field name="name">calendar.event.form</field>
+        <field name="model">calendar.event</field>
+        <field name="inherit_id" ref="calendar.view_calendar_event_form_popup" />
+        <field name="arch" type="xml">
+            <xpath expr="//field[@name='partner_ids']" position="after">
+                <field name="template_ids"
+                       widget="many2many_tags"
+                       can_create="true"
```
lasley: domain restrict to only service type products

product_service_duration/models/product_template.py
```diff
@@ -0,0 +1,29 @@
...
+
+from odoo import fields, models
+import odoo.addons.decimal_precision as dp
+
+
+class ProductTemplate(models.Model):
+
+    _inherit = 'product.template'
+
+    service_user_id = fields.Many2one(
```
lasley: This actually needs to be to a resource, not to a user. The resource is the user in the context of their working hours.

woodbrettm: That makes a lot more sense now haha

product_service_duration/models/product_template.py
```diff
@@ -0,0 +1,29 @@
+# -*- coding: utf-8 -*-
+# Copyright 2017 LasLabs Inc.
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+
+from odoo import fields, models
+import odoo.addons.decimal_precision as dp
+
+
+class ProductTemplate(models.Model):
```
lasley: This actually needs to be pushed back into the variant please. Think multiple doctors that perform the same exam types.

product_service_duration/views/calendar_event_view.xml
woodbrettm: Hmm why type products? I think earlier in the review you specified in another template ids field in the calendar popup form to do product type service. Which one should I do?

lasley: products with `type == 'service'` is what I meant in both areas.

woodbrettm:
@lasley thanks for the review

I updated some of the logic to prevent validation errors if an event has ended as well as if the event is selected as all_day.

With the above bit in mind, I also added some logic to prevent users from changing the minimum service time on products if they're causing conflicts in events.

A dependency on `sale` was added as well, as the product.product views only seem to be located in the product variants side menu.

Cheers!

lasley:
> With the above bit in mind, I also added some logic to prevent users from changing the minimum service time on products if they're causing conflicts in events.

Previously scheduled events should not be changing if the service time on the product is changing. It should only be relevant for new appointments

woodbrettm:
@lasley I think I may actually be misunderstanding, however to confirm, I'm now using the following logic for both the validations and compute_min_duration in calendar event:
```python
            if record.stop < today_time:
                continue

            if record.allday:
                continue
```
So the logic will only apply to currently running events and events in the future. 
Is that good?

lasley (approved):
Thanks @bmw95

lasley:
@tedsalmon please review

tedsalmon (approved):
LGTM 👍

PR #252, opened 2017-04-05
10.0 del product supplierinfo tree price info

mourad-ehm:
This module in no longer useful. In the new model product.supplierinfo the price is a simple field. The field pricelist_ids was removed.


bealdav:
:+1:

PR #254, opened 2017-04-18
[NEW] 10.0 product attribute group

gdgellatly:
Hi,

This is a new module which brings the add all option function from product variant multi up to date and inline with the new style variants while porting a bit of extra special variant functionality, the ability to mass update products just by changing the members of the attribute group.

The module looks simple but is very powerful for organisations that have extensive use of variants with replicated attributes across many products.


gdgellatly:
I've moved this over to https://github.com/odoonz/product/tree/10.0 . The v8 proposal was never reviewed after over a year, and this one is idle 3 months now.  So closing due to lack of interest even if I think its cool.

PR #257, opened 2017-04-20
[FIX] normalize version number

sbidoul:
I thought there was a pylint-odoo check for that. 

We must not be able to merge addons with invalid versions as this breaks downstream processes.


gdgellatly:
👍 That code was submitted for PR long before pylint-odoo or standardised version numbers. 

EDIT: Wow I just checked, nearly 2 years for a merge.  Maintainers might need to close and reopen every open PR prior to whenver standardised version numbers came into effect.

pedrobaeza (approved):

PR #258, opened 2017-05-01
[8.0][ADD] New module product_pricelist_cache.

NL66278:
Mainly intended to speed up POS loading, but possibly usefull in other scenario's were computing product prices takes a lot of time.


NL66278:
In one customer scenario cut CPU time for product loading in POS from 1913 seconds to 32.

Together with module to not load customer data (PR proposed on pos repo), not needed for most cash registers, cut POS loading time from over nine minutes to less then one.

product_pricelist_cache/models/product_price_cache.py
```diff
@@ -0,0 +1,62 @@
...
+
+class ProductPriceCache(models.Model):
+    _name = 'product.price.cache'
+    _order = 'pricelist_id, product_id'
+
+    pricelist_id = fields.Many2one(
+        comodel_name='product.pricelist',
+        string='Pricelist',
+        readonly=True,
+        index=True,
```
hbrunn: cascading deletes would be sensible here too, right?

product_pricelist_cache/models/product_price_cache.py
```diff
@@ -0,0 +1,62 @@
...
+                    'product_id': product.id,
+                    'price': product_price,
+                })
+
+    @api.model
+    def cache_pricelist_prices_all(self):
+        """Cache prices for all pricelists, for all products."""
+        product_model = self.env['product.product']
+        pricelist_model = self.env['product.pricelist']
+        all_products = product_model.search([('type', '=', 'sale')])
```
hbrunn: this looks like it should be the search on pricelists? and when searching for products, I think you can restrict them to `sale_ok==True`

hbrunn (changes requested):
small things

hbrunn (approved):

product_pricelist_cache/models/product_price_cache.py
```diff
@@ -0,0 +1,63 @@
...
+                product_cache.create({
+                    'pricelist_id': pricelist.id,
+                    'product_id': product.id,
+                    'price': product_price,
+                })
+
+    @api.model
+    def cache_pricelist_prices_all(self):
+        """Cache prices for all pricelists, for all products."""
+        product_model = self.env['product.product']
```
lasley, line 58: It'd be nice to see a test for this method

lasley (approved):
One note regarding an untested method, I won't block for it though

PR #259, opened 2017-05-04
10 mig product variante inactive

EBII:
Active, disables variant from list


product_variant_inactive/views/product_variant_view.xml
```diff
@@ -0,0 +1,35 @@
...
+
+    <record model="ir.ui.view" id="detail_field_custom_in_view">
+        <field name="model">product.product</field>
+        <field name="inherit_id" ref="product.product_product_tree_view"/>
+        <field name="arch" type="xml">
+            <xpath expr="//tree" position="inside">
+                <field name="active" invisible ="1"/>
+                <button type="object" name="button_activate" invisible="1"
+                        attrs="{'invisible': [('active','=',True)]}" icon="fa-ban"/>
+                <button type="object" name="button_deactivate" invisible="1"
```
gdgellatly, line 24: @EBII have you considered using a confirm string here in case of accidental click?  Without trying to second guess your reasons for this module and why you designed it this way, I kind of see 2 use cases.

1: Occasional deactivation of a few products
2: Deactivation of many products (>20)

Obviously your way trades screen space for explicitness to the user, but idk, for me I think it would be more natural to select the products I want to deactivate, then click more menu deactivate or activate.  It would also eliminate the need for all that fields_view_get code.

I'm also not sure, but do we just want regular users having access to this button of should it be limited to groups.

These are just thoughts.  I'll review better when I understand the rationale.

EBII: @bealdav, please answers him. 
Thanks

bealdav: Hi @gdgellatly,

Thanks for your comment,

We've done this PR but we'll no more use this module for our uses cases in 10 version. It's just a contribution (base on v8 production install) that works but improvement could be done by community. Modules are done to get additional feature over the time, but as it works well, especially in productivity side.

1/ Accidental click: It depends of your users, then decision to put xml confirm which cost a lot in productivity, must be define as fine tuning in usability custom module depending of your customers I think.

2/ Desactivation of many: you're right an additional action in more menu could be better to reach excellence of features

3/ A group in button could be a valuable additional feature.

I propose to put 2/ and 3/ in roadmap of the module but merge as it or maybe you could make a PR on our ?

gdgellatly: @bealdav As I said, not blocking, just trying to understand the use case. I've forced a rebuild on runbot just to test.

bealdav: ok @gdgellatly thanks a lot for your review and good ideas.

bealdav:
@gdgellatly now fixed. 

Thanks to ultime review.

gdgellatly:
@bealdav @EBII you'll need some tests to get this merged in this repo I'd think.  Only 3 functions, all very easily testable.  Not that I'm the ultimate authority here.  THat would be @lasley .

In any case functionally I'm OK with everything.

gdgellatly (approved):

hparfr (approved):

bealdav (approved):

PR #265, opened 2017-05-16
10.0 product attribute exclusion

gdgellatly:
The premise is that sometimes combination of attributes are not all available and so we want to exclude those on 1, many or all product templates. This allows you to quickly create an exclusion matrix, hand edit if desired and it will prevent those variants from being created, unlink existing ones if it can, or merely deactivate existing ones.

The module is complete aside from above mentioned concerns over cyclomatic complexity - IOW it works, just not sure how efficiently yet.


gdgellatly:
@lasley I have no idea why this fails.  The only string I even changed was the summary one from triple quoted to literal concat style.

product_combination_exclude/__manifest__.py
```diff
@@ -0,0 +1,23 @@
+# -*- coding: utf-8 -*-
+# Copyright 2017 Graeme Gellatly
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+{
+    'name': 'Product Combination Exclude',
+    'summary': "Allows the specification of incompatible product "
++              "combinations when creating products",
```
lasley: Looks like the copy/paste was from a git diff? There's a `+` at the beginning of this line

gdgellatly: hah, it was your diff.  My linter always shows the entire manifest file as an error so couldn't see it.

gdgellatly:
Well this is a new one for me.  Because I'm force pushing commits theres no diff but I just don't understand how I'm breaking either product_service_duration or the server in general.  It was green before I added the extra little fillin tests.

Weirdly on runbot product service durations isn't even listed to be installed.

pedrobaeza:
Have you rebased on top of 10.0 branch?

Is there any new default value that activates the exclusion behavior?

gdgellatly:
I don't really know how to do that.  Gave it a go, looks OK, lets see.  Nah, I was really careful with it.  It is very independent module.  All fresh demo data for testing even.  The only thing I can think is the override in create as I moved code out of the try/except clause.

lasley:
Hmmmm I think the rebase wasn't done quite right - I see some unrelated commits here, one of which looks to be related to product_service_duration (89a153f). I don't see the commits for the exclusion module now though?

gdgellatly:
@lasley @pedrobaeza OK I found the issue.  Was indentation leaving no path to super on create call without attributes

gdgellatly:
Closing this, it needs a total rethink as it is quite buggy in actual use and breaks badly in master plus it got rebase issues.  Easier to start from scratch.

gdgellatly:
Just a note on this.  Go the camptocamp way if looking to implement this.  In practice it just had too many subtle bugs and usability issues, completely abandoned now #275

PR #267, opened 2017-05-17
[8.0][IMP] Replicate sale order line names behavior

zamberjo:
Replicate behavior of *sale.order.line:name* field, set to customer product code + name instead of product code + name.

### DIFF
--- a/product_supplierinfo_for_customer/__openerp__.py
+++ b/product_supplierinfo_for_customer/__openerp__.py
@@ -28,6 +28,7 @@
     "depends": [
         "base",
         "product",
+        "sale",
     ],
     "data": [
         "views/product_view.xml",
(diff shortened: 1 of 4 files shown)


pedrobaeza:
This should be another module, as you can't change dependencies in a module in stable, and maybe you don't want this behavior by default. Please see also product_customer_code module for the same use.

zamberjo:
Hi @pedrobaeza 

After checking these modules - [`product_customer_code`](https://github.com/OCA/product-attribute/tree/8.0/product_customer_code) and [`product_customer_code_sale`](https://github.com/OCA/sale-workflow/tree/8.0/product_customer_code_sale), both flagged as `'installable': False` -, I'd replace them with a new module (maybe `product_supplierinfo_customer_code`?) that could take advantage of the base functionality of `product.supplierinfo` from `product` module.

pedrobaeza:
Yes, but take into account than on v10 people is going the other direction (migration product_customer_code*).

PR #270, opened 2017-05-19
[8.0][FIX] Use self.env instead of pooler.

NL66278:
Using pooler gives deprecated warnings on runbot.


pedrobaeza (approved):

customer_pricing_in_product_view/tests/test_customer_pricing_in_product_view.py
```diff
@@ -19,29 +19,21 @@
...
                 self.cr, self.uid, list_id)
         except Exception:
             self.fail("Pricelist name_get failed")
 
     def test_view_partner_pricing(self):
-        partner_obj = self.pool['res.partner']
-        partner_id = [self.env.ref('base.res_partner_1').id]
-        result = partner_obj.view_partner_pricing(
-            self.cr, self.uid, partner_id)
+        partner = self.env.ref('base.res_partner_1']
```
LoisRForgeFlow: fix the bracket ] -> )

customer_pricing_in_product_view/tests/test_customer_pricing_in_product_view.py
```diff
@@ -19,29 +19,21 @@
 #    along with this program.  If not, see <http://www.gnu.org/licenses/>.
 #
 ##############################################################################
```
LoisRForgeFlow: You could take the occasion to update the headers as well.

customer_pricing_in_product_view/tests/test_customer_pricing_in_product_view.py
NL66278: @lreficent I will, also the headers.

gdgellatly:
@NL66278 interesting, I never even recall doing this for v8 as its a version I never used.  Oh I just found the PR, it took 2 years to merge no wonder its full of deprecation.  Anyway, for v10 I had to pull the scope right back.  It might be worth checking if v8 has the same issue (essentially web client not using context change in search bar to unhide columns in tree views).

Anyway, I learnt something from this so I'll be making a PR on Monday on the V10 code to do a similar clean.

NL66278:
@gdgellatly I think this is ready for merge now. Interestingly the name_get of product.pricelist still needs to get patched in 8.0.

lasley (approved):

PR #274, opened 2017-06-21
[10.0][ADD] New module to add tags on products

benwillig:
New module to add tag on products.
Users can search products based on a tag name.


product_tags/README.rst
```diff
@@ -0,0 +1,51 @@
...
+=====
+
+* On the product view, you can add and create tags.
+* Tags will be shown in the kanban view.
+
+Bug Tracker
+===========
+
+Bugs are tracked on `GitHub Issues
+<https://github.com/OCA/{project_repo}/issues>`_. In case of trouble, please
```
grindtildeath: s/{project_repo}/product-attribute

product_tags/views/product_template_tag.xml
```diff
@@ -0,0 +1,58 @@
...
+        <field name="model">product.template.tag</field>
+        <field name="arch" type="xml">
+            <tree>
+                <field name="name"/>
+                <field name="products_count"/>
+            </tree>
+        </field>
+    </record>
+
+    <record model="ir.actions.act_window" id="product_template_tag_act_window">
```
grindtildeath: No menuitem for this action ?

product_tags/models/product_template.py
```diff
@@ -0,0 +1,15 @@
...
+
+from odoo import fields, models
+
+
+class ProductTemplate(models.Model):
+
+    _inherit = 'product.template'
+
+    tag_ids = fields.Many2many(
+        comodel_name='product.template.tag', string="Tags",
```
grindtildeath: IMO 'product tags' would be nicer as a string.

product_tags/security/product_template_tag.xml
```diff
@@ -0,0 +1,17 @@
...
+<odoo>
+
+    <record model="ir.model.access" id="product_template_tag_access_user">
+        <field name="name">product.template.tag access user</field>
+        <field name="model_id" ref="model_product_template_tag"/>
+        <field name="group_id" ref="base.group_user"/>
+        <field name="perm_read" eval="1"/>
+        <field name="perm_create" eval="1"/>
+        <field name="perm_write" eval="1"/>
+        <field name="perm_unlink" eval="1"/>
```
grindtildeath: Any user can create or delete tags ? IMO using a group would be better.

grindtildeath (commented):
Please update README to latest template (try on runbot, roadmap).
Is your module intended to add tags only on product template ? I think it should then be name product_template_tags.
Otherwise, my suggestion would be to allow the tags to be added on product.product as well, but it could be stated in Readme, with some adaptations (and the keep product_tag everywhere instead of product_template_tags).

product_tags/views/product_template_tag.xml
benwillig: No indeed, I was not really sure about where to put this menu. Maybe in the settings app but that would not make a lot of sense I think.

product_tags/security/product_template_tag.xml
benwillig: My opinion was the same than for the menu item.

benwillig:
I updated my PR based on your comments @grindtildeath. I didn't change the non existing menu item and ACLs because as we discussed, this module only depends on ```product```.

grindtildeath (approved):

leemannd (approved):
Thanks for the work!

product_template_tags/README.rst
```diff
@@ -0,0 +1,71 @@
...
+
+Configuration
+=============
+
+* Go on the product form view.
+* Create and edit tags using the form view
+
+.. figure:: path/to/local/image.png
+   :alt: alternative description
+   :width: 600 px
```
LoisRForgeFlow: remove

product_template_tags/__init__.py
```diff
@@ -0,0 +1 @@
+from . import models
```
LoisRForgeFlow, line 1: headers with the coding at least

LoisRForgeFlow (approved):
Code reivew LGTM, some non-blocking comments

product_template_tags/views/product_template.xml
```diff
@@ -0,0 +1,54 @@
+<?xml version="1.0" encoding="utf-8"?>
+<!-- Copyright 2017 ACSONE SA/NV
+     License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl). -->
+
+<odoo>
+
+    <record model="ir.ui.view" id="product_template_form_view">
+        <field name="name">product.template.form (in product_tags)</field>
```
legalsylvain: name are unnecessary. (just for your information, not a blocking point)

product_template_tags/models/product_template_tag.py
```diff
@@ -0,0 +1,26 @@
...
+
+    name = fields.Char(string="Name", required=True, translate=True)
+    color = fields.Integer(string="Color Index")
+    product_tmpl_ids = fields.Many2many(
+        comodel_name='product.template', string="Products",
+        relation='product_template_product_tag_rel',
+        column1='tag_id', column2='product_tmpl_id')
+    products_count = fields.Integer(
+        string="# of Products", compute='_compute_products_count', store=True)
+
```
legalsylvain: could you add a company_id field (optional) ? with an according ir.rule. (access OK if company_id = current company or False). (and displayed only if user is part of multicompany)
it will be usefull in all multi company context.

product_template_tags/views/product_template_tag.xml
```diff
@@ -0,0 +1,61 @@
...
+    </record>
+
+    <record model="ir.actions.act_window" id="product_template_tag_act_window">
+        <field name="name">Product Template Tag</field>
+        <field name="res_model">product.template.tag</field>
+        <field name="view_mode">tree,form</field>
+        <field name="view_type">form</field>
+    </record>
+
+    <!-- No menu item defined for this action because this module only
```
legalsylvain: why don't you provide some menu item to manage product tags ? As it, it is not possible to update / unlink a tag. Or did I missed something ?

product_template_tags/models/product_template_tag.py
```diff
@@ -0,0 +1,26 @@
...
+from odoo import api, fields, models
+
+
+class ProductTemplateTag(models.Model):
+
+    _name = 'product.template.tag'
+    _description = 'Product Tag'
+
+    name = fields.Char(string="Name", required=True, translate=True)
+    color = fields.Integer(string="Color Index")
```
legalsylvain, line 14: Where this field is used ? (I mean, how can we change the value ?)

product_template_tags/README.rst
legalsylvain: alternatively, you can add a screen shot to display product form view with the new field. (not a blocking point)

product_template_tags/README.rst
```diff
@@ -0,0 +1,71 @@
...
+=====================
+Product Template Tags
+=====================
+
+This addon allow to add tags on products.
+
+Installation
+============
+
+Just install it!
```
legalsylvain: I think that such readme part is unnecessary, and could be removed. (not a blocking point)

legalsylvain (changes requested):
Hi @benwillig. 
Thanks a lot for this module. Tags on product is a really missing feature.

Some comments and questions inline. 

kind regards.

product_template_tags/models/product_template_tag.py
benwillig: When you select a tag on a product, if you click on the selected tag, you will be able to change the color.

product_template_tags/models/product_template_tag.py
benwillig: Indeed, this could be a nice feature, I will add it.

product_template_tags/views/product_template_tag.xml
benwillig: This module only depends on product and there is no specific menu defined for this module. A "product tags" menu should be in the sale/purchase configuration, maybe we could create two modules named ```sale_product_template_tags``` and ```purchase_product_template_tags```, but thoses module would only add a menu item.

product_template_tags/models/product_template_tag.py
legalsylvain: Ow. Thanks ! I did'nt know.

product_template_tags/views/product_template_tag.xml
legalsylvain: Wow I see ! Odoo is worth in each new version about that topic. Indeed, I checked for product.category, and there is no menu item in product module. (only the action). This is stupid by design.

Well, as a workaround, what about to create an ir.ui.menu, in configuration settings (in Parameters) parts here : 

image

then, if OCA want another module (sale_... or purchase_...) it will be allways possible, but do not having this menu item by default when your usefull module is installed avoid users to make basic interesting actions. (deleting / duplicating / enabling tags..)

don't you think ?

legalsylvain: Note : I had a similar problem with barcode model, defined in barcodes module, but the menu is only on Point of Sale.

https://github.com/OCA/stock-logistics-barcode/tree/10.0/barcodes_generator_abstract.

I so added a menu item in configuration settings, to avoid to depend on point of sale.

https://github.com/OCA/stock-logistics-barcode/blob/10.0/barcodes_generator_abstract/views/menu.xml

regards.

benwillig: Good idea, I added a menu in the parameters section.

product_template_tags/__init__.py
benwillig: This is not necessary as this file will only contains import statement. If there would be a hook, I would understand but it's not the case.

product_template_tags/models/product_template_tag.py
benwillig: Multi company feature added!

legalsylvain (approved):
Thanks a lot ! 
now all is ok for me.

regards.

rafaelbn:
Hi @benwillig , thank you for this contribution! 😄 

Please add a menu for Product Tag I miss it (at least) in Inventory -> Configuration -> Products -> (missing) "Product Tags".

A place where you can see all your product tags

legalsylvain:
Hi @rafaelbn. guess you missed the latest comments. It is not possible to add a menu in inventory part, because this module does'nt depend on 'stock'. It depends only on 'product'. For that reason the menu is in the 'Setting' parts. 
The other solution is to make this module depending on 'stock' (or 'sale', or other') but it will force users to install there modules even if they don't need them. (for exemple, guys using only 'product' + 'account'). 

I don't think a perfect solution, due to bad design in Odoo.

What do you think ?

regards.

rafaelbn:
Thanks you @legalsylvain for the comment.

Well, testing:

- I've test you example `for exemple, guys using only 'product' + 'account` and in this case there is not any menu anywhere to create products! 😲 . So the only way to create productis is in the invoice lines.
- So if there isn't any place to see products: why do you need to tag them? 🤔 

If @benwillig propose this module is because  a customer a least is using one of the options: sales, stock or purchase. He said ` Users can search products based on a tag name.` : Where can the search products? In which view?

My logic say:

- if you need to tag products if because you have products so you need a view of products.
- You could have services or stockable products but usualy comapnies which have stockable products also have services products, but companies which services producits is not mandatory that they have stockable products. So I would discard in depend in stock.
- As if you install purchase you are installing stock by dependency, I discard also purchase.
- **My proposal**: because statistic and logic is to add dependency with sale and add the menu to manage tags. 

But I'm not blocking at all. You can leave it as is now and merge. You @legalsylvain are PSC so you can change label to needreview and merge!

legalsylvain:
@rafaelbn,

> guys using only 'product' + 'account' [...] So the only way to create products is in the invoice lines.

Odoo menus are definitivly bad designed ! Just a pity to add a dependency to a module, only for a menu item.  I was thinking that it could be great to separate 'sale' module by 'sale' and 'sale_menu' module (the second one just depend on base, and add menu item). It will avoid such changes...

Well, thanks for your analysis. I'm globally agree with you. 

@benwillig,
what about to : 
- make this module depend on 'sale'
- move the menu item into sale configuration part. 

Indeed I think too that it will be better user Experience.

(sorry for changing my mind during the review)

benwillig:
@rafaelbn Indeed, our customer need it in the Inventory menu.

>So if there isn't any place to see products: why do you need to tag them?

Then, why would you install this module?

As you said, this module allows to search products based on tag name and there isn't any view to use it (I didn't override the name_search method btw). But, if you search in ```product``` module, there is tons of view/action created but not associated to any menu item. This module follows the same logic. If you install ```sale``` or ```purchase```, you will be able to search products based on tag name because I inherit the base views of products.

I already added a menu item in the settings menu. It might not be the sexiest solution, but IMO this is the cleanest.

I won't make this module inherit of ```sale``` or ```purchase``` or any other module that is an APP. If I want tags and I'm not using sales but only purchases. I don't want to install sales and I don't want to pay more because I installed an other app.

legalsylvain:
> I don't want to install sales and I don't want to pay more because I installed an other app.

Paying for sale or purchase. quiet weird. (joke) You're right. I didn't thought that adding a dependencies could change the odoo price. (time are changing ...) 

> I already added a menu item in the settings menu. It might not be the sexiest solution, but IMO this is the cleanest.

That's true. LGTM for me.

@benwillig : could you just add a line in the readme file like "to configure tags, go to ... ... " as this place is not the "normal" places where users will search to configure products things.  (usually product settings are in sale / settings.)

I think we can merge after.

kind regards.

rafaelbn (approved):

legalsylvain:
@benwillig, Thanks for all the changes.
Well 5 :+1:. time to merge.

PR #275, opened 2017-06-21
[9.0][ADD] module product_variant_exclusion

grindtildeath:
Product Variant Exclusion
=========================

This module extends the functionality of product variants to allow you to
select which combination of product attribute values you want to exclude
from the product variants.

It's build with a focus on performance using lots of attribute values, as
it avoids creating unnecessary product.product should the combination of
attribute values be part of an exclusion.

### DIFF
--- a/product_variant_exclusion/README.rst
+++ b/product_variant_exclusion/README.rst
@@ -1,0 +1,67 @@
+.. image:: https://img.shields.io/badge/licence-AGPL--3-blue.svg
+   :target: http://www.gnu.org/licenses/agpl-3.0-standalone.html
+   :alt: License: AGPL-3
+
+=========================
+Product Variant Exclusion
+=========================
+
+This module extends the functionality of product variants to allow you to
+select which combination of product attribute values you want to exclude
+from the product variants.
+
+It's build with a focus on performance using lots of attribute values, as
+it avoids creating unnecessary product.product should the combination of
+attribute values be part of an exclusion.
+
+Installation
+============
+
+Just install it
+
+Usage
+=====
+
+To use this module, you need to activate Product variants.
+On product template's form, under the variants tab will appear a new tree view
+allowing you to add variant exclusions.
+
+.. image:: https://odoo-community.org/website/image/ir.attachment/5784_f2813bd/datas
+   :alt: Try me on Runbot
+   :target: https://runbot.odoo-community.org/runbot/135/9.0
+
+Bug Tracker
+===========
+
+Bugs are tracked on `GitHub Issues
+<https://github.com/OCA/product-attribute/issues>`_. In case of trouble, please
+check there if your issue has already been reported. If you spotted it first,
+help us smash it by providing detailed and welcomed feedback.
+
+Credits
+=======
+
+Images
+------
+
+* Odoo Community Association: `Icon <https://github.com/OCA/maintainer-tools/blob/master/template/module/static/description/icon.svg>`_.
+
+Contributors
+------------
+
+* Akim Juillerat <akim.juillerat@camptocamp.com>
+
+Maintainer
+----------
+
+.. image:: https://odoo-community.org/logo.png
+   :alt: Odoo Community Association
+   :target: https://odoo-community.org
+
+This module is maintained by the OCA.
+
+OCA, or the Odoo Community Association, is a nonprofit organization whose
+mission is to support the collaborative development of Odoo features and
+promote its widespread use.
+
+To contribute to this module, please visit https://odoo-community.org.
--- a/product_variant_exclusion/models/product.py
+++ b/product_variant_exclusion/models/product.py
@@ -1,0 +1,159 @@
+# -*- coding: utf-8 -*-
+# Copyright 2017 Camptocamp SA
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+from openerp import models, fields, api, _
+from openerp.exceptions import ValidationError, UserError
+import psycopg2
+
+
+class ProductTemplate(models.Model):
+
+    _inherit = 'product.template'
+
+    variant_exclusion_ids = fields.One2many(
+        'product.variant.exclusion', 'product_tmpl_id', 'Variants exlcusion'
+    )
+
+    allowed_value_ids = fields.Many2many('product.attribute.value',
+                                         compute='_compute_allowed_values')
+
+    @api.depends('attribute_line_ids.value_ids')
+    def _compute_allowed_values(self):
+        for template in self:
+            template.allowed_value_ids = template.attribute_line_ids.mapped(
+                'value_ids')
+
+    @api.multi
+    def write(self, vals):
+        res = super(ProductTemplate, self).write(vals)
+        if 'variant_exclusion_ids' in vals:
+            self.create_variant_ids()
+            self.delete_variant_ids()
+        return res
+
+    @api.multi
+    def delete_variant_ids(self):
+        exclusions = self.env['product.variant.exclusion'].search(
+            [('product_tmpl_id', '=', self.id)])
+        for template in self:
+            for product in template.product_variant_ids:
+                for excl in exclusions:
+                    if set(excl.attribute_value_ids.ids) <= set(
+                            product.attribute_value_ids.ids):
+                        try:
+                            with self.env.cr.savepoint():
+                                product.unlink()
+                        except (psycopg2.Error, ValidationError):
+                            product.write({'active': False})
+                        break
+
+
+class ProductAttributeLine(models.Model):
+
+    _inherit = 'product.attribute.line'
+
+    @api.multi
+    def write(self, vals):
+        attr_values = vals.get('value_ids')[0][2]
+        if len(attr_values) <= 1:
+            for line in self:
+                for exclusion in line.product_tmpl_id.variant_exclusion_ids:
+                    if set(attr_values) <= set(
+                            exclusion.attribute_value_ids.ids):
+                        raise UserError(_(
+                            'You cannot leave only one value on a line '
+                            'if this value is part of an exclusion'))
+        return super(ProductAttributeLine, self).write(vals)
+
+    @api.multi
+    def unlink(self):
+        for line in self:
+            for val_id in line.value_ids.ids:
+                if val_id in line.product_tmpl_id.variant_exclusion_ids.mapped(
+                        'attribute_value_ids').ids:
+                    raise UserError(_(
+                        'You cannot delete an attribute line, when '
+                        'one of its values is part of an exclusion'))
+
+
+class ProductVariantExclusion(models.Model):
+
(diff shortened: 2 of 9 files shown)


mdietrichc2c (approved):

product_variant_exclusion/models/product.py
```diff
@@ -0,0 +1,154 @@
...
+from openerp.exceptions import ValidationError, UserError
+import psycopg2
+
+
+class ProductTemplate(models.Model):
+
+    _inherit = 'product.template'
+
+    variant_exclusion_ids = fields.One2many(
+        'product.variant.exclusion', 'product_tmpl_id', 'Variants exlcusion'
```
vrenaville: minor typo exlcusion -> exclusion

grindtildeath:
@gdgellatly @lasley 
As I seen you both worked on an variant exclusion feature in 10.0, would be great if you can do some review on this.

product_variant_exclusion/README.rst
```diff
@@ -0,0 +1,67 @@
...
+   :alt: License: AGPL-3
+
+=========================
+Product Variant Exclusion
+=========================
+
+This module extends the functionality of product variants to allow you to
+select which combination of product attribute values you want to exclude
+from the product variants.
+
```
gdgellatly, line 12: @grindtildeath I did do a brief review of this when you asked, and thought things looked really good.  Certainly at a technical implementation level I was at the point of reconsidering my own work.

There are a few housekeeping issues, like having all the models in one file as opposed to seperate ones that will need dealing with.  Would be nice to see better tests as well, rather than just a single template creation.

I'll put a few more questions in as I go.  It won't be complete, its late.  Nothing blocking, just trying to understand.

product_variant_exclusion/models/product.py
```diff
@@ -0,0 +1,174 @@
...
+                            'You cannot leave only one attribute value on a '
+                            'line if this value is part of an exclusion'))
+        if self.value_ids.ids != attr_values:
+            new_values = self.env[
+                'product.attribute.value'].browse(attr_values)
+            exclusion_values = self.product_tmpl_id.variant_exclusion_ids\
+                .mapped('attribute_value_ids')
+            for new_val in new_values:
+                if new_val in exclusion_values:
+                    raise UserError(_('You cannot remove an attribute value '
```
gdgellatly, line 74: Why not? Could you not just unlink the related exclusions at the same time?

product_variant_exclusion/models/product.py
```diff
@@ -0,0 +1,174 @@
...
+    def write(self, vals):
+        attr_values = vals.get('value_ids', False) and vals.get(
+            'value_ids')[0][2]
+        if attr_values and len(attr_values) <= 1:
+            for line in self:
+                for exclusion in line.product_tmpl_id.variant_exclusion_ids:
+                    if set(attr_values) <= set(
+                            exclusion.attribute_value_ids.ids):
+                        raise UserError(_(
+                            'You cannot leave only one attribute value on a '
```
gdgellatly, line 65: This needs a usability test but I think its OK.  The scenario I'm thinking of is removing some attributes and adding others. 

But also the logic looks a bit wrong/brittle.  The obvious question is why not, if an attribute is only excluded in limited circumstances (i.e. dependent on another attribute value), why do we need this error? Would it not make more sense to unlink the corresponding attribute values of the exclusions (e.g. if we deleted Male from gender and size 40 was excluded from Female, unlink 40)?  But also I can see a scenario where due to a combination of exclusions this error wouldn't be triggered when it should. (when len(attr_values) > 1 but the remaining attr values are also part of an exclusion themself)

product_variant_exclusion/models/product.py
```diff
@@ -0,0 +1,174 @@
...
+        return super(ProductAttributeLine, self).write(vals)
+
+    @api.multi
+    def unlink(self):
+        for line in self:
+            for val_id in line.value_ids.ids:
+                if val_id in line.product_tmpl_id.variant_exclusion_ids.mapped(
+                        'attribute_value_ids').ids:
+                    raise UserError(_(
+                        'You cannot delete an attribute line when '
```
gdgellatly, line 86: Again why not?  Just unlink the exclusions as well couldn't you?  OK maybe in this one there is an issue if its a 3+ way exclusion, because you don't know if the equivalent 2 way should now be excluded.

I'm OK with all 3 errors on reflection, its not like people delete attribute lines that often.

product_variant_exclusion/models/product.py
```diff
@@ -0,0 +1,174 @@
...
+
+    product_tmpl_id = fields.Many2one('product.template', required=True)
+    attribute_value_ids = fields.Many2many('product.attribute.value',
+                                           required=True)
+
+    def _check_values_attributes_uniqueness(self, value_ids):
+        values = self.env['product.attribute.value'].browse(value_ids)
+        attributes = values.mapped('attribute_id')
+        if len(values) != len(attributes):
+            raise UserError(_('You cannot have multiple values from the '
```
gdgellatly, line 103: This is a design decision.  I agree with it, but understand that it severely limits scalability for the user. (We have 12 templates that combined would require 200 exclusions, a bit tedious to enter by hand)  Thats why in my module, although we made the same decision, we included the capability to automatically generate the combinations.

product_variant_exclusion/models/product.py
```diff
@@ -0,0 +1,174 @@
...
+                    value_ids):
+                raise UserError(_(
+                    'You cannot create an exclusion for an attribute value '
+                    'which is unique in its attribute line !'))
+        self._check_values_attributes_uniqueness(value_ids)
+        return super(ProductVariantExclusion, self).create(vals)
+
+    @api.multi
+    def write(self, vals):
+        value_ids = vals.get('attribute_value_ids')[0][2]
```
gdgellatly, line 126: Beware IndexError, its feasible an extension could allow just changing just the tmpl_id (e.g. to allow duplication) or even just add a text field or something.  Accept it can't happen in its present implementation.

product_variant_exclusion/models/product.py
```diff
@@ -0,0 +1,174 @@
...
+        if len(values) != len(attributes):
+            raise UserError(_('You cannot have multiple values from the '
+                              'same attribute in an exclusion.'))
+
+    @api.model
+    def create(self, vals):
+        value_ids = vals.get('attribute_value_ids')[0][2]
+        if len(value_ids) <= 1:
+            raise UserError(_(
+                'You cannot create an exclusion with only one value !'))
```
gdgellatly, line 111: For a lot of these errors, consider presenting them in the positive rather than the negative.

e.g Exclusions require more than one value.  Basically tell the user what they need to do, not what they don't need to.

product_variant_exclusion/models/product.py
```diff
@@ -0,0 +1,174 @@
...
+        if self._context.get('create_product_variant'):
+            products_to_write = self.env['product.product']
+            # Case 2: Setting active to True
+            if vals.get('active', False):
+                for product in self:
+                    product_value_ids = product.attribute_value_ids.ids
+                    if product.product_tmpl_id.variant_exclusion_ids:
+                        for excl in product.product_tmpl_id.\
+                                variant_exclusion_ids:
+                            if not set(excl.attribute_value_ids.ids) <= set(
```
gdgellatly, line 149: someone will know better than me, but isn't there a more new API way of doing these recordset comparisons?

product_variant_exclusion/models/product.py
```diff
@@ -0,0 +1,174 @@
...
+
+    @api.multi
+    def write(self, vals):
+        if self._context.get('create_product_variant'):
+            products_to_write = self.env['product.product']
+            # Case 2: Setting active to True
+            if vals.get('active', False):
+                for product in self:
+                    product_value_ids = product.attribute_value_ids.ids
+                    if product.product_tmpl_id.variant_exclusion_ids:
```
gdgellatly, line 146: I prefer your explicitness here, just in case someone writes you should use product.variant_exclusion_ids.

product_variant_exclusion/models/product.py
```diff
@@ -0,0 +1,174 @@
...
+from openerp import models, fields, api, _
+from openerp.exceptions import ValidationError, UserError
+import psycopg2
+
+
+class ProductTemplate(models.Model):
+
+    _inherit = 'product.template'
+
+    variant_exclusion_ids = fields.One2many(
```
gdgellatly, line 13: We're happy with the copy functionality when duplicating a product template?  Personally I'm fine with it, just making sure its considered.

product_variant_exclusion/models/product.py
gdgellatly: Also I wonder if a lot of these write and create overrides aren't better dealt with @api.constrains().  Maybe a technical person can enlighten it more, but these seem like constraints rather than changing anything.

product_variant_exclusion/views/product.xml
```diff
@@ -0,0 +1,20 @@
...
+    <record id="product_template_only_form_view" model="ir.ui.view">
+        <field name="name">product.template.product.form</field>
+        <field name="model">product.template</field>
+        <field name="inherit_id" ref="product.product_template_only_form_view"/>
+        <field name="arch" type="xml">
+            <xpath expr="//page[@name='variants']//field[@name='attribute_line_ids']" position="after">
+                <group>
+                    <field name="allowed_value_ids" invisible="1"/>
+                    <field name="variant_exclusion_ids" widget="one2many_list" context="{'show_attribute': True}" nolabel="1">
+                        <tree string="Exclusions" editable="bottom">
```
gdgellatly, line 12: Given that this is positioned under the attribute values, consider using top for editable, that way user can keep a view of attributes onscreen regardless of how long the exclusions get.

product_variant_exclusion/tests/test_product_variant_exclusion.py
```diff
@@ -0,0 +1,58 @@
...
+                'value_ids': [(6, False, self.attribute_size.value_ids.ids)],
+            }), (0, False, {
+                'attribute_id': self.attribute_colour.id,
+                'value_ids': [(6, False, self.attribute_colour.value_ids.ids)],
+            }), (0, False, {
+                'attribute_id': self.attribute_gender.id,
+                'value_ids': [(6, False, self.attribute_gender.value_ids.ids)],
+            })],
+            'variant_exclusion_ids': [(0, False, {
+                'attribute_value_ids': [(6, False, [v.id,
```
gdgellatly, line 53: Maybe a comment or a docstring for this, took a while to work out what was being excluded.

grindtildeath:
@gdgellatly Thank you very much for the review work. 
As this module has been abandoned by our customer, I will try to do these fixes when I'll have no more urgent tasks.

rousseldenis:
@grindtildeath What's the status of this ?

grindtildeath:
@rousseldenis Since the module wasn't used by our customer and it's v9.0, I guess we can safely close this.

PR #279, opened 2017-06-27
[IMP] product_brand: update form view for 10.0

oerp-odoo:
Old form view was out of order: form blocks misaligned, because it was
not using Odoo 10 views styling and layout.

Before change form looks like this: 

before-form

After change:

after-form

New form better use empty space in a view and form blocks are aligned properly.


Cedric-Pigeon (approved):

gdgellatly (approved):
Looks good - thanks - just reviewed the code and images, didn't worry about forcing a runbot rebuild.

lasley (approved):
Thanks @oerp-odoo

PR #276, opened 2017-06-23
[REF] product_customer_code: Migrate module product_customer_code to new api

ysantiago:

PR #282, opened 2017-06-28
[REF] product_customer_code, product_supplierinfo_for_customer: Changes in README

ysantiago:


moylop260 (approved):
👍 
This README update help us to avoid create changes that won't be approved.

(Like as us from https://github.com/OCA/product-attribute/pull/276)

moylop260:
@JulioSerna @elicoidal What do you think about?

JulioSerna (approved):

product_customer_code/README.rst
```diff
@@ -0,0 +1,27 @@
...
+    :alt: License: AGPL-3
+
+=====================
+Product customer code
+=====================
+
+This module extends of module product and allows to have different codes
+for the same product according to the customer.
+
+**NOTE**: This module is overlapped by `product_supplierinfo_for_customer
```
elicoidal: do you mean superseded?

product_customer_code/README.rst
```diff
@@ -0,0 +1,27 @@
...
+<../product_supplierinfo_for_customer>`_
```
elicoidal: I dont think the URL can work all the time. You need to specify absolute path for both.

elicoidal (changes requested):
some details

product_customer_code/README.rst
ysantiago: @elicoidal The suggested changes were applied.

elicoidal (approved):

PR #283, opened 2017-06-30
[REF] product_supplierinfo_for_customer_sale: Return all result in name_search, products found by default code and products found by customer code

ysantiago:


ysantiago:
@JulioSerna @moylop260 can you please review this change?

product_supplierinfo_for_customer_sale/models/product_product.py
```diff
@@ -13,14 +13,15 @@ class ProductProduct(models.Model):
...
+        if not name or not self._context.get('partner_id'):
             return res
         partner_id = self._context['partner_id']
         supplierinfo = self.env['product.supplierinfo'].search(
             [('name', '=', partner_id), '|', ('product_code', operator, name),
              ('product_name', operator, name)])
         product = self.search(
-            [('product_tmpl_id', '=', supplierinfo.product_tmpl_id.id)],
+            [('product_tmpl_id', 'in',
+              supplierinfo.mapped('product_tmpl_id.id'))],
```
moylop260: `s/supplierinfo.mapped('product_tmpl_id.id')/supplierinfo.mapped('product_tmpl_id').ids/g`

product_supplierinfo_for_customer_sale/models/product_product.py
```diff
@@ -13,14 +13,15 @@ class ProductProduct(models.Model):
...
             limit=limit)
-        res = product.name_get()
-        return res
+        # list(set()) to delete repeated records (if exist)
```
moylop260: Remove this comment (The code is obvious that delete repeated)

product_supplierinfo_for_customer_sale/models/product_product.py
```diff
@@ -13,14 +13,15 @@ class ProductProduct(models.Model):
...
+        return list(set(res + product.name_get()))
```
moylop260: Use `res.extend(product.name_get())` in order to use the same pointer

moylop260 (changes requested):
Please, add a test case where a result returns 2 or more items

ysantiago:
@moylop260 please, can you see the changes?

moylop260 (approved):

product_supplierinfo_for_customer_sale/models/product_product.py
```diff
@@ -13,14 +13,15 @@ class ProductProduct(models.Model):
...
              ('product_name', operator, name)])
         product = self.search(
-            [('product_tmpl_id', '=', supplierinfo.product_tmpl_id.id)],
+            [('product_tmpl_id', 'in',
+              supplierinfo.mapped('product_tmpl_id').ids)],
             limit=limit)
-        res = product.name_get()
-        return res
+        res.extend(product.name_get())
+        return list(set(res))
```
moylop260: Sorry, just one more change... 
in order to improvement the performance here we should use this transformation just if is really need it.
So IMHO is better use:
```python
if products:
    res = list(set(res + products.name_get()))
return res
```


We can save a lot of seconds by method executed, you could calculate it with:
```bash
python -m timeit -s "res=[(1, 'A'),(2, 'B'),(3, 'C'),(2, 'B'),]" "res=list(set(res))"
1000000 loops, best of 3: 0.596 usec per loop
```

product_supplierinfo_for_customer_sale/models/product_product.py
```diff
@@ -13,14 +13,15 @@ class ProductProduct(models.Model):
     def name_search(self, name='', args=None, operator='ilike', limit=80):
         res = super(ProductProduct, self).name_search(
             name, args=args, operator=operator, limit=limit)
-        if res or not self._context.get('partner_id'):
+        if not name or not self._context.get('partner_id'):
             return res
         partner_id = self._context['partner_id']
         supplierinfo = self.env['product.supplierinfo'].search(
             [('name', '=', partner_id), '|', ('product_code', operator, name),
              ('product_name', operator, name)])
```
moylop260:
```python
if not supplierinfo:
    return res
```

product_supplierinfo_for_customer_sale/models/product_product.py
moylop260: Or better...
Use a `product_ids = set(supplierinfo.mapped('product_tmpl_id').ids) - set([id for id, name in res])` from domain of search call method.

And reduce the limit from current res.
`limit=limit - len(res)`

And you will have unique products with exactly the number of records defined and you will save a lot of time reducing the query.

product_supplierinfo_for_customer_sale/tests/test_product_name_search.py
```diff
@@ -58,3 +64,23 @@ def test_10_find_product_customer_code(self):
         self.assertEquals(len(product_names), 1)
         self.assertEquals(self.product.id, product_names[0][0])
         self.assertEquals('[code_product] Name_product', product_names[0][1])
+
+        # Create in product_1 a customer code similar to product, then
+        # name_search must find two products
+        self.assertFalse(self.product_1.customer_ids)
+
+        self.supplierinfo_dict.update({
+            'product_code': 'code_test_1',
```
moylop260: Use the `product.default_code` to return 2 records
One from supplier info and other one from product default search.

moylop260 (changes requested):
sorry other changes

product_supplierinfo_for_customer_sale/tests/test_product_name_search.py
```diff
@@ -19,6 +19,7 @@ def setUp(self):
              'default_code': 'code_product', }).with_context(
                  {'partner_id': self.customer.id,
                   'supplierinfo_type': 'customer'})
+
```
moylop260: remove useless change

moylop260 (approved):
👍 
Thanks!

@JulioSerna 
What do you think about?

JulioSerna (approved):

moylop260:
@elicoidal 
What do you think about?

elicoidal (approved):
LGTM

PR #284, opened 2017-07-11
[8.0][product_dimension][FIX] Fix volume being incorrectly calculated due to rounding issues.

ivantodorovich:
Without this fix, volume was being incorrectly calculated in some cases.

For instance,

UOM: cm
Length: 57.00
Height: 8.00
Width: 19.50

Resulting volume: 0.005 when it should be 0.009 (0.008892, if not rounded)

### DIFF
--- a/product_dimension/__openerp__.py
+++ b/product_dimension/__openerp__.py
@@ -16,7 +16,7 @@
 #    along with this program.  If not, see <http://www.gnu.org/licenses/>.
 {
     'name': 'Product Dimension',
-    'version': '8.0.3.0.0',
+    'version': '8.0.2.0.1',
     'category': 'Product',
     'author':  'ADHOC SA,Camptocamp,Odoo Community Association (OCA)',
     'license': 'AGPL-3',
--- a/product_dimension/__openerp__.py
+++ b/product_dimension/__openerp__.py
@@ -16,7 +16,7 @@
 #    along with this program.  If not, see <http://www.gnu.org/licenses/>.
 {
     'name': 'Product Dimension',
-    'version': '8.0.2.0.0',
+    'version': '8.0.3.0.0',
     'category': 'Product',
     'author':  'ADHOC SA,Camptocamp,Odoo Community Association (OCA)',
     'license': 'AGPL-3',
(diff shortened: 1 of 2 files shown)


pedrobaeza (approved):

ivantodorovich:
ping @jjscarafia

gdgellatly (approved):

product_dimension/__openerp__.py
```diff
@@ -16,7 +16,7 @@
 #    along with this program.  If not, see <http://www.gnu.org/licenses/>.
 {
     'name': 'Product Dimension',
-    'version': '8.0.2.0.0',
```
lasley, line 19: This should only be increased a patch version. `8.0.2.0.1`

product_dimension/product.py
```diff
@@ -83,7 +84,8 @@ def convert_to_meters(self, measure, dimensional_uom):
         return self.env['product.uom']._compute_qty_obj(
             from_unit=dimensional_uom,
             qty=measure,
-            to_unit=uom_meters)
+            to_unit=uom_meters,
+            round=False)
```
lasley: Please add a trailing comma & put the `)` on the next line. Notice how you had to edit two lines to diff in order for one line of change? The trailing comma + `)` on its own line prevents this & reduces the possibility for conflicts.

lasley (approved):
Minor comments, otherwise 👍

lasley:
@moylop260 - the Runbot build on this is skipping for some reason. The only way I know of to identify the reasoning for this is in the server logs themselves - is there another way?

moylop260:
The .travis.yml looks like as a too old version.
Could you try updating from https://github.com/OCA/maintainer-quality-tools/blob/master/sample_files/.travis.yml template?

product_dimension/product.py
ivantodorovich: I'm not sure I understand. You mean like this?
```py
return self.env['product.uom']._compute_qty_obj(
             from_unit=dimensional_uom,
             qty=measure,
             to_unit=uom_meters,
             round=False,
)
```

ivantodorovich:
Let me know if it's ok now, and I'll squash commits.

lasley (approved):

lasley:
Thanks @ivantodorovich. No need on the squash, it's a merge option as long as we're squashing everything 👍

PR #286, opened 2017-07-18
Port to 10.0

ilyasProgrammer:


pedrobaeza:
There's already a migration of this module (which you don't reference it anyway) in https://github.com/OCA/product-attribute/pull/195, so please review that one instead of creating a new PR.

PR #292, opened 2017-09-29
[MIG][10.0]product_supplier_info_for_customer

AaronHForgeFlow:
## Use product supplier info also for customers

This modules allows to use supplier info structure, available in Inventory tab of the product form, also for defining customer information, allowing to define prices per customer and product.

Update: Added back the option to import supplierinfo prices to PL items

### Known issues
- As discussed here: #236 The module `product_customer_code` is deprecated in favor of this one.

- The product.supplierinfo form view is overwritten as the absence of name attributes make it impossible to inherit.

- This module only import prices from supplierinfo for customers not suppliers.


product_supplierinfo_for_customer/README.rst
```diff
@@ -0,0 +1,52 @@
...
+The product code / product name specified for the customer can be reflected
+on the sale orders using module `product_supplierinfo_for_customer_sale
+<https://github.com/OCA/product-attribute/tree/8.0/product_supplierinfo_for_customer_sale>`_
+
+Known issues / Roadmap
+======================
+
+* Product prices through this method are only guaranteed on the standard sale
+  order workflow. Other custom flows maybe don't reflect the price.
+* The minimum quantity will not also be applied on sale orders.
```
elicoidal: "neither will apply "

product_supplierinfo_for_customer/__init__.py
```diff
@@ -0,0 +1,5 @@
+# -*- coding: utf-8 -*-
+##############################################################################
```
elicoidal: No need for copyright notice in `__init__.py`

product_supplierinfo_for_customer/__manifest__.py
```diff
@@ -0,0 +1,26 @@
+# -*- coding: utf-8 -*-
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl.html).
```
elicoidal: Please use standard copyright headers

product_supplierinfo_for_customer/README.rst
```diff
@@ -0,0 +1,52 @@
...
+This modules allows to use supplier info structure, available in
+*Inventory* tab of the product form, also for defining customer information,
+allowing to define prices per customer and product.
+
+Configuration
+=============
+
+For these prices to be used in sale prices calculations, you will have
+to create a pricelist with a rule with option "Based on" with the value
+"Supplier prices on the product form" (although the text is not clear enough).
```
elicoidal: Why not change the name to "Partner prices on the product form"?

product_supplierinfo_for_customer/README.rst
```diff
@@ -0,0 +1,52 @@
...
+suppliers.
+
+.. image:: https://odoo-community.org/website/image/ir.attachment/5784_f2813bd/datas
+   :alt: Try me on Runbot
+   :target: https://runbot.odoo-community.org/runbot/188/10.0
+
+
+The product code / product name specified for the customer can be reflected
+on the sale orders using module `product_supplierinfo_for_customer_sale
+<https://github.com/OCA/product-attribute/tree/8.0/product_supplierinfo_for_customer_sale>`_
```
elicoidal: 10.0

product_supplierinfo_for_customer/README.rst
```diff
@@ -0,0 +1,52 @@
...
+
+Known issues / Roadmap
+======================
+
+* Product prices through this method are only guaranteed on the standard sale
+  order workflow. Other custom flows maybe don't reflect the price.
+* The minimum quantity will not also be applied on sale orders.
+* Computed fields in product.supplierinfo object won't properly work for
```
elicoidal: won't work properly
(although I do not really understand what the sentence means)

elicoidal (changes requested):
some details but overall LGTM
Functional tests are OK
Good job!

product_supplierinfo_for_customer/README.rst
AaronHForgeFlow: Sorry for the mistake. This text is obsolete as that options is not available in v10. Currently the prices in the pricelist are not passed to the SO. I'm wondering if adding this feature back is really needed or that can be implemented in other way :confused:

product_supplierinfo_for_customer/README.rst
AaronHForgeFlow: :+1: The module is not migrated yet, so I'll just remove this section. When migrating product_supplierinfo_for_customer_sale this readme will be updated

product_supplierinfo_for_customer/README.rst
AaronHForgeFlow: This is not valid in v10, so removed.

product_supplierinfo_for_customer/README.rst
elicoidal: Personally I feel it is redundant with the PL management system so I would not fight for it

product_supplierinfo_for_customer/__manifest__.py
elicoidal: Please add the copyright holder

elicoidal (changes requested):

AaronHForgeFlow:
@elicoidal Now there's no option in the pricelist to choose the price from the supplierinfo. I opened an issue here: https://github.com/OCA/product-attribute/issues/297 If it's ok to all I'll put that feature back

elicoidal:
I mean for customer function that links to supplierinfo table is redundant with standard pricelist (set up price x product).
@aheficent

AaronHForgeFlow:
@elicoidal But it's not possible to associate a pricelist to a specific customer, so if I need to specify a fix price for a customer it's not possible.

elicoidal:
@aheficent I am not quite sure what you mean by " it's not possible to associate a pricelist to a specific customer".
IMO you can:
1. Create a Pricelist called "Pricelist Agrolait"
2. add rules one by one per product / price or any other filter
3. Assign the PL to Agrolait (and not assign it to another customer for sure)

Maybe I did not understand your meaning

AaronHForgeFlow:
@elicoidal In this case step3 is totally manual, isn't it? Users have to search and select the proper PL in SO among other PL for other customers.

pedrobaeza:
@elicoidal it's not direct to have a fixed price per customer (although I don't see this as a viable way of doing business, but you know, customers demand...), so this will fill the gap.

AaronHForgeFlow:
Is ok then to add a partner_id field to product.pricelist.item model?

elicoidal:
@pedrobaeza @aheficent 

> In this case step3 is totally manual, isn't it? Users have to search and select the proper PL in SO among other PL for other customers.

As manual but far less than selecting n times the customer in supplierinfo table (in my case you have only one selection of the price list in the customer)
NB: you could add a partner_id in the PL and on_change to automatically assign the PL in the customer

> it's not direct to have a fixed price per customer (although I don't see this as a viable way of doing business, but you know, customers demand...), so this will fill the gap.

Not sure about your meaning here.

As said, I am not against the feature (and I am not blocking it) but I think it is redundant and I do not see really the advantage of your solution compared with the standard pricelist (or features that could not be done via pricelist).

AaronHForgeFlow:
As you can specify the PL in the customer I don't like the idea of putting the partner in the PL. I think that a wizard that allows to manually import the prices from the supplier.info to the PL it's better.

PS: From my side it's ok to leave the module as it is, however, the module will be losing some of the features it had in v8.

elicoidal:
@aheficent Anyway your added feature would apply for both suppliers and customer so actually I think adding it does add something that current version does not provide (PL function of product partner price).
So probably good idea to add it (up to you :tongue: )

AaronHForgeFlow:
@elicoidal Added the features but in button. Some issues found:
- It's not possible to create pricelist items in an onchange method. If this is ok I'll move the button to a wizard.
- I restricted the importing to customers because that was the purpose of this module, get the specific prices from customers.

elicoidal:
@aheficent Sorry but I am not following you. Which button and/or wizard are you talking about?

AaronHForgeFlow:
@elicoidal Button in the PL to import supplier.info into PL items

elicoidal:
@aheficent Here I am not sure to follow: this would duplicate the data (with possible sync erreur/mistake).
IMO this is not desirable: why not forward port the feature of "Supplier price in Product form" in the PL? It is not possible?

Otherwise I would suggest not to use prices in product for customers and hide them in the form.

All in all I have the feeling we are making a complex feature out of simple problem.

AaronHForgeFlow:
@elicoidal That's what I suggested first. To forward port the feature of "Supplier price in Product form" in the PL. That option was on PL items not in the PL itself because the product is selected in the PL item. I prefer this also.

elicoidal:
@aheficent In summary. options:
* forward port supplier in product: redundant (but acceptable IMO)
* No use the price in product (but then needs to be comfortable to use pricelist)

I dont like the button option (for duplicated data)

pedrobaeza:
@elicoidal It's not redundant. You are insisting each time that is redundant, but it isn't. Customer pricelists don't allow right now that behavior. Any way, if you see it acceptable although you think it's redundant, let's got for that approach.

AaronHForgeFlow:
Added back the feature on PL items to use the prices from the supplierinfo table. Currently only for customers. We can let also to import from vendors, but IMHO we have to separate that at PL item level, so in this case there would be two new options: Import from customers or import from vendors.

PS: all comments are welcome :smile:

ofonseca-pyming:
@aheficent is your desired behavior that return price 0 when there is no customer in product form ? ... or maybe i'm doing something wrong ?

AaronHForgeFlow:
@extrememicro No, in that case the list price should be taken. Isn't that happening?

ofonseca-pyming:
@aheficent No in my test.
if i have untouched the default global PL configured with sale price and create a new one with "Product price from product form", the products using the second cost 0 € on sales orders.

### Steps to reproduce it
* Activate in Sales config "Advanced prices based on formulas"
* Go to Sales > Pricelists, and create a new pricelist "Prices exceptions by customer" with an element like this:
Global pricelist based on formula "Product price from product form"
* Edit Default global pricelist based on sale price and add an element like this:
Category pricelist based on another pricelist.
In category choose "All" and in based pricelist, choose the previous one ("Product price from product form")
* Go to products, select "iPad Mini", and on Sales tab, Customers section, add an element line like this:
Customer "Agrolait"
Put a price line with 10€ and no quantity
* Go to Sales > Quotation and create a new one
* Select "Agrolait" and change pricelist to "Prices exceptions by customer"
* Add a line with "iPad Mini" product 
* Add another line with "Ipad Retina Display"

### Expected result
* "iPad Mini" price is 10 €, that corresponds to "Agrolait" price from product form.
* "Ipad Retina Display" price is 750 € that corresponds to sale price.

### Obtained result
* "iPad Mini" price is 10 €, that corresponds to "Agrolait" price from product form.
* "Ipad Retina Display" price is 0 € that corresponds to a non existent "Agrolait" price from product form.

Note that _get_price_from_supplierinfo return 0.0 when there is no supplierinfo.price.
Finally, i tried to add a new element to the default price list, the global one based on sale price, but not seems to work two global rules in the same pricelist.

AaronHForgeFlow:
@extrememicro thank you for the explanation! actually in that case the system should get the standard flow for this case. I'll check.

AaronHForgeFlow:
Did the changes to take the list_price when no supplierinfo found. Tests updated. I think this makes more sense as the default pricelist is based also in the list price

AaronHForgeFlow:
It seems to me that failing lint checks are due other modules. This is estrange because travis should ignore those. Do I need to edit the travis config file?

@mreficent you may face this issue before

MiquelRForgeFlow:
@aheficent  There is a few pylint errors in the repo that before were undetected. But they are easy to amend, so in order to have green this, could you fix those pylint red errors?

AaronHForgeFlow:
@mreficent would you help me here?

pedrobaeza:
See comments from v9.

MiquelRForgeFlow:
updated

MiquelRForgeFlow (approved):

AaronHForgeFlow:
@elicoidal IMO your comments were attended, is it possible to update your review? Thanks

elicoidal (approved):
LGTM. Merging given the feedback and time under review

pedrobaeza:
@elicoidal you have missed my comment https://github.com/OCA/product-attribute/pull/292#issuecomment-401154523. @aheficent please make another PR with the changes that you have already made on v9

elicoidal:
@pedrobaeza sorry about that: I had the feeling it was covered.

MiquelRForgeFlow:
@pedrobaeza your comments were covered.

PR #302, opened 2017-11-03
[8.0][FIX/MIG] product_manufacturer

MiquelRForgeFlow:
When @sbidoul updated the prefixes in https://github.com/OCA/product-attribute/commit/69cbb5a8a82895b3ded670e5c5ef99aef219501a, he should have updated also the name of the migration folder of the `product_manufacturer` module, migration done previously by @hbrunn in https://github.com/OCA/product-attribute/pull/94.

PR #303, opened 2017-11-03
[9.0][FIX/MIG] product_manufacturer

MiquelRForgeFlow:
Forward port of https://github.com/OCA/product-attribute/pull/302.


pedrobaeza:
You have to remove it instead of updating it.

MiquelRForgeFlow:
Is it necessary to be removed? I like to see all the old migration folders, like if it was a record history.

pedrobaeza:
This can cause problems migrating from v9 to v10 if you have a v8 migration scripts, as it will be also run.

pedrobaeza:
That's the reason each migration script should be only on the corresponding version.

MiquelRForgeFlow:
Ok.

MiquelRForgeFlow:
@pedrobaeza Done

pedrobaeza (approved):

PR #316, opened 2018-01-24
FIX override od base field in create

bealdav:
default price.type is ID 1, but according to circomstance you may have to deal with another price.type


florian-dacosta (approved):
It is more flexible this  way!
:+1:

mourad-ehm (approved):

Issue #320, opened 2018-01-30
[11.0] product_firmware_version


max3903:
This module allows you to store the latest/official/tested firmware version on an electronic product.

# Usage
- Go to Inventory > Master Data > Products
- Create or edit a product
- Enter the firmware version and save
- Check that the firmware version is saved and can be used to search

bodedra:
@max3903 

PR: #323  is ready for review.

PR #323, opened 2018-01-31
[ADD] product_firmware_version

bodedra:
#320


max3903 (changes requested):
@bodedra Please use the OCA logo from https://github.com/OCA/maintainer-tools/blob/master/template/module/static/description/icon.png

max3903 (approved):

product_firmware_version/views/product_view.xml
```diff
@@ -0,0 +1,29 @@
...
+    </record>
+    <!-- Product base search view -->
+    <record id="product_template_search_fm_version_view" model="ir.ui.view">
+        <field name="name">product.template.search.view</field>
+        <field name="model">product.template</field>
+        <field name="inherit_id" ref="product.product_template_search_view"/>
+        <field name="arch" type="xml">
+            <data>
+                <field name="name" position="attributes">
+                    <attribute name="filter_domain">['|','|','|',('default_code','ilike',self),('name','ilike',self),('barcode','ilike',self), ('firmware_version','ilike',self)]</attribute>
```
guewen: As soon as another addon will have the same idea about adding a field in this domain, only one of them will work. Wouldn't it be safer to add a new search filter? Or maybe customize the name_search?

The firmware version doesn't seem so important to me to be in the "name" search filter, am I wrong?

bodedra: Agree.

We have two options for these.

1. Adding new firmware version filter. (User need to select particular these filter option)

2. Override `name_search`. (only append domain filter)

What your suggestion ?

guewen (approved):
Thanks!

yvaucher (approved):

Issue #321, opened 2018-01-30
[11.0] stock_production_lot_firmware_version


max3903:
This module allows you to store the current firmware version on an serialized electronic item.

# Usage
- Go to Inventory > Master Data > Lot/Serial Numbers
- Create or edit a lot/serial number
- Enter the firmware version and save
- Check that the firmware version is saved and can be used to search

bodedra:
@max3903 

PR: #324 is ready for review.

PR #324, opened 2018-01-31
[ADD] stock_production_lot_firmware_version

bodedra:
#321


max3903 (changes requested):
@bodedra Please use the OCA logo from https://github.com/OCA/maintainer-tools/blob/master/template/module/static/description/icon.png

max3903 (approved):

stock_production_lot_firmware_version/views/stock_view.xml
```diff
@@ -0,0 +1,29 @@
...
+    </record>
+    <!-- Lots/Serial Numbers search view -->
+    <record id="search_product_lot_filter_fm_version" model="ir.ui.view">
+        <field name="name">Production Lots Filter</field>
+        <field name="model">stock.production.lot</field>
+        <field name="inherit_id" ref="stock.search_product_lot_filter"/>
+        <field name="arch" type="xml">
+            <data>
+                <field name="name" position="attributes">
+                    <attribute name="filter_domain">['|','|',('name','ilike',self),('ref','ilike',self),('firmware_version','ilike',self)]</attribute>
```
guewen: As soon as another addon will have the same idea about adding a field in this domain, only one of them will work. Wouldn't it be safer to add a new search filter? Or maybe customize the name_search?

The firmware version doesn't seem so important to me to be in the "name" search filter, am I wrong?

bodedra: Agree.

We have two options for these.

1. Adding new firmware version filter. (User need to select particular these filter option)

2. Override `name_search`. (only append domain filter)

What your suggestion ?

max3903: @bodedra As long as the user can search using the firmware version field in the advanced filters, we are good. No need for a specific filter, nor include it in the name_search.

guewen (approved):
Thanks!

yvaucher (approved):

Issue #322, opened 2018-01-30
[11.0] product_priority


max3903:
This module allows the user to specify a priority on a product: the higher the number, the lower the priority.
The priority can then be used in other modules for determining which products to use in different scenario.

# Usage
- Go to a Product
- Enter the Priority

bodedra:
@max3903 

PR #327 is ready for review.

PR #327, opened 2018-02-02
[ADD] product_priority

bodedra:
#322


product_priority/__init__.py
```diff
@@ -0,0 +1,4 @@
+# Copyright (C) 2017 - TODAY, Open Source Integrators
```
max3903: @bodedra `2018`

max3903 (changes requested):

max3903 (approved):

JordiBForgeFlow:
Why would you assign the priority to product.template and not product.product? As I see it, you might want to ship black t-shirt as a lower priority than yellow t-shirt when someone order the t-shirt that has 2 variants

max3903:
@jbeficent We need to order 2 different product templates, not product variants.

If the demand is for a small black 2018 t-shirt, you can't ship a large yellow 2018 one, but we could ship the small black 2017 t-shirt.

JordiBForgeFlow (approved):

JordiBForgeFlow:
Thanks for the clarification @max3903

guewen (approved):

Issue #319, opened 2018-01-30
[11.0] product_end_of_life


max3903:
This module allows you to set the End of Life Date of a Product.

Products with EoL Date in the past are excluded from the list of Products in Purchase Order so the product can not be ordered.

A scheduled action is created to know in advance the list of products whose EoL Date is approaching.
An email will be sent to the members of the Discussion Channel for Product Eol Notification Group.

# Configuration

## Discuss Channel

* Look for the the Private Channel -> Product End-Of-Life Channel.
* Click settings and add members on this channel.

## Scheduled Action

* Go to Settings > Technical > Automation > Scheduled Actions
* Look for Product: End-Of-Life Notification
* Set desired values. Run manually to test.
* Check email is sent out to the members of the Product End-Of-Life Channel.

# Usage

## Products

* Go to Inventory > Master Data > Products
* Create or edit a product
* Enter the End of Life Date and save

max3903:
In a second version of the module, implement @simahawk suggestions to improve the approaching_** fields handling (See details in #328).

bodedra:
Ready to close in favour of https://github.com/OCA/product-attribute/pull/328/

PR #328, opened 2018-02-08
[11.0] Product End-of-Life 

agyamuta:
#319


product_end_of_life/models/product.py
```diff
@@ -0,0 +1,136 @@
...
+            days = ir_cron.approaching_number * 7
+        elif ir_cron.approaching_type == 'months':
+            days = ir_cron.approaching_number * 30
+        else:
+            days = ir_cron.approaching_number
+
+        today = datetime.today().date()
+        # search products that are days approaching Eol in
+        # reference to today's date.
+        product_eol_ids = self.env['product.product'].search([
```
max3903: @agyamuta local variable 'product_eol_ids' is assigned to but never used

max3903 (changes requested):

bodedra (changes requested):
@agyamuta There are two local variables (apr_month and int_month) assigned to but never used.

bodedra (changes requested):
@agyamuta There are two local variables (apr_month and int_month) assigned to but never used.

product_end_of_life/models/product.py
```diff
@@ -0,0 +1,187 @@
...
+        ir_act_server = self.env['ir.actions.server'].search([
+            ("code", "=", "model.send_notify_product_eol()")],
+            limit=1)
+        ir_cron = self.env['ir.cron'].search([
+            ('ir_actions_server_id', '=', ir_act_server.id)],
+            limit=1)
+        # set approaching eol interval
+        if ir_cron.approaching_type == 'weeks':
+            apr_day = ir_cron.approaching_number * 7
+            apr_month = 0
```
bodedra: @agyamuta local variable `apr_month` is assigned to but never used

product_end_of_life/models/product.py
```diff
@@ -0,0 +1,187 @@
...
+        elif ir_cron.approaching_type == 'months':
+            apr_day = ir_cron.approaching_number * 30
+            apr_month = 0
+        else:
+            apr_day = ir_cron.approaching_number
+            apr_month = 0
+        # set execute every interval
+        if ir_cron.interval_type == 'weeks':
+            int_day = ir_cron.interval_number * 7
+            int_month = 0
```
bodedra: @agyamuta local variable `int_month` is assigned to but never used

product_end_of_life/models/product.py
```diff
@@ -0,0 +1,187 @@
...
+        data = "Run Date: %s" % str(run_date)
+        data += "<br/><br/><strong> [Product Name]    [Days Before Eol Date] \
+                    [End-of-Life Date] </strong>" 
+        for pid in product_eol_ids:
+           data += "<br/> [ %s ]" % pid.name
+           data += "[ %s ]" % str(datetime.strptime(pid.eol_date,\
+               "%Y-%m-%d").date() - run_date)
+           data += "[ %s ]" % str(pid.eol_date)
+        self.env['mail.message'].sudo(self._uid).create({
+            'model': 'mail.channel',
```
bodedra: remove white space after eol 47.

product_end_of_life/models/product.py
```diff
@@ -0,0 +1,187 @@
...
+        data += "<br/><br/><strong> [Product Name]    [Days Before Eol Date] \
+                    [End-of-Life Date] </strong>" 
+        for pid in product_eol_ids:
+           data += "<br/> [ %s ]" % pid.name
+           data += "[ %s ]" % str(datetime.strptime(pid.eol_date,\
+               "%Y-%m-%d").date() - run_date)
+           data += "[ %s ]" % str(pid.eol_date)
+        self.env['mail.message'].sudo(self._uid).create({
+            'model': 'mail.channel', 
+            'res_id': channel_id.id,
```
bodedra: remove white space after eol 48.

product_end_of_life/models/product.py
```diff
@@ -0,0 +1,187 @@
...
+            int_day = ir_cron.approaching_number * 30
+            int_month = 0
+        elif ir_cron.interval_type == 'days':
+            int_day = ir_cron.interval_number
+            int_month = 0
+        # for minutes/hours throw an exception
+        else:
+            int_day = 1
+            int_month = 0
+        run_day = datetime.strptime(ir_cron.nextcall, \
```
bodedra: These should be

    run_day = datetime.strptime(ir_cron.nextcall,
                                "%Y-%m-%d %H:%M:%S").date()

product_end_of_life/models/product.py
```diff
@@ -0,0 +1,187 @@
...
+        string='End of Life Date')
+
+    @api.depends('product_variant_ids', 'product_variant_ids.eol_date')
+    def _compute_eol_date(self):
+        unique_variants = self.filtered(
+            lambda template: len(template.product_variant_ids) == 1)
+        for template in unique_variants:
+            template.eol_date = template.product_variant_ids.eol_date
+        for template in (self - unique_variants):
+            template.eol_date = False
```
guewen: The last 2 lines are useless: in computed fields, when you don't assign a value to the computed field, it's automatically set to False

product_end_of_life/models/product.py
```diff
@@ -0,0 +1,187 @@
...
+            limit=1)
+        data = "Run Date: %s" % str(run_date)
+        data += "<br/><br/><strong> [Product Name]    [Days Before Eol Date] \
+                    [End-of-Life Date] </strong>" 
+        for pid in product_eol_ids:
+           data += "<br/> [ %s ]" % pid.name
+           data += "[ %s ]" % str(datetime.strptime(pid.eol_date,\
+               "%Y-%m-%d").date() - run_date)
+           data += "[ %s ]" % str(pid.eol_date)
+        self.env['mail.message'].sudo(self._uid).create({
```
guewen: Could you extract the preparation of the dict of values in another method so they can be customized by other addons? (the subject and body are likely values you would want to customize)

product_end_of_life/models/product.py
```diff
@@ -0,0 +1,187 @@
...
+            'model': 'mail.channel', 
+            'res_id': channel_id.id, 
+            'subject':'Product End-Of-Life Notification',
+            'body': data,
```
guewen: Shouldn't the subject and body be translatable?

product_end_of_life/models/product.py
```diff
@@ -0,0 +1,187 @@
...
+            'message_type':'comment',
+            'subtype_id':1})
```
guewen: Please use the xmlid, e.g. self.env.ref('mail.mt_note').id (I don't know which one is 1)

product_end_of_life/models/product.py
```diff
@@ -0,0 +1,187 @@
...
+
+    @api.model
+    def send_notify_product_eol(self):
+        # get the cronjob record
+        ir_act_server = self.env['ir.actions.server'].search([
+            ("code", "=", "model.send_notify_product_eol()")],
+            limit=1)
+        ir_cron = self.env['ir.cron'].search([
+            ('ir_actions_server_id', '=', ir_act_server.id)],
+            limit=1)
```
guewen: Why not use the xmlid of the cron job?

product_end_of_life/models/product.py
guewen: better: `fields.Datetime.from_string(ir_cron.nextcall).date()`

product_end_of_life/models/product.py
```diff
@@ -0,0 +1,187 @@
...
+        # send email to recipients from the Product Eol Discussion Channel.
+        self.email_product_eol_list(product_eol_ids, run_day)
+        return True
+
+
+class ProductVariants(models.Model):
+    _inherit = 'product.product'
+
+    eol_date = fields.Date(
+        string='End of Life Date'
```
guewen: `send_notify_product_eol` will do searches by this date, probably worth to add an index

product_end_of_life/models/product.py
```diff
@@ -0,0 +1,187 @@
...
+    )
+
+    @api.model
+    def name_search(self, name, args=None, operator='ilike', limit=100):
+        result = super(ProductVariants, self).name_search(
+            name=name,
+            args=args,
+            operator=operator,
+            limit=limit)
+        context = self._context or {}
```
guewen: `context = self.env.context`
It's never `None` so `or {}` is not needed

product_end_of_life/models/product.py
```diff
@@ -0,0 +1,187 @@
...
+        update_result = []
+        today = fields.Date.today()
+        if 'source_from' in context and \
+                context.get('source_from') == 'purchase_order':
```
guewen: `if context.get('source_from') == 'purchase_order'` is enough, no need to check if the key is there or not

product_end_of_life/models/product.py
```diff
@@ -0,0 +1,187 @@
...
+            if result:
```
guewen: Isn't `result` an empty list in that case? If yes, no need the `if result` as the loop `for res in result` will do nothing.

product_end_of_life/models/product.py
```diff
@@ -0,0 +1,187 @@
...
+                for res in result:
+                    for product in self.browse(res[0]):
```
guewen:
```
products = self.browse(res[0] for res in result)
for product in products:
```

Otherwise you'll issue a new `SELECT` for each product

product_end_of_life/models/product.py
```diff
@@ -0,0 +1,187 @@
...
+                        else:
+                            # Design product code, name
+                            def _name_get(d):
+                                name = d.get('name', '')
+                                code = \
+                                    context.get('display_default_code', True) \
+                                    and d.get('default_code', False) or False
+                                if code:
+                                    name = '[%s] %s' % (code, name)
+                                return (d['id'], name)
```
guewen: I'd move the function outside of the for loop (even remove this function altogether, see below)

product_end_of_life/models/product.py
```diff
@@ -0,0 +1,187 @@
...
+                        # Exclude expired products
+                        if product.eol_date and product.eol_date <= today:
+                            continue
+                        else:
+                            # Design product code, name
+                            def _name_get(d):
+                                name = d.get('name', '')
+                                code = \
+                                    context.get('display_default_code', True) \
+                                    and d.get('default_code', False) or False
```
guewen: Hard to read and `... and ... or ...` is to avoid.
why not:

```
if (context.get('display_default_code', True) and 
        d.get('default_code')):
    name = '[%s] %s' % (d['default_code'], name)
```

product_end_of_life/models/product.py
```diff
@@ -0,0 +1,187 @@
...
+                            variant = ", ".join([v.name for v in
+                                                 product.attribute_value_ids])
+                            name = variant and "%s (%s)" % \
+                                (product.name, variant) or product.name
+                            mydict = {
+                                'id': product.id,
+                                'name': name,
+                                'default_code': product.default_code,
+                            }
+                            update_result.append(_name_get(mydict))
```
guewen: This code seems really more complicated that it needs to be, there are too many indirections, isn't the following simpler (maybe I miss something though)?

```python
variant = ", ".join([v.name for v in
                     product.attribute_value_ids])
name = product.name
if variant:
    name = "%s (%s)" % (name, variant)
if (context.get('display_default_code', True) and 
        d.get('default_code')):
    name = '[%s] %s' % (d['default_code'], name)
update_result.append((id, name))
```

product_end_of_life/models/product.py
```diff
@@ -0,0 +1,187 @@
...
+
+    @api.model
+    def search_read(self, domain=None, fields=None, offset=0, limit=None,
+                    order=None):
+        res = super(ProductVariants, self).search_read(domain=domain,
+                                                       fields=fields,
+                                                       offset=offset,
+                                                       limit=limit,
+                                                       order=order)
+        today = str(datetime.today().date())
```
guewen: `today = fields.Date.today()`

product_end_of_life/models/product.py
```diff
@@ -0,0 +1,187 @@
...
+        context = self._context or {}
+        if 'source_from' in context and \
+                context.get('source_from') == 'purchase_order':
```
guewen: `if context.get('source_from') == 'purchase_order'` is enough

product_end_of_life/models/product.py
```diff
@@ -0,0 +1,187 @@
...
+            if not((res and domain) is None) and len(domain) > 0:
```
guewen: hmm, not sure exactly what you want, but it can be simpler, don't you want `if res and domain:`?

product_end_of_life/models/product.py
```diff
@@ -0,0 +1,187 @@
...
+                params = self.env.context and \
+                        self.env.context.get('params', False) or False
+                action = params and params.get('action', False) or False
```
guewen:
```python
params = self.env.context.get('params') or {}
action = params.get('action')
```

BTW you aliased `self.env.context` to `context` above so why not using it?

product_end_of_life/models/product.py
```diff
@@ -0,0 +1,187 @@
...
+                if not action:
```
guewen: It's not clear what purpose the params and action have and where do they come from, could you add a comment?

product_end_of_life/models/product.py
```diff
@@ -0,0 +1,187 @@
...
+                    final_res = []
+                    for product_product in res:
+                        if 'id' in product_product:
+                            for product in self.env['product.product'].browse(
+                                    product_product.get('id')):
```
guewen: Same problem than before, do not browse inside the loop.

Create a recordset first, then loop
```python
product_ids = [product['id'] for product in res if product.get('id')]
products = self.env['product.product'].browse(product_ids)
```

guewen (changes requested):

product_end_of_life/__manifest__.py
```diff
@@ -0,0 +1,21 @@
...
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+{
+    "name": "Product End-Of-Life Date",
+    "summary": "Alert users when products approach EoL Date",
+    "version": "11.0.1.0.0",
+    "license": "AGPL-3",
+    "author": "Open Source Integrators, Odoo Community Association (OCA)",
+    "category": "Product",
+    "website": "http://www.opensourceintegrators.com",
```
simahawk: https://github.com/OCA/product-attribute/

product_end_of_life/models/ir_cron.py
```diff
@@ -0,0 +1,18 @@
+# Copyright (C) 2018 - TODAY, Open Source Integrators
+# Integrators License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+from odoo import fields, models
+
+
+class IrCron(models.Model):
+    _inherit = 'ir.cron'
+
+    approaching_number = fields.Integer(
```
simahawk: :thinking: isn't this a global settings for the logic used below? I think should be move to ir.config.param

product_end_of_life/models/product.py
```diff
@@ -0,0 +1,187 @@
...
+        unique_variants = self.filtered(
+            lambda template: len(template.product_variant_ids) == 1)
+        for template in unique_variants:
+            template.eol_date = template.product_variant_ids.eol_date
+        for template in (self - unique_variants):
+            template.eol_date = False
+
+    @api.multi
+    def _inverse_eol_date(self):
+        self.ensure_one()
```
simahawk: this should stay multi

product_end_of_life/models/product.py
```diff
@@ -0,0 +1,187 @@
...
+        if len(self.product_variant_ids) == 1:
+            self.product_variant_ids.eol_date = self.eol_date
+
+    @api.model
+    def email_product_eol_list(self,product_eol_ids, run_date):
```
simahawk: run flake8 to spot linting issue ;)

product_end_of_life/models/product.py
```diff
@@ -0,0 +1,187 @@
...
+        if not product_eol_ids:
+           return
+        channel_id = self.env['mail.channel'].search([
+            ('name', '=', 'Product End-Of-Life Channel')],
```
simahawk: do not search by name, use `self.env.ref('my_nice_xmlid')`

product_end_of_life/models/product.py
```diff
@@ -0,0 +1,187 @@
...
+            limit=1)
+        data = "Run Date: %s" % str(run_date)
+        data += "<br/><br/><strong> [Product Name]    [Days Before Eol Date] \
```
simahawk: do not compose HTML, use a qweb template

product_end_of_life/models/product.py
```diff
@@ -0,0 +1,187 @@
...
+        ir_cron = self.env['ir.cron'].search([
+            ('ir_actions_server_id', '=', ir_act_server.id)],
+            limit=1)
+        # set approaching eol interval
+        if ir_cron.approaching_type == 'weeks':
+            apr_day = ir_cron.approaching_number * 7
+            apr_month = 0
+        # TBD: need to correctly compute based on months.
+        elif ir_cron.approaching_type == 'months':
+            apr_day = ir_cron.approaching_number * 30
```
simahawk: see my comment above about moving this to a config param

product_end_of_life/models/product.py
simahawk: I'd go for a `search_read(prod_ids,  [fields, that, you, need])`

simahawk (changes requested):

simahawk:
would be nice to have tests too. :wink: 
Small tip: add them before refactoring according to comments. In this way you will be sure that it respects the behavior of the initial design. It will also help you to understand which parts of the code should stay and should go away.

product_end_of_life/models/product.py
agyamuta: Corresponded. Thanks for pointing out.

product_end_of_life/models/ir_cron.py
max3903: @simahawk It's not global. It is specific to this scheduled action.

Having it here also makes it easier for the admin to configure everything in one place.

agyamuta: @simahawk thanks for the thought. In addition to @max3903 if we move it to ir.config.param it has provided only 2 fields - key/value and we need value and selections(days/weeks/months) for users to set it.

product_end_of_life/models/product.py
agyamuta: Got it.. Thanks.

product_end_of_life/models/product.py
agyamuta: Corresponded.. Thanks.

product_end_of_life/models/product.py
agyamuta: in this function when fields is used to get today's date an error 'AttributeError: 'list' object has no attribute 'Date'' since as you can see on the function parameter that field is being used.

product_end_of_life/models/product.py
agyamuta: thank you. simplified the code.

product_end_of_life/models/product.py
agyamuta: thank you. simplified the code.

product_end_of_life/models/ir_cron.py
simahawk: how many crons do you expect to have? I mean: is the user supposed to create cron records? 
Very likely not :)
... if this is true then the cron is not the place where to store this information IMHO.

Plus, if you want you can store more then one info in one single param separated by comma (or anything else).
In any case, it won't be that much hard to add a new method on the product that does:
```python
    @tools.ormcache('self')
    def _get_eol_settings(self):
        params = self.env["ir.config_parameter"].sudo()
        return {
            'approaching_days': params.get('product_eol.approaching_days'),
            'approaching_type': params.get('product_eol.approaching_type'),
        }
```
Easy and decoupled from the rest of the code. So, in the future, if you want to change the place where settings come from you deal only w/ this method.

simahawk: regarding 

> Having it here also makes it easier for the admin to configure everything in one place.

you can add the selection fields to res.config.settings and make stored on config param keys.[ This is done already in many places](https://github.com/odoo/odoo/blob/11.0/addons/auth_signup/models/res_config_settings.py#L35).

simahawk: another option, if you really like to config this from the cron, would be to pass them as keyword args to the cron call. Like:
```xml
<field name="code">model.send_notify_product_eol(approaching_type='days', approaching_number=1)</field>
```
so the user can still set everything from the cron. But, again, I think that this settings are one-shot setting and you don't play w/ them every day :)

max3903 (approved):

max3903:
@simahawk Thanks for your constructive review. Unfortunately we don't have the time to take all your comments into account. @agyamuta has other developments in the pipe.

I updated the issue #319 mentioning the remaining improvements and added the link in the Roadmap section of the module description. We can take care of it later and merge this PR in the mean time. Sounds good?

bodedra (approved):

p-tombez:
@max3903 I took over this PR in order to make the improvements and add the tests as suggested by @simahawk 
As I'm not aware of the context nor the technical specifications, could you provide me some insights about the actual needs (e.g. do we want to send a notification for each product each time the cron is executed or only the first time ? What is the cron behavior if the EoL date is changed on a product ?)
I'm asking you that because I discussed with @guewen and spotted some flaws and I think the code needs a little makeover.

max3903:
@p-tombez the frequency of the cron defines the width of the window. The approaching delay defines the start date of the window.

We want to be able to receive one notification for the product whose eol date is in the window.

Of course if the product eol date is changed to the future, you will get 2 notifications: each time the product eol date is in the window.

The approaching delay allows the user to have some time to find an alternative product. Otherwise he may receive a notification for product "dying" tomorrow.

p-tombez:
@agyamuta @max3903 Can you rebase 11.0 so when I do the pull-request, it will not be polluted with irrelevant changes ?

p-tombez:
I think the last operation kind of messed up the PR. There are some files I deleted that are here again (like models/ir_cron.py and views/ir_cron_view.xml)

max3903:
@p-tombez @simahawk @guewen I believe everything has been taken care of. Can you update your review?

product_end_of_life/data/mail_channel_data.xml
```diff
@@ -0,0 +1,17 @@
+<?xml version="1.0" encoding="utf-8"?>
+<odoo noupdate="1">
+        <record id="channel_product_eol" model="mail.channel">
+            <field name="name">Product End-Of-Life Channel</field>
+            <field name="description">Product End-Of-Life Group Notification.</field>
+            <field name="email_send">True</field>
+            <field name="public">private</field>
+            <field name="channel_last_seen_partner_ids" eval="[(4, ref('base.user_root'))]"/>
+        </record>        
+        <record id="mail_message_product_eol" model="mail.message">
```
simahawk: I don't think you need this. You can freeze the values in the method `_prepare_msg_values`

product_end_of_life/models/product.py
```diff
@@ -0,0 +1,101 @@
...
+
+        cron = self.env.ref('product_end_of_life.ir_cron_notify_product_eol')
+        cron.write({'interval_number': interval,
+                    'interval_type': uot,
+                    'nextcall': datetime.combine(next_win_date_start,
+                                                 datetime.min.time())})
+        return True
+
+    @api.model
+    def name_search(self, name, args=None, operator='ilike', limit=100):
```
simahawk: FYI to avoid duplicating this code you could override `_search` (name_search uses `_search` and `search_read` uses `search under the hood).

simahawk (approved):
LGTM. Approving but I'd like to drop the mail.message record which is not useful at all and might be deleted fro db.

product_end_of_life/models/product.py
```diff
@@ -0,0 +1,101 @@
+# Copyright (C) 2018 - TODAY, Open Source Integrators
+# Integrators License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+from datetime import datetime, date
+from dateutil.relativedelta import relativedelta
+from odoo import api, fields as fields_cls, models
```
leemannd: I see no reason to import `fields` as `fields_cls`

leemannd (changes requested):
Only one nitpicking question

product_end_of_life/models/product.py
p-tombez: Method on line https://github.com/OCA/product-attribute/pull/328/files#diff-cabc9687719397e1da647f8dc3ab2b25R92 has a named attribute ```fields```. Therefore, the line https://github.com/OCA/product-attribute/pull/328/files#diff-cabc9687719397e1da647f8dc3ab2b25R95 will fail

max3903:
@agyamuta There is still 2 comments from @simahawk that needs to be taken care of. Can you make the changes so we can merge this PR?

agyamuta:
@max3903 still working on it. Testing the latest code unfortunately, mail channel is not showing and posting of the message is also not working.

product_end_of_life/data/mail_channel_data.xml
agyamuta: @simahawk thanks for the comment. This is updated on the latest code.

product_end_of_life/models/product.py
agyamuta: @simahawk thanks for the comment. This is updated on the latest code.

max3903:
@guewen Can you update your review?

product_end_of_life/__manifest__.py
simahawk: ping?

product_end_of_life/tests/__init__.py
```diff
@@ -0,0 +1,5 @@
+# -*- coding: utf-8 -*-
```
simahawk: remove coding from *.py files

product_end_of_life/views/purchase_view.xml
```diff
@@ -0,0 +1,14 @@
+<?xml version="1.0"?>
+<odoo>
+    <record id="purchase_order_form_product_eol" model="ir.ui.view">
+        <field name="name">purchase.order.form.product.eol</field>
+        <field name="model">purchase.order</field>
+        <field name="inherit_id" ref="purchase.purchase_order_form"/>
+        <field name="arch" type="xml">
+            <xpath expr="//sheet//notebook//page//field//tree//field[@name='product_id']"
```
simahawk: `//sheet//notebook//page//field` which field? Order lines? If yes, use `//field[@name='order_line']//field[@name='product_id']`

simahawk (changes requested):

product_end_of_life/views/purchase_view.xml
```diff
@@ -5,9 +5,12 @@
         <field name="model">purchase.order</field>
         <field name="inherit_id" ref="purchase.purchase_order_form"/>
         <field name="arch" type="xml">
-            <xpath expr="//sheet//notebook//page//field//tree//field[@name='product_id']"
+            <xpath expr="//sheet//notebook//page//field[@name='order_line']//tree//field[@name='product_id']"
```
bodedra: Please change from `<xpath expr="//sheet//notebook//page//field[@name='order_line']//tree//field[@name='product_id']" ` to `<xpath expr="//field[@name='order_line']//field[@name='product_id']"` as @simahawk  mentioned. ;-)

bodedra (approved):

Issue #331, opened 2018-02-16
product_internal_reference_unique & product_internal_reference_mandatory


max3903:
# product_internal_reference_unique 

This module adds a constraint on the internal reference of the product to make it unique across the database.

# product_internal_reference_mandatory

This module adds a constraint on the internal reference of the product to make it mandatory.

max3903:
Ping @agyamuta

bodedra:
In favour of #333 and #334 , we can close these.

PR #333, opened 2018-02-28
[11.0] product_internal_reference_unique

agyamuta:
Correspondence in reference to issue https://github.com/OCA/product-attribute/issues/331


max3903:
@agyamuta Can you provide a demo data file with an internal reference for the demo products that doesn't have one? Thanks

pedrobaeza:
We are calling everywhere "internal reference" as "code", so I think you can name your module as *product_code_unique*

max3903 (approved):

max3903:
@agyamuta Can you apply the same logic as in #334 with the pre_init_hook?

agyamuta:
@max3903 Applied logic as in #334 . Thanks.

product_code_unique/__init__.py
```diff
@@ -0,0 +1,12 @@
+# Copyright (C) 2018 - TODAY, Open Source Integrators
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+from . import models
+
+
+def pre_init_product_code(cr):
```
feketemihai, line 7: You will have to treat templates as weel as the other module..

feketemihai:
Some tests would help..

feketemihai (changes requested):

product_code_unique/README.rst
```diff
@@ -0,0 +1,60 @@
+.. image:: https://img.shields.io/badge/licence-AGPL--3-blue.svg
+   :target: http://www.gnu.org/licenses/agpl-3.0-standalone.html
+   :alt: License: AGPL-3
+
+==================================
+Unique Product Internal Reference
+==================================
```
JayVora-SerpentCS: Need to remove one '=' to align with the last 'e' of reference

product_code_unique/models/product.py
```diff
@@ -0,0 +1,14 @@
...
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+from odoo import fields, models
+
+
+class ProductProduct(models.Model):
+    _inherit = 'product.product'
+
+    default_code = fields.Char('Internal Reference', index=True)
+
```
JayVora-SerpentCS: Was it needed to write the field definition again ?Cannot we add only the constraint?

agyamuta: Thanks for pointing out. Tested and it works the same. This is corresponded.  Unlike with lower odoo version if not redefined changes wont take effect.

product_code_unique/__init__.py
agyamuta: Product Templates are not mandatory but required when it has by default 1 variant only. So no need to initialize the value in this case.

agyamuta: image

max3903: @feketemihai 

FYI, @agyamuta will change the method to find and remove duplicate default_code in product_product.

We don't need to touch product_template as the default_code of a product template with many variants will be empty:
https://github.com/odoo/odoo/blob/11.0/addons/product/models/product_template.py#L256

The uniqueness constraint is maintained with empty values.

product_code_unique/models/product.py
```diff
@@ -0,0 +1,12 @@
+# Copyright (C) 2018 - TODAY, Open Source Integrators
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+from odoo import fields, models
```
JayVora-SerpentCS: remove fields, as its unused

product_code_unique/README.rst
```diff
@@ -0,0 +1,60 @@
...
+This module adds a constraint on the internal reference of the product 
+to make it unique across the database.
+
+Usage
+=====
+* Unable to save a product when a new internal reference or default_code value is
+  the same with an existing record.
+* A pre_init_hook process is initiated when there exist records without an internal reference(default_code).
+  A default value is generated to populate empty field as a temporary value.
+
```
JayVora-SerpentCS, line 18: Add Runbot link.

JayVora-SerpentCS (changes requested):

max3903:
@JayVora-SerpentCS @feketemihai Can you update your review please?

JayVora-SerpentCS (approved):

PR #334, opened 2018-02-28
[11.0] product_internal_reference_mandatory

agyamuta:
Correspondence in reference to issue https://github.com/OCA/product-attribute/issues/331


max3903:
@agyamuta Same comment as in https://github.com/OCA/product-attribute/pull/333#issuecomment-369245794

max3903:
@agyamuta You need to define a `pre_init_hook` to load the data file before installation. See at the bottom of this page:
https://www.odoo.com/documentation/11.0/reference/module.html

agyamuta:
Thanks for the comment @max3903. I have the following issues in using pre_init_hook.
a. With the current setup, module is first attempted to be installed prior to load the demo data. Installation failed  due to the existing data thats violates Null data for internal reference.
b. In the pre_init_hook, how to identify that the DB instance is configured to load Demo Data.
image

max3903:
@agyamuta You can check in the module table on one of the dependency. There is a boolean field `demo`.

bodedra (approved):
LGTM. Thanks.

max3903 (approved):

product_code_mandatory/__init__.py
```diff
@@ -0,0 +1,16 @@
...
+
+from . import models
+
+
+def pre_init_product_code(cr):
+    cr.execute("""UPDATE product_template
+        SET default_code = 'DEFAULT' || nextval('ir_default_id_seq')
+        WHERE default_code is NULL
+        OR LENGTH(default_code) = 0""")
+    cr.execute("""UPDATE product_product
```
feketemihai, line 12: I would update the product.product with a reference from template, let's say the product template has DEFAULT100, and on variants associated you get DEFAULT100-1, DEFAULT100-2, when you search by template code on products, to have the variantss from it...

product_code_mandatory/README.rst
```diff
@@ -0,0 +1,62 @@
+.. image:: https://img.shields.io/badge/licence-AGPL--3-blue.svg
+    :target: http://www.gnu.org/licenses/agpl-3.0-standalone.html
+    :alt: License: AGPL-3
+
+======================================================
+Set as Mandatory field for  Product Internal Reference
```
feketemihai: Use module name here..

product_code_mandatory/README.rst
```diff
@@ -0,0 +1,62 @@
+.. image:: https://img.shields.io/badge/licence-AGPL--3-blue.svg
```
feketemihai: Update with png image

product_code_mandatory/__manifest__.py
```diff
@@ -0,0 +1,21 @@
+# Copyright (C) 2018 - TODAY, Open Source Integrators
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+{
+    "name": "Product Internal Reference as Mandatory",
```
feketemihai: Use Required instead on Mandatory

product_code_mandatory/models/product.py
```diff
@@ -0,0 +1,15 @@
+# Copyright (C) 2018 - TODAY, Open Source Integrators License AGPL-3.0
+# or later (http://www.gnu.org/licenses/agpl).
+
+from odoo import fields, models
+
+
+class ProductProduct(models.Model):
```
feketemihai, line 7: by default you create a product template, and you can add variants, the pre_init_hook is adding a default code to templates first, so here i think you should use the same, the product.template model.

feketemihai (changes requested):

product_code_mandatory/__init__.py
agyamuta: @feketemihai thanks for the comment. 

Let me quote Max' comment:
We don't need to touch product_template as the default_code of a product template with many variants will be empty:

https://github.com/odoo/odoo/blob/11.0/addons/product/models/product_template.py#L256-L266

With Odoo's default behavior when a product.template is created it creates a product.product record in which the product.product.default_code is initialized with a unique value. This value then is pushed back to product.template.default_code.

max3903: @agyamuta Be careful my comment was referring to the case of uniqueness. In this case, the field cannot be empty.

product_code_mandatory/models/product.py
agyamuta: The pre_init_hook process in this module is to make sure that all existing records(product_product table) would have an initial value and not empty or blank. This is only done once upon installation of the module.

max3903:
@feketemihai Can you update your review?

product_code_mandatory/__manifest__.py
```diff
@@ -0,0 +1,21 @@
...
+    "category": "Product",
+    "website": "http://www.opensourceintegrators.com",
+    "depends": ["product"],
+    "data": [
+        "views/product_view.xml",
+        "views/product_code_seq.xml",
+    ],
+    'demo': [
+    ],
+    "pre_init_hook": 'pre_init_product_code',
```
MeetKD: Can you please remove empty key ?

PR #306, opened 2017-11-26
Migrate product_dimension v11.0

awa-si:

PR #337, opened 2018-03-12
11.0 mig product dimension

ghost:


pedrobaeza:
Please include #339 and squash a bit commit history by logical set/author.

leemannd (changes requested):
They are steps missing from a migration 9.0 -> 10.0 
https://github.com/OCA/maintainer-tools/wiki/Migration-to-version-10.0

product_dimension/__init__.py
```diff
@@ -0,0 +1,5 @@
+# -*- coding: utf-8 -*-
+# © 2016 Camptocamp SA
```
leemannd: You can remove copyright from `init` files https://github.com/OCA/maintainer-tools/blob/master/template/module/__init__.py

product_dimension/__manifest__.py
```diff
@@ -0,0 +1,17 @@
+# -*- coding: utf-8 -*-
+# © 2018 brain-tec AG (http://www.braintec-group.com)
+# © 2015-2016 Camptocamp SA
+# © 2015 ADHOC SA  (http://www.adhoc.com.ar)
```
leemannd: Copyright symbol can be replaced s/©/Copyright/

product_dimension/models/product.py
```diff
@@ -0,0 +1,79 @@
+# -*- coding: utf-8 -*-
+# © 2015 ADHOC SA  (http://www.adhoc.com.ar)
+# © 2015-2016 Camptocamp SA
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+
+from openerp import models, fields
```
leemannd: All `openerp` import should be replaced by `odoo`

leemannd (changes requested):

leemannd:
Sidenote: there is an ongoing migration for V11. But it is not going forward => https://github.com/OCA/product-attribute/pull/306
There is also a migration for V10.0 that we should take care of. With a discussion about the views https://github.com/OCA/product-attribute/pull/195/files

ghost:
@leemannd thanks for your input
I removed dependency to stock and added suggested improvement aswell as german translation

product_dimension/__init__.py
```diff
@@ -0,0 +1,4 @@
+# -*- coding: utf-8 -*-
```
asaunier: Minor point but this line is no longer required in v11 (Python 3), see https://github.com/OCA/maintainer-tools/wiki/Migration-to-version-11.0#tasks-to-do-in-the-migration
(ditto for the other py files)

product_dimension/README.rst
```diff
@@ -0,0 +1,47 @@
+.. image:: https://img.shields.io/badge/licence-AGPL--3-blue.svg
+    :alt: License: AGPL-3
+
+Product Dimension
+=================
+Add dimensions (length, width and height) to products. Find the volume
+automatically when you change one of these dimensions.
+
+This module was previously hosted on github.com/ingadhoc/odoo-addons and
```
asaunier: Using the full URL https://github.com/adhoc-dev/odoo-addons would make the link clickable.

product_dimension/README.rst
```diff
@@ -0,0 +1,47 @@
...
+
+
+Credits
+=======
+
+Contributors
+------------
+* Juan Jose Scarafia <jjs@ingadhoc.com>
+* Leonardo Pistone <leonardo.pistone@camptocamp.com>
+* Denis Leemann <denis.leemann@camptocamp.com>
```
asaunier: You can add your name here :)

product_dimension/i18n/de.po
```diff
@@ -0,0 +1,63 @@
+# Translation of Odoo Server.
```
asaunier, line 1: Nice!
But aren't [the other languages available in v9](https://github.com/OCA/product-attribute/tree/9.0/product_dimension/i18n) missing?
Isn't a "OCA Transbot updated translations from Transifex" commit missing, providing the translations done in Transifex?
See https://github.com/OCA/maintainer-tools/wiki/Merge-commits-in-pull-requests#mergesquash-the-commits-generated-by-transifex

ghost: good point: I created migration based on 10.0 branch and there the i18n folder is missing.. 
https://github.com/OCA/product-attribute/tree/10.0/product_dimension

how can I add the missing files?

ghost:
@asaunier thanks for the review. I've added the changes.. the i18n files are still missing.. any idea how they can be added so the git history is intact?

asaunier:
@BT-kaberer Thanks for the changes!

It's indeed a bit weird that the i18n files have not been integrated in branch 10.0 !? I don't know if there was a reason, or maybe they were simply forgotten?

To get them back you could for instance try to create a new dummy migration branch from branch 9.0 and run the ``git format-patch`` as described in https://github.com/OCA/maintainer-tools/wiki/Migration-to-version-10.0#technical-method-to-migrate-a-module-from-90-to-100-branch
Then you squash the "OCA Transbot" commits and cherry-pick the resulting commit to your existing branch (you might have a conflict with your ``de.po`` file).

Alternately it is maybe acceptable to copy-paste the PO files by hand in your branch and commit them yourself, since anyway [it is recommended to squash the i18n commits](https://github.com/OCA/maintainer-tools/wiki/Merge-commits-in-pull-requests#mergesquash-the-commits-generated-by-transifex) when migrating a module, thus not preserving the translations history.

product_dimension/__manifest__.py
```diff
@@ -0,0 +1,16 @@
...
+    'name': 'Product Dimension',
+    'version': '11.0.1.0.0',
+    'category': 'Product',
+    'author': 'brain-tec AG, ADHOC SA, Camptocamp, '
+              'Odoo Community Association (OCA)',
+    'license': 'AGPL-3',
+    'depends': ['product'],
+    'data': ['views/product_view.xml'],
+    'installable': True,
+    'images': [],
```
ghost: as we have `product_dimension/static/description/icon.png` then maybe we should add it to images?
```
    'images': [
        'static/description/icon.png',
    ],
```

product_dimension/models/product.py
```diff
@@ -0,0 +1,78 @@
+# Copyright 2015 ADHOC SA  (http://www.adhoc.com.ar)
+# Copyright 2015-2016 Camptocamp SA
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+
+from odoo import models, fields
+from odoo import api
```
ghost: can be in one row )

product_dimension/__manifest__.py
```diff
@@ -0,0 +1,16 @@
+# Copyright 2018 brain-tec AG (http://www.braintec-group.com)
+# Copyright 2015-2016 Camptocamp SA
+# Copyright 2015 ADHOC SA  (http://www.adhoc.com.ar)
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+{
+    'name': 'Product Dimension',
+    'version': '11.0.1.0.0',
+    'category': 'Product',
+    'author': 'brain-tec AG, ADHOC SA, Camptocamp, '
```
ghost: `Camptocamp SA`

product_dimension/models/product.py
```diff
@@ -0,0 +1,78 @@
...
+
+
+class ProductTemplate(models.Model):
+    _inherit = 'product.template'
+
+    @api.onchange('length', 'height', 'width', 'dimensional_uom_id')
+    def onchange_calculate_volume(self):
+        if (not self.length or not self.height or not self.width or
+                not self.dimensional_uom_id):
+            return False
```
ghost: If we check if all required values are present
```
if (self.length and self.height and self.width and self.dimensional_uom_id):
```
and put `volume` calculation under this `if` then we should increase code coverage.

Same for `'product.product'`

leemannd (approved):

product_dimension/models/product.py
```diff
@@ -3,23 +3,24 @@
...
+        if self.length and self.height and self.width and \
+                self.dimensional_uom_id:
+            length_m = self.convert_to_meters(
+                self.length, self.dimensional_uom_id)
+            height_m = self.convert_to_meters(
+                self.height, self.dimensional_uom_id)
+            width_m = self.convert_to_meters(
+                self.width, self.dimensional_uom_id)
+            self.volume = length_m * height_m * width_m
+        return False
```
asaunier: With this change, the function will always return ``False``. Is it really the wanted behaviour?
Wasn't the function already OK?
Ditto for the function in ``ProductTemplate``.

asaunier (commented):
A question about your changes in the product model.
Thanks for adding the translations back!

product_dimension/README.rst
```diff
@@ -0,0 +1,48 @@
+.. image:: https://img.shields.io/badge/licence-AGPL--3-blue.svg
+    :alt: License: AGPL-3
+
+Product Dimension
```
simahawk: update readme according to https://github.com/OCA/maintainer-tools/blob/master/template/module/README.rst

product_dimension/README.rst
```diff
@@ -0,0 +1,48 @@
+.. image:: https://img.shields.io/badge/licence-AGPL--3-blue.svg
```
simahawk: .png

product_dimension/models/product.py
```diff
@@ -0,0 +1,81 @@
...
+            ('category_id', '=', self.env.ref('product.uom_categ_length').id)
+        ]
+
+    length = fields.Float()
+    height = fields.Float()
+    width = fields.Float()
+    dimensional_uom_id = fields.Many2one(
+        'product.uom',
+        'Dimensional UoM',
+        domain=_get_dimension_uom_domain,
```
simahawk: `domain=lambda self: self._get_dimension_uom_domain(),`

product_dimension/models/product.py
```diff
@@ -0,0 +1,81 @@
...
+
+from odoo import models, fields, api
+
+
+class Product(models.Model):
+    _inherit = 'product.product'
+
+    @api.onchange('length', 'height', 'width', 'dimensional_uom_id')
+    def onchange_calculate_volume(self):
+        if self.length and self.height and self.width and \
```
simahawk: these 2 methods are repeated into prod template: move them to simple functions and call them where needed

simahawk (changes requested):

product_dimension/models/product.py
ghost: OK correct. I changed to previous behavior. Function was re-written to increase code coverage

product_dimension/models/product.py
ghost: I added your improvements except this one: 
1) these functions are already simple functions. 
one could, if it makes sense, encapsulate further and move the function on product.product and call it from product.template

2) the PR has already been approved by @leemannd

leemannd: @simahawk @BT-kaberer It'm not really in favour of too much refactoring nor improvements during migrations.
But in this one seems legit. It could be great if you take the time to consider @simahawk 's improvement.

nomadic-coding: @simahawk @leemannd 
done

ghost: @leemannd @simahawk 
done

product_dimension/README.rst
```diff
@@ -1,37 +1,65 @@
...
+This module extends the functionality of product to support
+dimensions (length, width and height). Find the volume
 automatically when you change one of these dimensions.
 
-This module was previously hosted on https://github.com/ingadhoc/odoo-addons and
-before that on https://launchpad.net/~ingenieria-adhoc.
+This module was previously hosted on https://github.com/ingadhoc/odoo-addons
+and before that on https://launchpad.net/~ingenieria-adhoc.
+
+Installation
```
simahawk: you can drop these sections if they don't provide additional info.

product_dimension/models/product.py
```diff
@@ -50,18 +37,25 @@ def _get_dimension_uom_domain(self):
...
-                self.length, self.dimensional_uom_id)
-            height_m = self.convert_to_meters(
-                self.height, self.dimensional_uom_id)
-            width_m = self.convert_to_meters(
-                self.width, self.dimensional_uom_id)
-            self.volume = length_m * height_m * width_m
-        return False
+        volume = self._calc_volume(
+            self.length, self.height, self.width, self.dimensional_uom_id)
+        if isinstance(volume, bool):
```
simahawk: `if volume is False:`

product_dimension/models/product.py
```diff
@@ -0,0 +1,75 @@
...
+            width_m = self.convert_to_meters(width, uom_id)
+            volume = length_m * height_m * width_m
+
+        return volume
+
+    @api.onchange('length', 'height', 'width', 'dimensional_uom_id')
+    def onchange_calculate_volume(self):
+        volume = self._calc_volume(
+            self.length, self.height, self.width, self.dimensional_uom_id)
+        if isinstance(volume, bool):
```
simahawk: `if volume is False:` is more explicit.
Also, since this is an onchange, when volume is not valued don't you want to reset the value?
If parameters do not match then the volume should be updated as well. No?

product_dimension/README.rst
```diff
@@ -0,0 +1,76 @@
...
+=================
+
+This module extends the functionality of product to support
+dimensions (length, width and height). Find the volume
+automatically when you change one of these dimensions.
+
+This module was previously hosted on https://github.com/ingadhoc/odoo-addons
+and before that on https://launchpad.net/~ingenieria-adhoc.
+
+Installation
```
simahawk: drop empty sections

simahawk (changes requested):

product_dimension/models/product.py
leemannd: Thanks

product_dimension/models/product.py
ghost: yes you are right. i changed it. It was just so the function behaved like before migration. but now make more sense to me

asaunier (approved):

simahawk (approved):

leemannd:
@oleksandrpaziuk Could you update your review?

ghost (approved):
LGTM!
But newline required in the end of *.po files

ivantodorovich (approved):
LGTM

jjscarafia (approved):

pedrobaeza:
Can you please squash your migration commits together before merging?

florian-dacosta (approved):
Seems good.
One question though, wouldn't it be better to put the volume field as a computed stored field?
This module is based only on the onchange mechanism, which is a bit weak for this kind of data no?
For instance, if I always update my products by csv file import, if I change the dimension, the volume won't be recomputed right?

ivantodorovich:
@florian-dacosta the thing is that if we use a calculated field, you then loose the posibility to edit the volume. Maybe we can achieve the same with and inverse function and a hidden field.

I agree that using onchange for this is not ideal. I've seen some weird behaviours with this module in previous versions, sadly I don't remember them to replicate

LoisRForgeFlow:
As a migration effort, isn't this ready to merge? Any extra discussion/improvement belong to a separate issue/PR at this point I think.

asaunier:
@lreficent +1
6 approvals and no request for change it seems. Is there still something missing before the PR is merged?

rvalyi (approved):

rvalyi:
dear contributors, so after approving this PR myself too yesterday I went with the merge. But I let you know that today I made a new PR to migrate product_dimension further to v12 here
https://github.com/OCA/product-attribute/pull/485
So your review is welcome...

PR #338, opened 2018-03-12
[9.0][product_dimension][FIX] Fix volume being incorrectly calculated due to rounding issues (#284)2

ivantodorovich:
This fix from 8.0 was not migrated to 9.0


pedrobaeza (approved):
Fast-tracking this as it was previously reviewed

PR #339, opened 2018-03-12
[10.0][product_dimension][FIX] Fix volume being incorrectly calculated due to rounding issues (#284)

ivantodorovich:
I don't know if this module was already migrated or not. But, just in case..


pedrobaeza (approved):
Fast-tracking this as it was previously reviewed

PR #341, opened 2018-03-20
[11.0][MIG/IMP] product_sequence

LoisRForgeFlow:
Done in this PR:
* Migration to v11.
* Enhancement to support the definition different sequences for product categories. 

Product Sequence
================

This module allows to associate a sequence to the product reference.
The reference (default code) is unique (SQL constraint) and required.

You can optionally specify different sequences for different product
categories.

Usage
=====

To specify a different sequence for a product category proceed as follows:

#. Go to the a Product Category form view.
   (**note:** you will need to install Inventory app to be able to access to
   the form view, *Inventory > Configuration > Products > Products Categories*;
   or create a menuitem manually).
#. Fill the *Prefix for Product Internal Reference* as desired.


JordiBForgeFlow (approved):
:+1: Technical and functional review

MiquelRForgeFlow (changes requested):
Needs more tests

product_sequence/models/product_product.py
```diff
@@ -47,20 +47,20 @@ def write(self, vals):
...
-                sequence = product.categ_id.sequence_id
+        for product in self:
+            if vals.get('default_code', '') == '/':
+                category_id = vals.get('categ_id', product.categ_id.id)
+                category = self.env['product.category'].browse(category_id)
+                sequence = category.sequence_id
                 if not sequence:
                     sequence = self.env.ref(
                         'product_sequence.seq_product_auto')
                 ref = sequence.next_by_id()
```
MiquelRForgeFlow: could be added a logger info to show that ref was changed. Like "Internal Reference XX of product  YY was changed to ZZ"

MiquelRForgeFlow (approved):

PR #201, opened 2016-11-10
[MIG][10.0] product_multi_image

lasley:
This is WIP & served to rule out old API wire crossing as the 9.0 upgrade issue in #199. We're getting the same effect here, so that's not the case.

I'll fix this PR up once the issue is figured out in the 9.0 version. 

Depends:
- [x] #199 
- [x] OCA/server-tools#600
- [x] OCA/server-tools#629
- [x] OCA/server-tools#677
- [x] OCA/server-tools#709

PR #343, opened 2018-03-28
[10.0] Migrate product_categ_image

alexis-via:
The v8 version of product_categ_image is available here https://github.com/OCA/product-attribute/pull/201

### DIFF
--- a/product_categ_image/models/product_category.py
+++ b/product_categ_image/models/product_category.py
@@ -1,0 +1,13 @@
+# -*- coding: utf-8 -*-
+#
+# © 2016 Akretion (http://www.akretion.com)
+# @author Alexis de Lattre <alexis.delattre@akretion.com>
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+
+from openerp import models, fields
+
+
+class ProductCategory(models.Model):
+    _inherit = 'product.category'
+
+    image = fields.Binary(string='Image')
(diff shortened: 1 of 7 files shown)


bguillot (approved):

LRovira-PlanetaTIC (approved):

product_categ_image/models/product_category.py
```diff
@@ -0,0 +1,12 @@
...
+# @author Alexis de Lattre <alexis.delattre@akretion.com>
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+
+from odoo import fields, models
+
+
+class ProductCategory(models.Model):
+    _inherit = 'product.category'
+
+    image = fields.Binary(string='Image')
```
pedrobaeza, line 12: Shouldn't this have `attachment=True` for a better DB space management?

alexis-via: Probably. But this is already used in production, so maybe we could do that when migrating to v11.

pedrobaeza: You can add a migration script right now if you want. There's a method in openupgradelib for performing the extract. Any way, at least put a TODO here.

FFernandez-PlanetaTIC: :+1: to put a TODO

Please, could we merge? Thanks

pedrobaeza:
@FFernandez-PlanetaTIC can you superseed this adding extra `attachment=True` attribute and fixing possible Travis lint checks? (I think they will go out when you rebase)

PR #344, opened 2018-04-05
[I18N] Update translation base, add finnish translation

aisopuro:


pedrobaeza:
No POT should include now in OCA modules. It's handled automatically, and you should make the translation in Transifex: https://odoo-community.org/page/translate.

I have uploaded this time the full file in Transifex for not forcing you to replicate the translation, but please take this into account for other times.

PR #355, opened 2018-05-30
Don't format the ID

### LINKED ISSUE #354: [11.0] product_brand logo not showing up in kanban view
The kanban view of the product_brand page, found by visiting `Sales > Configuration > Product Brands`, is not displaying the brand `logo`. Instead you see the broken image icon.

Here's a screenshot of what it looks like:

broken-image

The odoo backend raises the following error for each image it tries to fetch:

```
ValueError: invalid literal for int() with base 10: '3,033'
```

Right away the comma (in *3,033*) spelled out an issue with the values that were being passed to the image url. So, doing some further investigation I tracked down the issue to the arguments passed to the image url. Here's the outstanding code (from [product-attribute/product_brand/views/product_brand_view.xml](https://github.com/OCA/product-attribute/blob/11.0/product_brand/views/product_brand_view.xml)).
From the record with id `view_product_brand_kanban`:

```
<img t-att-src="kanban_image('product.brand', 'logo', record.id.value)"
```

The outstanding error here is the use of `record.id.value`. This is returning the id as a formatted integer. 
 (ex 3,033 instead of 3033). Instead, if you use `record.id.raw_value`, the issue is fixed.

Pull request coming shortly

wrhansen:
Lookup fails when the ID is formatted. The unformatted version of the
data is located under `raw_value`.

Closes #354

### DIFF
--- a/product_brand/__manifest__.py
+++ b/product_brand/__manifest__.py
@@ -8,7 +8,7 @@
 
 {
     'name': 'Product Brand Manager',
-    'version': '11.0.1.0.0',
+    'version': '11.0.1.0.1',
     'category': 'Product',
     'summary': "Product Brand Manager",
     'author': 'NetAndCo, Akretion, Prisnet Telecommunications SA'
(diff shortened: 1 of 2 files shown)


pedrobaeza:
Please change module version number according https://github.com/OCA/maintainer-tools/blob/master/CONTRIBUTING.md#version-numbers

pedrobaeza (approved):
Fast-tracking this as obvious

PR #359, opened 2018-06-13
[10.0] [MIG] product_multi_image (supersede)

FFernandez-PlanetaTIC:
Hi,

Supersedes #201 and improves uninstall hook to move images from multi to single mode.

Requires of OCA/server-tools#1286 to do it.

Thanks!


FFernandez-PlanetaTIC:
Hi,

This PR requires OCA/server-tools#1286 but the latter is pending reviewers.

Those interested can collaborate in the revision.

Thanks!

pedrobaeza:
Now that the other module has been merged, please review this one and fix lint and CI errors

FFernandez-PlanetaTIC (commented):
846709c8bfd4d0773a5dcaea670ac67969affa55 avoids resizing large image to small/medium when saved from single image. This fix is required to pass product_default_image tests.

FFernandez-PlanetaTIC (commented):
922f31ad11ddd53835529b8a5d433033f04cb533 fixes error in product_default_image install hook when product_multi_image changes image field to computed and non-store

FFernandez-PlanetaTIC:
@LRovira-PlanetaTIC Please, could you review this? Thanks

LRovira-PlanetaTIC (approved):
OK, it works correctly. Now, you can uninstall the module and the main image is saved as a product image.

product_multi_image/models/product_template.py
```diff
@@ -22,3 +22,30 @@ class ProductTemplate(models.Model):
...
+            elif 'image_small' in vals:
+                vals['image'] = vals['image_small']
+                del vals['image_small']
+        return super(ProductTemplate, self).write(vals)
+
+    @api.multi
+    def _set_multi_image_main_medium(self):
+        # on save product module resizes large image to medium
+        # medium image should not overwrite the large
+        pass
```
pedrobaeza, line 45: I'm not sure about this inheritance break, as you are not calling super here in any case. Can you discriminate any way when to not call super? What is the use case for always inhibiting this method?

pedrobaeza:
I have squashed a bit the commit history before reviewing.

product_multi_image/models/product_template.py
FFernandez-PlanetaTIC: There was an issue saving an image as the main of a product, it was **resized to small** before saving:

This happens because:
1. *product* module resizes an image to all sizes before saving it.
  https://github.com/OCA/OCB/blob/7a824c376ffe0ef6326c7c03a12e9105f9ce9f51/addons/product/models/product_template.py#L289
  https://github.com/OCA/OCB/blob/7a824c376ffe0ef6326c7c03a12e9105f9ce9f51/odoo/tools/image.py#L266
2. *product_multi_image* redefines image fields for all sizes as related from main image.
3. *base_multi_image* saves main image for all sizes in the same field:
  https://github.com/OCA/server-tools/blob/201129d9905f1312405bf4f3e08bf6e082312ac9/base_multi_image/models/owner.py#L19

Consequently the images for all sizes are saved in the same field and the last one, in this case the small one, wins :hatched_chick::trophy:

As an image is always saved for all sizes and, to not lose quality, I always save it as the largest, I fix the issue not saving it for the rest of sizes.

I can not think of a better solution... :confused:

product_multi_image/tests/test_product_multi_image.py
```diff
@@ -0,0 +1,194 @@
...
+
+    def test_uninstall_hook_product(self):
+        """ It should remove ``image_ids`` associated with products """
+        hooks.uninstall_hook(self.env.cr, self.registry)
+        images = self.env['base_multi_image.image'].search(
+            [("owner_model", "=", 'product.product')],
+        )
+        self.assertFalse(len(images))
+
+    def test_uninstall_hook_teplate(self):
```
thomaspaulb, line 188: typo here.

product_multi_image/models/product_template.py
thomaspaulb: Have studied this also. I think the solution is good because it's an exception for products only, necessary because products have already this built-in resizing on write. As [base_multi_image mentions clearly](https://github.com/OCA/server-tools/blob/10.0/base_multi_image/models/owner.py#L53), the `_set_multi_image_main_*` family of functions is meant only as a compatibility layer. Product IMO could be an exception to this rule - the images are still saved correctly and without loss.

thomaspaulb (approved):

fuentes73 (approved):
Please can you rebase and resolve conflicts?

FFernandez-PlanetaTIC:
@fuentes010 , done.

Regards! :wave:

FFernandez-PlanetaTIC:
Can we merge this?

product_multi_image/__manifest__.py
```diff
@@ -0,0 +1,33 @@
...
+
+{
+    "name": "Multiple Images in Products",
+    "version": "10.0.1.0.0",
+    "author": "Antiun Ingeniería, "
+              "Tecnativa, "
+              "LasLabs, "
+              "Odoo Community Association (OCA)",
+    "license": "AGPL-3",
+    "website": "https://www.tecnativa.com",
```
nikul-serpentcs: @FFernandez-PlanetaTIC  URL should be `https://github.com/OCA/product-attribute`

nikul-serpentcs (commented):
Nitpicking

product_multi_image/__manifest__.py
FFernandez-PlanetaTIC: Done, thanks!

PR #363, opened 2018-07-09
[10.0][FIX]  product_supplierinfo_for_customer: remove some views

MiquelRForgeFlow:
This module adds
```python
<menuitem id="product_supplierinfo_sale_menu" parent="sale.menu_product"
              action="product_supplierinfo_action" />
```
so it also depends of `sale` module.


pedrobaeza:
The problem is more to add that menu. Why?

MiquelRForgeFlow:
@aheficent

AaronHForgeFlow:
Yes, this is a mistake, this should not depend on sale as previous versions do not depend on it.

Better to move the menu and the actions to https://github.com/OCA/sale-workflow/pull/625

AaronHForgeFlow:
No, I was wrong. This menu exists in previous versions, we cannot remove it. The parent menu is the wrong thing here.

pedrobaeza:
But that's because previous version added sales menu on base. In this case, I think it's correct to put it on the sales part in OCA/sale-workflow#625

MiquelRForgeFlow:
Yeah, in v9 that menu was in `product`, but in commit https://github.com/odoo/odoo/commit/9e57185449a0bd958ad38eddc280d146e2d60e1e of v10 was moved to `sale`.

AaronHForgeFlow (changes requested):
Cannot create pricelist based on supplier.info. There's a constraint that has to be overridden:
image

MiquelRForgeFlow:
Ups, I will fix that

MiquelRForgeFlow:
Hey @aheficent this error is due to the `product_pricelist_tier` that is not "compatible" with the `product_supplierinfo_for_customer`. If someone wants to use the two modules, then than someone should create a glue module. But this should not impede to merge this fix.

@pedrobaeza you can merge this PR

AaronHForgeFlow (approved):
Yes, a glue module will be needed to use both. Ok this is ok to me then. :+1:

MiquelRForgeFlow:
@pedrobaeza ?

MiquelRForgeFlow:
If we don't merge this ASAP, some people (who don't have `sale` installed) will bother about failing installation of the module.

pedrobaeza:
OK, merging blindly trusting you :wink:

MiquelRForgeFlow:
thanks, but it's not blindly, you can check in https://github.com/OCA/product-attribute/pull/363.

pedrobaeza:
That's why I trust you

PR #367, opened 2018-07-13
[11.0][MIG] product_supplierinfo_for_customer

sergio-teruel:
cc @Tecnativa

PR #368, opened 2018-07-16
[10.0][REF] product_supplierinfo_for_customer: views

MiquelRForgeFlow:
This PR includes:

-  Refactoring of views, in order to not overwrite views and use inherit and xpath.
-  Only users in group 'product.group_pricelist_item' should be able to view prices in the customer supplierinfo form.

### DIFF
--- a/product_supplierinfo_for_customer/views/product_view.xml
+++ b/product_supplierinfo_for_customer/views/product_view.xml
@@ -1,29 +1,36 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <odoo>
-    <record model="ir.ui.view" id="product_supplierinfo_extended_form_view">
-    <!-- Copying and modifying the standard as the absence of names makes it impossible to inherit-->
-        <field name="name">product.supplierinfo.extended.form</field>
+    <record id="product_supplierinfo_form_view" model="ir.ui.view">
+        <field name="name">product.supplierinfo.form.view</field>
         <field name="model">product.supplierinfo</field>
+        <field name="priority" eval="100"/>
+        <field name="inherit_id" ref="product.product_supplierinfo_form_view" />
         <field name="arch" type="xml">
-           <form string="Partner Information">
+            <xpath expr="//field[@name='name']" position="before">
+               <field name="type" invisible="1"/>
+           </xpath>
+        </field>
+    </record>
+
+    <record id="product_supplierinfo_customer_form_view" model="ir.ui.view">
+        <field name="name">product.supplierinfo.customer.form.view</field>
+        <field name="model">product.supplierinfo</field>
+        <field name="priority" eval="100"/>
+        <field name="arch" type="xml">
+            <form string="Customer Information">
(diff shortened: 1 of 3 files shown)


MiquelRForgeFlow:
@aheficent please review ASAP

product_supplierinfo_for_customer/views/product_view.xml
```diff
@@ -95,18 +69,25 @@
...
+                           context="{'default_search_type':'customer','default_type':'customer','default_product_tmpl_id':id,'form_view_ref':'product_supplierinfo_customer_form_view'}">
+                    </field>
+                </div>
             </div>
         </field>
     </record>
 
+    <record id="product.product_supplierinfo_type_action" model="ir.actions.act_window">
+        <field name="name">Partner Pricelists</field>
+        <field name="res_model">product.supplierinfo</field>
```
AaronHForgeFlow, line 106: why is this needed?

AaronHForgeFlow (changes requested):
:-1: purchase pricelist menu requires vendor domain

product_supplierinfo_for_customer/views/product_view.xml
MiquelRForgeFlow: to change name of "Vendor Pricelists" to "Partner Pricelists"

Also, I now updated it to pass the supplier domain

AaronHForgeFlow: ok, now it's good

AaronHForgeFlow (approved):
:+1: thank you for his fix!

AaronHForgeFlow:
ping @jbeficent

pedrobaeza:
In #367 we have also reworked several things for better usability. I'm not sure as I haven't reviewed this if they are in the same wave or not.

MiquelRForgeFlow:
They are in a similar wave, luckily. Even so, if not, I wouldn't accept a "we can't accept this PR because we have already open a new v11 PR with a bit different approach".

pedrobaeza:
Hey, I'm not saying that! I'm saying to exchange ideas for making both the same.

MiquelRForgeFlow:
Ok. In essence, they are the same. They are different only in naming some views. I would prefer merge this PR, and then update the commit history of v11 PR to include this. This way, it would be easier to track differences.

pedrobaeza:
I prefer to keep both as is, as with your option would mean that we have to solve a lot of conflicts. If features of both PRs match, I think it's good to let it as is.

MiquelRForgeFlow:
This PR includes
```python
<xpath expr="//form/group/group[2]" position="attributes">
    <attribute name="groups">product.group_pricelist_item</attribute>
</xpath>
```
An improvement that the v11 PR should include.

AaronHForgeFlow:
:+1: Functional test and working good

JordiBForgeFlow (approved):
:+1: Functional and technical review.

PR #369, opened 2018-07-17
[10.0][ADD] product_code_unique module

chafique-delli:


OpenCode (approved):

OpenCode (commented):
It doesn't work on runbot

bodedra (changes requested):
runbot is  :red_circle: @chafique-delli may you fix it.

chafique-delli:
@OpenCode , @bodedra , Ok, I'm looking to fix that.
Thank you,

OpenCode:
@chafique-delli I tested the code on a local computer. When I install the modulo I can't find the constraint in the database on product_product table. The problem is the pre_init_product_code function. You set all the product without code as "DEFAULT_CODE" before module install but when Odoo try to create constraint in database, it found product with some code and constraint creation is interrupted. The query in pre_init_product_code function is wrong. Please, look at https://github.com/OCA/product-attribute/blob/11.0/product_code_unique/__init__.py#L9 and fix your code

chafique-delli:
@OpenCode , You're right, I did not be careful to the query in pre_init_product_code function, it was incomplete.
I fix that.
Thank you,

OpenCode:
@chafique-delli Travis and Runbot are red because the query is wrong. Please, try the changes on your locale installation before push them.

OpenCode:
I restarted Travis. Now it's green. Code tested on Runbot and it's ok for me.

bodedra (approved):
LGTM. :+1:

OpenCode:
@OCA/product-maintainers Can you merge it?

bodedra:
@pedrobaeza

lfreeke:
I get warning when enter and trying to save a product with a same product_code.

I notice that when I sneakly put a space after a same product_code it will save the product with out a unique warning. 

I still will give it a thumbs up, because it works as it should be and the space is a corner case that can be solved later if necessary. 

:+1:  (test)

PR #373, opened 2018-07-24
[10.0][FIX] product_supplierinfo_for_customer: assure partner is passed in the context

MiquelRForgeFlow:
When creating tests for the `product_supplierinfo_for_customer_sale` module (https://github.com/OCA/sale-workflow/pull/625), I found that sometimes the partner in the context is passed as 'partner' ([see](https://github.com/OCA/OCB/blob/10.0/addons/sale/models/sale.py#L890), [see](https://github.com/OCA/OCB/blob/10.0/addons/product/models/product_pricelist.py#L274)) and sometimes is passed as 'partner_id' ([see](https://github.com/OCA/OCB/blob/10.0/addons/sale/models/sale.py#L868)).

Please, let's merge this


AaronHForgeFlow (approved):
:+1: fast-track please

MiquelRForgeFlow:
:green_apple:

JordiBForgeFlow (approved):

PR #376, opened 2018-08-27
[10.0][ADD] product_assortment

rousseldenis:
Adds a method to filter to enhance it.


rousseldenis:
@OCA/product-maintainers Needs review

product_assortment/models/ir_filters.py
```diff
@@ -0,0 +1,86 @@
...
+    record_count = fields.Integer(compute='_compute_record_count')
+
+    is_assortment = fields.Boolean(
+        default=lambda x: x._get_default_is_assortment())
+
+    @api.multi
+    def _get_eval_domain(self):
+        res = super(ProductAssortment, self)._get_eval_domain()
+
+        if self.whitelist_product_ids and res:
```
lmignon, line 44: @rousseldenis @Cedric-Pigeon In this code whitelist and blacklist seems to be mutually exclusive... but nothing prevents the  user to fill values for both... (A constrains ? or ...)

lmignon (changes requested):
A little comment.

product_assortment/models/ir_filters.py
Cedric-Pigeon: @lmignon why do you think they are mutually exclusive. This is not the case. The goal is to include manually some products which are not selected by the filter and/or exclude product from the filter

lmignon: oups sorry, I misread the code.

lmignon (approved):

lmignon:
@OCA/product-maintainers  This one is ready to merge and works in production for months.

lmignon:
No one to merge this one?

lmignon:
Thank you @pedrobaeza !

pedrobaeza:
You're welcome, man!

PR #377, opened 2018-09-05
[11.0][NEW] product_secondary_unit: New module for set a second units for a product

sergio-teruel:
cc @Tecnativa


pedrobaeza:
We talk about `product_secondary_unit`, not `product_second_unit`

hieulucky111 (approved):

rousseldenis:
Is this not a duplicate of product packaging uom module? Because it does exactly the same and is more flexible as it takes advantages of product packaging (you have as many units as you want). See stock logistics warehouse repo.

lmignon:
I share the same feeling as @rousseldenis

lukebranch:
@sergio-teruel @rousseldenis ,

Sorry to ask questions on a PR, however i'd like to know, Is there a particular use-case this module was designed for?

I have an industry-specific requirement for a company that manufactures aluminium cans (eg. for Coca Cola, John West Tuna, etc.). This Industry will always quote a unit price for cans as a per 1000 Units (ie. 1000 x Unit), but Sale and Purchase UoM will remain Unit (ie. x1).

In this case I need to define a secondary UoM on product. Is this module or product packaging uom module designed for this sort of use-case?

I have tried this PR in the runbot, but am not sure of how to use it. Is it to allow for defining a UoM in more than one UoM Category (eg. Sale and Purchase in Unit, and Secondary UoM in Weight)?

rousseldenis:
There are several modules that work together.

packaging_uom, sale_packaging and purchase_packaging.

Packaging uom 'simply' add an uom on product packagings.

The other one let you use them in sale and purchases flows.

Let's say you store your coca with 1L bottles, the unit will be (1X Unit x 1L). But you have also the packaging that say 6x Unit x 1L which is the pack you use in purchases.

Or a pallet of them and so on

lukebranch:
@sergio-teruel ,

Thanks for the explanation, that makes sense.

rousseldenis:
@sergio-teruel Have you seen remarks ?

Cedric-Pigeon:
@sergio-teruel I agree with @rousseldenis. You are not taking a good way. Packaging modules are more powerful and not limited to 2 units of measure. Please take a look.

product_secondary_unit/security/ir.model.access.csv
```diff
@@ -0,0 +1,4 @@
+id,name,model_id:id,group_id:id,perm_read,perm_write,perm_create,perm_unlink
+access_product_secondary_unit_user,access_product_second_unit_user,model_product_secondary_unit,base.group_user,1,0,0,0
+access_product_secondary_unit_user_stock,access_product_second_unit_user_stock,model_product_secondary_unit,stock.group_stock_user,1,1,1,0
+access_product_secondary_unit_manager,access_product_second_unit_manager,model_product_secondary_unit,stock.group_stock_manager,1,1,1,1
```
yajo: You should either depend on `stock` or remove these permission lines.

I'm getting this error while trying to install the addon:

```
Exception: Module loading product_secondary_unit failed: file product_secondary_unit/security/ir.model.access.csv could not be processed:
 No matching record found for external id 'stock.group_stock_user' in field 'Group'
No matching record found for external id 'stock.group_stock_manager' in field 'Group'
```

yajo (changes requested):

rousseldenis:
cc @Tecnativa Before reviewing, would you enter discussion about module utility and redundancy?

sergio-teruel:
Excuse me everybody...

Hi, my first idea was use the packaging conversion modules but my collegue @pedrobaeza said to me that the core packaging module will be enterprise, so I decided use a new data model (product.secondary.unit).
The other reason to creatre new modules is that packaging module does not allow using units of measure of different categories, p.e I have conversion beetwen units and Kg with a approximate conversion factor.
And this module has not any effect on logistics package. P.E I can sell 5 pieces of a product that they weight (product_qty in sale order) are 10 Kg, but I can do it in different packages.
I hope your comprension..

rousseldenis:
@sergio-teruel What @pedrobaeza certainly talked about is the 'product.ul' model removal. Which is completely different. This was the package where you put products inside (box, pallet, ...) for delivery. And reintroduced through @alexis-via PR https://github.com/OCA/stock-logistics-tracking/pull/18/

The model 'product.packaging' is still on v11 and master. This is the one used by 'packaging_uom'.

Moreover, Odoo still uses it in master since commit https://github.com/odoo/odoo/commit/da00534e5b7a3e569038e59cd5c6596d3bec6cba

Do you need more information about those ?

rousseldenis:
@sergio-teruel For the unit categories conversion, the sale packaging and purchase packaging does what you need:
image


Did it on runbot !

sergio-teruel:
Sorry, but not, with packaging modules you can not sell in a secondary unit writing directly the second unit.
With secondary_unit modules if you have very hight unit conversion combinations you defines it in product directly, do not need a lot unit of measures.
Other reason, with secondary unit modules I can set directly in sale order line qty in secondary unit or primary unit (E. g. I sell n units but my stock and invoicing unit is Kg).
I attach this [video](https://www.youtube.com/watch?v=Du7-yiAKjQY) of it.

seleccion_022

rousseldenis:
Ok, but that's more a 'conversion factor helper' than a secondary unit. I understand your needs.

There is a PR for barcodes management in packagings (that's used in real life) here https://github.com/OCA/stock-logistics-barcode/pull/153 that can help to call the good one.

Indeed that implies creating as many packagings as multiple units you have, but the process uses the standard uom model and can be called at any place. That's the advantage.

Why not improving the existing one to fullfill your requirements (adding the shortcuts you have created or ...) ?

rafaelbn:
Hi @rousseldenis ,

Let's introduce concept "Caliber / Size" which doesn't' have and relation with Packages.

- You sell "legs of Iberian hams" by kg (€/kg). The product is "Leg of Iberiam hams" with main unit kg. And the price is per Kg.
- You don't know how is the weight of each leg (they are weighed in the delivery)
- They are transported in wooden boxes, always the same box and in each box they enter 1, 2 or 3 legs. The boxes do not matter for sale just for transportation.
- Customer ask you for 2 legs and you sent quotation in €/kg because usually (caliber) 1 leg is 6 kilos and each kg sale price is 70€.
- You fill the product in the SO and user fill second uom with 2 legs. With this module Odoo knows that it is 12 kg and units are filled. So you get 12 kg * 70€ = 840€ + taxes
- In inventory the picking is processed with 13,5 kg (because is the real weight is the 2 legs) and packaged in 1 wood box. This 13,5 kg are know in the SO line to invoice.

Make the same example case with 3 legs. Kg are weight but the box is the same.

1. One (1) leg is not Packaging Type 
2. One (1) leg is "Unit" uom category, not Weight (kg)
3. Could you record us a video how to do this with your approach? I've tried and it's not possible



4. We could improve the existing but it bring to us: (4.1) disadvantages of product packaging (4.2) Depend on product packaging modules which it's not necessary in the example if you don't need Packages and only "conversion" between units.
5. UX / Usability

Furthermore, with this PR we know stock in the main unit (kg) plus the approximate number of legs (kg/caliber). What happens with Stock in packages.

What do you think?

Thanks
Regards

rousseldenis:
@rafaelbn I will come with explanations and video but I've succeeded on runbot selling 10 pieces of 5kg. Delivering 60kg. On sale order, the delivered quantity is correct (12 x 5kg).

The stock is managed by kg and the quantity is decreased by 60 x. Quants are ok.

> Depend on product packaging modules which it's not necessary in the example if you don't need Packages and only "conversion" between units.

You introduce a new model, so using existing one is better

rousseldenis:
I didn't test the invoicing completely but I think this one could be useful https://github.com/OCA/sale-workflow/pull/642

rousseldenis:
image

rousseldenis:
image

rousseldenis:
image

I've forced availability for test purpose.

The last two quants are concerned. One of -50 kg for original move. One of -10 for the exceeding delivered quantity

rafaelbn:
Hi @rousseldenis ,

> You introduce a new model, so using existing one is better

Using existing one is better if we don't mix apples with oranges. The main thing here is requirement + this requirements are not for a single customer but a sector.

We Tecnativa like to make thinkg simpe as possible and reuse always, we maintain version by version all modules.

After two years working in Odoo using what Odoo and OCA have this is the time to improve what is missing.

Please take in care that we haven''t (as @sergio-teruel said) just developed without testing the modules you said. There is a consider big amount hours of analysis before taking this decision and 2 years working in this sector.

You proposal is used by us in other customers, of course we know to use conversion between units in the same category but it not the solution of this case. Absolutely in this case not. Our customer will introduce one sale.order line per second (usually nobody think on UX), just opening the pop-up and closing it plus finding fields will be 5 seconds. Note: and Odoo packaging modules are (sorry IMHO) awful!

If we use Packages for conversion unit of **different categories** the we cannot use packaging for in the case real packaging (in the case).

- You recommend us to use 5 o more modules that don't fit requirements and we should make a big work (plus maintenance! no module is in v11) instead of creating a new one.
- You proposal is not operative for user's customer work

Thanks you for your example! and time! :heart: but what you did it's not correct for us and doesn't fit customer requirement because:

https://github.com/OCA/product-attribute/pull/377#issuecomment-422866861

- Unit price is incorrect, the price (I already said) it is not by unit (in your case 5kg is the name) is by kg. The price you show is per 1 unit (in your case 5kg is the name) of  Ice Cream and not per 1 kg of Ice cream
- Ordered Qty is incorrect, the quantity should be kg.
- Delivered Qty. is incorrect, should be kg
- Invoiced Qty, is incorrect, should be kg. 
- So in this first screenshot: (a) Which is the price per kg? How many kg. of Ice Cream will you deliver?. I can answer by you, you don't know. You only know you will deliver 10 Ice Cream and each Ice Cream cost 500€. Plus (+) report, pivot, etc..

Think that the main unit is kg. always and everywhere AND a second unit of measure must be show in everywhere always, SO, Picking, Stock, Invoice, Accounting. The solution of course is not the simple functionality of creating unit for ALL the products with a ratio... :smile: (that should be easy!). Unit like 0.6-0.8, 1.5, 1.2-1.4, and hundred more..

Your proposal is from where we come. We work with Odoo packaging and some customs from v8, 3 years ago.. https://github.com/OCA/product-attribute/pull/377#issuecomment-422330898

- https://github.com/OCA/sale-workflow/pull/221
- https://github.com/OCA/sale-workflow/pull/302

rousseldenis (changes requested):
I don't have too much time now to dig into.

I proposed a table at Louvain Sprint for logistics topics.

I think we could make converging efforts to develop a solution as strong as possible and fill in functional and UX gaps. Did you plan to work on such topics ?

sergio-teruel:
I agree about the code sprint table..
I want you to keep in mind that in this case I do not want control logistics aspects...
:smile:

rousseldenis:
Of course, this one is depending on product only !  But one rules them all :smile:

rousseldenis (approved):
If you want to avoid product packaging complexity management with different uoms, take this one.

Code review and tested.

rafaelbn:
Please don't merge until @sergio-teruel notice us thanks

@Yajo https://github.com/OCA/product-attribute/pull/377#pullrequestreview-156270953 is solved

yajo (approved):

PR #378, opened 2018-09-05
[11.0][MIG] product_supplierinfo_for_customer

MiquelRForgeFlow:
Superseeds https://github.com/OCA/product-attribute/pull/367 to update commit history of v10.

@sergio-teruel I know that maybe I should have asked first, but I wouldn't want to bother you or giving you more work to do. I have maintained all your changes except two:

- readme file: You deleted content of readme, but @pedrobaeza says that is better to keep it, so I kept it when adding readme folder.
- product_view.xml: When you did the migration, the v10 at that moment was doing "extended" versions of views, and your migration improved that. But later in v10, it was abandoned that approach and taken approach of create new views (non inheriting customer views). I maintained that approach.


product_supplierinfo_for_customer/__manifest__.py
```diff
@@ -0,0 +1,25 @@
+# Copyright 2015 OdooMRP team
```
elicoidal, line 1: probably need to update the dates as you are modifying at present

product_supplierinfo_for_customer/demo/product_demo.xml
```diff
@@ -0,0 +1,12 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<odoo noupdate="1">
```
elicoidal, line 4: add copyright

product_supplierinfo_for_customer/views/product_view.xml
```diff
@@ -0,0 +1,114 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<odoo>
```
elicoidal, line 6: copyright

product_supplierinfo_for_customer/views/product_view.xml
```diff
@@ -0,0 +1,114 @@
...
+
+    <record id="product_supplierinfo_customer_form_view" model="ir.ui.view">
+        <field name="name">product.supplierinfo.customer.form.view</field>
+        <field name="model">product.supplierinfo</field>
+        <field name="priority" eval="9999"/>
+        <field name="arch" type="xml">
+            <form string="Customer Information">
+                <group>
+                    <group string="Customer">
+                        <field name="name" string="Customer" context="{'default_supplier': 0, 'search_default_customer': 1, 'default_customer': 1}"/>
```
elicoidal: Please improve the readbility of the xml with one attribute per line

elicoidal (changes requested):
some details

thomaspaulb:
@elicoidal Have your reviews been successfully addressed?

product_supplierinfo_for_customer/i18n/de.po
```diff
@@ -0,0 +1,128 @@
...
+#: view:product.template:product_supplierinfo_for_customer.product_template_extended_form_view
+msgid ""
+"{'default_search_type':'supplier','default_type':'supplier','default_product_tmpl_id':id}"
+msgstr ""
+"{'default_search_type':'supplier','default_type':'supplier','default_product_tmpl_id':id}"
+
+#. module: product_supplierinfo_for_customer
+#: view:product.supplierinfo:product_supplierinfo_for_customer.product_supplierinfo_extended_form_view
+msgid "{'select_type': type}"
+msgstr "{'select_type': type}"
```
thomaspaulb, line 128: Quite a number of translations are for attributes that don't need translation, such as [this one](https://github.com/OCA/product-attribute/pull/378/files#diff-3f11b17f61198057bc5aca5ce4810fd8R131). Could you re-export the German translation to see whether they disappear, or otherwise, add `t-translation="off"` ([link](https://github.com/OCA/OCB/blob/11.0/odoo/tools/translate.py#L211)) to the relevant attributes.

thomaspaulb (approved):
Translation fix is optional IMO, the tests run well so I think this can be merged.

product_supplierinfo_for_customer/__manifest__.py
```diff
@@ -0,0 +1,26 @@
...
+    "depends": [
+        "purchase",
+    ],
+    "data": [
+        "views/product_view.xml",
+    ],
+    "demo": [
+        "demo/product_demo.xml",
+    ],
+    "installable": True,
```
mpanarin: non-blocking but i find this line to be redundant.

The `installable:False` was not used from migrations to v 9.0

product_supplierinfo_for_customer/models/product_product.py
```diff
@@ -0,0 +1,50 @@
...
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl.html).
+from odoo import api, models
+
+
+class ProductProduct(models.Model):
+    _inherit = 'product.product'
+
+    @api.multi
+    def _get_price_from_supplierinfo(self, partner_id):
+        if partner_id:
```
mpanarin: i think you can `assert` this in order to keep it flatter

product_supplierinfo_for_customer/models/product_product.py
```diff
@@ -0,0 +1,50 @@
...
+            self, price_type, uom=False, currency=False, company=False):
+        if price_type == 'partner':
+            partner = self.env.context.get('partner_id', False) or \
+                self.env.context.get('partner', False)
+            for product in self:
+                price = product._get_price_from_supplierinfo(partner)
+                if not price:
+                    return super(ProductProduct, self).price_compute(
+                        'list_price', uom, currency, company)
+                prices = dict.fromkeys(self.ids, 0.0)
```
mpanarin: :stop_sign:  you are overwriting the dict on every iteration
as well as you will only have one iteration, as you return `prices` inside the cycle

product_supplierinfo_for_customer/models/product_supplierinfo.py
```diff
@@ -0,0 +1,36 @@
...
+        selection=[
+            ('customer', 'Customer'),
+            ('supplier', 'Supplier'),
+        ], string='Type',
+        default='supplier')
+
+    @api.multi
+    @api.onchange('type')
+    def onchange_type(self):
+        if self.type == 'supplier':
```
mpanarin: this is broken due to multi, and will fail if you have recordset of >1 values

Also, as you use this onchange only in tests it is better to leave a comment about that in the method as it is confusing )

product_supplierinfo_for_customer/tests/test_product_supplierinfo_for_customer.py
```diff
@@ -0,0 +1,110 @@
+# Copyright 2015 OdooMRP team
+# Copyright 2015 AvanzOSC
+# Copyright 2015 Tecnativa
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl.html).
+import odoo.tests.common as common
+
+
+class TestProductSupplierinfoForCustomer(common.TransactionCase):
```
mpanarin: Consider changing this to SavepointCase. As you have a lot of test cases it will be a lot quicker

mpanarin (changes requested):

elicoidal (approved):
LGTM

product_supplierinfo_for_customer/views/product_view.xml
```diff
@@ -73,15 +81,26 @@
...
+                    }</attribute>
                 <attribute name="domain">[('type','=','supplier')]</attribute>
                 <attribute name="invisible">1</attribute>
             </field>
-            <field name="seller_ids" position="after">
-                <field name="supplier_ids"
-                       context="{'default_search_type':'supplier','default_type':'supplier','default_product_tmpl_id':id}"
-                       domain="[('type','=','supplier')]" />
-            </field>
+            <xpath expr="//field[@name='seller_ids']/.." position="after">
```
SimoRubi, line 102: Why did you remove `domain="[('type','=','supplier')]"`?

product_supplierinfo_for_customer/views/product_view.xml
```diff
@@ -94,7 +113,12 @@
                 <div name="pricelist_customer">
                     <separator string="Customers" />
                     <field name="customer_ids"
-                           context="{'default_search_type':'customer','default_type':'customer','default_product_tmpl_id':id,'form_view_ref':'product_supplierinfo_for_customer.product_supplierinfo_customer_form_view','tree_view_ref':'product_supplierinfo_for_customer.product_supplierinfo_customer_tree_view'}">
+                           context="{
```
SimoRubi, line 141: Why did you remove `'default_product_tmpl_id':id`?

product_supplierinfo_for_customer/views/product_view.xml
```diff
@@ -104,7 +128,12 @@
     <record id="product.product_supplierinfo_type_action" model="ir.actions.act_window">
         <field name="name">Partner Pricelists</field>
         <field name="res_model">product.supplierinfo</field>
-        <field name="context">{'default_search_type':'supplier','default_type':'supplier','default_product_tmpl_id':id,visible_product_tmpl_id':False}</field>
+        <field name="context">{
```
SimoRubi, line 158: Why did you remove `'default_product_tmpl_id':id`?

product_supplierinfo_for_customer/views/product_view.xml
```diff
@@ -73,15 +81,26 @@
         <field name="inherit_id" ref="purchase.view_product_supplier_inherit"/>
         <field name="arch" type="xml">
             <field name="seller_ids" position="attributes">
-                <attribute name="context">{'default_search_type':'supplier','default_type':'supplier','default_product_tmpl_id':id}</attribute>
+                <attribute name="context">{
```
SimoRubi, line 84: Why did you remove `'default_product_tmpl_id':id`?

product_supplierinfo_for_customer/models/product_product.py
```diff
@@ -0,0 +1,50 @@
...
+    def _get_price_from_supplierinfo(self, partner_id):
+        if partner_id:
+            for product in self:
+                supplierinfo = self.env['product.supplierinfo'].search(
+                    ['|', ('product_tmpl_id', '=', product.product_tmpl_id.id),
+                     ('product_id', '=', product.id),
+                     ('type', '=', 'customer'),
+                     ('name', '=', partner_id)])
+                if supplierinfo:
+                    return supplierinfo.price
```
SimoRubi: Ok, you want to return the first supplierinfo that you find. But what if you find more than one? I suggest adding `limit=1` to the search arguments

product_supplierinfo_for_customer/models/product_product.py
```diff
@@ -0,0 +1,50 @@
...
+    @api.multi
+    def price_compute(
+            self, price_type, uom=False, currency=False, company=False):
+        if price_type == 'partner':
+            partner = self.env.context.get('partner_id', False) or \
+                self.env.context.get('partner', False)
+            for product in self:
+                price = product._get_price_from_supplierinfo(partner)
+                if not price:
+                    return super(ProductProduct, self).price_compute(
```
SimoRubi: Returning from inside the loop doesn't seem to be a good idea unless you want to return the _first you find_ like in `_get_price_from_supplierinfo`.
I suggest to get the prices from the `super(ProductProduct, self).price_compute(price_type, uom, currency, company)`; then overwrite the result with what you find for the supplier
and then overwrite the entries for which `_get_price_from_supplierinfo` returns something different from `0.0`.

SimoRubi (changes requested):
Thanks for your contribution!

product_supplierinfo_for_customer/README.rst
```diff
@@ -0,0 +1,70 @@
+.. image:: https://img.shields.io/badge/license-AGPL--3-blue.png
```
tarteo, line 1: Can you generate the readme using https://github.com/OCA/maintainer-tools/blob/master/tools/gen_addon_readme.py

product_supplierinfo_for_customer/models/product_supplierinfo.py
```diff
@@ -0,0 +1,36 @@
...
+
+
+class ProductSupplierinfo(models.Model):
+    _inherit = 'product.supplierinfo'
+
+    type = fields.Selection(
+        selection=[
+            ('customer', 'Customer'),
+            ('supplier', 'Supplier'),
+        ], string='Type',
```
tarteo: string here is not needed Odoo automatically capitalizes field names.

product_supplierinfo_for_customer/models/product_product.py
luistorresm: What do you think about change the logic here:

```
        if not partner_id:
             return 0
        supplierinfo = self.env['product.supplierinfo'].search(
           ['|', ('product_tmpl_id', '=', product.product_tmpl_id.id),
            ('product_id', '=', product.id),
            ('type', '=', 'customer'),
            ('name', '=', partner_id)])
        return supplierinfo.price
```

Whit this you gain a indentation

And maybe add a `self.onsere_one()` in the method, you has the return in the `for`

tarteo (changes requested):
Just some minor comments (besides the suggestions already made by the other reviewers)

rafaelbn:
https://github.com/OCA/product-attribute/pull/367

sergio-teruel:
@mreficent Can you generate the new readme?

MiquelRForgeFlow:
Thank you for all the reviews. I will do the changes in the next few days.

pedrobaeza:
@mreficent can you please finish this?

product_supplierinfo_for_customer/tests/test_product_supplierinfo_for_customer.py
MiquelRForgeFlow: I suppose this is an optional change, right?

product_supplierinfo_for_customer/models/product_supplierinfo.py
MiquelRForgeFlow: I will just delete `@api.multi`

product_supplierinfo_for_customer/tests/test_product_supplierinfo_for_customer.py
pedrobaeza: Well, it's a well received optimization. The problem both have is that when testing on parallel with other modules that tests reports, both will hang, so `HttpCase` is the only that saves that problem.

product_supplierinfo_for_customer/README.rst
MiquelRForgeFlow: no, I can't :(

product_supplierinfo_for_customer/views/product_view.xml
MiquelRForgeFlow: error, not intended

product_supplierinfo_for_customer/views/product_view.xml
MiquelRForgeFlow: is not removed

product_supplierinfo_for_customer/views/product_view.xml
MiquelRForgeFlow: is not removed

product_supplierinfo_for_customer/views/product_view.xml
MiquelRForgeFlow: is not removed

product_supplierinfo_for_customer/i18n/de.po
MiquelRForgeFlow: I won't touch translations :S

use transifex please

MiquelRForgeFlow:
please @rafaelbn remove `duplicate` tag

@pedrobaeza all changes done plus some improvements

MiquelRForgeFlow:
:green_apple:

product_supplierinfo_for_customer/README.rst
tarteo: Ok, then we'll wait for the cron when it's merged.

Issue #381, opened 2018-09-11
product_pricelist_fixed_price failing tests


daramousk:
On 0d09b6e2 and specifically on product_pricelist_fixed_price/tests/test_product_pricelist_fixed_price.py:45 there seems to be an attempt to set the `product.pricelist` type to `purchase` which fails because the `purchase` type is inserted on the `odoo.addons.purchase` module and `product_pricelist_fixed_price` does not depend on it.

Is this a bug as I understand?

pedrobaeza:
Yes, it seems so

PR #382, opened 2018-09-11
[FIX] product_pricelist_fixed_price set conditional running if `purchase` module is installed.

daramousk:
See #381

### DIFF
--- a/product_pricelist_fixed_price/tests/test_product_pricelist_fixed_price.py
+++ b/product_pricelist_fixed_price/tests/test_product_pricelist_fixed_price.py
@@ -15,6 +15,7 @@ class TestProductPricelistFixedPrice(common.TransactionCase):
             * for purchase price list
         """
         item_obj = self.env['product.pricelist.item']
+        module_obj = self.env['ir.module.module']
         plv = self.env.ref('product.ver0')
         base_sale = item_obj._get_default_base({'type': 'sale'})
         vals = {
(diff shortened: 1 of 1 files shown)


pedrobaeza:
I don't think this is the best option. That module is not needed for the feature to work. It's more on the test part. You can make that tests conditional if `purchase` is installed.

daramousk:
@pedrobaeza Fixed

product_pricelist_fixed_price/tests/test_product_pricelist_fixed_price.py
```diff
@@ -40,24 +41,27 @@ def test_onchange_base_ext(self):
         self.assertEqual(item.price_discount, -1.0)
 
         # change the type of the pricelist => purchase
-        plv.pricelist_id.type = 'purchase'
-        base_pur = item_obj._get_default_base({'type': 'purchase'})
-        self.assertNotEqual(base_sale, base_pur)
+        if module_obj.search_count([
```
pedrobaeza: It's better for the diff to reverse the condition `if not...` and return early if so.

NL66278: You might also consider splitting the method in one testing sales prices and one testing purchase prices (following the principles of clean coding).  Then in the test for purchase prices do the early return as recommended by Pedro. 

If you want to clean up the code some more, you can extract a method that does the testing for a certain pricetype, because the code under the check... commented lines contain a lot of repetition.

NL66278 (approved):
:+1: LGTM

pedrobaeza (approved):

PR #383, opened 2018-09-12
make it depends on purchase module

zeroheure:
This module is unusable without purchase module.


pedrobaeza:
That's not true. The information can be provided without that module. These are technical dependencies, not functional one. Closing this as no valid.

PR #384, opened 2018-09-12
Fix tab name in product form

zeroheure:
Tab name is "Inventory" instead of "Purchase"

### DIFF
--- a/product_supplierinfo_revision/README.rst
+++ b/product_supplierinfo_revision/README.rst
@@ -30,7 +30,7 @@ To use this module:
 #. By clicking on Apply button, a new supplier pricelist will be created with
    a price increased accordingly to the percent entered. Old supplier
    pricelist will have a date end from which pricelist will be out of date.
-#. Once the Inventory Management module (stock) is installed, you can check 
+#. Once the Inventory Management (*stock*) module is installed, you can check 
    supplier pricelists as well as variation percent in the product form, by 
    clicking on *Inventory* tab.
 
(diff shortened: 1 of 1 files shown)


pedrobaeza:
Well, there are multiple paths for checking the products, so maybe the explanation would be larger specifying that it would be available if module X is installed.

zeroheure:
Added

pedrobaeza (approved):

zeroheure:
Travis failed for external reasons : unable to downdload Python. How can I relaunch it ?

pedrobaeza:
Relaunched

zeroheure:
ping

carlosdauden (approved):

product_supplierinfo_revision/README.rst
```diff
@@ -23,8 +30,9 @@ To use this module:
 #. By clicking on Apply button, a new supplier pricelist will be created with
    a price increased accordingly to the percent entered. Old supplier
    pricelist will have a date end from which pricelist will be out of date.
-#. You can check supplier pricelists as well as variation percent in the
-   product form, by clicking on *Purchase* tab.
+#. Once the Inventory Management (*stock*) module is installed, you can check 
+   supplier pricelists as well as variation percent in the product form, by 
+   clicking on *Inventory* tab.
```
carlosdauden, line 35: On client instance I look suppliers in Purchase tab, but if you has check...

zeroheure: I don't remember exactly but I think that the location depends on installed modules. 
I've checked again and if Purchase Management is install, suppliers are located in product's Purchase tab. I will look further, don't merge.

PR #398, opened 2018-10-25
[11.0][ADD] product_secondary_unit: Move sale_secondary_unit to sale_order_secondary_unit module

sergio-teruel:
cc @Tecnativa

### DIFF
--- a/product_secondary_unit/i18n/es.po
+++ b/product_secondary_unit/i18n/es.po
@@ -6,8 +6,8 @@ msgid ""
 msgstr ""
 "Project-Id-Version: Odoo Server 11.0\n"
 "Report-Msgid-Bugs-To: \n"
-"POT-Creation-Date: 2018-09-05 10:03+0000\n"
-"PO-Revision-Date: 2018-09-05 12:05+0200\n"
+"POT-Creation-Date: 2018-10-26 10:00+0000\n"
+"PO-Revision-Date: 2018-10-26 12:01+0200\n"
 "Last-Translator: \n"
 "Language-Team: \n"
 "Language: es_ES\n"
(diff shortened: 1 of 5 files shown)


product_secondary_unit/views/product_views.xml
```diff
@@ -14,6 +14,11 @@
                        domain="[('product_tmpl_id', '=', id)]"
                        options="{'no_create': True}"/>
             </xpath>
+            <xpath expr="//field[@name='uom_po_id']" position="after">
```
nikul-serpentcs: IMO here Direct use `<field name="uom_po_id" position="after">`

sergio-teruel:
@nikul-serpentcs Changes done!! Thanks

nikul-serpentcs (approved):
Code Review LGTM

product_secondary_unit/models/product_template.py
```diff
@@ -16,3 +16,7 @@ class ProductTemplate(models.Model):
         comodel_name='product.secondary.unit',
         string='Default unit sale',
     )
+    purchase_secondary_uom_id = fields.Many2one(
```
lmignon: @sergio-teruel Why do we have a purchase field into an addon that does not depend on 'purchase'?  This field should be declared into `purchase_order_secondary_unit`

lmignon (changes requested):

product_secondary_unit/models/product_template.py
lmignon: The same comment applies to `sale_secondary_uom_id` This one should be declared into `sale_order_secondary_unit`. Do I miss something?

sergio-teruel: I desagree with that. As yo see here (https://github.com/OCA/OCB/blob/11.0/addons/product/models/product_template.py#L104) is in produt module where is declared this field, so I have done the same..

lmignon: I'll not block but I also find not logic to have these 'sale' ant 'purchase' fields of odoo core declared into product. If we can avoid the same mistake...

sergio-teruel:
@lmignon Changes done.. I change each field in your module...

lmignon (approved):
Thank you @sergio-teruel !

rousseldenis:
@sergio-teruel Is the title of this PR relevant towards changes ?

sergio-teruel:
Noo... ups!! Ok I removed the sale secondary unit. I moved it to sale_order_secondary_unit module...

sergio-teruel:
@rousseldenis Title changed!! Thanks

product_secondary_unit/i18n/es.po
```diff
@@ -39,6 +39,14 @@ msgstr "Creado en"
 msgid "Default Secondary Unit of Measure."
 msgstr "Segunda unidad de medida por defecto"
 
+#. module: product_secondary_unit
```
rousseldenis, line 42: @sergio-teruel I think this is not necessary anymore

rousseldenis (approved):

PR #400, opened 2018-10-31
[11.0][ADD] product_pricelist_supplierinfo: Supplier info prices in sales pricelists

pedrobaeza:
This module allows you to create a sales pricelist based on product supplierinfo prices. If you want, you can bypass minimum quantity in pricelist item.

Installation
============
Although this module doesn't depend technically on **Sales Management**, you must install it for configuring and seeing the effects of it.

Configuration
=============

To configure pricelists with the new feature of this module, you need to:

* Go to *Sales > Configuration > Settings* and check "Multiple Sales Prices per Product" option and   "Prices computed from formulas" after that. You must have correct permissions and you must install **Sales Management** app (sale) to see these settings.
* Create or edit a Sales Pricelist at *Sales > Catalog > Pricelists*.
* Add or edit a pricelist item and check "Formula" option in "Price Computation" section.
* You will see the new option "Prices based on supplier info".
* If you want to bypass the "Min.Quantity" field of the supplier info and always select the lowest quantity price, check the option "Ignore Supplier Info Min. Quantity".

Usage
=====

For adding supplier info:

* Go to *Sales > Catalog > Products*
* Open or create a product.
* Go to "Purchase" page.
* On "Vendors" section, add the supplier and prices.
* You can drag and drop for reordering these lines.

Check the remark in known issues about the supplier info line selection.

For checking pricelists in action, you can (with `sale` module installed):

* Go to *Sales > Orders > Quotations*
* Create or edit a quotation.
* Add a line.
* Select a product with the criteria to match the pricelist from supplier info.
* See the proper price appears in the line.

Known issues / Roadmap
======================

* First line that matches by minimum quantity criteria, independently from the supplier, will be the one that provides the price for the pricelist.
* There's no mechanism for selecting the supplier from possible sources like sales orders.

cc @Tecnativa


pedrobaeza:
Oops, I have pushed temp branch to OCA instead of own remote. Sorry.

product_pricelist_supplierinfo/models/product_pricelist.py
```diff
@@ -0,0 +1,41 @@
...
+    def _compute_price_rule(self, products_qty_partner, date=False,
+                            uom_id=False):
+        """Recompute price after calling the atomic super method for
+        getting proper prices when based on supplier info.
+        """
+        rule_obj = self.env['product.pricelist.item']
+        result = super(ProductPricelist, self)._compute_price_rule(
+            products_qty_partner, date, uom_id)
+        for product, qty, _partner in products_qty_partner:
+            rule = rule_obj.browse(result[product.id][1])
```
lmignon, line 23: The call to `browse` into a loop should be avoided. (potential perf issue). It would be better to collect all the product's ids before the loop and call browse for the whole list. Moreover, it appears that you are only interested by `supplieinfo`rules. Why not a call to `search`?

product_pricelist_supplierinfo/models/product_product.py
```diff
@@ -0,0 +1,71 @@
...
+
+
+class ProductProduct(models.Model):
+    _inherit = 'product.product'
+
+    def _get_supplierinfo_pricelist_price(self, rule, date=None,
+                                          quantity=None):
+        """Method for getting the price from supplier info."""
+        self.ensure_one()
+        domain = [('id', 'in', self.seller_ids.ids)]
```
lmignon: To get the value for `seller_ids.ids` the orm must do a first sql search query on the `product_supplier_info` table. To avoid this query, you should replace this first part of the domain by `['|', ('product_id', '=', self.id), ('product_tmpl_id', '=', self.product_tmpl_id)]`
An other way is to try to keep advantage of the cache by doing all the logic with the orm (no `search`)

product_pricelist_supplierinfo/tests/test_product_supplierinfo.py
```diff
@@ -0,0 +1,107 @@
+# Copyright 2018 Tecnativa - Vicent Cubells
+# Copyright 2018 Tecnativa - Pedro M. Baeza
+# License AGPL-3 - See http://www.gnu.org/licenses/agpl-3.0.html
+
+from odoo.tests import common
+
+
+class TestProductSupplierinfo(common.SavepointCase):
+    at_install = False
```
lmignon: Please remove `at_install` and `post_install`

lmignon (changes requested):

product_pricelist_supplierinfo/tests/test_product_supplierinfo.py
pedrobaeza: I can't remove it because I have the problem when creating the products that if `purchase`is installed, it fails by the SQL constraint.

lmignon: @pedrobaeza Tests are green without the post_install. That must be removed. https://travis-ci.org/acsone/product-attribute/jobs/449732838

pedrobaeza: OK, I will apply odoo/odoo#27883 in the systems where I put this module for not fighting you each time.

product_pricelist_supplierinfo/models/product_product.py
pedrobaeza: @lmignon accessing `ids` doesn't require to make the select under the hook. This is the only value that is already pre-stored when you make a browse or access a relational field. It's when you access other fields that prefetching machinery starts to work, make the select for the whole recordset and cache data for accessing. Can you please recheck?

I'm using this domain technique for readability, as using `filtered` or `if` conditions would affect it. The drawback is not too much or it's even better than the other way, as we only have 2 queries (search and select), reading one record (that it can even be on cache). On contrary, the other way you have to select all supplierinfo records, which might spend more time than the other 2 queries.

product_pricelist_supplierinfo/models/product_pricelist.py
pedrobaeza: They are already in cache for sure as `super` has used these rules for selecting them, but OK, I add a line for prefetching them if needed.

pedrobaeza:
@lmignon changes or comments done. Can you please review again?

product_pricelist_supplierinfo/models/product_product.py
lmignon: @pedrobaeza AFAIK, `ids` requires to make the select under the hook. Indeed, the SQL FK between product_supplierinfo and product_template is defined on the table product_supplierinfo. Therefore to get the list of ids of `product.supplierinfo` from `product.product` odoo must make an sql query `select .... from product_supplierinfo where product_tmpl_id = %s`
I agree that to get the `id` of a Many2one field no query is done under the hook since the value is stored as a column of the current model table. But it's not the case for the others kinds of relation fields.  That being said, in this case we can hope that the seller_ids will be retrieved with 1 sql query thanks to the prefetch mechanism into the orm. If the relation is already prefetched you should maybe replace the search by a loop on the recordset. What do you think about?

CasVissers-360ERP:
@pedrobaeza 
Functional test, I think the minimum quantity process is not correct:
Two supplier info lines:
qty 1 = 35EU
qty 10 = 34EU
When creating an SO for:
qty 1 = 35EU
qty 2-9 = 34EU 
qty 10-x = 0EU
What  I would expect:
qty 1-9=35EU
qty 10-x=34EU

AaronHForgeFlow:
Hi @pedrobaeza, This is containing part of the functionality of the product_supplierinfo_for_customer module, isn't it? I like the idea of splitting that module

pedrobaeza:
Not exactly: you can use this module without `product_supplierinfo_for_customer`, which is the case for my customer requirement: they have supplierinfo pricelists with their "cost price", and they want to base their sale prices on these costs with a percentage margin, so they use this new option and put a negative discount for having an increase on it.

But indeed this can be used also for that module for pricelists, although it's less important now that Odoo allows to put specific prices per pricelist in the product form, and you should probably adjust the case where the same partner acts for supplier and for customer, or even without that, due to the restricted search which is done by default, so a glue module or a tweak at least should be done.

Said that, there's one remaining improvement from @lmignon's comments that I have to make here, which I'll do shortly.

product_pricelist_supplierinfo/models/product_product.py
pedrobaeza: @lmignon you're totally right about the extra query needed. What I was saying was for many2one indeed, not for many2many. Changed in next commit by the search. As this method is isolated and we don't know about cache status, let's go for the `search` and let ORM the rest.

CasVissers-360ERP:
@pedrobaeza What about the minimal qty calculation?

pedrobaeza:
I'm checking now.

product_pricelist_supplierinfo/models/product_product.py
lmignon: @pedrobaeza Thank you for the change. This part of the code looks better now. I'm still wonder if we could make the method `_get_supplierinfo_pricelist_price` works with a multi recordset. I'll comment on the method.

pedrobaeza: What is the interest of making this multi-record aware? Having into account that `_compute_price_rule` is already for one record, it's difficult to handle this is a proper way I think, but I haven't tried.

product_pricelist_supplierinfo/models/product_product.py
```diff
@@ -0,0 +1,75 @@
...
+                '|',
+                ('date_start', '=', False),
+                ('date_start', '<=', date),
+                '|',
+                ('date_end', '=', False),
+                ('date_end', '>=', date),
+            ]
+        # We use a different default order because we are interested in getting
+        # the price for lowest minimum quantity
+        price = self.env['product.supplierinfo'].search(
```
lmignon: @pedrobaeza what prevents us from doing this search for all products at once and then doing the loop? (nitpicking :smirk:)
In a lot of cases, price is computed for a lot of product with qty=1. The Idea would be to write the domain with 
```python
            '|',
            ('product_id', 'in', self.ids),
            ('product_tmpl_id', 'in', self.product_tmpl_id.ids),
```
 order the search by `order= 'product_id, product_tmpl_id, sequence, min_qty, price'`

The price can be computed each time the product change into the loop.

```python
result = dict.from_keys(self.ids, None)
psis = self.env['product.supplierinfo'].search(domain, order= 'product_id, product_tmpl_id, sequence, min_qty, price')
if psis:
    product_id = psis[0].product_id
    for psi in psis:
        if product_id != psi.product_id:
            # here we put all the logic from line 41 to line 62
            ...
            result[product_id.id] = price
return result
```

pedrobaeza: I don't think this deserves to be done as too convoluted: you come from the other method where each argument is a tuple that have to be evaluated for the proper rule. Then we would need to store all products and their corresponding quantities, and call this multi-record aware method with that new structure, and deal with this properly.

Uf, too much effort being this call normally call for one record when retrieving the price in the sales order line.

product_pricelist_supplierinfo/models/product_product.py
lmignon: @pedrobaeza The main interest is that all the supplier info could be retrieved with one call to search for all the products of a same price list rule. 
You are right, the code could be a little bit more complex to read but could scale more efficiently. As I said, it's a details and you already have done a great work.  If I or someone else has a performance problem with this module, there will still be time to adapt it at that time. At this point, I think the bottleneck will first be the native odoo code before this one.

product_pricelist_supplierinfo/models/product_product.py
pedrobaeza: Exactly, that's the main point I think: optimizing this will have 0 benefit with the super code we are already handling. Take into account that I'm already dealing with several passes over same tuple list for recomputing what is not correct, but before there has been 2 useless computations already. We have to push for a proper refactoring on master.

pedrobaeza:
@CasVissers the problem with quantities should be fixed now

lmignon (approved):
Thank you @pedrobaeza. LGTM (Code review + functional tests)

product_pricelist_supplierinfo/models/product_product.py
```diff
@@ -0,0 +1,79 @@
...
+            ]
+        # We use a different default order because we are interested in getting
+        # the price for lowest minimum quantity if no_supplierinfo_min_quantity
+        supplierinfos = self.env['product.supplierinfo'].search(
+            domain, order='sequence, min_qty, price',
+        )
+        if rule.no_supplierinfo_min_quantity:
+            price = supplierinfos[:1].price
+        else:
+            price = supplierinfos[-1:].price
```
AaronHForgeFlow, line 43: Why we take the last one here? To take the most expensive one?

pedrobaeza: No, on contrary, take the price for the maximum "minimum quantity".

If you have these supplierinfo:

1. Min. qty: 1 - Price: 10
2. Min. qty: 5 - Price: 9
3. Min. qty: 10 - Price: 8

Querying for quantity=5, it will return 1 and 2. The price to apply is 2.
Querying for quantity=10, it will return 1, 2 and 3. The price to apply is 3.

AaronHForgeFlow: Ok. If not minimum qty, take it. If not, the last one. Ok.

pedrobaeza: Yeah, in the case you check the mark, always record 1 is returned. Why? This is a requirement of my customer for several products for not propagation escalated prices from supplier to customer. They get better prices from supplier if purchasing big quantities, but the customer pays the same always, no matter the sold quantity.

AaronHForgeFlow: It makes sense

AaronHForgeFlow (approved):
:+1: Functional review

pedrobaeza:
@CasVissers with your final approval to quantities part we merge this.

CasVissers-360ERP:
@pedrobaeza going to test now.

CasVissers-360ERP:
@pedrobaeza 
My test:
Product x 
Supplierinfo - supplier A - min qty 1 - price 30
Supplierinfo - supplier A - min qty 10 - price 35
Supplierinfo - supplier A - min qty 15 - price 40
Pricelist line - Product X - 50% discount
_I know a bit illogical_

Create sale order, add product x:
qty 1-9 = 15,00 OK (thus fixed)
qty 10-15 = = 17,50 OK (thus fixed)
**qty 15-x = 17,50 Not OK...**

pedrobaeza:
@CasVissers I'm not able to reproduce the problem you state:

* Product configuration:
  seleccion_001

* Pricelist configuration:
  seleccion_004

* Quotation:
  seleccion_005

And a general advise: for proper testing, you should isolate the most possible variables. Discount change doesn't modify here anything and only introduces noise. You should make first the test without this change, and if the problem happens also without that change, then you should leave out that variable from the explanation. If the problem arises due to that change, then the problem is for sure related. But well, as said, it's not reproducible in both cases.

CasVissers-360ERP (approved):
Functional test OK. Nice module!

CasVissers-360ERP:
@pedrobaeza You are right, must have been something wrong in my test. Sorry for wasting your time. Great job on the module.

We just created something to base a sales price list item on supplier products, a dirty solution so not fitting for OCA (not possible in a clean way). But in combination with this module we can create a sale price list line for all products of a supplier based on the supplier purchase price. Great functionality.

PR #401, opened 2018-11-02
remove post_install

lmignon:
post_install are not required for travis on OCA repos. Tests are green without post_install.


pedrobaeza:
@lmignon as always, I talk when making integration tests that interacts with all the modules in that system. Here in the repo there are some modules that are not going to be installed. Any way, please don't open new PRs and let's continue in my main PR. I don't understand why do you have now this fight against `post_install`...

PR #404, opened 2018-11-06
[ADD] product_weight_through_uom: New module

chienandalu:
This module allows to use the unit of measure as a reference for the product
weight. So a product with a UoM of Kg will have a weight of 1.0 Kg. and a
product with a UoM of grams will have a weight of 0.001 Kg.

It still allows to adjust the total weight manually with an extra field.

cc @Tecnativa


product_weight_through_uom/models/product_template.py
```diff
@@ -0,0 +1,62 @@
...
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+
+from odoo import api, fields, models
+from odoo.addons import decimal_precision as dp
+
+
+class ProductTemplate(models.Model):
+    _inherit = "product.template"
+
+    extra_weight = fields.Float(
```
pedrobaeza, line 11: Put this field at product.product level, not here.

product_weight_through_uom/models/product_template.py
```diff
@@ -0,0 +1,62 @@
...
+    @api.onchange('extra_weight')
+    def _onchange_extra_weight(self):
+        if self.is_weight_uom:
+            self.weight = self.uom_id.factor_inv + self.extra_weight
+        else:
+            self.weight = self.weight + self.extra_weight
+
+    @api.multi
+    def write(self, vals):
+        if vals.get('uom_id') or vals.get('extra_weight'):
```
pedrobaeza, line 62: Why we need to overwrite both methods in template/product?

pedrobaeza (changes requested):

product_weight_through_uom/models/product_template.py
chienandalu: What happens if there are not variants enabled. The products are edited through the `product.template` model and the proper values are set to their `product.product` counterpart: https://github.com/odoo/odoo/blob/11.0/addons/product/models/product_template.py#L243

Then maybe the way is to use a similar compute/inverse mechanism with this field.

product_weight_through_uom/models/product_template.py
chienandalu: A similar problem with this. Template and product could have a dependant behavior or not.

product_weight_through_uom/models/product_template.py
pedrobaeza: Yes, good idea.

chienandalu:
Let's see now

chienandalu:
@pedrobaeza What do you think about initializing the weight of the affected products?

product_weight_through_uom/models/product_template.py
```diff
@@ -19,6 +21,21 @@ class ProductTemplate(models.Model):
         default=False,
     )
 
+    @api.depends('product_variant_ids', 'product_variant_ids.weight')
+    def _compute_extra_weight(self):
+        unique_variants = self.filtered(
+            lambda x: len(x.product_variant_ids) == 1)
+        for template in unique_variants:
+            template.extra_weight = template.product_variant_ids.extra_weight
+        for template in (self - unique_variants):
```
pedrobaeza: No need of this. ORM already initializes to 0 all values before calling compute method. Any way, a better alternative would be:
```python
(self - unique_variants).update({'extra_weight': 0.0})
```

product_weight_through_uom/models/product_template.py
```diff
@@ -0,0 +1,79 @@
...
+    def _inverse_extra_weight(self):
+        for template in self.filtered(
+                lambda x: len(x.product_variant_ids) == 1):
+            template.product_variant_ids.extra_weight = template.extra_weight
+
+    @api.depends('uom_id.category_id')
+    def _compute_is_weight_uom(self):
+        product_uom_categ_kgm = self.env.ref('product.product_uom_categ_kgm')
+        for product in self:
+            if product.uom_id.category_id == product_uom_categ_kgm:
```
pedrobaeza:
```python
product.is_weight_uom = (
    product.uom_id.category_id == product_uom_categ_kgm
)
```

product_weight_through_uom/models/product_template.py
```diff
@@ -0,0 +1,79 @@
...
+        for template in (self - unique_variants):
+            template.extra_weight = 0.0
+
+    @api.multi
+    def _inverse_extra_weight(self):
+        for template in self.filtered(
+                lambda x: len(x.product_variant_ids) == 1):
+            template.product_variant_ids.extra_weight = template.extra_weight
+
+    @api.depends('uom_id.category_id')
```
pedrobaeza:
```suggestion
    @api.depends('uom_id', 'uom_id.category_id')
```
It's more important to take into account that someone changes UoM in the product that a category is changed in the UoM :wink:

product_weight_through_uom/models/product_template.py
```diff
@@ -0,0 +1,79 @@
...
+        # When a UoM is in Weight category, we set the product weight to its
+        # corresponding conversion to Kg.
+        if self.is_weight_uom:
+            self.weight = self.uom_id.factor_inv + self.extra_weight
+
+    @api.onchange('extra_weight')
+    def _onchange_extra_weight(self):
+        if self.is_weight_uom:
+            self.weight = self.uom_id.factor_inv + self.extra_weight
+        else:
```
pedrobaeza, line 57: You shouldn't change weight value if it's not weight. I would say more: you should hide this field and put 0 in extra weight when changing UoM to a non weight one.

pedrobaeza (changes requested):

chienandalu:
What about https://github.com/OCA/product-attribute/pull/404#issuecomment-439080613 ?

pedrobaeza:
I don't think we should alter that data, because they can be already correct counting with the extra weight. Resetting them on module installation can lead to data loss. Other thing is that we migrate our customers with a special script for initializing things, but we can't assume it fits all the users of the module.

The only formula I think on is to initialize the field if the weight is currently 0. Deal?

chienandalu:
Great :smile:

chienandalu:
Ok, done

product_weight_through_uom/hooks.py
```diff
@@ -0,0 +1,15 @@
...
+
+def post_init_hook(cr, registry, vals=None):
+    """For brand new installations"""
+    env = api.Environment(cr, SUPERUSER_ID, {})
+    # Change only those with no weight already set
+    products_init = env['product.product'].search([
+        ('weight', '=', 0),
+    ]).filtered('is_weight_uom')
+    for product in products_init:
+        product._onchange_uom()
```
pedrobaeza: As this method overwrites the standard one, it's performing a lot of things that can be undesirable, so the options are:

* Don't call here this method and put directly the assignment.
* Create another onchange method with a different name (`_onchange_uom_product_weight_through_uom`) for performing this task.

product_weight_through_uom/models/product.py
```diff
@@ -0,0 +1,53 @@
...
+    )
+
+    @api.onchange('extra_weight')
+    def _onchange_extra_weight(self):
+        if self.is_weight_uom:
+            self.weight = self.uom_id.factor_inv + self.extra_weight
+        else:
+            self.extra_weight = 0
+
+    @api.onchange('uom_id', 'uom_po_id')
```
pedrobaeza: Are you changing the scope of the onchange?

pedrobaeza (changes requested):

product_weight_through_uom/models/product.py
chienandalu: No

pedrobaeza: OK. If so, we need a new onchange.

chienandalu:
Changed

pedrobaeza (approved):
Proceeding with this after the long way to get it fit

PR #405, opened 2018-11-14
FIX README [product_supplierinfo_for_customer]

rafaelbn:
Update README with real functionality

Please @sergio-teruel review

cc @tecnativa

### DIFF
--- a/product_supplierinfo_for_customer/README.rst
+++ b/product_supplierinfo_for_customer/README.rst
@@ -27,7 +27,7 @@ Product Supplierinfo for Customers
 
 This modules allows to use supplier info structure, available in
 *Inventory* tab of the product form, also for defining customer information,
-allowing to define prices per customer and product.
+allowing to define "Customer Product Name" and "Customer Product Code".
 
 **Table of contents**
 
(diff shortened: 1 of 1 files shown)


rafaelbn:
@mreficent you migrate this module in https://github.com/OCA/product-attribute/pull/378 maybe you could review please

product_supplierinfo_for_customer/README.rst
```diff
@@ -45,18 +45,12 @@ Usage
...
-supplierinfo at the product form.
+where you can define records for customers.
 
 Known issues / Roadmap
 ======================
 
-* Product prices through this method are only guaranteed on the standard sale
-  order workflow. Other custom flows maybe don't reflect the price.
-* The minimum quantity will neither apply on sale orders.
+*
```
MiquelRForgeFlow, line 53: Then just remove this `Known issues / Roadmap` section if becomes empty.

product_supplierinfo_for_customer/README.rst
```diff
@@ -27,7 +27,7 @@ Product Supplierinfo for Customers
 
 This modules allows to use supplier info structure, available in
 *Inventory* tab of the product form, also for defining customer information,
-allowing to define prices per customer and product.
+allowing to define "Customer Product Name" and "Customer Product Code".
```
AaronHForgeFlow, line 30: Well, you can define prices for the customer, even if they are not used elsewhere. If we remove this here I'll put something in the Known Issues section like "The prices and minimum quantitties defined there are just information"

rousseldenis:
@sergio-teruel

rousseldenis:
@rafaelbn Can you attend comments ?

rafaelbn:
What I mean @aheficent @mreficent is that if we cannot set prices we shouldn't write in the README that we can. See:



Do you understand? :smile:

product_supplierinfo_for_customer/README.rst
rafaelbn: See my screenshot, you cannot :thinking: 
https://user-images.githubusercontent.com/8736623/51282091-8d937580-19e4-11e9-8955-e590f73a7982.png

AaronHForgeFlow:
hmmm I think you should see the same view as for suppliers. Maybe something change between 10.0 and 11.0. Need to recheck. BTW we can keep this issue open IMO.

PR #406, opened 2018-11-15
[FIX] 10.0 product_profile : do not customize product easy form view

PierrickBrun:
Before this fix:
Product.product.easy.form view does not work (odoo tries to use the field product_id which is not loaded)

After this fix:
The view works as expected and isn't customized by the product_profile module.
It woul be complex to implement the profile customization into the easy form view while being of little interest. There is already a view with every fields that handle it properly.

### DIFF
--- a/product_profile/models/product.py
+++ b/product_profile/models/product.py
@@ -267,7 +267,11 @@ class ProductProduct(models.Model):
     @api.model
     def fields_view_get(self, view_id=None, view_type='form',
                         toolbar=False, submenu=False):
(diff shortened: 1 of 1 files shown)


bealdav (approved):

chafique-delli (approved):

hparfr (approved):

mourad-ehm (approved):

bealdav:
ping @pedrobaeza could you merge please. Thanks

PR #416, opened 2018-12-06
[12.0][MIG] Migrated product_sequence to v12.

sudhir-erpharbor:
Please review and give feedback.


atchuthan:
@pedrobaeza 
We have the CLA document for SodexisTeam in https://github.com/OCA/OpenUpgrade/blob/12.0/doc/cla/corporate/sodexis.md

what am I missing?

pedrobaeza:
@atchuthan that one is Odoo CLA, not OCA CLA. Did you send OCA ECLA or ICLA? There's also an issue with used emails.

atchuthan:
> @atchuthan that one is Odoo CLA, not OCA CLA. Did you send OCA ECLA or ICLA? There's also an issue with used emails.

OCA ECLA

scigghia (approved):
Tested on Odoo 12 CE and works

product_sequence/models/product_product.py
```diff
@@ -0,0 +1,76 @@
...
+        return super(ProductProduct, self).create(vals)
+
+    @api.multi
+    def write(self, vals):
+        """To assign a new internal reference, just write '/' on the field.
+        Note this is up to the user, if the product category is changed,
+        she/he will need to write '/' on the internal reference to force the
+        re-assignment."""
+        for product in self:
+            if vals.get('default_code', '') == '/':
```
BT-aestebanez, line 54: this `if` could be out of the loop for improving the performance

product_sequence/models/product_product.py
```diff
@@ -0,0 +1,76 @@
...
+                category_id = vals.get('categ_id', product.categ_id.id)
+                category = self.env['product.category'].browse(category_id)
```
BT-aestebanez, line 56: what if there is no category?

product_sequence/models/product_product.py
```diff
@@ -0,0 +1,76 @@
...
+                sequence = category.sequence_id
+                if not sequence:
+                    sequence = self.env.ref(
+                        'product_sequence.seq_product_auto')
+                ref = sequence.next_by_id()
+                vals['default_code'] = ref
```
BT-aestebanez, line 62: are you sure you want to update the main dict? I would say you should update only the values to write in the record of the loop

BT-aestebanez (commented):
Thanks for your contribution.
I reviewed the code (translations not reviewed) and I only have three small comments :)
Alvaro

product_sequence/models/product_product.py
```diff
@@ -0,0 +1,76 @@
...
+
+    default_code = fields.Char(
+        required=True,
+        default='/',
+        track_visibility='onchange',
+        help="Set to '/' and save if you want a new internal reference "
+             "to be proposed."
+    )
+
+    _sql_constraints = [
```
celm1990, line 21: Because add sql contraints????

this functionality is added on module [product_code_unique](https://github.com/OCA/product-attribute/blob/12.0/product_code_unique/models/product.py#L10)

I prefered keep into 2 modules

celm1990 (changes requested):

product_sequence/models/product_product.py
pedrobaeza: Yes, I agree

MiquelRForgeFlow:
This PR should be closed. Superseed by https://github.com/OCA/product-attribute/pull/507.

PR #420, opened 2018-12-24
[11.0][ADD] product_restricted_type

AdriaGForgeFlow:
Add module product_restricted_type to v11

This module adds 'restricted_product_type' to product_category model.
With this optional selection field 'restricted_product_type' should
have the same values as in the product.template, 'type' field.

A constrain is also established in the product so that when the 'type'
defined in the product does not match with the 'restricted_product_type'
defined in the product category it raises a Validation Error.

Also, in the product category one would not be able to change the field
'restricted_product_type' if the category has been assigned at least to
one product that already has this category, but has a different 'type' value.


LoisRForgeFlow:
@ageficent Test are failing in travis, can you have a look?

JordiBForgeFlow (approved):
The PR seems to have non-related commits. Perhaps you should rebase?

JordiBForgeFlow:
@lreficent Can you review again?

LoisRForgeFlow (approved):
LGTM

PR #351, opened 2018-04-24
[MIG] product_uom: migration to 11.0

hprasetyou:

PR #423, opened 2019-01-08
[11.0][MIG] product_uom

MiquelRForgeFlow:
Superseeds https://github.com/OCA/product-attribute/pull/351.


pedrobaeza:
The name doesn't reflect correctly the purpose of the module. I would rename it to `product_uom_extra_data`, and move it to OCA/community-data-files.

PR #424, opened 2019-01-11
[10.0][ADD] product_origin

rousseldenis:


pedrobaeza:
Shouldn't it be `product_origin_country` or `product_source_country`?

rousseldenis:
@pedrobaeza It's 'country of origin of the product'. So, if we put product in front : 'product's country of origin. So : product_country_origin. That's my reflexion :smiley:

pedrobaeza:
Well, I see the sorting a bit artificial on module's name, as in English a noun that specifies more another noun without preposition should go before. You say "origin country" or "country of origin", but no "country origin". @elicoidal as you are very well-known on English usages, can you tell us about this? Or any other English native speaker...

rousseldenis:
@pedrobaeza https://en.wikipedia.org/wiki/Country_of_origin

I'll change to product_country_of_origin. That's better definitely

legalsylvain (changes requested):
Hi Denis. Thanks for this module.

What about to introduce in the module also state (like odoo)
In that case the name of the module could be simply product_origin 

And add onchange and contraint. You can find all the code here if you want.

https://github.com/grap/grap-odoo-business/blob/8.0/sale_food/models/product_product.py

Regards.

rousseldenis:
@legalsylvain I hope this will attend everyone expectations :smile:

product_origin/README.rst
```diff
@@ -0,0 +1,98 @@
...
+
+**Table of contents**
+
+.. contents::
+   :local:
+
+Usage
+=====
+
+Go to product form, then fill in the country and/or state of origin of the
```
elicoidal, line 41: I usually prefer enumerated list for this kind of description
```
#. Go to product form
#. Click on 'General Information' tab.
#. Fill in the country and/or state of origin of the product
```

product_origin/models/product_template.py
```diff
@@ -0,0 +1,34 @@
...
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+from odoo import api, fields, models
+
+
+class ProductTemplate(models.Model):
+
+    _inherit = 'product.template'
+
+    country_id = fields.Many2one(
```
elicoidal, line 12: This should not be modifiable otherwise you get discrepancies with the state.
IMO should be calculated/readonly

product_origin/readme/HISTORY.rst
```diff
@@ -0,0 +1,4 @@
+10.0.1.0.0 (2019-01-11)
+~~~~~~~~~~~~~~~~~~~~~~~
+
+* First version
```
elicoidal: be more explicit...

elicoidal (commented):
Be careful that country is tied to state. IMO current design might bring inconsistency in the data.
"Country of Origin" looks better

product_origin/models/product_template.py
```diff
@@ -0,0 +1,34 @@
...
+    )
+
+    state_id = fields.Many2one(
+        comodel_name='res.country.state',
+        string='Country State of Origin',
+        ondelete='restrict',
+        index=True,
+    )
+
+    @api.onchange('country_id')
```
daramousk: Could this domain be possibly set on the xml? ex. `<field name="state_id" domain="[('country_id', '=', country_id)]"/>` or on the field's definition so to avoid the `onchange`?

setup/product_origin/setup.py
```diff
@@ -0,0 +1,6 @@
+import setuptools
```
daramousk, line 1: Maybe encoding needed here as well as the other init files?

product_origin/models/product_template.py
```diff
@@ -0,0 +1,34 @@
...
+
+class ProductTemplate(models.Model):
+
+    _inherit = 'product.template'
+
+    country_id = fields.Many2one(
+        comodel_name='res.country',
+        string='Country of Origin',
+        ondelete='restrict',
+        index=True,
```
daramousk, line 16: Is `index=True` here really needed?

daramousk (commented):
Some comments

legalsylvain (approved):
Thanks

Cedric-Pigeon:
@rousseldenis Could you review comments please?

product_origin/models/product_template.py
rousseldenis: I'd prefer let them for searches

setup/product_origin/setup.py
rousseldenis: In setup, this is not needed. This is generated by https://github.com/acsone/setuptools-odoo

rousseldenis:
@Cedric-Pigeon commentes attended

Cedric-Pigeon (approved):

Cedric-Pigeon:
@rousseldenis Travis is :red_circle:  . :-(

PR #432, opened 2019-02-06
[11.0][IMP] product_sequence: Remove default_code sql constraint

sergio-teruel:
The sql restriction is added in product_code_unique, so if you need this restriction, please install this module.

cc @Tecnativa


product_sequence/README.rst
```diff
@@ -1,17 +1,41 @@
...
+    :target: https://translation.odoo-community.org/projects/product-attribute-11-0/product-attribute-11-0-product_sequence
+    :alt: Translate me on Weblate
+.. |badge5| image:: https://img.shields.io/badge/runbot-Try%20me-875A7B.png
+    :target: https://runbot.odoo-community.org/runbot/135/11.0
+    :alt: Try me on Runbot
+
+|badge1| |badge2| |badge3| |badge4| |badge5| 
+
 This module allows to associate a sequence to the product reference.
 The reference (default code) is unique (SQL constraint) and required.
```
pedrobaeza: Change this text for advising the installation of the other module.

product_sequence/readme/INSTALL.rst
```diff
@@ -0,0 +1,7 @@
+Prior to installing this module, if you have any existing products you should
```
pedrobaeza: This should be removed also.

sergio-teruel:
@pedrobaeza changes done!

product_sequence/readme/INSTALL.rst
```diff
@@ -0,0 +1,2 @@
+Products with a default_code of '/' or empty will automatically be assigned a
```
pedrobaeza, line 1: Is this mechanism still needed now that the constraint has been removed?

pedrobaeza:
Sorry to insist, but I still don't see the reference to install the other module as said.

Cedric-Pigeon:
@sergio-teruel Would not it be better to put product_code_unique in dependencies to replace the sql constrains. Removing the constrains as this will change the behaviour....

Cedric-Pigeon (changes requested):
See above

pedrobaeza:
But the idea is to not break existing installations adding a dependency. Moreover, we are doing this because we don't want that constraint. It's not the aim of the module, and for our use case, it's wasting our customer DB.

Cedric-Pigeon:
@pedrobaeza This addon aims to link a sequence to the product default code. I don't see any cases in which a sequence can provide duplicates.

pedrobaeza:
There are several, like obsolescence of certain products that are replaced by others with the same internal code, mistakes with the product UoM that requires to create a new product, but they want to have both with the code for searching, etc.

Any way, this module shouldn't extra-limit its purpose. You have the other one for the constraint.

Cedric-Pigeon:
@pedrobaeza Indeed. I agree. Than it would be useful to add a history section in the readme to explain the change and to let people who will migrate from previous versions what they have to do to keep the constrain

pedrobaeza:
Well, that's why I was insisting to @sergio-teruel to put that information in the README. Put it as HISTORY can be other option indeed.

sergio-teruel:
@pedrobaeza @Cedric-Pigeon changes done

pedrobaeza (approved):

AaronHForgeFlow (approved):
:+1: LGTM + Tested

PR #433, opened 2019-02-11
[12.0] [FIX] Patch product_brand

willmakenoise:
Fixed error when installing product_brand module:
```
File "odoo12/custom/addons/product_brand/reports/sale_report.py", line 21, in _query
    with_clause, following = query_str.split('SELECT')
ValueError: too many values to unpack (expected 2)
```

Refactor code to use similar style of updating query as standard odoo modules do.

### DIFF
--- a/product_brand/reports/sale_report.py
+++ b/product_brand/reports/sale_report.py
@@ -14,21 +14,6 @@ class SaleReport(models.Model):
 
     # pylint:disable=dangerous-default-value
     def _query(self, with_clause='', fields={}, groupby='', from_clause=''):
-        query_str = super()._query(
-            with_clause=with_clause, fields=fields, groupby=groupby,
-            from_clause=from_clause)
-        # Split query
-        with_clause, following = query_str.split('SELECT')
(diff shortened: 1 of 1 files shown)


pedrobaeza (approved):
Good idea to simplify code.

pedrobaeza:
Please sign CLA.

willmakenoise:
I've now signed CLA and e-mailed it.

Feels good finally being a part of OCA :)

willmakenoise:
Flake8 seems to fail because of a line length being 89.
Should I fix this? 

I just watched Raymond Hettinger's "Beyond PEP 8" talk... so I'm now unsure if this would be the right thing to do ;)

pedrobaeza:
About PEP8, we use stricter one, and it's not really a limitation, but a good way to avoid other problems (like cyclomatic complexity). Check full guidelines at https://github.com/OCA/odoo-community.org/blob/master/website/Contribution/CONTRIBUTING.rst

chienandalu (approved):
Looks good. Thanks!

PR #434, opened 2019-02-11
12.0 mig product tags

hegenator:
Migration of `product_template_tags` from 10.0 to 12.0


daramousk (commented):
Mostly suggestions

product_template_tags/models/product_template.py
```diff
@@ -0,0 +1,14 @@
...
+
+
+class ProductTemplate(models.Model):
+
+    _inherit = 'product.template'
+
+    tag_ids = fields.Many2many(
+        comodel_name='product.template.tag', string="Product Tags",
+        relation='product_template_product_tag_rel',
+        column1='product_tmpl_id', column2='tag_id')
```
daramousk, line 14: Is there a reason for giving the `relation`, `column1` and `column2` arguments? As I understand they will be automatically set.

product_template_tags/models/product_template_tag.py
```diff
@@ -0,0 +1,32 @@
...
+from odoo import api, fields, models
+
+
+class ProductTemplateTag(models.Model):
+
+    _name = 'product.template.tag'
+    _description = 'Product Tag'
+
+    name = fields.Char(string="Name", required=True, translate=True)
+    color = fields.Integer(string="Color Index")
```
daramousk, line 13: There needs to be a constraint here since the colours that can be shown have specific numbers. Or maybe it could be a selection field?

product_template_tags/models/product_template_tag.py
```diff
@@ -0,0 +1,32 @@
...
+    product_tmpl_ids = fields.Many2many(
+        comodel_name='product.template', string="Products",
+        relation='product_template_product_tag_rel',
+        column1='tag_id', column2='product_tmpl_id')
```
daramousk, line 17: Same here, do you need the `relation`, `column1`, and `column2` attributes to be set explicitly?

product_template_tags/models/product_template_tag.py
```diff
@@ -0,0 +1,32 @@
...
+    products_count = fields.Integer(
+        string="# of Products", compute='_compute_products_count', store=True)
+    company_id = fields.Many2one(
+        comodel_name='res.company', string="Company",
+        default=lambda self: self._default_company())
```
daramousk, line 22: Maybe this could be `default= lambda self: self.env.user.company_id` so that you could avoid defining a function.

product_template_tags/tests/test_product_tags.py
```diff
@@ -0,0 +1,21 @@
+from odoo.tests import common
+
```
daramousk, line 2: Maybe here (and in some init files, you maybe want your copyright plus the licence comment (actually I think on the init files only the licence)

daramousk: And maybe some tests trying to create tags with users from different companies, and tag colour testing

product_template_tags/views/product_template.xml
```diff
@@ -0,0 +1,52 @@
...
+
+    <record model="ir.ui.view" id="product_template_form_view">
+        <field name="name">product.template.form (in product_template_tags)</field>
+        <field name="model">product.template</field>
+        <field name="inherit_id" ref="product.product_template_form_view"/>
+        <field name="arch" type="xml">
+
+            <xpath expr="//page[@name='general_information']" position="inside">
+                <group>
+                    <field name="tag_ids" widget="many2many_tags" options="{'color_field': 'color'}"/>
```
daramousk, line 15: This seems a little bit off in the bottom of the page and occupying 8 columns. Maybe it would be better to be under one of the 2 columns before the tabs?

product_template_tags/views/product_template.xml
```diff
@@ -0,0 +1,52 @@
...
+    </record>
+
+    <record model="ir.ui.view" id="product_template_search_view">
+        <field name="name">product.template.search (in product_template_tags)</field>
+        <field name="model">product.template</field>
+        <field name="inherit_id" ref="product.product_template_search_view"/>
+        <field name="arch" type="xml">
+
+            <field name="categ_id" position="after">
+                <field name="tag_ids"/>
```
daramousk, line 45: What is this for?

product_template_tags/views/product_template_tag.xml
```diff
@@ -0,0 +1,73 @@
...
+    <record model="ir.ui.view" id="product_template_tag_search_view">
+        <field name="name">product.template.tag.search (in product_template_tags)</field>
+        <field name="model">product.template.tag</field>
+        <field name="arch" type="xml">
+            <search>
+                <field name="name"/>
+                <field name="company_id"/>
+                <group expand="0">
+                    <filter name="group_by_company" string="Company"
+                            domain="[]" context="{'group_by': 'company_id'}"
```
daramousk, line 40: I think the domain can be omitted.

product_template_tags/views/product_template_tag.xml
```diff
@@ -0,0 +1,73 @@
...
+        </field>
+    </record>
+
+    <record model="ir.ui.view" id="product_template_tag_tree_view">
+        <field name="name">product.template.tag.tree (in product_template_tags)</field>
+        <field name="model">product.template.tag</field>
+        <field name="arch" type="xml">
+            <tree>
+                <field name="name"/>
+                <field name="company_id" widget="selection" groups="base.group_multi_company"/>
```
daramousk, line 53: Do we need a widget `selection` on a non editable list view?

product_template_tags/views/product_template_tag.xml
```diff
@@ -0,0 +1,73 @@
...
+<!-- Copyright 2017 ACSONE SA/NV
+     License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl). -->
+
+<odoo>
+
+    <record model="ir.ui.view" id="product_template_tag_form_view">
+        <field name="name">product.template.tag.form (in product_template_tags)</field>
+        <field name="model">product.template.tag</field>
+        <field name="arch" type="xml">
+            <form>
```
daramousk, line 11: As I understand it, we want the user to be able to set tags/colours for products per company. If that is true, then there should be the `colour` field here, and the user should not be able to add their own tags/colours from the product form. Also, the rights should be different (`group_user` should be able to only read, and possibly only some stock managers be able to create/write/unlink)

product_template_tags/views/product_template_tag.xml
```diff
@@ -0,0 +1,73 @@
...
+    </record>
+
+    <record model="ir.actions.act_window" id="product_template_tag_act_window">
+        <field name="name">Product Template Tag</field>
+        <field name="res_model">product.template.tag</field>
+        <field name="view_mode">tree,form</field>
+        <field name="view_type">form</field>
+    </record>
+
+    <record model="ir.ui.menu" id="product_template_tag_config_menu">
```
daramousk, line 66: Maybe this would make more sense under Stock -> Products -> Configuration? But of course then you would need stock. Would it be a bad idea to set it as a dependency? Probably. Might as well keep the menu here, still it seems a little bit out of place.

emagdalenaC2i:
Please close this PR because it is already migrated in #448

hegenator:
Closing the PR since it has already been migrated elsewhere.

PR #435, opened 2019-02-18
[ADD] Adds product_code_cpv module.

daramousk:


pedrobaeza:
Maybe this should go to OCA/community-data-files, and if this is a standard nomenclature, include as data the records for the given nomenclature.

product_code_cpv/__init__.py
```diff
@@ -0,0 +1,4 @@
+# -*- coding: utf-8 -*-
+# Copyright 2019 bitwise.solutions <https://bitwise.solutions>
```
NL66278, line 2: No copyright in init files.

product_code_cpv/__init__.py
```diff
@@ -0,0 +1,4 @@
...
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
```
NL66278, line 3: https: on license files.

product_code_cpv/models/product_template_cpv.py
```diff
@@ -0,0 +1,12 @@
+# -*- coding: utf-8 -*-
+# Copyright 2019 bitwise.solutions <https://bitwise.solutions>
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+from odoo import fields, models
+
+
+class ProductTemplateCPV(models.Model):
+    _name = 'product.template.cpv'
+
+    name = fields.Char()
```
NL66278, line 10: All fields should be both required (therefore not NULLable in SQL) and readonly.

NL66278 (changes requested):
:+1: Nice addition, but some changes would be welcome. See other comments. And you might look into Pedro's suggestion on the right repository.

PR #436, opened 2019-02-20
[11.0][FIX] product_pricelist_supplierinfo: Only works fine with product.product

carlosdauden:
Add product.template compatibility

@Tecnativa


product_pricelist_supplierinfo/models/product_pricelist_supplierinfo_mixin.py
```diff
@@ -0,0 +1,73 @@
...
+
+
+class ProductPricelistSupplierinfoMixin(models.AbstractModel):
+    _name = 'product.pricelist.supplierinfo.mixin'
+
+    def _get_supplierinfo_pricelist_price(self, rule, date=None,
+                                          quantity=None):
+        """Method for getting the price from supplier info."""
+        self.ensure_one()
+        if self._name == 'product.template':
```
pedrobaeza: This mixin is not correct if you have to put here this kind of code, and the extra model is not justified only for this. For this specific case, I would:

1. Define one new method in `product.template` with the same logic as `_get_supplierinfo_pricelist_price`, adding an extra keyword argument called `extra_domain` or similar. Add the logic for seeing if `extra_domain` has value, then sum it to the computed domain. Remove this part of code in which I'm writing over now.
2. Define `_get_supplierinfo_pricelist_price` in `product.template`, calling this method with the extra domain logic for product.template.
3. Do the same on `product.product`, but adding the extra domain for product.

product_pricelist_supplierinfo/models/product_product.py
```diff
@@ -1,72 +1,13 @@
...
-            if rule.price_min_margin:
-                price_min_margin = convert_to_price_uom(rule.price_min_margin)
-                price = max(price, price_limit + price_min_margin)
-            if rule.price_max_margin:
-                price_max_margin = convert_to_price_uom(rule.price_max_margin)
-                price = min(price, price_limit + price_max_margin)
-        return price
-
-    @api.multi
     def price_compute(self, price_type, uom=False, currency=False,
```
pedrobaeza, line 15: Is this still needed as you have put the same in `product.template`?

pedrobaeza (changes requested):

rafaelbn (approved):
Functional :+1:

product_pricelist_supplierinfo/models/product_pricelist_supplierinfo_mixin.py
carlosdauden: I can't add extra domain because self.id has distinct values and is assigned to diferent fields:
https://github.com/OCA/product-attribute/pull/436/files#diff-394e2468e18fe6e03a642a868c33bc75R18
https://github.com/OCA/product-attribute/pull/436/files#diff-394e2468e18fe6e03a642a868c33bc75R23

carlosdauden: Mixed removed

product_pricelist_supplierinfo/models/product_product.py
carlosdauden: Yes, because is defined too in:
https://github.com/odoo/odoo/blob/19558129f0e2d5c9940f6fa95534840f8f025547/addons/product/models/product.py#L531
and this code try skip the defined method to it don'f fails when price_type == 'supplierinfo':
https://github.com/odoo/odoo/blob/19558129f0e2d5c9940f6fa95534840f8f025547/addons/product/models/product.py#L548

carlosdauden:
Untested code is a copy of:
https://github.com/odoo/odoo/blob/11.0/addons/product/models/product_pricelist.py#L224-L240

pedrobaeza (approved):

PR #437, opened 2019-02-25
11.0][IMP] product_template_tags: No create edit tags

alvaro-domatix:
Added the no_create_edit option to the tags in order to do not allow to create duplicates.


leemannd (commented):
IMO it is too restrictive. You may want to allow users to create tags from this view

PR #439, opened 2019-03-12
[ADD] 12.0 module product_base_mass_add

PierrickBrun:
@sebastienbeau @bealdav @mourad-ehm


pedrobaeza:
README is very confusing. Please specify better the aim of this module.

pedrobaeza:
And the name doesn't help either...

product_base_mass_add/i18n/fr.po
```diff
@@ -0,0 +1,82 @@
+# Translation of Odoo Server.
+# This file contains the translation of the following modules:
+#	* quick_purchase
```
legalsylvain: I think this file is wrong. (reference to quick_purchase in all the doc). Should be regenerated.

PierrickBrun:
Readme should be better now.

Also you can look at these PRs to better understand what the module is for:
https://github.com/OCA/purchase-workflow/pull/653
https://github.com/OCA/stock-logistics-workflow/pull/510 

Indeed it is hard to name these kind of module, I think this is the less bad choice.

PierrickBrun:
@pedrobaeza 
About the naming maybe product_base_quick_add could be better.
I'm OK with making the change but could you propose something quickly ? product_base_quick_add or something else

bealdav:
@pedrobaeza about the module naming here, I propose to decide today because I've a go in production tomorrow.

Thanks a lot of for your advice here.

product_base_mass_add/README.rst
```diff
@@ -0,0 +1,95 @@
...
+.. |badge5| image:: https://img.shields.io/badge/runbot-Try%20me-875A7B.png
+    :target: https://runbot.odoo-community.org/runbot/135/12.0
+    :alt: Try me on Runbot
+
+|badge1| |badge2| |badge3| |badge4| |badge5| 
+
+This module is abstract and can't be used as is.
+
+It provides functions in order to show a product grid from another model.
+You can then add products and quantities en masse to the model you are working with.
```
pedrobaeza: s/en masse/in batch

pedrobaeza:
Thanks for extra information on the README. Now I have clear the aim of the module. For the name, I would go to `base_product_mass_edition`:

* `base` as first word due to current guidelines - second point in https://github.com/OCA/odoo-community.org/blob/master/website/Contribution/CONTRIBUTING.rst#11modules
* `mass_edition` as according the description, it serves for adding or editing.

I have put a comment in `purchase_quick` about an already existing module we developed for similar purposes (the quick edition of purchase lines), but pre-filtering the shown products (https://github.com/OCA/purchase-workflow/tree/11.0/purchase_order_product_recommendation). We have the same for sales: https://github.com/OCA/sale-workflow/tree/11.0/sale_order_product_recommendation. They don't have any base module, as the code is very straight-forward and due to the particularities of each model (sale.order.line vs purchase.order.line), although there are some pieces of duplicated code.

I know it maybe be a bit late for reconducting this, but I advise you to check if that can fit your needs (or with an extension/integration), or at least for taking ideas from there. We put a lot of emphasis on performance. About the "onchanges", we took the direct approach of calling the main ones instead of propagating all of them.

PierrickBrun:
@pedrobaeza 

Thanks for the quick answer.
I find `base_product_mass_edtion` to be confusing: it looks like we are editing the products while we are editing product relations (ie. purchase.order.line or stock.move)
Also editing is not the nominal case, this is needed to ensure consistency but the main goal of the module is to add products, editing the lines is not the nominal use case.
I'd prefer going with `base_product_mass_addition` or `base_product_mass_add`, which one do you prefer ?

Thank you for you input on purchase_quick, these modules are close you're right. 
I think 'x_x_product_recommandation' could inherit from this one with a different list provided than from `x_quick` module.
I'm going to mention 'purchase_order_product_recommendation' into 'purchase_quick' 's readme in order to let users decide which one fits best their use case.

pedrobaeza:
OK with `base_product_mass_addition`

base_product_mass_addition/models/product_product.py
```diff
@@ -0,0 +1,107 @@
...
+
+    def _get_quick_line(self, parent):
+        return None
+
+    def _add_quick_line(self, parent, line_model):
+        vals = self._prepare_quick_line(parent)
+        vals = self._complete_quick_line_vals(parent, vals)
+        if not vals.get('price_unit'):
+            vals['price_unit'] = 0.0
+        self.env[line_model].create(vals)
```
bealdav: If your line_model doesn't contains `price_unit` field, you may have a log warning.
May be, you could test with `getattr()`

base_product_mass_addition/models/product_product.py
```diff
@@ -0,0 +1,107 @@
...
+        help="Set this quantity to create a new line "
+             "for this product or update the existing one."
+    )
+
+    def _prepare_quick_line(self, parent):
+        res = self._get_quick_line_qty_vals()
+        res.update({'product_id': self.id})
+        return res
+
+    def _get_quick_line(self, parent):
```
bealdav: Not sure `quick` is really self explained word ?
- _get_original_line
- _get_parent_line

base_product_mass_addition/models/product_product.py
```diff
@@ -0,0 +1,107 @@
...
+    _inherit = 'product.product'
+
+    qty_to_process = fields.Float(
+        compute='_compute_process_qty',
+        inverse='_inverse_set_process_qty',
+        help="Set this quantity to create a new line "
+             "for this product or update the existing one."
+    )
+
+    def _prepare_quick_line(self, parent):
```
bealdav: The same for `quick` word here as below

base_product_mass_addition/models/product_product.py
```diff
@@ -0,0 +1,107 @@
...
+            vals = self._complete_quick_line_vals(parent, vals)
+            line.write(vals)
+        else:
+            line.unlink()
+
+    def _get_quick_line_qty_vals(self):
+        return None
+
+    def _complete_quick_line_vals(self, parent, vals,
+                                  parent_key='', lines_key=''):
```
bealdav: Not blocking for me but seems the common way to define default with no value is `None`

https://stackoverflow.com/questions/13075044/initialize-parameter-of-method-with-default-value

and never `[]` FYI
https://effbot.org/zone/default-values.htm

bealdav (changes requested):
Thanks for you nice contribution

Small comments

base_product_mass_addition/models/product_product.py
```diff
@@ -0,0 +1,107 @@
...
+        res.update({'product_id': self.id})
+        return res
+
+    def _get_quick_line(self, parent):
+        return None
+
+    def _add_quick_line(self, parent, line_model):
+        vals = self._prepare_quick_line(parent)
+        vals = self._complete_quick_line_vals(parent, vals)
+        if not vals.get('price_unit'):
```
mourad-ehm: Not modify vals just before creation. Add this part in  _complete_quick_line_vals method or _prepare_quick_line to easily change it by inheritance.

base_product_mass_addition/models/product_product.py
PierrickBrun: The goal is to make sure there are no clash with another module, _get_parent_line could be used in a lot of contexts

PierrickBrun:
I think this is stable enough now, as you can see, both the implementations's tests are passing

Can I get some reviews ?

base_product_mass_addition/models/product_product.py
PierrickBrun: This is already done, I think you are reviewing an older commit

mourad-ehm: OK

mourad-ehm (approved):

bealdav (approved):

bealdav:
Tests of this module are implemented in these PR

https://github.com/OCA/purchase-workflow/pull/653

https://github.com/OCA/stock-logistics-workflow/pull/510

base_product_mass_addition/views/product_view.xml
```diff
@@ -0,0 +1,37 @@
...
+        </field>
+    </record>
+    <record id="product_search_form_view" model="ir.ui.view">
+        <field name="model">product.product</field>
+        <field name="inherit_id" ref="product.product_search_form_view"/>
+        <field name="arch" type="xml">
+            <xpath expr="//filter[last()]" position="after">
+                <filter name="filter_for_current_supplier"
+                    string="For current supplier" domain="[]"
+                    context="{'use_only_supplied_product': True}"
```
sebastienbeau: Should be move in purchase ?

base_product_mass_addition/views/product_view.xml
```diff
@@ -0,0 +1,37 @@
...
+                <field name="qty_to_process"/>
+                <field name="incoming_qty"/>
+                <field name="qty_available"/>
+                <button name="button_quick_open_product" string="Open" type="object"/>
+            </tree>
+        </field>
+    </record>
+    <record id="product_search_form_view" model="ir.ui.view">
+        <field name="model">product.product</field>
+        <field name="inherit_id" ref="product.product_search_form_view"/>
```
sebastienbeau, line 21: change the type of view use primary !

sebastienbeau (changes requested):

sebastienbeau (approved):

PR #441, opened 2019-03-13
Header is added twice with this view

Martronic-SA:
Header already exists in the product view (product.template.common.form)
Remove the added header tag and add status bar inside the existing one


kos94ok-3D:
Hi, @Nikul-Chaudhary. Can you review it?

nikul-serpentcs (approved):
Good Catch, Code Review LGTM :+1:

PR #445, opened 2019-04-15
[11.0][IMP] product_secondary_unit: Add security group

sergio-teruel:
cc @Tecnativa

### DIFF
--- a/product_secondary_unit/migrations/11.0.1.0.2/post-migration.py
+++ b/product_secondary_unit/migrations/11.0.1.0.2/post-migration.py
@@ -1,0 +1,13 @@
+#  Copyright 2019 Tecnativa - Sergio Teruel
+#  License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl
+from openupgradelib import openupgrade
+
+
+@openupgrade.migrate(use_env=True)
+def migrate(env, version):
+    group_secondary_unit = env.ref(
+        'product_secondary_unit.group_secondary_unit')
+    users = env['res.users'].with_context(active_test=False).search([])
+    users.write({
+        'groups_id': [(4, group_secondary_unit.id)]
+    })
--- a/product_secondary_unit/views/product_views.xml
+++ b/product_secondary_unit/views/product_views.xml
@@ -7,7 +7,7 @@
         <field name="name">Product template Secondary Unit</field>
         <field name="model">product.template</field>
         <field name="inherit_id" ref="product.product_template_form_view"/>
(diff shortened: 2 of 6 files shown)


product_secondary_unit/migrations/11.0.1.0.2/post-migration.py
```diff
@@ -0,0 +1,13 @@
+#  Copyright 2019 Tecnativa - Sergio Teruel
+#  License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl
+from openupgradelib import openupgrade
+
+
+@openupgrade.migrate(use_env=True)
+def migrate(env, version):
+    group_secondary_unit = env.ref(
+        'product_secondary_unit.group_secondary_unit')
+    users = env['res.users'].search([])
```
pedrobaeza: Add `with_context(active_test=False)`

pedrobaeza (commented):
Maybe you should add the group in settings screen

product_secondary_unit/migrations/11.0.1.0.2/post-migration.py
sergio-teruel: Why?

pedrobaeza: For writing the group also on eventually deactivated users (that can be activated later, and being without the group).

sergio-teruel: Ok. Changes done

chienandalu (approved):
Looks good. Also tested on runbot :+1:

rafaelbn (approved):
Functionally tested in runbot :+1: 

Thanks @sergio-teruel

product_secondary_unit/views/product_views.xml
```diff
@@ -10,7 +10,7 @@
         <field name="groups_id" eval="[(4, ref('product.group_uom'))]"/>
         <field name="arch" type="xml">
             <xpath expr="//page[@name='general_information']" position="inside">
-                <group string="secondary unit of measure">
+                <group string="secondary unit of measure" groups="product_secondary_unit.group_secondary_unit">
```
pedrobaeza: This shouldn't be done here, but on `groups_id` in the view (where you have right now the `product.group_uom` one).

pedrobaeza (changes requested):

sergio-teruel:
Oks. changes done

pedrobaeza (approved):

PR #446, opened 2019-04-16
[12.0][MIGR] product_supplierinfo_for_customer

renatonlima:


pedrobaeza:
With this module, I wanted to change a bit the implementation, doing an inheritance of `product.supplierinfo`, but changing the name, and thus, we avoid the hack in `search` method plus the modification of supplier info views. What do you think about this change? I can do it if you prefer.

Check also the red status.

renatonlima:
@pedrobaeza,

I will push some fixes to red status for this PR. About your changes in this module, I agree, maybe we can create one object inherited from product.supplierinfo with name product.customerinfo and could we rename this module to product_customerinfo? What do you thing? I can do this :)

pedrobaeza:
I don't see the need of renaming the module, as this denotes the source of the data, but thanks for doing the refactoring.

product_supplierinfo_for_customer/models/product_product.py
```diff
@@ -0,0 +1,55 @@
...
+                self.env.context.get('partner', False)
+            prices = super(ProductProduct, self).price_compute(
+                'list_price', uom, currency, company)
+            for product in self:
+                price = product._get_price_from_supplierinfo(partner)
+                if not price:
+                    continue
+                prices[product.id] = price
+                if not uom and self._context.get('uom'):
+                    uom = self.env['product.uom'].browse(self._context['uom'])
```
youring, line 41: Replace product.uom with uom.uom, price list based on partner functional test ok

MiquelRForgeFlow: @renatonlima please

rafaelbn:
Dear @renatonlima could you please re-take this PR a little bit? You have comments from @youring and @mreficent and @pedrobaeza weill give last feedback, me too.

Also @murtuzasaleh could review this PR and help as is interested for continuing with his PR https://github.com/OCA/sale-workflow/pull/841 :smile: 

OCA force! Let's go!

pedrobaeza:
Yes, I'm not sure if @renatonlima understood what I wanted to do. Basically it's:

```python
class ProductCustomerinfo(models.Model):
    _inherit = "product.supplierinfo"
    _name = "product.customerinfo"

    _inherit = "product.template"

    customer_ids = fields.One2many(
        comodel_name="product.customerinfo"
        ...
```
and the rest of the needed stuff.

MiquelRForgeFlow:
Superseded in https://github.com/OCA/product-attribute/pull/515.

PR #449, opened 2019-04-26
[MIG][12.0] product_weight

tbaden:
Migration of product_weight to v12


tbaden:
added missing commit history

tbaden:
@pedrobaeza runbot is getting killed after 40 minutes without valid logs to check.
do you have any idea?

pedrobaeza:
There's a problem right now with the sizing of the servers. Let's see if @gurneyalex can fix it soon.

Meanwhile, force a rebuild to see if you can get it.

florian-dacosta (changes requested):
Hi,
Thanks for the PR.
There is a bug in case of multi level bom if the bom lines have a quantity different than 1
More explanation and the fix here : https://github.com/tbaden/product-attribute/pull/1

tbaden:
@florian-dacosta thanks for the information, I'll attend these changes at the end of this week.

pedrobaeza:
Any news on this?

tbaden:
@pedrobaeza is that the right way to merge the fix into the PR?
btw. thanks for the reminder, somehow missed this one

pedrobaeza:
Yes, but I see one line that can be removed commented on the original PR.

tbaden:
fixed it.

pedrobaeza (approved):
Amended commit history and changes seems OK, so counting with implicit @florian-dacosta approval, I'm merging.

PR #451, opened 2019-05-03
10.0 migration of pricelist per product

wpichler:


pricelist_per_product/models/product_pricelist_item.py
```diff
@@ -0,0 +1,47 @@
...
+    def create(self, vals):
+        if vals.get('product_id'):
+            vals['related_sequence'] = self._get_sequence_price_grid(vals)
+        return super(ProductPricelistItem, self).create(vals)
+
+    @api.multi
+    def write(self, vals):
+        for item in self:
+            if vals.get('product_id'):
+                vals['related_sequence'] = item._get_sequence_price_grid(vals)
```
umiphos, line 45: If you are writing more than one item, you are replacing the vals['related_sequence'] with the new item.
If you are expecting only 1 record to be written, then why not use ```self.ensureone()```

wpichler: write must be a @api.multi function - so ensure_one is not supported here.
If it gets called with more than one record - then it iterates over the records - for every record it does recalc the related_sequence - change the vals dict - does call the super write function with only the current item - and then processes the next item.
So it does create a @api.one out of the @api.multi call to support different related_sequence on write.

Issue #452, opened 2019-05-04
Travis Cleanup 10.0


tbaden:
- [x] product_weight @tbaden https://github.com/OCA/product-attribute/pull/455
- [x] product_default_image @tbaden https://github.com/OCA/product-attribute/pull/456
- [x] product_supplierinfo_for_customer @tbaden https://github.com/OCA/product-attribute/pull/457
- [x] product_brand @tbaden https://github.com/OCA/product-attribute/pull/458
- [x] product_sequence @tbaden https://github.com/OCA/product-attribute/pull/459
- [x] product_state @tbaden https://github.com/OCA/product-attribute/pull/460
- [x] product_assortment @tbaden https://github.com/OCA/product-attribute/pull/461

tbaden:
************* Module product_weight
product_weight/wizard/product_weight_update_view.xml:8: [W7910(wrong-tabs-instead-of-spaces), ]  Use wrong tabs indentation instead of four spaces
product_weight/wizard/product_weight_update_view.xml:9: [W7910(wrong-tabs-instead-of-spaces), ]  Use wrong tabs indentation instead of four spaces
product_weight/wizard/product_weight_update_view.xml:10: [W7910(wrong-tabs-instead-of-spaces), ]  Use wrong tabs indentation instead of four spaces
product_weight/wizard/product_weight_update_view.xml:11: [W7910(wrong-tabs-instead-of-spaces), ]  Use wrong tabs indentation instead of four spaces
product_weight/wizard/product_weight_update_view.xml:12: [W7910(wrong-tabs-instead-of-spaces), ]  Use wrong tabs indentation instead of four spaces
product_weight/wizard/product_weight_update_view.xml:13: [W7910(wrong-tabs-instead-of-spaces), ]  Use wrong tabs indentation instead of four spaces
product_weight/wizard/product_weight_update_view.xml:14: [W7910(wrong-tabs-instead-of-spaces), ]  Use wrong tabs indentation instead of four spaces
product_weight/wizard/product_weight_update_view.xml:15: [W7910(wrong-tabs-instead-of-spaces), ]  Use wrong tabs indentation instead of four spaces
product_weight/wizard/product_weight_update_view.xml:16: [W7910(wrong-tabs-instead-of-spaces), ]  Use wrong tabs indentation instead of four spaces
product_weight/wizard/product_weight_update_view.xml:17: [W7910(wrong-tabs-instead-of-spaces), ]  Use wrong tabs indentation instead of four spaces
[...]

PR #458, opened 2019-05-04
[FIX][10.0] product_brand travis warnings

tbaden:
issue https://github.com/OCA/product-attribute/issues/452 :
fix of:
************* Module product_brand.models.product_brand
product_brand/models/product_brand.py:31: [C8108(method-compute), ProductBrand] Name of compute method should start with "compute"
product_brand/models/product_brand.py:36: [W8104(api-one-deprecated), ProductBrand._get_products_count] api.one deprecated

### DIFF
--- a/product_brand/models/product_brand.py
+++ b/product_brand/models/product_brand.py
@@ -28,13 +28,14 @@ class ProductBrand(models.Model):
     )
     products_count = fields.Integer(
         string='Number of products',
-        compute='_get_products_count',
+        compute='_compute_products_count',
     )
 
-    @api.one
+    @api.multi
     @api.depends('product_ids')
-    def _get_products_count(self):
-        self.products_count = len(self.product_ids)
+    def _compute_products_count(self):
+        for this in self:
+            this.products_count = len(this.product_ids)
 
 
 class ProductTemplate(models.Model):


product_brand/models/product_brand.py
```diff
@@ -28,13 +28,14 @@ class ProductBrand(models.Model):
...
     )
 
-    @api.one
+    @api.multi
     @api.depends('product_ids')
-    def _get_products_count(self):
-        self.products_count = len(self.product_ids)
+    def _compute_products_count(self):
+        for this in self:
+            self.products_count = len(self.product_ids)
```
pedrobaeza: s/self/this

tbaden: thanks, I somehow missed that...
fixed and pushed

product_brand/models/product_brand.py
```diff
@@ -28,13 +28,14 @@ class ProductBrand(models.Model):
...
+        compute='_compute_products_count',
     )
 
-    @api.one
+    @api.multi
     @api.depends('product_ids')
-    def _get_products_count(self):
-        self.products_count = len(self.product_ids)
+    def _compute_products_count(self):
+        for this in self:
```
pedrobaeza, line 37: FYI, we don't use too much `this` as variable name on Odoo, but `record`, `rec`, `one` or something significant according the model (`partner`, `product`...). But no need to change it here.

pedrobaeza (approved):

product_brand/models/product_brand.py
tbaden: Thanks, I'll keep this in mind for the next one

PR #474, opened 2019-05-14
[10.0] [NEW] Add module product_code_remove.

NL66278:
This module offers customers the possibility to not add product codes to the name used for order lines, invoice lines, etc. Searching on code is not affected.


pedrobaeza:
This already exits through https://github.com/OCA/sale-workflow/tree/10.0/sale_order_line_description and https://github.com/OCA/purchase-workflow/tree/10.0/purchase_order_line_description

NL66278:
@pedrobaeza It is not quite the same, although there is a limited overlap in purpose. The sale-workflow and purchase-workflow modules only work on the description_sale and description_purchase fields, that many customers (at least of our customers) do not use. Apart from that, this module acts on the product, causing it to be effective wherever products are used, like for instance also in invoices.

I am not saying the other modules are in any way not good, on the contrary, they seem to be nicely programmed and effective, only that the focus and purpose are different.

pedrobaeza:
That modules also remove the code if you see. The question with this module is that you are removing from the name_get the code, but not from the name_search, so it's a bit misleading that you are able to search a product by code, but don't see the code. I know that this code seems more simplistic, but it cause several inconsistencies.

NL66278:
@pedrobaeza I do not see the inconsistencies. There are many places in the system where we can search a model on some field, without seeing that field in the result. One example: searching a partner on chamber of commerce in the partner_coc module.

The module proposed is exactly what some of our customers want, and we want to make it available for others who want this functionality. It is not a replacement for the modules you mentioned, those remain totally valid for the customers that want that functionality, nor would they even conflict when all three of the modules were to be installed.

pedrobaeza:
OK, if you see it correct, go ahead.

ovnicraft:
LTGM and test passed ok.

ovnicraft (approved):

nikul-serpentcs:
@NL66278 Add README Section Or Can you please Migrate other version

lfreeke (approved):
Functional test :+1: 

@NL66278 can you add the readme as requested?

Issue #476, opened 2019-05-16
product_state [11.0] 'application': True, is better to change to False


andreampiovesana:
It is NOT an application right?

tbaden:
@andreampiovesana  No it's not. It seems this tag is there since the very beginning of this module. Would you like to create a PR, otherwise I'll create one.

edit:
created PR: https://github.com/OCA/product-attribute/pull/477

PR #477, opened 2019-05-16
[FIX] [11.0] product_state app false

tbaden:
Fix for issue: https://github.com/OCA/product-attribute/issues/476
As the default for application is false, I removed it from the manifest.


pedrobaeza (approved):
Fast-tracking this

PR #479, opened 2019-05-20
[12.0] [MIG] product_end_of_life

tbaden:


product_end_of_life/__manifest__.py
```diff
@@ -0,0 +1,22 @@
+# Copyright (C) 2018 - TODAY, Open Source Integrators
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+{
+    "name": "Product End-Of-Life Date",
+    "summary": "Alert users when products approach EoL Date",
+    "version": "11.0.1.0.0",
```
max3903:
```suggestion
    "version": "12.0.1.0.0",
```

tbaden:
will push the history again. seems like I messed up my git user for the patch.

tbaden:
found a missing newline... will fix it. otherwise it can be reviewed @max3903

AdriaGForgeFlow (approved):
Code review and functional test in runbot (haven't tested mailing feature). LGTM :+1:

product_end_of_life/views/product_view.xml
```diff
@@ -0,0 +1,50 @@
...
+        </field>
+    </record>
+    <!-- Product Variant form view -->
+    <record id="product_variant_easy_edit_eol_view" model="ir.ui.view">
+        <field name="name">product.variant.easy.edit.eol.view</field>
+        <field name="model">product.product</field>
+        <field name="inherit_id" ref="product.product_variant_easy_edit_view"/>
+        <field name="arch" type="xml">
+            <xpath expr="//group[@name='weight']" position="after">
+                <group></group>
```
nikul-serpentcs, line 21: @tbaden The empty group is needed?

tbaden: @nikul-serpentcs as I did not change this part I'm not 100% sure, but most likely for a correct placement inside the view.

nikul-serpentcs: @tbaden Ok, Can you please check if need then keep as it is, otherwise remove it.

Thanks

simahawk:
@tbaden anything else to do? Ping reviewers if you think you are done.

claudiagn (approved):
LGTM!

tbaden:
@simahawk any idea why the merge failed?

product_end_of_life/models/product.py
```diff
@@ -79,10 +80,23 @@ def send_notify_product_eol(self):
...
+        args = args or []
+        domain = []
+        if name:
+            domain = [
+                ('name', operator, name)
+            ]
+            if operator in expression.NEGATIVE_TERM_OPERATORS:
+                domain = ['&', '!'] + domain[1:]
+        products = self.search(domain + args, limit=limit)
+        return products.name_get()
```
huguesdk, line 94: when i stumbled on this, i did not understand where this came from, and why it had to be added when migrating to 12.0. i then saw that in 11.0, `product.product` redefined `.name_get()`, and this called `.search()`, but this has disappeared in 12.0.

this implementation is quite different from the default `.name_get()`. isn’t there another way? if not, at least a comment explaining why it’s done like this and its tradeoffs would be nice.

huguesdk (changes requested):
`test_search_read()` fails if the module `product_model_viewer` is installed. i don’t know which module causes the error, though.

PR #483, opened 2019-05-25
[10.0][IMP] unified unique code implementation

ovnicraft:
`product_code_unique` module already did unique, depend from it on `product_sequence`

### DIFF
--- a/product_sequence/__manifest__.py
+++ b/product_sequence/__manifest__.py
@@ -7,7 +7,7 @@
     "name": "Product Sequence",
     "version": "10.0.2.0.0",
     "author": "Zikzakmedia SL,Sodexis,Odoo Community Association (OCA)",
-    "website": "http://www.zikzakmedia.com",
+    "website": "https://github.com/OCA/product-attribute",
     "license": "AGPL-3",
     "category": "Generic Modules/Inventory Control",
     "depends": ["product_code_unique"],
--- a/product_sequence/__manifest__.py
+++ b/product_sequence/__manifest__.py
@@ -4,19 +4,15 @@
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
 
 {
-    'name': 'Product Sequence',
-    'version': '10.0.1.0.0',
-    'author': "Zikzakmedia SL,Sodexis,Odoo Community Association (OCA)",
-    'website': 'http://www.zikzakmedia.com',
-    'license': 'AGPL-3',
-    'category': 'Generic Modules/Inventory Control',
-    'depends': [
-        'product',
-    ],
-    'data': [
-        'data/product_sequence.xml',
-    ],
-    'pre_init_hook': 'pre_init_hook',
-    'auto_install': False,
-    'installable': True,
+    "name": "Product Sequence",
+    "version": "10.0.2.0.0",
+    "author": "Zikzakmedia SL,Sodexis,Odoo Community Association (OCA)",
+    "website": "http://www.zikzakmedia.com",
+    "license": "AGPL-3",
+    "category": "Generic Modules/Inventory Control",
+    "depends": ["product_code_unique"],
+    "data": ["data/product_sequence.xml"],
+    "pre_init_hook": "pre_init_hook",
+    "auto_install": False,
+    "installable": True,
 }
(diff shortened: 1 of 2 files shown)


product_sequence/models/product_product.py
```diff
@@ -7,32 +7,23 @@
 
 
 class ProductProduct(models.Model):
-    _inherit = 'product.product'
+    _inherit = "product.product"
```
moylop260, line 10: Why this change is needed?

product_sequence/__manifest__.py
```diff
@@ -4,19 +4,15 @@
...
-    'data': [
-        'data/product_sequence.xml',
-    ],
-    'pre_init_hook': 'pre_init_hook',
-    'auto_install': False,
-    'installable': True,
+    "name": "Product Sequence",
+    "version": "10.0.2.0.0",
+    "author": "Zikzakmedia SL,Sodexis,Odoo Community Association (OCA)",
+    "website": "http://www.zikzakmedia.com",
```
nikul-serpentcs: URL should be `https://github.com/OCA/product-attribute`

nikul-serpentcs (commented):
nitpicking

product_sequence/models/product_product.py
ovnicraft: @moylop260 i use black :-)

ovnicraft:
@moylop260 @nikul-serpentcs please review it

product_sequence/models/product_product.py
moylop260: IMHO you should disable black for this case.
Since that we don't have discussed about using double quotes
However, you can start a new thread to propose define it:
 - https://github.com/OCA/odoo-community.org/blob/master/website/Contribution/CONTRIBUTING.rst

For now, if we don't have define it use double-quotes or single-quotes then other people could change from double to single and other people could change single to double.

The problem is that these kind of changes without previous definition makes:
 - big diffs hard to review PR
 - Issues to merge back-ports or foward-ports
 - Conflicts if there are 2 or more Pull Requests for same module

ovnicraft: Could be but now there is no another PR, is not a big diff for this case.

moylop260: It is a big diff for this case
Since that the title talks about one change of depends but the diff has 11 additions and 15 deletions for manifest and 9 additions and 20 deletions for product_product.py where I expected just 1 or 2 lines changed.

It is valid if you have created first the rule to global guidelines because other people has different custom styles but not OCA ones.

There are developers using the following way
```python
    field_1          = fields.Datetime()
    field_name_large = fields.Integer()
    field_3          = fields.Char()
```

Then, start a war of big diff changing custom style 1 to custom style 2 (not OCA ones)

# Using Custom Style for each developer:

Commit1 Custom Style A - "Adding `required=True` to field1"
```diff
-    field_1          = fields.Datetime()
+    field_1 = fields.Datetime(required=True)
    field_name_large = fields.Integer()
-    field_3          = fields.Char()
+    field_3 = fields.Char()
```

Commit2 Custom Style B - "Adding readonly to field_name_large"
```diff
-    field_1 = fields.Datetime()
+    field_1          = fields.Datetime()
-    field_name_large = fields.Integer()
+    field_name_large = fields.Integer(readonly=True)
-    field_3 = fields.Char()
+    field_3          = fields.Char()
```

# Using all developers the same styles ruled by OCA guidelines:

Commit1 Style OCA - "Adding `required=True` to field1"
```diff
-    field_1 = fields.Datetime()
+    field_1 = fields.Datetime(required=True)
    field_name_large = fields.Integer()
    field_3 = fields.Char()
```

Commit2 Custom Style B - "Adding readonly to field_name_large"
```diff
    field_1 = fields.Datetime(required=True)
-    field_name_large = fields.Integer()
+    field_name_large = fields.Integer(readonly=True)
    field_3 = fields.Char()
```

Small diff without conflicts.

If you want create a new rule then you need first create a PR to guidelines document before.

nikul-serpentcs (approved):
@ovnicraft Squash Commit's

Only code Review LGTM :+1:

pedrobaeza:
I agree that we shouldn't change style in this kind of PRs. You should deactivate automatic formatters when doing changes. If you introduce a new module, it's ok to be done with automatic formatting as it's the first diff.

We can discuss about applying black or similar formatters automatically via a bot, balancing the noise in the diff with the benefits of this homogeinity. I personally don't value the homogeneity in the string delimiter. It doesn't add any value for me.

ovnicraft:
I have no time to this changes.
Closing it.

PR #485, opened 2019-05-25
12.0 mig product dimension

rvalyi:
standard migration followed by these changes:

1. product.uom -> uom.uom
2. changed some xml_id refs from product... to uom...
3. added required readonly=False on the product.template length, height and width related fields to enable the edition, just like it's now done in addons/product for variantes fields that are editable from the template when there is a unique variant (like when the variant setting is set to False).


product_dimension/__manifest__.py
```diff
@@ -0,0 +1,18 @@
...
+# Copyright 2015-2016 Camptocamp SA
+# Copyright 2015 ADHOC SA  (http://www.adhoc.com.ar)
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+{
+    'name': 'Product Dimension',
+    'version': '12.0.1.0.0',
+    'category': 'Product',
+    'author': 'brain-tec AG, ADHOC SA, Camptocamp SA, '
+              'Odoo Community Association (OCA)',
+    'license': 'AGPL-3',
```
tbaden, line 11: please add the website tag with value: https://github.com/OCA/product-attribute

product_dimension/views/product_view.xml
```diff
@@ -0,0 +1,36 @@
...
+            <field name="dimensional_uom_id"/>
+            <field name="length" attrs="{'readonly':[('type','=','service')]}"/>
+            <field name="height" attrs="{'readonly':[('type','=','service')]}"/>
+            <field name="width" attrs="{'readonly':[('type','=','service')]}"/>
+          </group>
+        </xpath>
+      </field>
+    </record>
+
+    <record model="ir.ui.view" id="product_template_only_form_view">
```
tbaden, line 20: optional change:
id before model

use original id (product_template_form_view)

product_dimension/views/product_view.xml
```diff
@@ -0,0 +1,36 @@
+<?xml version="1.0"?>
+<odoo>
+
+    <record model="ir.ui.view" id="product_normal_form_view">
```
tbaden, line 4: optional change:
id before model

use original id (product_variant_easy_edit_view)

tbaden (changes requested):
thanks for the work 👍 
I did a code review: some minor changes
please use the fragment [readme](https://github.com/OCA/maintainer-tools/wiki/Migration-to-version-12.0)

bealdav (approved):
Tested and code review

flotho:
Hi, I didn't checked furthermore this PR, but is there any possible conflicts with #491 ?

Regards

product_dimension/views/product_view.xml
```diff
@@ -0,0 +1,36 @@
...
+        </xpath>
+      </field>
+    </record>
+
+    <record model="ir.ui.view" id="product_template_only_form_view">
+      <field name="name">product_template_form_view</field>
+      <field name="model">product.template</field>
+      <field name="inherit_id" ref="product.product_template_form_view"/>
+      <field name="arch" type="xml">
+        <xpath expr="//page[@name='general_information']" position="inside">
```
ivantodorovich: I think it makes more sense to place this group after the Logistic group.

image

In that case the fields would be already under the Inventory page that's hidden if the product type is service, so you can even remove the 

`attrs="{'readonly':[('type','=','service')]`

product_dimension/views/product_view.xml
```diff
@@ -0,0 +1,36 @@
...
+          <group string="Dimensions" name="dimensions" colspan="2" attrs="{'invisible': [('product_variant_count', '&gt;', 1)]}">
+            <field name="dimensional_uom_id" attrs="{'invisible': [('product_variant_count', '&gt;', 1)]}"/>
+            <field name="length" attrs="{'readonly':[('type','=','service')],'invisible': [('product_variant_count', '&gt;', 1)]}"/>
```
ivantodorovich: Also, the `attrs="{'invisible': [('product_variant_count', '&gt;', 1)]}"` is making this field invisible if you use the full variant view (that's generated based on the product.template view), if the variant count is > 1.

But it's not that easy to fix. In fact this module is not working good with multiple variants.


In product.template, the volume fields are related fields to the variant attribute with readonly=False.

That means the values will be related to the ones of the first variant. And when you edit, you'll be editing only the values for the first variant.

So in this case I think there are 2 possible solutions:

**1. Split this view in two: one for product.template and one for product.product.**
So, instead of extending `product.product_template_form_view` (that's shared among product.template and product.product), you'll have to extend:
- `product.product_template_only_form_view` for product.template
- `product.product_normal_form_view` for product.product

In the product.template you will show the fields only if variant_count = 1
In product.product you will show the field always

**2. Use the odoo-way approach instead, the one it's being used for volume and weight fields**
In this case volume fields won't be related fields. Instead they'll be computed fields with inverse function. (check: https://github.com/odoo/odoo/blob/12.0/addons/product/models/product_template.py#L247)

Basically, the fields will always be shown on the views, so you can keep the extending `product.product_template_form_view` approach. But instead the values will be 0 if the product.template has more than 1 variant.

IMO option 2 is less intuitive for the user... but it's the odoo-way, meaning that in the end it's consistent user experience

ivantodorovich (changes requested):

product_dimension/views/product_view.xml
clementmbr: Thank you very much @ivantodorovich for your comments. I've just improved Raphael's PR with some of your ideas. It would great if you can review it.

For the `product.product` side I inherit the `product_normal_form_view` and for the `product_template` side I inherit the `product_template_only_form_view` which allows to add the `attrs="{'invisible': [('product_variant_count', '&gt;', 1)]}"` condition without impacting the product variants views.

I finally didn't use the idea you've mentioned about defining the dimension fields as _computed fields_ instead of _related fields_ and I let them as _related fields_.
In fact, as the dimension fields are not appearing anymore in theproduct_template views, they don't need to be computed from the product variants fields, do you agree ?

I also added a `readonly=False` on the `dimensional_uom_id` field on the `product_template` objects to be able to modify it from the product template without accessing the product variant in case of one variant only, and I put the dimension fields in the _Inventory_ group as you suggested.

cc @rvalyi @flotho

alvaro-domatix (approved):

ivantodorovich (approved):
Thanks @clementmbr !!
Somehow I missed this notification, just saw this now.. sorry about that.

:+1: LGTM

rvalyi:
could somebody merge this if it's all ok now?

rvalyi:
up cc @florian-dacosta @jjscarafia @lepistone @guewen @pedrobaeza @grindtildeath @jbaudoux

florian-dacosta (approved):

PR #493, opened 2019-07-15
[12.0][IMP] product_brand: save logo as attachment

GSLabIt:


pedrobaeza:
This change can't be done without migration script. See an example here: https://github.com/OCA/bank-payment/commit/640f31e0277199f8b5cec7c0026db456aee55630

GSLabIt:
@pedrobaeza ooopss ..sorry will add the script.

GSLabIt:
@pedrobaeza not sure if done correctly ..but at least done 😬

pedrobaeza:
You must change module version number (to `12.0.2.0.0`), and name the folder that way

GSLabIt:
thought version number change would be done while merging ..probably i misunderstood ..will do

pedrobaeza:
Well, I can do that bump on merge, but it won't work any way as you have put the migrations folder with `12.0.1.0.0`, so better you do all together.

GSLabIt:
@pedrobaeza do it work now?

pedrobaeza:
Well, I have asked you to put `12.0.2.0.0`, as it's a huge change that requires a migration script (see https://github.com/OCA/odoo-community.org/blob/master/website/Contribution/CONTRIBUTING.rst#version-numbers)

GSLabIt:
@pedrobaeza sorry didn't get it ..i'm getting quite confused with all this rules .. 😬

pedrobaeza:
It's just to follow my comment :wink:

GSLabIt:
@pedrobaeza 3rd attempt ... 😬

pedrobaeza (approved):
And here it goes!

pedrobaeza:
Well, commit message puts 11.0 instead of 12.0, but I'll squash on merge removing the version.

GSLabIt:
@pedrobaeza oh sorry i was initially doing for 11 because i back ported import image from url to 11 😬

PR #495, opened 2019-08-07
[12.0][MIG] product_standard_price_tax_included

legalsylvain:
Port from V8, product_standard_price_tax_included that allow to have pricelist based on Cost price, in a B2C configuration. (with price VAT Included).

**sample**
You have a product :
- Cost : 10€ (Vat Excl)
- Sale Price (Vat incl) : 20€
- VAT : 20% (Price with Tax included)

You have a marginless pricelist based on the cost

**without this module**

sale_order_form_without_module

**With this module**
sale_order_form

thanks for your review.

CC : @quentinDupont


legalsylvain:
CC : original reviewers : @moylop260, @pedrobaeza

rafaelbn:
Hi @legalsylvain , maybe this is possible in v12 only with configuration. As you can work with or without Taxed, both at the same time.

legalsylvain:
Hi @rafaelbn !
Thanks a lot for your review. I don't get what you propose exactly. What are the new options (w/o taxed) ?

The problem of Odoo (AFAIK) is that if you want to have Vat included prices. (because you're working in B2C) **AND** if you want to have some pricelists based on cost (for exemple price = Cost +20%), the unit price is not correct.

but maybe I missed a new option...

rafaelbn:
Hi @legalsylvain,

OK. So if you work B2C you configure SALE taxes to be included in the price but why to configure also PURCHASE taxes to be included?

I mean that you can work with taxes excluded in the price but activate the options to show included. Because it has been always a headache when working both ways. Look:



Rounding is awful and in Odoo never is the same "price + tax = total_price" and "total_price - tax = price". Always mistakes in rounding...

legalsylvain:
Hi @rafaelbn. I'm not sure to understand your proposal. but if the question is : why don't I use tax excluded for sale price, it is not a choice. My user are B2C users. So they are talking price VAT included.  For them, the price of the beer if 2,50 VAT included, not 2,08 VAT excluded that has no sense. If check vat excl, as proposed, I force the saler to set price vat excl, that is not an option.
kind regards.

tonygalmiche (approved):
Hi @legalsylvain,
I tested this module on my personal computer.
I checked the code.
Everything is ok for me.
Thank you for this migration.

legalsylvain:
Weird, I'm declared as the maintainer of this module ...

pedrobaeza:
That's because it was prevented to self-merge a PR that adds the maintainer, and this migration is somehow that. This was done for avoiding someone adding himself/herself as maintainer in one PR and being able to self-merge that PR. Thus, the requirement is that the maintainer is declared on the main branch, not here. The side effect is that for migrations, other in the PSC must merge the PR.

/ocabot merge nobump

PR #498, opened 2019-08-22
[12.0] [ADD] product_pack

ernestotejeda:
Cc @Tecnativa


product_pack/i18n/en.po
```diff
@@ -0,0 +1,207 @@
+# Translation of Odoo Server.
```
pedrobaeza: Remove this offending file

product_pack/models/product_pack_line.py
```diff
@@ -0,0 +1,57 @@
...
+        ('product_uniq', 'unique(parent_product_id, product_id)',
+         'Product must be only once on a pack!'),
+    ]
+
+    @api.constrains('product_id')
+    def check_relations(self):
+        """
+        Check assisted packs don't have packs as child
+        """
+        for rec in self:
```
pedrobaeza:
```suggestion
        for rec in self.filtered('product_id.pack'):
```
and you remove one nested level

product_pack/models/product_product.py
```diff
@@ -0,0 +1,117 @@
...
+    )
+    used_pack_line_ids = fields.One2many(
+        'product.pack.line',
+        'product_id',
+        'On Packs',
+        help='List of packs where product is used.'
+    )
+
+    @api.multi
+    def separete_pack_products(self):
```
pedrobaeza: typo: s/separete/separate

product_pack/models/product_product.py
```diff
@@ -0,0 +1,117 @@
...
+        recordsets.
+        :return: [packs, no_packs]
+        """
+        packs = self.filtered(lambda p: p.pack and p.pack_type in [
+            'totaliced_price',
+            'none_detailed_assisted_price',
+            'none_detailed_totaliced_price',
+        ])
+
+        # for compatibility with website_sale
```
pedrobaeza, line 41: Put a TODO here to check why this is needed

product_pack/models/product_product.py
```diff
@@ -0,0 +1,117 @@
...
+                'outgoing_qty': 0,
+                'virtual_available': (
+                    pack_virtual_available and
+                    max(min(pack_virtual_available), 0) or False),
+            }
+        return res
+
+    @api.constrains('pack_line_ids')
+    def check_recursion(self):
+        """
```
pedrobaeza: Put this only in one line

product_pack/models/product_product.py
```diff
@@ -0,0 +1,117 @@
...
+                'incoming_qty': 0,
+                'outgoing_qty': 0,
+                'virtual_available': (
+                    pack_virtual_available and
+                    max(min(pack_virtual_available), 0) or False),
+            }
+        return res
+
+    @api.constrains('pack_line_ids')
+    def check_recursion(self):
```
pedrobaeza: Move this check recursion at line level when you change product_id, to cover both cases: create and write.

product_pack/models/product_template.py
```diff
@@ -0,0 +1,99 @@
...
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+
+from odoo import fields, models, api, _
+from odoo.exceptions import ValidationError
+
+
+class ProductTemplate(models.Model):
+    _inherit = 'product.template'
+
+    pack_type = fields.Selection([
```
pedrobaeza, line 11: Why is this configured at template level and the pack lines at variant level? This seems totally inconsistent

product_pack/models/product_template.py
```diff
@@ -0,0 +1,99 @@
...
+                            'Pack"'))
+
+    @api.constrains('company_id', 'product_variant_ids')
+    def check_pack_line_company(self):
+        """
+        Check packs are related to packs of same company
+        """
+        for rec in self:
+            for line in rec.pack_line_ids:
+                if line.product_id.company_id != rec.company_id:
```
pedrobaeza: Consider here if the company is empty.

product_pack/models/product_template.py
```diff
@@ -0,0 +1,99 @@
...
+                    raise ValidationError(_(
+                        'Pack lines products company must be the same as the '
+                        'parent product company'))
+            for line in rec.used_pack_line_ids:
+                if line.parent_product_id.company_id != rec.company_id:
```
pedrobaeza: Consider here if the company is empty.

product_pack/readme/CONTRIBUTORS.rst
```diff
@@ -0,0 +1,4 @@
+* `Tecnativa <https://www.tecnativa.com>`_:
```
pedrobaeza, line 7: You should add here Ing. Adhoc contributors

product_pack/readme/USAGE.rst
```diff
@@ -0,0 +1,5 @@
+To use this module, you need to:
+
+#. Go to product and set "Product Pack", then set the products has to include
```
pedrobaeza: Put a line to where exactly in the menu path you should go for "Go to product".

product_pack/security/ir.model.access.csv
```diff
@@ -0,0 +1,3 @@
+id,name,model_id:id,group_id:id,perm_read,perm_write,perm_create,perm_unlink
+access_product_pack_line_employee,product.pack.line (se module README on why sale user is used here),model_product_pack_line,base.group_user,1,1,1,1
```
pedrobaeza: This mustn't use sale_user. Check permissions for products, and put them accordingly. And add new one on `sale_product_pack` if needed.

product_pack/security/ir.model.access.csv
```diff
@@ -0,0 +1,3 @@
...
+access_product_pack_line_all,product.pack.line,model_product_pack_line,,1,0,0,0
```
pedrobaeza: This NEVER should be global. At least, it must be for base user. Portal and public users must be studied to check if we should put them or not (annotate it on the README).

pedrobaeza (changes requested):
Migration scripts will be needed for moving data from old field names to new ones.

product_pack/models/product_template.py
ernestotejeda: Why could this be inconsistent?
I suppose product_type is in the template because in that field both the template and the attributes must have the same value, which seems fine to me.
On the other hand, it seems that there should be a list of pack lines for each product variant, that may be correct in that way, what do you think?

product_pack/demo/product_product_demo.xml
```diff
@@ -0,0 +1,93 @@
...
+<odoo noupdate="0">
+    <record id="product_pack_cpu_detailed_components" model="product.product">
+        <field name="name">Pack CPU (Detailed Components Price)</field>
+        <field name="categ_id" ref="product.product_category_4"/>
+        <field name="pack" eval="True"/>
+        <field name="pack_type">components_price</field>
+        <field name="standard_price">20.5</field>
+        <field name="list_price">30.75</field>
+        <field name="type">service</field>
+        <field name="allow_modify_pack" eval="False" />
```
zaoral: Does we need this in allow_modify_pack = False y demo data?

product_pack/views/product_pack_line_views.xml
```diff
@@ -0,0 +1,33 @@
...
+    <record model="ir.ui.view" id="product_pack_line_form">
+        <field name="name">product.pack.line.form</field>
+        <field name="model">product.pack.line</field>
+        <field name="type">form</field>
+        <field name="arch" type="xml">
+            <form string="Pack Product Line">
+                <group>
+                    <field name="product_id"/>
+                    <field name="quantity"/>
+                    <field name="discount" groups="sale.group_discount_per_so_line" invisible="context.get('pack_type') == 'fixed_price'"/>
```
zaoral: refrence sale group but now dependecy to sale module

product_pack/views/product_pack_line_views.xml
```diff
@@ -0,0 +1,33 @@
...
+
+    <record model="ir.ui.view" id="product_pack_line_tree">
+        <field name="name">product.pack.line.tree</field>
+        <field name="model">product.pack.line</field>
+        <field name="type">tree</field>
+        <field name="arch" type="xml">
+            <tree string="Pack Product Lines" editable="bottom">
+                <field name="product_id"/>
+                <field name="quantity"/>
+                <field name="discount" groups="sale.group_discount_per_so_line" invisible="context.get('pack_type') == 'fixed_price'"/>
```
zaoral: refrence sale group but now dependecy to sale module

product_pack/i18n/es.po
```diff
@@ -0,0 +1,422 @@
...
+"MIME-Version: 1.0\n"
+"Content-Type: text/plain; charset=UTF-8\n"
+"Content-Transfer-Encoding: \n"
+"Language: es\n"
+"Plural-Forms: nplurals=2; plural=(n != 1);\n"
+
+#. module: product_pack
+#: model:ir.model.fields,help:product_pack.field_product_product_pack_type
+#: model:ir.model.fields,help:product_pack.field_product_template_pack_type
+#: model:ir.model.fields,help:product_pack.field_sale_order_line_pack_type
```
zaoral: update the translation file, the sale_order* references to files / views / models does not exist anymore in this module

product_pack/migrations/12.0.1.0.0/pre-migration.py
```diff
@@ -0,0 +1,14 @@
...
+from openupgradelib import openupgrade
+
+
+@openupgrade.migrate()
+def migrate(env, version):
+    openupgrade.logged_query(
+        env.cr, """
+        UPDATE product_template
+        SET pack_type = pack_price_type
+        """,
```
zaoral: Since the selection values has been fixed (misspelling error) we need to also add this here in order to do not lost the saved values.

* old "totalice_price" change to "totaliced_price"
* old "none_detailed_assited_price" to "none_detailed_assisted_price"

zaoral (changes requested):
Hi there thanks for the work in this module.
Look good to me +1
Only leave some changes

About the missing CLA,

* Juan Jose already added to the github user.
* Nmr Nicolas is currently in vacation, maybe this can be do later.
* About adhoc-transbot  this a bot for translations, do you need we add this one to the ECLA?

Best Regards

product_pack/migrations/12.0.1.0.0/pre-migration.py
zaoral: Almost I miss, also the allow_modify_pack field has change from selection to boolean, Not sure if we need to add something here of automatically change to False if not set and True if one of the two old selections were selected.

product_pack/views/product_pack_line_views.xml
ernestotejeda: @zaoral thanks for the review.
I have a question regarding that field. Why that field? Those discounts shouldn't be done using pricelists?

zaoral: Hi @ernestotejeda 

We use the discount field is a optional field where you can define an extra discount for a product just when is been purchase in a pack, This is to make it more attractive to the customer that want to purchase a product  to also purchase another products that will have a discount if they are purchased together.

Example: you sell 3 products with this prices, if they are purchased separately

* Product A: 100
* Product B: 120
* Product C: 130

Total 350

But when selling Pack Product you are offering: Product A + Product B + Product C, there you can make an extra discount to each product (if you want) just to encourage the customer to purchase the 3 products instead of 1:

* Product A: 100 - 10% = 90
* Product B: 120 = 120
* Product C: 130 - 5% = 123.5

Pack Total 333,5

pedrobaeza: But I see a bit confusing to use both specific price unit and discount. If you want an specific price, put in on pack line price unit, and if you want to apply a discount over the corresponding sales price  (after applying customer pricelist), then that's the way to go, but not both...

IMO, we should remove specific price, put the discount on the sales part, and follow this last strategy. What do you think?

zaoral: >> we should remove specific price,

I am not sure to remove the price field because there is the case that we like to have a specific price for a product when is sell as part of a product pack. @jjscarafia what do you think about this?

>> But I see a bit confusing to use both specific price and discount

IMHO is good to have both, the price and the discount, because this way we show the customer the original price and the discount we are making.

pedrobaeza: The problem with specific prices is that they are not maintainable over time (you can't modify them easily on batch). And it's a very rough edge, as the usual case is to define a price in the pack product, not to specify discounts per product. That's why I think it's better even to leave out both for now.

jjscarafia: Hi @pedrobaeza . 
1. We don't have any price_unit field on the "product.pack.line", we only have a discount field used, as @zaoral said, for discounts to be applied only when selling packs.
2. I would keep this field (discount) on product_pack module just in case it can be used in a possible new purchase_product_pack module. But, if you still prefer to move it to a the sale_product_pack module, no problem
3. The only price_unit field we create is [this one](https://github.com/ingadhoc/product/blob/d8eba5f94631e5e8da99f807c96ead694ef89dc2/product_pack/models/sale_order_line_pack_line.py#L25) that is used when working with "none_detailed_assited_price" option. Check [this demo video](https://drive.google.com/file/d/1Q0rzePCdQ2S5SVI1YFt5eef42SkwAN9I/view) of this functionality

pedrobaeza: Thank you very much for the video! Yeah, now I have it more clear, and of course I have mixed the `price_unit` at sale line pack level with product pack line.

Any way, you can see that you have packed a lot of different features for a lot of use cases, and I would unpack them, unfolding these features through successive modules. For the "assisted" feature (which is not a correct English name IMO for what you have showcased me in the video, but I don't have right now a better one), it's clearly an extension module for sale_product_pack, adding the option in pack type and incorporating the logic in sales order, so better to rip off now. What are the advantages of doing so? We will get smaller pieces that can be migrated, tested, etc faster. We avoid also possible side effects, cleaner code (we avoid a lot of `if` in the same base code). Do you agree in this unfolding? (we will take care).

About the discount, let's keep it, but on `sale_product_pack` (and properly labelled that is for sales), because if we share the same discount for sales and purchases, on merchandise (purchase and sell) businesses, we won't want to have same discount on both operations.

jjscarafia: Perfect Pedro. It's ok for me to change the "assisted" name for anything better than that and to do it on a separate module extending the functionality.

ernestotejeda:
Changes done

zaoral:
Thank you for the changes @ernestotejeda 

I have tested run an upgrade at it seams there is failing

zaoral:
sorry, I made enter before pasting the log.
Here is the error in the log.

Best Regards
```sh
2019-08-28 20:40:23,031 40 INFO pack-new odoo.modules.loading: loading 109 modules...
2019-08-28 20:40:24,775 40 INFO pack-new odoo.modules.migration: module product_pack: Running migration [>12.0.1.1.0] pre-migration
2019-08-28 20:40:25,425 40 INFO pack-new OpenUpgrade: product_pack: pre-migration script called with version 12.0.1.0.0
2019-08-28 20:40:25,428 40 ERROR pack-new odoo.sql_db: bad query: b"\n        UPDATE product_template\n        SET pack_type = (\n            CASE\n                WHEN pack_price_type='none_detailed_assited_price'\n                    THEN 'none_detailed_assisted_price'\n                WHEN pack_price_type='totalice_price'\n                    THEN 'totaliced_price'\n                ELSE\n                    pack_price_type\n                END\n        )\n        "
ERROR: column "pack_type" of relation "product_template" does not exist
LINE 3:         SET pack_type = (
                    ^

2019-08-28 20:40:25,428 40 ERROR pack-new OpenUpgrade: Error running b"\n        UPDATE product_template\n        SET pack_type = (\n            CASE\n                WHEN pack_price_type='none_detailed_assited_price'\n                    THEN 'none_detailed_assisted_price'\n                WHEN pack_price_type='totalice_price'\n                    THEN 'totaliced_price'\n                ELSE\n                    pack_price_type\n                END\n        )\n        "
2019-08-28 20:40:25,428 40 ERROR pack-new OpenUpgrade: product_pack: error in migration script product_pack/migrations/12.0.1.1.0/pre-migration.py: column "pack_type" of relation "product_template" does not exist
LINE 3:         SET pack_type = (
                    ^

2019-08-28 20:40:25,428 40 ERROR pack-new OpenUpgrade: column "pack_type" of relation "product_template" does not exist
LINE 3:         SET pack_type = (
                    ^
  File "/usr/local/lib/python3.5/dist-packages/openupgradelib/openupgrade.py", line 1495, in wrapped_function
    if use_env2 else cr, version)
  File "/opt/odoo/custom/repositories/product-attribute-12.0-add-product_pack/product_pack/migrations/12.0.1.1.0/pre-migration.py", line 26, in migrate
    """,
  File "/usr/local/lib/python3.5/dist-packages/openupgradelib/openupgrade.py", line 979, in logged_query
    cr.execute(query, args)
  File "/usr/local/lib/python3.5/dist-packages/odoo/sql_db.py", line 155, in wrapper
    return f(self, *args, **kwargs)
  File "/usr/local/lib/python3.5/dist-packages/odoo/sql_db.py", line 232, in execute
    res = self._obj.execute(query, params)
psycopg2.ProgrammingError: column "pack_type" of relation "product_template" does not exist
LINE 3:         SET pack_type = (
                    ^

```

product_pack/migrations/12.0.1.0.0/pre-migration.py
```diff
@@ -0,0 +1,28 @@
...
+        UPDATE product_template
+        SET pack_type = (
+            CASE
+                WHEN pack_price_type='none_detailed_assited_price'
+                    THEN 'none_detailed_assisted_price'
+                WHEN pack_price_type='totalice_price'
+                    THEN 'totaliced_price'
+                ELSE
+                    pack_price_type
+                END
```
Yajo:
```suggestion
            END
```

product_pack/models/product_pack_line.py
```diff
@@ -0,0 +1,65 @@
+# Copyright 2019 Tecnativa - Ernesto Tejeda
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+
+from odoo import api, fields, models, _
+import odoo.addons.decimal_precision as dp
+from odoo.exceptions import ValidationError
+
+
+class ProductPack(models.Model):
```
yajo:
```suggestion
class ProductPackLine(models.Model):
```

product_pack/models/product_pack_line.py
```diff
@@ -0,0 +1,65 @@
...
+        """
+        Check assisted packs don't have packs as child
+        """
+        for rec in self.filtered('product_id.pack'):
+            parent = rec.parent_product_id
+            if parent.pack_type == 'none_detailed_assisted_price':
+                raise ValidationError(_(
+                    'You can not set this product because it is a pack '
+                    'product and this line is part of a '
+                    '"None Detailed - Assisted Price Pack"'))
```
yajo:
```suggestion
                    '"None Detailed - Assisted Price" pack'))
```

product_pack/models/product_pack_line.py
```diff
@@ -0,0 +1,65 @@
...
+
+    @api.constrains('product_id')
+    def check_recursion(self):
+        """ Check recursion on packs """
```
yajo:
```suggestion
        """Check recursion on packs"""
```

product_pack/models/product_pack_line.py
```diff
@@ -0,0 +1,65 @@
...
+        for line in self:
+            parent_product = line.parent_product_id
+            pack_lines = line
+            while pack_lines:
+                if parent_product in pack_lines.mapped('product_id'):
+                    raise ValidationError(_(
+                        'Error! You cannot set recursive packs.\n'
```
yajo:
```suggestion
                        'You cannot set recursive packs.\n'
```

It's already obvious that it is an error :blush:

product_pack/models/product_product.py
```diff
@@ -0,0 +1,102 @@
...
+
+
+class ProductProduct(models.Model):
+    _inherit = 'product.product'
+
+    pack_line_ids = fields.One2many(
+        'product.pack.line',
+        'parent_product_id',
+        'Pack Products',
+        help='List of products that are part of this pack.'
```
yajo:
```suggestion
        help='Products that are part of this pack.'
```

This might be a list, or it could be a kanban... Well, they're products anyway.

product_pack/models/product_product.py
```diff
@@ -0,0 +1,102 @@
...
+    )
+    used_pack_line_ids = fields.One2many(
+        'product.pack.line',
+        'product_id',
+        'On Packs',
+        help='List of packs where product is used.'
```
yajo:
```suggestion
        help='Packs where product is used.'
```

product_pack/models/product_product.py
```diff
@@ -0,0 +1,102 @@
...
+    pack_line_ids = fields.One2many(
+        'product.pack.line',
+        'parent_product_id',
+        'Pack Products',
+        help='List of products that are part of this pack.'
+    )
+    used_pack_line_ids = fields.One2many(
+        'product.pack.line',
+        'product_id',
+        'On Packs',
```
yajo:
```suggestion
        'Found in packs',
```

Most likely the product is inside the pack and not on top of it. :stuck_out_tongue:

product_pack/models/product_product.py
```diff
@@ -0,0 +1,102 @@
...
+        help='List of packs where product is used.'
+    )
+
+    @api.multi
+    def separate_pack_products(self):
+        """ Divide the products and the pack products into two separate
```
yajo:
```suggestion
        """Split products and the products in 2 separate recordsets.
```

Remember the docstrings standard! :blush: https://www.python.org/dev/peps/pep-0257/

product_pack/models/product_product.py
```diff
@@ -0,0 +1,102 @@
...
+        recordsets.
+        :return: [packs, no_packs]
+        """
+        packs = self.filtered(lambda p: p.pack and p.pack_type in [
```
yajo: Always use `set`s when using `in`. It's a lot faster.
```suggestion
        packs = self.filtered(lambda p: p.pack and p.pack_type in {
```

product_pack/models/product_product.py
```diff
@@ -0,0 +1,102 @@
...
+            'totaliced_price',
+            'none_detailed_assisted_price',
+            'none_detailed_totaliced_price',
+        ])
```
yajo:
```suggestion
        })
```

product_pack/models/product_product.py
```diff
@@ -0,0 +1,102 @@
...
+    )
+    used_pack_line_ids = fields.One2many(
+        'product.pack.line',
+        'product_id',
+        'On Packs',
+        help='List of packs where product is used.'
+    )
+
+    @api.multi
+    def separate_pack_products(self):
```
yajo:
```suggestion
    def split_pack_products(self):
```

product_pack/models/product_product.py
```diff
@@ -0,0 +1,102 @@
...
+                sub_qty = subproduct.quantity
+                if sub_qty:
+                    pack_qty_available.append(math.floor(
+                        subproduct_stock.qty_available / sub_qty))
+                    pack_virtual_available.append(math.floor(
+                        subproduct_stock.virtual_available / sub_qty))
+            # TODO correctly calculate pack virtual available for negatives
+            res[product.id] = {
+                'qty_available': (
+                    pack_qty_available and min(pack_qty_available) or False),
```
yajo:
```suggestion
                    min(pack_qty_available) if pack_qty_available else False),
```

Although the `and ... or` pattern works fine, it's less readable than this, and the result and complexity is the same, so better go for the most readable alternative.

product_pack/models/product_product.py
```diff
@@ -0,0 +1,102 @@
...
+                'incoming_qty': 0,
+                'outgoing_qty': 0,
+                'virtual_available': (
+                    pack_virtual_available and
+                    max(min(pack_virtual_available), 0) or False),
```
yajo:
```suggestion
                    max(min(pack_virtual_available), 0)
                    if pack_virtual_available else False),
```

product_pack/models/product_template.py
```diff
@@ -0,0 +1,101 @@
...
+    )
+    pack_line_ids = fields.One2many(
+        related='product_variant_ids.pack_line_ids',
+        readonly=True,
+    )
+    used_pack_line_ids = fields.One2many(
+        related='product_variant_ids.used_pack_line_ids',
+        readonly=True,
+    )
+    allow_modify_pack = fields.Boolean()
```
yajo:
```suggestion
    modifiable_pack = fields.Boolean()
```

product_pack/models/product_template.py
```diff
@@ -0,0 +1,101 @@
...
+        products = self.filtered(
+            lambda x: x.allow_modify_pack and
+            x.pack_type and x.pack_type != 'components_price')
+        for rec in products:
+            rec.allow_modify_pack = False
+
+    @api.constrains('pack_type')
+    def check_relations(self):
+        """
+        Check assisted packs don't have packs a child
```
yajo:
```suggestion
        """Check assisted packs don't have packs a child."""
```

See https://www.python.org/dev/peps/pep-0257/#one-line-docstrings

product_pack/models/product_template.py
```diff
@@ -0,0 +1,101 @@
...
+        """
+        for rec in self:
+            if rec.pack_type == 'none_detailed_assisted_price':
+                child_packs = rec.mapped(
+                    'pack_line_ids.product_id').filtered('pack')
+                if child_packs:
+                    raise ValidationError(_(
+                        'A "None Detailed - Assisted Price Pack" can not have '
```
yajo:
```suggestion
                        'A "None Detailed - Assisted Price" pack cannot have '
```

product_pack/models/product_template.py
```diff
@@ -0,0 +1,101 @@
...
+                        'a pack as a child!'))
+            if rec.pack:
+                used_lines = rec.product_variant_ids.used_pack_line_ids
+                for product in used_lines.mapped('parent_product_id'):
+                    if product.pack_type == 'none_detailed_assisted_price':
+                        raise ValidationError(_(
+                            'You can not set this product as pack because it '
```
yajo:
```suggestion
                            'You cannot set this product as pack because it '
```

product_pack/models/product_template.py
```diff
@@ -0,0 +1,101 @@
...
+                            'is part of a "None Detailed - Assisted Price '
```
yajo:
```suggestion
                            'is part of a "None Detailed - Assisted Price" '
```

product_pack/models/product_template.py
```diff
@@ -0,0 +1,101 @@
...
+                            'Pack"'))
```
yajo:
```suggestion
                            'pack'))
```

product_pack/models/product_template.py
```diff
@@ -0,0 +1,101 @@
...
+
+    @api.constrains('company_id', 'product_variant_ids')
+    def check_pack_line_company(self):
+        """
+        Check packs are related to packs of same company
```
yajo:
```suggestion
        """Check packs are related to packs of same company."""
```

product_pack/models/product_template.py
```diff
@@ -0,0 +1,101 @@
...
+                if (line.product_id.company_id and rec.company_id) and \
+                        line.parent_product_id.company_id != rec.company_id:
+                    raise ValidationError(_(
+                        'Pack lines products company must be the same as the '
+                        'parent product company'))
+
+    @api.multi
+    def write(self, vals):
+        """
+        We remove from prod.prod to avoid error
```
yajo:
```suggestion
        """We remove from prod.prod to avoid error."""
```

product_pack/models/product_template.py
```diff
@@ -0,0 +1,101 @@
...
+        """
+        if vals.get('pack_line_ids', False):
+            self.product_variant_ids.write(
+                {'pack_line_ids': vals.pop('pack_line_ids')})
```
yajo: :warning: :fearful: This code triggers some internal alerts. What is this supposed to be doing? Are you conscientious it's altering the dict? Is it on purpose?

yajo (changes requested):

product_pack/models/product_template.py
ernestotejeda: Yes, I'm conscientious it's altering the dict, the purpose of who did this in previous versions was to be able to manage pack lines of product.product from product.template (when there are no variants). I kept it in that way because I didn't see that it could cause any problem. Do you think that would affect something else?

ernestotejeda:
Changes done

product_pack/models/product_product.py
```diff
@@ -0,0 +1,101 @@
...
+            to_uom = self.env['uom.uom'].browse([self._context['uom']])
+
+        for product in packs:
+            list_price = product.price_compute('list_price').get(product.id)
+            if to_uom:
+                list_price = product.uom_id._compute_price(
+                    list_price, to_uom)
+            product.lst_price = list_price + product.price_extra
+
+    def _compute_quantities_dict(
```
sergio-teruel: This method is defined in stock module (https://github.com/odoo/odoo/blob/12.0/addons/stock/models/product.py#L92), so I think that mybe you should create a new module for extend this or add stock dependency 
cc @pedrobaeza What do you think about this??

sergio-teruel (changes requested):

product_pack/models/product_product.py
```diff
@@ -0,0 +1,101 @@
...
+                product_line_price = pack_line.product_id.price
+                pack_price += (product_line_price * pack_line.quantity)
+            prices[product.id] = pack_price
+        return prices
+
+    @api.depends('list_price', 'price_extra')
+    def _compute_product_lst_price(self):
+        packs, no_packs = self.split_pack_products()
+        super(ProductProduct, no_packs)._compute_product_lst_price()
+
```
sergio-teruel: Remove empty lines

sergio-teruel (changes requested):

product_pack/views/product_template_views.xml
```diff
@@ -0,0 +1,26 @@
...
+    <!-- Product view modification -->
+    <record model="ir.ui.view" id="product_template_only_form_view">
+        <field name="name">product.template.pack.form</field>
+        <field name="model">product.template</field>
+        <field name="inherit_id" ref="product.product_template_only_form_view" />
+        <field name="arch" type="xml">
+            <field name="type" position="after">
+                <field name="pack"/>
+                <field name="pack_type" attrs="{'invisible':[('pack', '=', False)], 'required':[('pack','=',True)]}" context="{'pack_type': pack_type}"/>
+                <field name="modifiable_pack" attrs="{'invisible':[('pack', '!=', False), ('pack_type', '!=', 'components_price')]}"/>
```
sergio-teruel: If pack is True modifiable_pack is invisible?? Sure??

product_pack/models/product_template.py
yajo: I'm not really sure this is wrong, but it feels like there's something wrong in it, so please, just to be sure:

1. Add a screenshot/screencast here so we see this code's real effect in odoo.
1. Just removing `readonly=True` from the field would be the same? At least it seems so...
1. In case it's OK, I think you should use a different dict, to avoid unexpected alterations:

   ```python
   _vals = vals.copy()
   _vals.pop('pack_line_ids')
   return super().write(_vals)
   ```

product_pack/views/product_template_views.xml
ernestotejeda: pack is True and pack_type == components_price.
If you unchecke `pack`, then `pack_type` is left with the value it had, that is, there is no onchange method that sets `pack_type` to False when `pack` is unchecked. that's why the condition have not only `pack_type == components_price`, it has also `pack is True`

product_pack/models/product_product.py
ernestotejeda: Maybe add a module named stock_product_pack or something like that, which depends on product_pack and stock and with autoinstall true. The only rare thing I see is that this module will only have that method

product_pack/models/product_template.py
ernestotejeda: @Yajo , I don't understand step 1. Anyway, what I did was remove the `readonly = true` that has no effect because in the product.template view is set readonly again with a condition. I also used a _vals variable as you state in step 3 to avoid unexpected alterations.

ernestotejeda:
Changes done

rafaelbn (commented):
You cannot modify a pack if it is not pack

rafaelbn (commented):
@zaoral @jjscarafia once you set is pack then all the options IMHO should be moved to the tab "Pack"

Check:

product_pack/migrations/12.0.1.0.0/pre-migration.py
rafaelbn: @ernestotejeda @sergio-teruel , here @zaoral says that the boolean is not needed! It was a selection ... I don't undestand the field, purpose, etc...  explain for a basic user and attende the @zaoral comment. Please @zaoral could you add a screenshot? Thanks

rafaelbn (commented):
I think we should add list_price in the pack tab , in the list for companents

product_pack/demo/product_product_demo.xml
rafaelbn: Answer @ernestotejeda please

rafaelbn:
Could pleas explain this options in the readme? maybe @zaoral could help in this texts :smile:

product_pack/README.rst
```diff
@@ -0,0 +1,95 @@
...
+* `ADHOC SA <https://www.adhoc.com.ar>`_:
+
+  * Juan José Scarafía
+  * Nicolas Mac Rouillon
+  * Katherine Zaoral
+* `NaN·TIC <http://www.nan-tic.com>`_
+* `Tecnativa <https://www.tecnativa.com>`_:
+
+  * Ernesto Tejeda
+  * Pedro M. Baeza
```
rafaelbn, line 94: Add Sergio y Jairo please

ernestotejeda:
- I removed the modifiable_pack field because it is not used anywhere in the code.
- I added an explanation in the readme of the options of the pack type field, but I think it would be good @zaoral to help me with that
- I added Sergio and Jairo in contributors section in the readme 

For the other suggestions maybe @zaoral , @jjscarafia or some one else could give an opinion

product_pack/models/product_pack_line.py
```diff
@@ -0,0 +1,65 @@
...
+    # need to enforce this condition
+    _sql_constraints = [
+        ('product_uniq', 'unique(parent_product_id, product_id)',
+         'Product must be only once on a pack!'),
+    ]
+
+    @api.constrains('product_id')
+    def check_relations(self):
+        """
+        Check assisted packs don't have packs as child
```
yajo: Please check https://www.python.org/dev/peps/pep-0257/#one-line-docstrings and follow the standard. :blush:

product_pack/models/product_pack_line.py
```diff
@@ -0,0 +1,65 @@
...
+
+    # because on expand_pack_line we are searching for existing product, we
+    # need to enforce this condition
+    _sql_constraints = [
+        ('product_uniq', 'unique(parent_product_id, product_id)',
+         'Product must be only once on a pack!'),
+    ]
+
+    @api.constrains('product_id')
+    def check_relations(self):
```
yajo:
```suggestion
    def _check_relations(self):
```

Constraint methods are private always.

product_pack/models/product_pack_line.py
```diff
@@ -0,0 +1,65 @@
...
+        for rec in self.filtered('product_id.pack'):
+            parent = rec.parent_product_id
+            if parent.pack_type == 'none_detailed_assisted_price':
+                raise ValidationError(_(
+                    'You can not set this product because it is a pack '
+                    'product and this line is part of a '
+                    '"None Detailed - Assisted Price" Pack'))
+
+    @api.constrains('product_id')
+    def check_recursion(self):
```
yajo:
```suggestion
    def _check_recursion(self):
```

product_pack/models/product_pack_line.py
```diff
@@ -0,0 +1,65 @@
...
+    @api.constrains('product_id')
+    def check_relations(self):
+        """
+        Check assisted packs don't have packs as child
+        """
+        for rec in self.filtered('product_id.pack'):
+            parent = rec.parent_product_id
+            if parent.pack_type == 'none_detailed_assisted_price':
+                raise ValidationError(_(
+                    'You can not set this product because it is a pack '
```
yajo:
```suggestion
                    'You cannot set this product because it is a pack '
```

product_pack/models/product_product.py
```diff
@@ -0,0 +1,99 @@
...
+class ProductProduct(models.Model):
+    _inherit = 'product.product'
+
+    pack_line_ids = fields.One2many(
+        'product.pack.line',
+        'parent_product_id',
+        'Pack Products',
+        help='Products that are part of this pack.'
+    )
+    used_pack_line_ids = fields.One2many(
```
yajo:
```suggestion
    used_in_pack_line_ids = fields.One2many(
```

I think this way the field usage is more self-explained. :thinking:

product_pack/models/product_product.py
```diff
@@ -0,0 +1,99 @@
...
+        'product.pack.line',
+        'product_id',
+        'Found in packs',
+        help='Packs where product is used.'
+    )
+
+    @api.multi
+    def split_pack_products(self):
+        """Split products and the pack in 2 separate recordsets.
+        :return: [packs, no_packs]
```
yajo, line 34:
```suggestion

        :return: [packs, no_packs]
```

product_pack/models/product_template.py
```diff
@@ -0,0 +1,88 @@
...
+             "Do not assist to get pack price using pack lines."
+    )
+    pack = fields.Boolean(
+        'Pack?',
+        help='Is a Product Pack?',
+    )
+    pack_line_ids = fields.One2many(
+        related='product_variant_ids.pack_line_ids',
+    )
+    used_pack_line_ids = fields.One2many(
```
yajo:
```suggestion
    used_in_pack_line_ids = fields.One2many(
```

product_pack/models/product_template.py
```diff
@@ -0,0 +1,88 @@
...
+        related='product_variant_ids.used_pack_line_ids',
+        readonly=True,
+    )
+
+    @api.constrains('pack_type')
+    def check_relations(self):
```
yajo:
```suggestion
    def _check_relations(self):
```

product_pack/models/product_template.py
```diff
@@ -0,0 +1,88 @@
...
+                used_lines = rec.product_variant_ids.used_pack_line_ids
+                for product in used_lines.mapped('parent_product_id'):
+                    if product.pack_type == 'none_detailed_assisted_price':
+                        raise ValidationError(_(
+                            'You cannot set this product as pack because it '
+                            'is part of a "None Detailed - Assisted Price" '
+                            'pack'))
+
+    @api.constrains('company_id', 'product_variant_ids')
+    def check_pack_line_company(self):
```
yajo:
```suggestion
    def _check_pack_line_company(self):
```

product_pack/models/product_template.py
```diff
@@ -0,0 +1,88 @@
...
+            for line in rec.used_pack_line_ids:
+                if (line.product_id.company_id and rec.company_id) and \
+                        line.parent_product_id.company_id != rec.company_id:
+                    raise ValidationError(_(
+                        'Pack lines products company must be the same as the '
+                        'parent product company'))
+
+    @api.multi
+    def write(self, vals):
+        """We remove from prod.prod to avoid error"""
```
yajo:
```suggestion
        """We remove from product.product to avoid error"""
```

product_pack/readme/CONTRIBUTORS.rst
yajo: This seems resolved.

product_pack/views/product_product_views.xml
```diff
@@ -0,0 +1,22 @@
...
+
+<odoo>
+    <record model="ir.ui.view" id="nan_product_pack_form">
+        <field name="name">product.product.pack.form</field>
+        <field name="model">product.product</field>
+        <field name="inherit_id" ref="product.product_normal_form_view" />
+        <field name="arch" type="xml">
+            <field name="type" position="after">
+                <field name="pack"/>
+                <field name="pack_type" attrs="{'invisible':[('pack', '=', False)], 'required':[('pack','=',True)]}" context="{'pack_type': pack_type}"/>
```
yajo: IMHO it would make more sense for the user if you put these 2 fields into the notebook page you create a few lines below....

product_pack/views/product_product_views.xml
```diff
@@ -0,0 +1,22 @@
...
+            </field>
+            <notebook position="inside">
+                <page string="Pack" attrs="{'invisible':[('pack', '=', False)]}">
```
yajo: ... and make the page always visible; just the fields in it become or not visible.

product_pack/views/product_template_views.xml
```diff
@@ -0,0 +1,25 @@
...
+        <field name="name">product.template.pack.form</field>
+        <field name="model">product.template</field>
+        <field name="inherit_id" ref="product.product_template_only_form_view" />
+        <field name="arch" type="xml">
+            <field name="type" position="after">
+                <field name="pack"/>
+                <field name="pack_type" attrs="{'invisible':[('pack', '=', False)], 'required':[('pack','=',True)]}" context="{'pack_type': pack_type}"/>
+            </field>
+            <notebook position="inside">
+                <page string="Pack" attrs="{'invisible': ['|', ('product_variant_count', '>', 1), ('pack', '=', False)]}">
```
yajo: ... and the same here.

product_pack/models/product_template.py
```diff
@@ -0,0 +1,88 @@
...
+             "* Detailed - Totaliced Price: Detail lines totalicing "
+             "lines prices on pack (don't show component prices).\n"
+             "* Detailed - Fixed Price: Detail lines and use product"
+             " pack price (ignore line prices).\n"
+             "* None Detailed - Assisted Price: Do not detail lines. "
+             "Assist to get pack price using pack lines./n"
+             "* None Detailed - Totaliced Price: Do not detail lines. "
+             "Do not assist to get pack price using pack lines."
+    )
+    pack = fields.Boolean(
```
yajo: Do we really need this field? :thinking: 

You can consider a product without `pack_type` as a non-pack product, and then get rid of this field. I also think it would be even more easy to use.

yajo (changes requested):

yajo:
Oh and please mark resolved conversations as resolved; it makes it easier to see what's pending. Actually you still had pending points that maybe you've missed because of this... :wink:

product_pack/readme/USAGE.rst
```diff
@@ -0,0 +1,21 @@
+To use this module, you need to:
+
+#. Go to *Sales > Products > Products*, create a product and set
+   "Product Pack".
```
zaoral: use  `set "Pack?"` instead of  `set "Product Pack"` which is the proper name of the field show in the view.

product_pack/readme/USAGE.rst
```diff
@@ -0,0 +1,21 @@
...
+#. Then set the products has to include in the product pack of the product.
+#. Set *Pack Type*.
+
+`Product pack` is a base module for `sale_product_pack` and others modules that
+needs to use packs. `Pack type` is used to define the behavior that the
+packs will have when it is selected in the sales order lines
+(if `sale_product_pack` module is installed). The optios of this field are the
+followings:
+
+  * Detailed - Components Prices: Detail pack lines with prices.
```
zaoral: The description looks good to me
I leave you here some suggestion if you would like to add it to your usage file.

  * Detailed - Components Prices: will show each components and its prices, including the pack product itself with its price.
  * Detailed - Totaliced Price: will show each component but will not show components prices. The pack product will be the only one that has price and this one will be the sum of all the components prices.
  * Detailed - Fixed Price: will show each components but will not show components prices. The pack product will be the only one that has price and this one will be the price set in the pack product.
  * None Detailed - Totaliced Price: Will not show the components information, only the pack product and the price will be the sum of all the components 
* None Detailed - Assisted Price: Will not show the components, only show the pack product. By default will show the pack price as the sum of the components, but will add a button in the sale order line that help you to change the pack price modifying the components prices, quantities, and discounts. This information let the user to play with the pack price depending of customer needs and agreements. and be able to consult/change this information without showing this information in the sale order lines.

zaoral (commented):
Hi @ernestotejeda 
Here I leave you some comments about the Readme you ask
Also make some comments in @Yajo comments in order to now you both point of views.

Last but not least, about the modifiable_pack field. please could you consider to add it again? I understand that this field was removed because is now only been used in sale_pack_product module, but If this is reason then maybe we should also move the pack_type field and leave this module as simpler module.

NOTE: @ernestotejeda I think that the migration scripts still failing

product_pack/views/product_product_views.xml
zaoral: @ernestotejeda @Yajo @rafaelbn 

I am agreed with @rafaelbn in this case

IMHO I think will be better to have the "Pack" boolean field as checkbox along side with "​Can be Purchased" and "Can be Sold". And if this field is show/hide the "Pack" page.

This way we can put all the pack related fields inside the "Pack" page and we do not need to apply invisible or not attributes by field.

What do you think?

product_pack/demo/product_product_demo.xml
zaoral: I think this has been already solved

product_pack/migrations/12.0.1.0.0/pre-migration.py
zaoral: @rafaelbn @ernestotejeda 

yes, here is a funcional example. For the Detailed kind Pack Types, when the pack is added to a sale order and we save, we will see one line for the product pack and new sale order lines one for each pack components.

At the momento the salesman can edit the sale order lines and change the quantites of components of remove some of them. The Modificable Pack field is the one that was making this possible or not. If the field is set True then the pack can be edited and can be changed whaterver we want (quantities, remove products, change predefined discounts). If the field is set False then we avoid that the salesman can change the pack: the pack is forced to be sell as it is defined can not be modificable.

Modificable Pack field was before a selection field because we want to add the same funcionality in two scopes, backend and e-commerce. So a pack could be:

 * Not modificable: its forced to be sell as it defined.
 * Modificable only in backend: can be modificable by the internal user the salesman in the backend. This is if the salesman have any agreements with the customer and the company has the policy to let some flexibility to change the pack components/quantities/remove them.
 * Modificable in backend and e-commerce: the portal user can select a pack product and add it to their shop cart, and then change the quantities of the products (less/more) of the components and can also remove components from the pack.

In this case since the new product_pack module is a generic one that does not depends from sale or website, then this field was change to a simpler boolean (which it looks good to me).

Now I see that the modificable pack field was removed, I think this field should be leave in this module because can be used for sale_product_pack as well by the purchase_product_pack in the future.

product_pack/models/product_product.py
pedrobaeza: Is this related to only sale? If so, there should be a module `sale_stock_product_pack` (as we can sell services pack and don't need `stock` for that). If it can be in both sale and purchase, then we need a module `stock_product_pack`, and probably their corresponding glue modules.

product_pack/models/product_template.py
pedrobaeza: He is asking for a video showcasting what this piece of code is doing.

product_pack/models/product_template.py
pedrobaeza: From UX point of view, it's better to have this flag, as it's a switch for the rest. I see correct to keep it, although I understand your reasoning. Another UX option would be to have an option `No pack` as default and make `pack_type` as required, but then that advanced selection will be outside the page and out of context.

product_pack/readme/USAGE.rst
pedrobaeza: Then better to put the label in the field instead only on view. I would go for `Is Pack?` instead.

product_pack/readme/USAGE.rst
pedrobaeza: I think `totaliced` is not a proper word. It should be `totalized` instead (https://www.linguee.com/english-spanish/translation/totalized+data.html)

About `assisted` (@ernestotejeda, remember we talked about extracting it into a separate module), we can called it `guided`, although the other word can be correct as well. But as said, we should leave it outside this first PR for getting this merged ASAP, and then prepare the other module on top of this one.

product_pack/views/product_product_views.xml
yajo: Hmm... ok, it makes sense too. I gues the bool field should be named `pack_ok` for consistency.

ernestotejeda: I agree with @zaoral too, since `"Pack?"` field will determine the visibility of the "Pack" page that is inside the notebook just like "Can be Purchased" and "Can be Sold" with their corresponding pages.
And I agree with @Jairo to name this field pack_ok

zaoral: Perfect, @ernestotejeda please could you implement this change and let us know.
change the field to pack_ok LGTM

ernestotejeda:
Changes done.
@zaoral, @pedrobaeza , @Yajo , @sergio-teruel , @jjscarafia , @rafaelbn  I would be grateful if you could review these latest changes to see if the suggestions were resolved correctly, Thank you.

product_pack/migrations/12.0.1.0.0/pre-migration.py
```diff
@@ -0,0 +1,35 @@
...
+                        THEN NULL
+                    WHEN pack_price_type='totalice_price'
+                        THEN 'totalized_price'
+                    WHEN pack_price_type='none_detailed_totaliced_price'
+                        THEN 'none_detailed_totalized_price'
+                    ELSE
+                        pack_price_type
+                END
+            ),
+            pack_ok = pack,
```
yajo: This column won't exist at the time you're executing this script, right? This should produce a failure in the migrations. I see these options:

1. Use `oldname` parameter in the `pack_ok` field. I'm not sure if this is deprecated today, but otherwise it should handle the rename flawlessly for you.
1. Do this conversion in post-migration.
1. Precreate the column with [`add_fields()`](https://github.com/OCA/openupgradelib/blob/614161f47bc7c1592fa9985650b4cbc7429983a2/openupgradelib/openupgrade.py#L2027). TBH I'm not sure if an `ALTER TABLE` command would be enough and let Odoo handle the rest... 

I think you should check if the same happens with the other fields you're using in this script. You'll have to test & debug this migration.

product_pack/migrations/12.0.1.0.0/pre-migration.py
pedrobaeza: Are there any pack type that is not compatible with this flag? If so, then maybe this should be unfolded inside pack type. I think non detailed packs should have this option, isn't it? Then I think it should be:

- Detailed ... - Non modifiable
- Detailed ... - Modifiable

ernestotejeda:
@zaoral , I have a question in this part of the code in the original module:
https://github.com/ingadhoc/product/blob/0e9d8979efeeec5aae2aa051499af2155b5ccc76/product_pack/models/product_product.py#L51

I would like you could help me. I'm extracting the redefinition of this method (_compute_quantities_dict) to a new module as commented here https://github.com/OCA/product-attribute/pull/498#discussion_r320351413: 
why here it is forced the 'virtual_available' key of the dictionary to a positive value or False?, why it cannot be negative?.
Wouldn't it be right to do something similar to what is done with the 'qty_available' key? I mean do this:
```
'virtual_available': (
                    pack_virtual_available and
                    min(pack_virtual_available) or False),
```

yajo:
I think that one of the most confusing parts of the module is that dropdown that modifies the pack behavior, possibly because it's a single field modifying many things, and we could make it more simple if we remove it and replace it with several fields:

- Show components in sale order (bool `pack_component_show`): Indicates if the components should be displayed individually in the sale order.
- Component prices (select `pack_component_price`): Indicates what to do with individual component prices. It should have these options:

  1. Detailed per component (`detailed`): It would include each component's price into its corresponding line. Constrained to pack_component_show=True.
  1. Merged in main product (`merged`): Pack price = main product price + sum of all component prices. If pack_component_show=True, individual component prices will be always 0.
  1. Ignored (`ignored`): Component prices are set to 0 and only the main product price is used.

What do you think about it?

Also another question: do any of these fields have any meaning without the `sale` module installed? :thinking: I don't think so, so all of them should go into `sale_product_pack`, right?

pedrobaeza:
> do any of these fields have any meaning without the sale module installed?

The idea is to reuse them on a future `purchase_product_pack`.

legalsylvain:
Hi all.
first, thanks for this module I just discovered right now !
I'm very interested to install and use this V12.0 module.
 
1) Also interested to develop a ``pos_product_pack`` to make the feature available in the PoS. Is this work planed by other people ?

2) About the feature, one of my needs is not covered : In some cases, the content of the pack is changing during the year. (depending of date). In my shops, we are selling sometime packs of various products including fruits or vegetables. Of course, the list of fruits and vegetables depend of dates, but the pack is always the same.

to cover this extra feature, I propose to make a PR against your branch, and include the following changes : 
- add a ``date_start`` and a ``date_stop`` fields on the pack.line model. (field optional)
- add a function in the pack model named ``_get_pack_lines(date=False)`` that return the content of the pack (product list), depending of the date. (if no date, today is used).
- call the function in the different part of this module. (replacing ``self.mapped('pack_line_ids.product_id'))``)
- add a group to make optional all that features. (something like "Pack Content Seasonality"). So, it is an extra feature, that will not make this module more complex, in most cases.

@ernestotejeda,  all, what about this proposal ?

I thought to add an extra module that depend on this module, but the function should be in this module, to make all the other ones working.

I can work on that point very quickly, and propose a PR this week.

Thanks for your comments !

pedrobaeza:
Hi, @legalsylvain thanks for your interest and your feature is interesting, although maybe it's better to include it in a new module for having that isolated. It can be included here the function you need though, returning by default the `mapped`, and being overwritten in that module. But my position is not strong in this if you want to have it on main module, but I think it's going to be more difficult with all the group stuff, and this module won't need other glue modules for any of the extensions (sale, purchase, stock, pos...).

About `pos_product_pack`, it's not yet planned in our side, but the idea of the module split is to allow things like that.

Now I'm wondering if we should create a repository OCA/pack for this or spread them like now.

legalsylvain:
Hi @pedrobaeza : Thanks for your quick reply !

> It can be included here the function you need though, returning by default the mapped, and being overwritten in that module. 

That's is the "must have" point for me, indeed.

> But my position is not strong in this if you want to have it on main module, but I think it's going to be more difficult with all the group stuff, and this module won't need other glue modules for any of the extensions (sale, purchase, stock, pos...).

Having one module : 
- pro : Less work, in each migration. (one per year is a lot, I see it as a big OCA problem, for the time being). 
- cons : what you say.

Maybe we should make one module, if most of the people think it could be interesting to have this option, and make two module if it's very specific case ? 
In food context, it's a must have feature, but I don't know for other context / verticalization. Let's see what other OCA people are saying ?
any way, not a blocking point for me.

> Now I'm wondering if we should create a repository OCA/pack for this or spread them like now.

for the time being : 
- product_pack
- product_pack_seasonality (maybe !)
- sale_pack
- sale_stock_pack ?
- purchase_pack
- pos_pack

Indeed, it maybe deserves a specific repository. :+1: 

Do you think it's relevant to make this change for 12.0, or to wait the super fast, super easy and super simple 13.0 coming soon version ?

rafaelbn:
@ernestotejeda please add test and let's merge this.

@zaoral @jjscarafia do we have your bless? https://youtu.be/bk_02-9OBmM

We cannot invest hundred hours in this module and we will maintain it for sure :smile:

legalsylvain:
> @ernestotejeda please add test and let's merge this.

please don't, before I added the function ``_get_pack_lines``.

I'll do this stuff in a few moment.

pedrobaeza:
Planned module names are:

* product_pack
* sale_product_pack
* stock_product_pack
* whatever_product_pack

Why? Because `pack` is a very generic word, used also on logistics and other places, so `product_pack` seems better. The repository must be OCA/product-pack in fact. I think it's better to do it now than later and have to move issues, modules, and so on.

rafaelbn:
We used `product_bundle` in the past: https://github.com/OCA/product-attribute/pull/180

It was beacause Odoo have already packages, packaging... :face_with_head_bandage:

legalsylvain:
@pedrobaeza I introduced the required function ``get_pack_products()``.
PR available here. https://github.com/Tecnativa/product-attribute/pull/1

finally, the problem I see if we set the seasonability feature in an extra module, that will require a lot of glue modules. I mean : 
- ``product_pack_seasonability`` : add date_start, date_stop + overload get_pack_products() to use a context value ``pack_date``)
- ``sale_product_pack_seasonability`` : to pass the date of the sale order in the context
- ``purchase_product_pack_seasonability`` : to pass the date of the purchase order in the context
- etc...

It will generate a lot of glue modules, only for a python line.

Solution 1 :  A lot of glue modules. (I'd like to avoid that).
Solution 2 : put seasonability features in product_pack module.
Solution 3 : in the ``sale_product_pack`` module (and other modules) call product_pack function with a ``with_context(pack_date=xxxx)`` that will be handled in the ``product_pack_seasonability``. (Lazy dependency)

What do you think ?

pedrobaeza:
@legalsylvain I don't see the need of that glue modules: `sale_product_pack` is going to call the method `get_pack_products` with the order date in the base. What changes is that the standard `product_pack` ignores that argument and returns always the same (the `mapped`), but your dependent module will change the return value according it.

pedrobaeza:
I have created OCA/product-pack: https://github.com/OCA/product-pack. @ernestotejeda please move PRs to that repository.

zaoral:
@ernestotejeda for me looks good.
Thank for the work.

About what you ask for the virtual_available, I have tested and discuss internally with the team, and for be honest we do not remember why we force to virtual_available to be positive, we were not able to found a user case, feel free to change it to

```python
'virtual_available': (
                    pack_virtual_available and
                    min(pack_virtual_available) or False),
```

About the migration script, thank you for change it

Best Regards

product_pack/models/product_template.py
```diff
@@ -0,0 +1,73 @@
...
+                        'parent product company'))
+            for line in rec.used_in_pack_line_ids:
+                if (line.product_id.company_id and rec.company_id) and \
+                        line.parent_product_id.company_id != rec.company_id:
+                    raise ValidationError(_(
+                        'Pack lines products company must be the same as the '
+                        'parent product company'))
+
+    @api.multi
+    def write(self, vals):
```
daramousk, line 66: What kind of error are you trying to avoid here? What if you set `store=True` to `pack_line_ids` or provide an inverse?

product_pack/security/ir.model.access.csv
```diff
@@ -0,0 +1,3 @@
+id,name,model_id:id,group_id:id,perm_read,perm_write,perm_create,perm_unlink
```
daramousk, line 1: Perhaps the security settings should mimic the settings in `product` module?

daramousk (commented):
Rather interesting to see how this module will interact with the rest. Probably a lot of changes have to be done (reporting, stock valuation etc) to take (or not take) into consideration the products of type pack. (Perhaps adding a product type would be better than the `pack_ok`)

What is the use case for this?
I understand that the products that are marked as Packs are supposed to contain other products, they are not comprised by these products.
If this is a picking issue, maybe some sort of grouping of stock.moves would be better.

pedrobaeza:
I think more changes will be needed if you create a new product type, as all the domains, filters, etc works over existing ones. The bridge for stock operations is https://github.com/OCA/stock-logistics-warehouse/pull/718. 

All of this must be moved though to the new OCA/product-pack repo.

rousseldenis:
Hi all, thanks for this.

Sorry to arrive late here but what is the difference with BOM's ?

pedrobaeza:
@rousseldenis there are plenty of reasons for having this:

* Don't depend on `mrp`.
* Having several prices options (like computing price from components + discount).
* Components display options (detailed/non detailed/assisted...).
* Expand it to other parts (like purchase or PoS).
* Seasonability like @legalsylvain proposes.
* etc.

zaoral:
Hi @ernestotejeda 
This PR can be closed? making the references to the PR that supersedes this one?

Best Regards

pedrobaeza:
Yes, superseded by https://github.com/OCA/product-pack/pull/1

PR #499, opened 2019-08-29
[11.0][IMP] product_pricelist_direct_print: don't hide category

ernestotejeda:
- [IMP] Don't hide category when 'last ordered product' is greather than zero

- [FIX] Show just the corresponding products by category when 'last ordered product' is greather than zero

Cc @Tecnativa

### DIFF
--- a/product_pricelist_direct_print/views/report_product_pricelist.xml
+++ b/product_pricelist_direct_print/views/report_product_pricelist.xml
@@ -80,7 +80,7 @@
                            t-as="category">
                             <t t-set="category_products"
                                t-value="products and products.with_context(categ_id=category.id).filtered(lambda x: x.categ_id.id == x.env.context['categ_id']) or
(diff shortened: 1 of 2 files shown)


rafaelbn (approved):

chienandalu (commented):
When you have categories in the filter and the last ordered limit, shouldn't the limit apply to each category instead of globally?

sergio-teruel (approved):
Functional test ok

ernestotejeda:
> When you have categories in the filter and the last ordered limit, shouldn't the limit apply to each category instead of globally?

@chienandalu applied to each category seems good to me, but I don't see that globally is bad, I think it depends on the circumstances. What do you think?

Tardo (approved):
LGTM :+1:

pedrobaeza:
@chienandalu @ernestotejeda please agree then on the final solution.

chienandalu (approved):
:+1:

PR #500, opened 2019-09-03
packaging_uom: fix self used in loop

guewen:


pedrobaeza (approved):
Fast-tracking obvious patch

PR #501, opened 2019-09-03
[FIX] product_pricelist_direct_print: empty categ. in last ordered products

chienandalu:
When categories are set on the filter and and last ordered products is
checked, if one of the categories was empty it causes an error due to
the t-value returning an empty dict instead of an empty recordset.

The fix returns an empty recordset of either product.template o
product.product model.

cc @Tecnativa


pedrobaeza (commented):
I think you can remove or part directly

sergio-teruel (approved):

chienandalu:
ok @pedrobaeza

pedrobaeza:
Sorry, David, it's needed. Please revert to previous one.

chienandalu:
Ok, there we go again :)

pedrobaeza (approved):
Fast tracking

PR #509, opened 2019-09-30
[WIP][12.0] Add module product_disc - alpha

grindtildeath:
This module aims to manage music records as product.

It introduces following models:

 - product.artist (Artist)
 - product.label (Record label)


ntsirintanis (commented):
This is super useful, thank you. I'll test it and probably come up with some ideas too, in the form of a PR. One thing that would be great to have is getting genre of each album automatically from services such as lastfm.

lfreeke:
Hello @grindtildeath I miss the record_format field in the view, is that on purpose?

PR #515, opened 2019-10-15
[12.0][MIG] product_supplierinfo_for_customer

MiquelRForgeFlow:
Supersede of https://github.com/OCA/product-attribute/pull/446.


product_supplierinfo_for_customer/models/product_template.py
```diff
@@ -10,25 +10,25 @@ class ProductTemplate(models.Model):
...
     variant_customer_ids = fields.One2many(
-        comodel_name='product.supplierinfo',
+        comodel_name='product.customerinfo',
         inverse_name='product_tmpl_id',
-        string='Vairant Customer',
-        domain=[('supplierinfo_type', '=', 'customer')])
+        string='Variant Customer',
+    )
 
     supplier_ids = fields.One2many(
```
pedrobaeza: Do you need to keep the definition of these fields?

product_supplierinfo_for_customer/__manifest__.py
```diff
@@ -17,11 +17,11 @@
         "purchase",
```
pedrobaeza: Why this one depends on purchase? I don't think the module must depend on it, and more wanting to use it for customers

product_supplierinfo_for_customer/security/ir.model.access.csv
```diff
@@ -0,0 +1,4 @@
+id,name,model_id:id,group_id:id,perm_read,perm_write,perm_create,perm_unlink
+access_product_customerinfo_user,product.customerinfo.user,model_product_customerinfo,base.group_user,1,0,0,0
+access_product_customerinfo_manager,product.customerinfo.manager,model_product_customerinfo,base.group_system,1,1,1,1
+access_product_customerinfo_purchase_manager,product.customerinfo purchase_manager,model_product_customerinfo,purchase.group_purchase_manager,1,1,1,1
```
pedrobaeza: This should be removed when stripping purchase dependency.

product_supplierinfo_for_customer/views/product_view.xml
```diff
@@ -120,11 +97,8 @@
                 <group attrs="{'invisible': [('product_variant_count','&lt;=',1)]}">
```
pedrobaeza: I think you are able to remove most of these overwrites with the refactoring. Take also into account the Odoo paradigm of not splitting variants supplierinfo with template ones.

product_supplierinfo_for_customer/migrations/12.0.1.0.0/post-migration.py
```diff
@@ -0,0 +1,68 @@
...
+            env, ids, 'product.supplierinfo', 'product.customerinfo')
+        openupgrade.logged_query(
+            # Set for the new table the next id val as the maximum of the
+            # other table + 1. This way, we don't have any of the old IDs
+            # duplicated in new table.
+            env.cr,
+            "SELECT setval('product_customerinfo_id_seq', "
+            "(SELECT MAX(id) FROM product_supplierinfo))"
+        )
+        env['product.supplierinfo'].browse(ids).unlink()
```
pedrobaeza: Do it on SQL

product_supplierinfo_for_customer/migrations/12.0.1.0.0/post-migration.py
```diff
@@ -0,0 +1,68 @@
...
+        ('ir_model_data', 'model', 'res_id'),
+    ]
+    for rename in renames:
+        openupgrade.logged_query(
+            env.cr, """
+            UPDATE %s
+            SET %s = %s
+            WHERE %s IN %s AND %s = %s""",
+            (
+                AsIs(rename[0]),
```
pedrobaeza: Use new `sql.SQL` syntax. It's much better than this. Or at least use a dictionary, as this is very hard to be understood with indexes.

pedrobaeza (changes requested):

product_supplierinfo_for_customer/models/product_customerinfo.py
```diff
@@ -0,0 +1,24 @@
+# Copyright 2019 Tecnativa - Pedro M. Baeza
+# Copyright 2019 Eficent Business and IT Consulting Services S.L.
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl.html).
+from odoo import api, fields, models, _
+
+
+class CustomerInfo(models.Model):
```
pedrobaeza: According guidelines, it should be `ProductCustomerInfo`

product_supplierinfo_for_customer/models/product_customerinfo.py
```diff
@@ -0,0 +1,24 @@
...
+    _inherit = "product.supplierinfo"
+    _name = "product.customerinfo"
+    _description = "Customer Pricelist"
+
+    name = fields.Many2one(
+        'res.partner', 'Customer',
```
pedrobaeza: We don't need to redefine all the attributes. Put only the redefined ones (string, domain and help).

product_supplierinfo_for_customer/models/product_template.py
```diff
@@ -10,25 +10,21 @@ class ProductTemplate(models.Model):
...
         inverse_name='product_tmpl_id',
-        string='Vairant Customer',
-        domain=[('supplierinfo_type', '=', 'customer')])
+        string='Variant Customer',
+    )
 
-    supplier_ids = fields.One2many(
-        comodel_name='product.supplierinfo',
-        inverse_name='product_tmpl_id',
+    seller_ids = fields.One2many(
```
pedrobaeza: Why do we need to keep this? I think you shouldn't override the label and leave the standard one, which is "Vendors". Even more, this one in singular is not appropriated IMO.

product_supplierinfo_for_customer/models/product_template.py
```diff
@@ -10,25 +10,21 @@ class ProductTemplate(models.Model):
...
         string='Supplier',
-        domain=[('supplierinfo_type', '=', 'supplier')])
+    )
 
-    variant_supplier_ids = fields.One2many(
-        comodel_name='product.supplierinfo',
-        inverse_name='product_tmpl_id',
+    variant_seller_ids = fields.One2many(
```
pedrobaeza: Same reasoning about label for this one

product_supplierinfo_for_customer/migrations/12.0.1.0.0/post-migration.py
```diff
@@ -0,0 +1,71 @@
...
+            # Set for the new table the next id val as the maximum of the
+            # other table + 1. This way, we don't have any of the old IDs
+            # duplicated in new table.
+            env.cr,
+            "SELECT setval('product_customerinfo_id_seq', "
+            "(SELECT MAX(id) FROM product_supplierinfo))"
+        )
+        openupgrade.logged_query(
+            env.cr,
+            "DELETE"
```
pedrobaeza: Missing spaces between words

product_supplierinfo_for_customer/migrations/12.0.1.0.0/post-migration.py
```diff
@@ -0,0 +1,71 @@
...
+                D=rename[2], E=tuple(ids), F=rename[1], G=old_model,
+            ),
+        )
+
+
+def fill_product_customerinfo(env):
+    # In pre-migration the product_customerinfo table still doesn't exist
+    other_fields = ''
+    if openupgrade.table_exists(env.cr, 'purchase_requisition_line'):
+        other_fields += ' purchase_requisition_line_id,'
```
pedrobaeza: This is not the way to do it. See below.

product_supplierinfo_for_customer/migrations/12.0.1.0.0/post-migration.py
```diff
@@ -0,0 +1,71 @@
...
+            # duplicated in new table.
+            env.cr,
+            "SELECT setval('product_customerinfo_id_seq', "
+            "(SELECT MAX(id) FROM product_supplierinfo))"
+        )
+        openupgrade.logged_query(
+            env.cr,
+            "DELETE"
+            "FROM product_supplierinfo"
+            "WHERE id in %s", (tuple(ids), )
```
pedrobaeza: Delete directly by the original condition (type = 'customer'), not by IDs

product_supplierinfo_for_customer/migrations/12.0.1.0.0/post-migration.py
```diff
@@ -0,0 +1,71 @@
...
+
+def fill_product_customerinfo(env):
+    # In pre-migration the product_customerinfo table still doesn't exist
+    other_fields = ''
+    if openupgrade.table_exists(env.cr, 'purchase_requisition_line'):
+        other_fields += ' purchase_requisition_line_id,'
+    # you must customize 'other_fields' to add fields from your custom modules
+    openupgrade.logged_query(
+        env.cr, """
+        INSERT INTO product_customerinfo (
```
pedrobaeza: You mustn't insert specific fields, but copy full structure and then move full record. This way, you don't need to worry by any possible extension to the model.

See https://github.com/OCA/contract/blob/a0e118c97570dc1c3cef7df236c714d30ee547b6/contract/migrations/12.0.4.0.0/pre-migration.py#L90-L103 as example for this.

product_supplierinfo_for_customer/migrations/12.0.1.0.0/post-migration.py
```diff
@@ -0,0 +1,71 @@
...
+        product_tmpl_id, delay, create_uid, create_date, write_uid, write_date
+        FROM product_supplierinfo
+        WHERE supplierinfo_type = 'customer'
+        RETURNING id""", (other_fields, other_fields),
+    )
+    ids = [x[0] for x in env.cr.fetchall()]
+    if ids:
+        _move_model_in_data(
+            env, ids, 'product.supplierinfo', 'product.customerinfo')
+        openupgrade.logged_query(
```
pedrobaeza: Due to the new approach, this code is needed for initializing sequence:
```python
    cr.execute("CREATE SEQUENCE IF NOT EXISTS product_customerinfo_id_seq")
    cr.execute("SELECT setval('product_customerinfo_id_seq', "
               "(SELECT MAX(id) FROM product_customerinfo))")
    cr.execute("ALTER TABLE product_customerinfo ALTER id "
               "SET DEFAULT NEXTVAL('product_customerinfo_id_seq')")
```

pedrobaeza (changes requested):

pedrobaeza (changes requested):
The statement
```
CREATE TABLE ... (LIKE ... INCLUDING ALL)
```
is missing. Without it, no constraints are copied, so the table is half built. Please include it.

MiquelRForgeFlow:
@pedrobaeza But the table has been created before during install. The migration file is post-migration, not pre-migration.

pedrobaeza:
You must do this in pre-migration, not post

AaronHForgeFlow (approved):
Functional tests with several customers and prices  LGTM :+1:

pedrobaeza (approved):

PR #516, opened 2019-10-15
[MIG][13.0] product brand

smurf-U:


pedrobaeza:
There's now a new repository called OCA/brand that should host this module as well IMO cc @sbejaoui

sbejaoui:
I agree with you @pedrobaeza on th fact that this module should move to OCA/brand repository.

IMO, this migration should unify the base model for brands. Maybe introduce `base_brand` module?

smurf-U:
@pedrobaeza @sbejaoui I'll close this PR and Open in OCA/brand repository.

PR #521, opened 2019-10-25
[12.0][FIX]  product_supplierinfo_for_customer: assure you get partner id

MiquelRForgeFlow:
Minor fix. Please, fast-tracking.

### DIFF
--- a/product_supplierinfo_for_customer/models/product_product.py
+++ b/product_supplierinfo_for_customer/models/product_product.py
@@ -27,12 +27,14 @@ class ProductProduct(models.Model):
     def price_compute(
             self, price_type, uom=False, currency=False, company=False):
         if price_type == 'partner':
-            partner = self.env.context.get('partner_id', False) or \
+            partner_id = self.env.context.get('partner_id', False) or \
                 self.env.context.get('partner', False)
+            if partner_id and isinstance(partner_id, models.BaseModel):
+                partner_id = partner_id.id
             prices = super(ProductProduct, self).price_compute(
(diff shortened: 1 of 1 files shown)


product_supplierinfo_for_customer/models/product_product.py
```diff
@@ -29,6 +29,8 @@ def price_compute(
         if price_type == 'partner':
             partner = self.env.context.get('partner_id', False) or \
                 self.env.context.get('partner', False)
+            if partner and not isinstance(partner, (int,)):
```
pedrobaeza: Instead of looking for not being integer, better to check if is `BaseModel`

MiquelRForgeFlow: done!

product_supplierinfo_for_customer/models/product_product.py
```diff
@@ -29,6 +29,8 @@ def price_compute(
         if price_type == 'partner':
             partner = self.env.context.get('partner_id', False) or \
                 self.env.context.get('partner', False)
+            if partner and isinstance(partner, models.BaseModel):
+                partner = partner.id
```
pedrobaeza: But if this is an ID, why the variable is called `partner`? It should be `partner_id` (I know that the code may not be yours, but we have now the opportunity to fix this).

MiquelRForgeFlow: done

pedrobaeza (approved):

PR #522, opened 2019-10-28
[12.0][FIX] product_supplierinfo_for_customer: name_get should also search for customerinfo

MiquelRForgeFlow:
This commit also includes `_name_search` customization from `product_supplierinfo_for_customer_sale`, because it makes sense to put in here instead of in the other module.


AaronHForgeFlow (approved):
It works fine. Searching products by supplierinfo is possible now, and also the displaying. :+1:

product_supplierinfo_for_customer/models/product_product.py
```diff
@@ -10,6 +10,125 @@
 class ProductProduct(models.Model):
     _inherit = 'product.product'
 
+    @api.multi
+    def name_get(self):
+
+        def _name_get(d):
```
pedrobaeza: Don't do nested methods that are not inheritable

product_supplierinfo_for_customer/models/product_product.py
```diff
@@ -10,6 +10,125 @@
...
+
+        # Prefetch the fields used by the `name_get`
+        self.sudo().read(
+            ['name', 'default_code', 'product_tmpl_id',
+             'attribute_value_ids', 'attribute_line_ids'], load=False)
+
+        product_template_ids = self.sudo().mapped('product_tmpl_id').ids
+
+        if partner_ids:
+            supplier_info = self.env['product.supplierinfo'].sudo().search([
```
pedrobaeza: You mustn't search for supplierinfo here. Better to pass a context here and on search of the supplierinfo, mix results of both supplierinfo + customerinfo.

pedrobaeza (changes requested):

product_supplierinfo_for_customer/models/product_product.py
MiquelRForgeFlow: but, it's the same as odoo method :S

see in https://github.com/OCA/OCB/blob/12.0/addons/product/models/product.py#L425

pedrobaeza: That doesn't mean it should be the best way. And what I don't like is that you make a verbatim copy of everything. See the trick I propose above.

product_supplierinfo_for_customer/models/product_product.py
MiquelRForgeFlow: odoo doesn't let you mix model records. It will say you "mixing orange with apples" message.

pedrobaeza: I know, but you only need to return an iterable, so deal with that doing a list of recordsets of one element.

MiquelRForgeFlow: no, in the odoo product `name_get` code, after doing the `search` in the supplierinfo model, it does a `read` on the result. If the result is a list as you propose, then it will break when trying to do the `read`. see: https://github.com/OCA/OCB/blob/12.0/addons/product/models/product.py#L458

pedrobaeza: OK, and what about returning new supplierinfo recordsets, but with customerinfo fields? Does the read works on them?

pedrobaeza: You might trick the `read` also if not.

product_supplierinfo_for_customer/models/product_supplierinfo.py
```diff
@@ -0,0 +1,37 @@
...
+    _inherit = "product.supplierinfo"
+
+    @api.model
+    def search(self, args, offset=0, limit=None, order=None, count=False):
+        res = super(ProductSupplierInfo, self).search(
+            args, offset=offset, limit=limit, order=order, count=count)
+        if self.env.context.get('customerinfo') and \
+                self._name == 'product.supplierinfo':
+            res2 = self.env['product.customerinfo'].search(
+                args, offset=offset, limit=limit, order=order, count=count)
```
pedrobaeza: You must adjust the limit for only returning remaining ones

product_supplierinfo_for_customer/models/product_product.py
```diff
@@ -10,6 +10,38 @@
...
     _inherit = 'product.product'
 
+    @api.multi
+    def name_get(self):
+        res = super(ProductProduct, self.with_context(customerinfo=True)).\
+            name_get()
+        return res
+
+    @api.model
+    def _name_search(self, name='', args=None, operator='ilike', limit=80):
```
pedrobaeza: Why this method doesn't follow a similar approach than the other?

pedrobaeza (changes requested):

product_supplierinfo_for_customer/models/product_product.py
MiquelRForgeFlow: because `name_get` uses `search`, but `_name_search` uses `_search` instead. Thus, it means I would need to also hack `_search`, but then my head would explode, trying to making compatible the hack of `search` with the hack of `_search` hahaha

MiquelRForgeFlow: `_search` returns ids, so here I cannot do same hack as done in `search`. See https://github.com/OCA/OCB/blob/12.0/addons/product/models/product.py#L532

AaronHForgeFlow:
descarga (1)

Apologies, couldn't help it.

pedrobaeza:
Let's merge it as it's good enough and I'm not able to review deeply

/ocabot merge patch

PR #353, opened 2018-05-28
[MIG][11.0] Migration of l10n_eu_product_adr_report to v11

BT-rmartin:

PR #523, opened 2019-11-01
[12.0] create product_dangerous_good

i-vyshnevska:
Create new module to add functionality to handle dangerous goods according to ADR standard 
https://www.unece.org/trans/danger/publi/adr/adr2011/11ContentsE.html


pedrobaeza:
I think this should go to OCA/community-data-files (there are already there some UNECE things - should it go with that name in the module?). There's also an ADR report that we developed for 8.0 in https://github.com/odoomrp/odoomrp-utils/tree/8.0/l10n_eu_product_adr_report that might be interesting for you, with a tentative migration to 11.0 in https://github.com/OCA/product-attribute/pull/353

i-vyshnevska:
@pedrobaeza  thanks for reference, some of mentioned addons have  much simpler logic, other pretty same, I think I can pick some logic from there

i-vyshnevska:
closed in favor https://github.com/OCA/community-data-files/pull/77

PR #524, opened 2019-11-05
[12.0][MIG] product_pricelist_revision: Migration to 12.0

ernestotejeda:
Cc @Tecnativa TT20175


product_pricelist_revision/models/pricelist.py
```diff
@@ -28,8 +29,21 @@ class ProductPricelistItem(models.Model):
 
     @api.model
     def _search_name(self, operator, value):
-        domain = []
-        return [('applied_on', '=', '3_global'),]
```
ernestotejeda: I have changed this because I think that as it was can produce unexpected results when searching by the name field.
What I have done here is perhaps slow for a large data input, but I suppose that in most cases there are not many objects of product.pricelist.item

product_pricelist_revision/models/pricelist.py
```diff
@@ -0,0 +1,55 @@
...
+    variation_percent = fields.Float(
+        compute='_compute_variation_percent',
+        store=True,
+        digits=dp.get_precision('Product Price'),
+        string='Variation %',
+    )
+
+    @api.model
+    def _search_name(self, operator, value):
+        def name_matching(item):
```
carlosdauden: Can you try this?
```suggestion
        return [
            '|', '|',
            ('categ_id', operator, value),
            ('product_tmpl_id', operator, value),
            ('product_id', operator, value),
        ]
```
It is not as "perfect" but it is more efficient.

Based on name generation method:
https://github.com/odoo/odoo/blob/35d4a8ccce8f3a34701d231e17a324cbbf5abfab/addons/product/models/product_pricelist.py#L480-L487

carlosdauden (approved):

product_pricelist_revision/README.rst
```diff
@@ -0,0 +1,93 @@
...
+.. |badge4| image:: https://img.shields.io/badge/weblate-Translate%20me-F47D42.png
+    :target: https://translation.odoo-community.org/projects/product-attribute-12-0/product-attribute-12-0-product_pricelist_revision
+    :alt: Translate me on Weblate
+.. |badge5| image:: https://img.shields.io/badge/runbot-Try%20me-875A7B.png
+    :target: https://runbot.odoo-community.org/runbot/135/12.0
+    :alt: Try me on Runbot
+
+|badge1| |badge2| |badge3| |badge4| |badge5| 
+
+This module allow create revisions of pricelist prices.
```
pedrobaeza:
```suggestion
This module allows to create revisions of pricelist prices.
```

product_pricelist_revision/README.rst
```diff
@@ -0,0 +1,93 @@
...
+
+.. contents::
+   :local:
+
+Usage
+=====
+
+To use this module, you need to:
+
+#. Go to *Sales > Products > Pricelists* and create a new Price List.
```
pedrobaeza, line 51: Add Configuration section saying that you need to enable "Multiple Sales Prices per Product" and select option "Prices computed from formulas (discounts, margins, roundings)" in Sales > Configuration > Settings.

product_pricelist_revision/README.rst
```diff
@@ -0,0 +1,93 @@
...
+#. Add a Pricelist Items to that Price List. Set 'Fix Price' value on the field
+   'Compute Price' and set a fixed price of 100.
```
pedrobaeza: Don't put an specific price, but give generic instructions:
```suggestion
   'Compute Price' and set a fixed price.
```

product_pricelist_revision/README.rst
```diff
@@ -0,0 +1,93 @@
...
+#. Go to the new menu *Sales > Products > Pricelist Items*, select the
+   previously created item and click on *Action > Duplicate Item*.
+#. In the wizard set 'Date Start', set 'Date End' and set 'Variation %' to
+   '50'.
```
pedrobaeza:
```suggestion
  a value different from 100.
```

product_pricelist_revision/README.rst
```diff
@@ -0,0 +1,93 @@
...
+#. Click on 'Apply' and you will see a new Price List Item that is a copy of
+   the previous item. In that new Item the value of the 'Fixed Price' field
+   will be 150 (50% more than the Fixed Price of the previous Item) and the
```
pedrobaeza:
```suggestion
   will be the previous price plus that price multiplied by the variation percentage and the value of the 'Variation %' will contain that value.
```

product_pricelist_revision/__manifest__.py
```diff
@@ -0,0 +1,20 @@
+# Copyright 2018 Carlos Dauden - Tecnativa
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+{
+    "name": "Product Pricelist Revision",
+    "summary": "Product Pricelist Revision",
+    "version": "12.0.1.0.0",
+    "category": "Product",
+    "website": "https://www.github.com/OCA/product-attribute",
+    "author": "Tecnativa S.L., Odoo Community Association (OCA)",
```
pedrobaeza:
```suggestion
    "author": "Tecnativa, Odoo Community Association (OCA)",
```

product_pricelist_revision/models/pricelist.py
pedrobaeza: But why do we need to overwrite this?

product_pricelist_revision/views/pricelist_view.xml
```diff
@@ -0,0 +1,73 @@
...
+            </field>
+            <field name="min_quantity" position="after">
+                <field name="compute_price" readonly="1"/>
+                <field name="fixed_price"/>
+                <field name="variation_percent"/>
+            </field>
+        </field>
+    </record>
+
+    <record id="product_pricelist_item_view_search" model="ir.ui.view">
```
pedrobaeza, line 36: Does this belong to the module feature?

pedrobaeza (changes requested):
My feeling here is that there are 2 features in this module: the revision one, which is OK with some comments, and a hidden one that is listing pricelist items, that must go into a separate module. @carlosdauden why is this in the module?

ernestotejeda:
This module adds a Tree view of 'Pricelist items' so that users can select one or more 'Pricelist items' and then create copies of them with a variation in their 'fixed price', so maybe there are not 2 features in this module. The problem here is that the search view of 'Pricelist items' defined here includes the 'name' field which is a computed field. As it is a computed field we have specify a method in the search argument in the field definition https://github.com/OCA/product-attribute/blob/825bb3749f94880ca0514ac6f66c7046fe742bb3/product_pricelist_revision/models/pricelist.py#L11

pedrobaeza:
OK, I see. Then that should be explained better in the README. Please add also coverage for the uncovered parts like that search.

product_pricelist_revision/models/pricelist.py
carlosdauden: Name is a computed field without search:
https://github.com/odoo/odoo/blob/35d4a8ccce8f3a34701d231e17a324cbbf5abfab/addons/product/models/product_pricelist.py#L457-L459

ernestotejeda: > Can you try this?
> 
> It is not as "perfect" but it is more efficient.
> 
> Based on name generation method:
> https://github.com/odoo/odoo/blob/35d4a8ccce8f3a34701d231e17a324cbbf5abfab/addons/product/models/product_pricelist.py#L480-L487

I think it could be a good balance, do you agree @pedrobaeza ?

pedrobaeza: Yeah, it's OK.

pedrobaeza (approved):

PR #531, opened 2019-11-28
[12.0] Add product_packaging_dimension

grindtildeath:
This module allows to store dimensions (length, width, height), weight and
volume of product packagings.

### DIFF
--- a/product_packaging_dimension/__manifest__.py
+++ b/product_packaging_dimension/__manifest__.py
@@ -1,0 +1,18 @@
+# Copyright 2019 Camptocamp SA
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl)
+{
+    'name': 'Product Packaging Dimension',
+    'summary': 'Manage packaging dimensions and weight',
+    'version': '12.0.1.0.0',
+    'category': 'Product',
+    'author': 'Camptocamp, Odoo Community Association (OCA)',
+    'license': 'AGPL-3',
+    'depends': [
+        'product',
+    ],
+    'website': 'https://github.com/OCA/product-attribute',
+    'data': [
+        'views/product_packaging.xml',
+    ],
+    'installable': True,
+}
--- a/product_packaging_dimension/models/product_packaging.py
+++ b/product_packaging_dimension/models/product_packaging.py
@@ -1,0 +1,28 @@
+# Copyright 2019 Camptocamp SA
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl)
+from odoo import api, fields, models
+
+
+class ProductPackaging(models.Model):
+    _inherit = "product.packaging"
+
+    max_weight = fields.Float('Weight (kg)')
+    # lngth IS NOT A TYPO https://github.com/odoo/odoo/issues/41353
+    lngth = fields.Integer("Length (mm)", help="length in millimeters")
+    width = fields.Integer('Width (mm)', help='width in millimeters')
+    height = fields.Integer('Height (mm)', help='height in millimeters')
+    volume = fields.Float(
+        'Volume (m³)',
+        digits=(8, 4),
+        compute='_compute_volume',
+        readonly=True,
+        store=False,
+        help='volume in cubic meters',
+    )
+
+    @api.depends('lngth', 'width', 'height')
+    def _compute_volume(self):
+        for pack in self:
+            pack.volume = (
+                pack.lngth * pack.width * pack.height
+            ) / 1000.0 ** 3
--- a/product_packaging_dimension/__init__.py
+++ b/product_packaging_dimension/__init__.py
@@ -1,0 +1,1 @@
+from . import models
--- a/product_packaging_dimension/models/__init__.py
+++ b/product_packaging_dimension/models/__init__.py
@@ -1,0 +1,1 @@
+from . import product_packaging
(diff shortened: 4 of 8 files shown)


product_packaging_dimension/models/product_packaging.py
```diff
@@ -0,0 +1,27 @@
+# Copyright 2019 Camptocamp SA
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl)
+from odoo import api, fields, models
+
+
+class ProductPackaging(models.Model):
+    _inherit = "product.packaging"
+
+    max_weight = fields.Float()
```
Tonow-c2c: is't not better to add one string ?

product_packaging_dimension/models/product_packaging.py
```diff
@@ -0,0 +1,27 @@
...
+        digits=(8, 4),
+        compute='_compute_volume',
+        readonly=True,
+        store=False,
+        help='volume in cubic meters',
+    )
+
+    @api.depends('length', 'width', 'height')
+    def _compute_volume(self):
+        for pack in self:
```
Tonow-c2c, line 25: add a warning this compute work only if we stay with `'length', 'width', 'height'` in mm :wink: 

and I will now why we use length ... in mm and volumen not in mm ?

product_packaging_dimension/__manifest__.py
```diff
@@ -0,0 +1,19 @@
+# Copyright 2019 Camptocamp SA
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl)
+{
+    'name': 'Product Packaging Dimension',
+    'summary': 'Manage packaging dimensions and weight',
+    'version': '12.0.1.0.0',
+    'category': 'Product',
+    'author': 'Camptocamp, Odoo Community Association (OCA)',
+    'license': 'AGPL-3',
+    'conflicts': ['delivery'],
```
grindtildeath: To drop (conflict solved by same `fields.Integer` instead of `fields.Float`)

jgrandguillaume (approved):
LGMT

product_packaging_dimension/models/product_packaging.py
```diff
@@ -0,0 +1,27 @@
+# Copyright 2019 Camptocamp SA
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl)
+from odoo import api, fields, models
+
+
+class ProductPackaging(models.Model):
+    _inherit = "product.packaging"
+
+    weight = fields.Float('Weight (kg)')
+    length = fields.Integer('Length (mm)', help='length in millimeters')
```
grindtildeath: :warning: https://github.com/odoo/odoo/issues/41353

product_packaging_dimension/models/product_packaging.py
```diff
@@ -6,7 +6,7 @@
 class ProductPackaging(models.Model):
     _inherit = "product.packaging"
 
-    max_weight = fields.Float()
+    weight = fields.Float()
```
grindtildeath: :warning: We might have to drop commits 4d877ab and 6bfbea7 because `max_weight` is defined in `delivery`: https://github.com/odoo/odoo/blob/12.0/addons/delivery/models/product_packaging.py#L13

product_packaging_dimension/models/product_packaging.py
```diff
@@ -0,0 +1,27 @@
+# Copyright 2019 Camptocamp SA
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl)
+from odoo import api, fields, models
+
+
+class ProductPackaging(models.Model):
+    _inherit = "product.packaging"
+
+    weight = fields.Float('Weight (kg)')
```
grindtildeath: :warning: We might have to drop commits 4d877ab and 6bfbea7 because `max_weight` is defined in `delivery`: https://github.com/odoo/odoo/blob/12.0/addons/delivery/models/product_packaging.py#L13

gurneyalex: AFAIK having 2 modules defining the same field is not really a problem.

product_packaging_dimension/models/product_packaging.py
rousseldenis: Following https://en.wikipedia.org/wiki/International_System_of_Units, I would say meters for length and m3 for volume.

grindtildeath: The issue in using meters is that we'd need a Float field type as meter is too big of a uom for parcel dimensions.
However odoo already defines these fields as Integer in `delivery` !
As there is no functional dependency between this module and `delivery` we must use the same field type in order to avoid conflicts...

hparfr: What are the drawbacks using float with at least 4 decimals ?

hparfr: I think it's better to not enforce any UoM because it depends too much of the business cases or even locales (non-metric systems). Adding a length_uom_id can solve this debate. 


Please also have a look on this PR https://github.com/odoo/odoo/pull/50252 to make these uom_id first class citizen.

grindtildeath: @hparfr It's been a while I had read your comments and forgot about it until I got back to having to handle uom for weight and volume. 
To me, your PR on odoo looks like the right direction, and we should probably update our modules to consider at least the `ir.config_parameter` instead of having uom in string descriptions.

rvalyi:
Hello @jgrandguillaume  I think it has been closed by mistake  It was merged in v13 but still not in v12...

rvalyi:
Nope my bad I switched to the wrong PR by mistake. This one should really stay closed, it is superseded by https://github.com/OCA/product-attribute/pull/593

PR #525, opened 2019-11-07
[ADD] module 'product_packaging_type' and 'product_packaging_type_required'

damdam-s:

PR #536, opened 2019-12-13
[13.0] Migrate product_packaging_type to 13.0

guewen:
Port of #525

### DIFF
--- a/product_packaging_type/models/product_packaging_type.py
+++ b/product_packaging_type/models/product_packaging_type.py
@@ -1,12 +1,13 @@
 # Copyright 2019 Camptocamp (<http://www.camptocamp.com>).
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html)
 
-from odoo import fields, models
+from odoo import api, fields, models
 
 
 class ProductPackagingType(models.Model):
     _name = "product.packaging.type"
     _description = "Type management for product.packaging"
+    _order = 'sequence, code'
 
     name = fields.Char(required=True, translate=True)
     code = fields.Char(required=True)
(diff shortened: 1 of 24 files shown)


gurneyalex (approved):

simahawk (approved):

emagdalenaC2i:
Is it a migration or a new module?
I did not find the previous product_packaging_type

product_packaging_type/models/product_packaging_type.py
```diff
@@ -0,0 +1,92 @@
...
+            msg = _('There must be one product packaging type set as "Is Default".')
+        elif default_count > 1:
+            msg = _('Only one product packaging type can be set as "Is Default".')
+        if msg:
+            raise ValidationError(msg)
+
+
+class ProductPackaging(models.Model):
+    _inherit = "product.packaging"
+    _order = "product_id, type_sequence"
```
jbaudoux, line 35: I would order by default based on the contained quantity
```suggestion
    _order = "product_id, qty, type_sequence"
```

guewen: It makes sense, but then I wonder if the `type_sequence` is really useful?

guewen:
@emagdalenaC2i link for the previous version is in the PR's description (it's an unmerged PR).

product_packaging_type/models/product_packaging_type.py
simahawk: @jbaudoux ping :)

rousseldenis:
@guewen @simahawk Beta/Alpha module does not mean no tests, no convention...

Could you add tests and split models per file please ? Thanks

PR #540, opened 2019-12-19
[12.0][FIX] product_sequence: Keep default_code when copy products if dictionary contains the default_code key.

sergio-teruel:
cc @Tecnativa

### DIFF
--- a/product_sequence/tests/test_product_sequence.py
+++ b/product_sequence/tests/test_product_sequence.py
@@ -103,3 +103,13 @@ class TestProductSequence(TransactionCase):
         categ_car.sequence_id = False
         categ_car.write({'code_prefix': 'KIA'})
         self.assertEqual(categ_car.sequence_id.prefix, "KIA")
+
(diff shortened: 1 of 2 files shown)


product_sequence/tests/test_product_sequence.py
```diff
@@ -1,10 +1,11 @@
 # Copyright 2016 Sodexis
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
 
-from odoo.tests.common import TransactionCase
+from odoo.tests.common import TransactionCase, tagged
 from ..hooks import pre_init_hook
 
 
+@tagged('post_install', '-at_install')
```
pedrobaeza: Why do you need this?

pedrobaeza (approved):

product_sequence/tests/test_product_sequence.py
sergio-teruel: fixed!

Tardo (approved):

PR #544, opened 2019-12-23
[12.0][ADD] product_pricelist_form

dalonsod:
This module allows full pricelist management for products and variants from
their own forms, including percentage and formula computation types, the same
way that pricelist item form does


dalonsod:
@jgrandguillaume could you take a look at this? I'm not sure if this improvement should be a module, or directly an Odoo modification.

Thanks!

rousseldenis:
@dalonsod Maybe you could add some screenshots to help understanding what this module does? Moreover, a more 'functional' name could be better. Like product_pricelist_form, or... (full does not make sense).

Could you also avoid 'replace' in views in order to not break inheritance ?

dalonsod:
> 
> 
> @dalonsod Maybe you could add some screenshots to help understanding what this module does? Moreover, a more 'functional' name could be better. Like product_pricelist_form, or... (full does not make sense).
> 
> Could you also avoid 'replace' in views in order to not break inheritance ?

Thanks for the review @rousseldenis, I agree with your comments. 
- I'll include some screeenshots in `README` for better understanding.
- No problem changing addon name. As you suggested, `product_pricelist_form` should be a good option.
- I'm not sure if the replace use can be avoided, but I'll try.

dalonsod:
@rousseldenis addon name changed, `USAGE` section added with more information and screenshots.

Unfortunately, I'm not able to avoid replace usage within a view. The goal is disabling an inline form definition for the pricelist `item_ids` of a product template, and using an already defined form view instead. Anyway, I put a very low priority for the affected view. Maybe it's possible to make a different approach to get the same bevahior. Any ideas? Thanks!

dalonsod:
> Unfortunately, I'm not able to avoid replace usage within a view. The goal is disabling an inline form definition for the pricelist `item_ids` of a product template, and using an already defined form view instead. Anyway, I put a very low priority for the affected view. Maybe it's possible to make a different approach to get the same bevahior. Any ideas? Thanks!

@rousseldenis could you take a look at this? Or the priority set might be enough for this? Thanks!!!

PR #549, opened 2020-01-07
[ADD] product_taxes

p-tombez:
Add module which allow to specify additional taxes per variant (e.g. variant with/without battery will not have the same taxes)


pedrobaeza:
Several comments:

- Shouldn't this go to OCA/product-variant?
- Name shouldn't contain plurals and be significative. I suggest `product_variant_specific_tax`.

p-tombez:
Closed in favor of https://github.com/OCA/product-variant/pull/161

PR #551, opened 2020-01-08
[13.0][MIG] product_packaging_dimension

grindtildeath:
Forward port of #531

### DIFF
--- a/product_packaging_dimension/models/product_packaging.py
+++ b/product_packaging_dimension/models/product_packaging.py
@@ -6,23 +6,21 @@ from odoo import api, fields, models
 class ProductPackaging(models.Model):
     _inherit = "product.packaging"
 
-    max_weight = fields.Float('Weight (kg)')
+    max_weight = fields.Float("Weight (kg)")
     # FIXME https://github.com/odoo/odoo/issues/41353
-    length = fields.Integer('Length (mm)', help='length in millimeters')
-    width = fields.Integer('Width (mm)', help='width in millimeters')
-    height = fields.Integer('Height (mm)', help='height in millimeters')
+    length = fields.Integer("Length (mm)", help="length in millimeters")
+    width = fields.Integer("Width (mm)", help="width in millimeters")
+    height = fields.Integer("Height (mm)", help="height in millimeters")
     volume = fields.Float(
-        'Volume (m³)',
+        "Volume (m³)",
         digits=(8, 4),
-        compute='_compute_volume',
+        compute="_compute_volume",
         readonly=True,
         store=False,
-        help='volume in cubic meters',
+        help="volume in cubic meters",
     )
 
-    @api.depends('length', 'width', 'height')
+    @api.depends("length", "width", "height")
     def _compute_volume(self):
         for pack in self:
-            pack.volume = (
-                pack.length * pack.width * pack.height
-            ) / 1000.0 ** 3
+            pack.volume = (pack.length * pack.width * pack.height) / 1000.0 ** 3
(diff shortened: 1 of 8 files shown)


product_packaging_dimension/models/product_packaging.py
```diff
@@ -0,0 +1,26 @@
...
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl)
+from odoo import api, fields, models
+
+
+class ProductPackaging(models.Model):
+    _inherit = "product.packaging"
+
+    max_weight = fields.Float("Weight (kg)")
+    # lngth IS NOT A TYPO https://github.com/odoo/odoo/issues/41353
+    lngth = fields.Integer("Length (mm)", help="length in millimeters")
```
Tonow-c2c, line 11:
```suggestion
    pack_length = fields.Integer("Length (mm)", help="length in millimeters")
```
and idem for other variable to be coherent

product_packaging_dimension/models/product_packaging.py
```diff
@@ -0,0 +1,26 @@
+# Copyright 2019 Camptocamp SA
```
Tonow-c2c:
```suggestion
# Copyright 2020 Camptocamp SA
```

idem for other

product_packaging_dimension/models/product_packaging.py
```diff
@@ -0,0 +1,26 @@
...
+
+
+class ProductPackaging(models.Model):
+    _inherit = "product.packaging"
+
+    max_weight = fields.Float("Weight (kg)")
+    # lngth IS NOT A TYPO https://github.com/odoo/odoo/issues/41353
+    lngth = fields.Integer("Length (mm)", help="length in millimeters")
+    width = fields.Integer("Width (mm)", help="width in millimeters")
+    height = fields.Integer("Height (mm)", help="height in millimeters")
```
Tonow-c2c, line 15: I am OK for https://github.com/OCA/product-attribute/pull/531#discussion_r355553682 but can you add docsting to prevent unit change and explain the choice :wink:

Tonow-c2c (commented):
comment idem as comment idem as https://github.com/OCA/product-attribute/pull/551/files#diff-7204ef8da825976a99b3672ada1c75c8

product_packaging_dimension/models/product_packaging.py
grindtildeath: Here we don't want to use `pack_height` or `pack_width` because we want to redefine the fields from `delivery` module. So it makes more sense to simply use `lngth` instead of `length` as explained in the linked issue.

product_packaging_dimension/models/product_packaging.py
grindtildeath: Done in 14a0a32

product_packaging_dimension/models/product_packaging.py
simahawk: what about this https://github.com/OCA/OCB/commit/25616b9f2bbca3f03dbde1dc81a57986241c1aab ?

simahawk:
@grindtildeath what about this OCA/OCB@25616b9 ?

grindtildeath:
@simahawk The commit is from 2016 so it seems the issues is always the same.
IMO, let's not rely on a modified lib and use a field name that prevents any JS issue.

guewen:
> what about this [OCA/OCB@25616b9](https://github.com/OCA/OCB/commit/25616b9) ?

AFAIS this is a different part of the code, the issue reported in odoo/odoo#41353 being about [onchanges](https://github.com/odoo/odoo/blob/bae07bce4ad375c9eca71663a8cc6996f402a6d7/addons/web/static/src/js/views/basic/basic_model.js#L1594) and the correction above is in `reload_record`.

guewen (approved):

simahawk (approved):
@grindtildeath please squash :)

PR #559, opened 2020-01-29
[IMP] product_supplierinfo_for_customer: Clean old stored action domain

sergio-teruel:
cc @Tecnativa TT21755
@chienandalu @Tardo review please


pedrobaeza (approved):
Fast-tracking as this is for migration

PR #565, opened 2020-02-05
[10.0][ADD] product_description_form

rousseldenis:


rousseldenis:
This adds the 'description' field on form view.

This will be dead for versions >= 12 as Odoo reintroduced it.

PR #566, opened 2020-02-05
[10.0][FIX] travis: Update to latest template

rousseldenis:


rousseldenis:
@pedrobaeza This seems to solve travis issues

pedrobaeza (approved):
I'm not anymore OCA's community manager for being the wildcard for this kind of things, but OK to do it at least once more :)

rousseldenis:
> I'm not anymore OCA's community manager for being the wildcard for this kind of things, but OK to do it at least once more :)

That's because for all those kind of PR's you had used to be the first on it.

Copy that. I will ping the PSC next time. Thanks !

PR #578, opened 2020-02-25
[MIG] base_product_mass_addition: Migration to 13.0

phuctranfxvn:


HviorForgeFlow:
@phuctranerp I think a rebase is required

> git fetch origin 13.0 # (OCA remote)
> git rebase origin/13.0
> git push phuctranerp -f

phuctranfxvn:
Hi @HviorForgeFlow done

HviorForgeFlow (approved):

phuctranfxvn:
any news on this MR @HviorForgeFlow ?

AdriaGForgeFlow (approved):
Code review, no major changes

AaronHForgeFlow (approved):
Code review :+1:

MiquelRForgeFlow (approved):

HviorForgeFlow:
@OCA/product-maintainers can we move this forward? Thanks!

base_product_mass_addition/models/product_product.py
```diff
@@ -0,0 +1,46 @@
...
+            parent = self.env[parent_model].browse(parent_id)
+            for product in self:
+                quick_line = parent._get_quick_line(product)
+                if quick_line:
+                    parent._update_quick_line(product, quick_line)
+                else:
+                    parent._add_quick_line(product, quick_line._name)
+
+    def _compute_process_qty(self):
+        if not self.env.context.get("parent_id"):
```
rousseldenis: Isn't it possible to get this with a parameter instead of context ?

PierrickBrun: It is called from `compute=` on a field, I don't think this would be possible (or I missed something ?)

dalonsod: @phuctranerp I was trying to migrate `stock_picking_quick` using this branch for the needed dependency in order to test it, and I've found that this method causes an error. `stock_picking_quick` already implements its own `_compute_process_qty` method, and even if inside it `qty_to_process` is always filled, an error is fired. Only if `_compute_process_qty` code in `base_product_mass_additions` is changed in order to write an initial value in `qty_to_process` for every record the error is not fired.

Since the computed field is not stored, every implementation of computed method should assign a value for every record (v13 requirement), even for the `super()` method, I think.

dalonsod: You can test it using https://github.com/solvosci/stock-logistics-workflow/tree/13.0-mig-stock_picking_quick

phuctranfxvn: Hello @dalonsod
Thanks for your feedbacks, I think the implementation of this method is originally not a "computation" method, so the author didnt implement the computation method. This allows to trigger the _inverse_set_process_qty method only. The migration is v12 as-is. Do you have any recommendation ?

dalonsod: In case of `stock_picking_quick` `_compute_process_qty` is finally used. Making this change in `base_product_mass_addition` method worked for me (making a trivial assignment):

```suggestion
        for product in self:
                product.qty_to_process = 0
```
I don't really know if this change could fire an unexpected behavior

kittiu:
@phucngta will v14 helps? https://github.com/OCA/product-attribute/pull/755

hailangvn:
Good day @kittiu. I will work on it.

hailangvn:
Good day @dalonsod, thanks for your comments. I updated this PR to clarify about `_compute_process_qty` and verified with your migration. Please help to not call `super(...)._compute_process_qty(...)` as there is no logic without knowledge about parent_model and parent_id.
Good day @kittiu, please help to review and let me know if anything need to be updated.

base_product_mass_addition/__manifest__.py
```diff
@@ -0,0 +1,15 @@
...
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+
+{
+    "name": "Base Product Mass Addition",
+    "version": "13.0.1.0.0",
+    "author": "Akretion, Odoo Community Association (OCA)",
+    "website": "https://github.com/OCA/product-attribute",
+    "license": "AGPL-3",
+    "category": "Product",
+    "depends": ["stock", "onchange_helper"],
```
rousseldenis, line 12: IMHO, you should depends on product only as people maybe want to use it having only sale installed for instance (and don't want to install stock one). And on the same idea, adds the qty_available related fields in your example 'stock_picking_quick'

hailangvn: Thanks @rousseldenis for your review. We will have separated PR to fix in 13.0 then port it to 14.0.
@dalonsod, I updated this PR again so that you can keep super(...) logic so please ignore my previous comment. Sorry for the confusion. And, please note about rousseldenis's comment.
> And on the same idea, adds the qty_available related fields in your example 'stock_picking_quick'

PR #580, opened 2020-02-26
[12.0] FIX product_pricelist_supplierinfo: no_supplierinfo_min_quantity should not influence supplierinfo order + Supplier filter

eLBati:
Steps:

 - Create 2 supplierinfo for 1 product, only setting their price
 - Create a pricelist "based on supplier info"
 - Create a sale order using that pricelist and selecting that product

The last supplierinfo is used, instead of the first one.

In general, no_supplierinfo_min_quantity should only influence the domain, because the correct order is already handled by default: 'sequence, min_qty desc, price'

**New funzionality**

Allow to only match prices from selected supplier

### DIFF
--- a/product_pricelist_supplierinfo/tests/test_product_supplierinfo.py
+++ b/product_pricelist_supplierinfo/tests/test_product_supplierinfo.py
@@ -38,7 +38,6 @@ class TestProductSupplierinfo(common.SavepointCase):
         })
         cls.pricelist = cls.env['product.pricelist'].create({
             'name': 'Supplierinfo Pricelist',
-            'discount_policy': 'without_discount',
             'item_ids': [
                 (0, 0, {
                     'compute_price': 'formula',
@@ -98,6 +97,15 @@ class TestProductSupplierinfo(common.SavepointCase):
         self.pricelist.item_ids[0].no_supplierinfo_min_quantity = True
         self.assertAlmostEqual(
             self.pricelist.get_product_price(self.product, 1, False), 50,
+        )
+
+    def test_pricelist_supplier_filter(self):
+        self.assertAlmostEqual(
+            self.pricelist.get_product_price(self.product, 5, False), 50,
+        )
+        self.pricelist.item_ids[0].filter_supplier_id = self.supplier2.id
+        self.assertAlmostEqual(
+            self.pricelist.get_product_price(self.product, 5, False), 10,
         )
 
     def test_pricelist_dates(self):
--- a/product_pricelist_supplierinfo/tests/test_product_supplierinfo.py
+++ b/product_pricelist_supplierinfo/tests/test_product_supplierinfo.py
@@ -26,26 +26,16 @@ class TestProductSupplierinfo(common.SavepointCase):
                     'name': cls.supplier1.id,
                     'min_qty': 5,
                     'price': 50,
+                    'sequence': 1,
                 }),
                 (0, 0, {
                     'name': cls.supplier2.id,
                     'min_qty': 1,
                     'price': 10,
+                    'sequence': 2,
                 }),
             ],
         })
-        cls.product.write({'seller_ids': [
-            (0, 0, {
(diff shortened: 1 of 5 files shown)


pedrobaeza:
That was done on purpose, and you can see by the Travis fault why in the test that exercises this.

eLBati:
Hi @pedrobaeza , I also fixed the tests.

Could you explain the rationale behind the previous implementation?
Thanks

pedrobaeza:
Isn't there a flag for selecting if taking into account minimum quantity or not?

eLBati:
Yes it is: `no_supplierinfo_min_quantity`. In the implementation of this PR, it is just used to exclude prices according to the quantity.

In the previous implementation, it is also used to change the order of the results, according to a logic that seems incorrect to me.

pedrobaeza:
I repeat: the test case is clear for exercising why that's done that way, and the prove is that you have needed to change it for getting it correct, so :-1: to this change.

For your use case, just uncheck `no_supplierinfo_min_quantity`.

eLBati:
In the use case described in this PR, `no_supplierinfo_min_quantity` is not checked.

I had to change the test case because I think the functionality was wrongly implemented or designed.

Otherwise, we need an answer to https://github.com/OCA/product-attribute/pull/580#issuecomment-591524272

eLBati:
Please also note that the implementation of this PR follows the logic of odoo itself, implemented in the method `_select_seller`.
This is what the module should do: to compute the sale price of a product supposing to buy that product following the logic that will be indeed used to purchase it.

@pedrobaeza if you don't have time to analyze this or provide an answer for https://github.com/OCA/product-attribute/pull/580#issuecomment-591524272 , can anyone else do it? @carlosdauden ?

Thanks

mrcast (commented):
I tested the PR and I think that it's correct, because it follows the logic of Odoo
Without this PR I'm not able to find a logic
thanks

eLBati:
@simahawk hi, this needs review actually, could you update the label?

marcelsavegnago (approved):

product_pricelist_supplierinfo/tests/test_product_supplierinfo.py
```diff
@@ -26,26 +26,16 @@ def setUpClass(cls):
                     'name': cls.supplier1.id,
                     'min_qty': 5,
                     'price': 50,
+                    'sequence': 1,
```
pedrobaeza, line 29: Why do you need to add this sequence?

pedrobaeza (changes requested):

product_pricelist_supplierinfo/tests/test_product_supplierinfo.py
eLBati: In order to have an explicit order of `seller_ids` items

eLBati: As the price used is taken from the first supplierinfo found in `_get_supplierinfo_pricelist_price`

pedrobaeza: But if you add the lines on UI, do they have different sequence?

eLBati: They do: `<field name="sequence" widget="handle"/>`

eLBati:
Hello, added the functionality to filter  suppliers to get prices from

mrcast (approved):
Functional test ok

rousseldenis (approved):
LGTM

rousseldenis:
@pedrobaeza Could you update your review ?

eLBati:
@pedrobaeza ping

eLBati:
First issue seems resolved with https://github.com/OCA/product-attribute/pull/667

New functionality added with https://github.com/ruuter/product-attribute/pull/1

Closing this

PR #582, opened 2020-02-28
[FIX] Remove isort seed

yajo:
This is a test implementation of the hypothesis explained in https://github.com/OCA/maintainer-quality-tools/pull/629#issuecomment-588131706, which aims to avoid unnecessary and constant git conflicts in `.isort.cfg` file.

@Tecnativa @sergio-teruel

### DIFF
--- a/.isort.cfg
+++ b/.isort.cfg
@@ -9,4 +9,4 @@ line_length=88
 known_odoo=odoo
 known_odoo_addons=odoo.addons
 sections=FUTURE,STDLIB,THIRDPARTY,ODOO,ODOO_ADDONS,FIRSTPARTY,LOCALFOLDER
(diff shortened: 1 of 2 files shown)


sergio-teruel (approved):
Tested

pedrobaeza:
How can we test this is working? Any way, this should be done massively through the bot that spreads this across all OCA repositories if working.

yajo:
It's tested here.

You can see that the bots are :heavy_check_mark:. That means pre-commit is happy and all source files stay the same (so isort works exactly the same), but they don't need the `known_third_party=psycopg2,setuptools` part.

This PR proves this works.

pedrobaeza (approved):
Understood. I think this is going to be injected massively, but we can go ahead with it for not waiting that moment.

PR #587, opened 2020-03-07
[FIX][12.0] product_dimension rename length to product_length

emagdalenaC2i:
Fix travis warning
************* Module product_dimension.models.product
product_dimension/models/product.py:22: [W8105(attribute-deprecated), Product] attribute "length" deprecated
product_dimension/models/product.py:72: [W8105(attribute-deprecated), ProductTemplate] attribute "length" deprecated

### DIFF
--- a/product_dimension/models/product.py
+++ b/product_dimension/models/product.py
@@ -8,10 +8,12 @@ from odoo import models, fields, api
 class Product(models.Model):
     _inherit = 'product.product'
 
-    @api.onchange('length', 'height', 'width', 'dimensional_uom_id')
+    @api.onchange(
+        'product_length', 'product_height', 'product_width', 'dimensional_uom_id')
     def onchange_calculate_volume(self):
         self.volume = self.env['product.template']._calc_volume(
-            self.length, self.height, self.width, self.dimensional_uom_id)
+            self.product_length, self.product_height,
+            self.product_width, self.dimensional_uom_id)
 
     @api.model
     def _get_dimension_uom_domain(self):
@@ -19,9 +21,9 @@ class Product(models.Model):
             ('category_id', '=', self.env.ref('uom.uom_categ_length').id)
         ]
 
-    length = fields.Float()
-    height = fields.Float()
-    width = fields.Float()
+    product_length = fields.Float('length')
+    product_height = fields.Float('height')
+    product_width = fields.Float('width')
     dimensional_uom_id = fields.Many2one(
         'uom.uom',
         'Dimensional UoM',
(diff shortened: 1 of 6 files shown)


emagdalenaC2i:
Should be also renamed height --> product_height and width --> product_width?

sebastienbeau:
Hi Eduardo

Thanks for the fix indeed length is now deprecated (to avoid issue with javascript as length already mean something).
In order to be consistent I think we should also rename width and height

To avoid issue for existing customer we also need to migrate the data.
Here an example : https://github.com/OCA/manufacture/blob/10.0/mrp_production_request/migrations/10.0.1.0.0/pre-migration.py
Thanks for your help to clean the travis issue !

emagdalenaC2i:
I think it is ready to review.
Travis is red due to other module

product_dimension/models/product.py
```diff
@@ -8,25 +8,25 @@
...
     def _get_dimension_uom_domain(self):
         return [
             ('category_id', '=', self.env.ref('uom.uom_categ_length').id)
         ]
 
-    length = fields.Float()
-    height = fields.Float()
-    width = fields.Float()
+    product_length = fields.Float('length')
+    product_height = fields.Float()
```
sebastienbeau: Can you change the label here?

product_dimension/models/product.py
```diff
@@ -8,25 +8,25 @@
...
+    product_width = fields.Float()
     dimensional_uom_id = fields.Many2one(
         'uom.uom',
         'Dimensional UoM',
         domain=lambda self: self._get_dimension_uom_domain(),
-        help='UoM for length, height, width',
+        help='UoM for length, product_height, product_width',
```
sebastienbeau: no need of change here

product_dimension/migrations/12.0.1.0.0/pre-migration.py
```diff
@@ -0,0 +1,15 @@
+# Copyright 2019 C2i Change 2 improve - Eduardo Magdalena <emagdalena@c2i.es>
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+from openupgradelib import openupgrade
+
+field_renames = [
+    ('product.template.length', 'product.template.product_length',
```
sebastienbeau: invalid syntax
see : https://github.com/OCA/openupgradelib/blob/master/openupgradelib/openupgrade.py#L587

should be [
 ('product.template', 'product_template', 'length', 'product_length'),
 ('product.template', 'product_template', 'width', 'product_width'),
...

sebastienbeau (changes requested):

emagdalenaC2i:
Done

rvalyi (approved):

product_dimension/migrations/12.0.1.0.0/pre-migration.py
```diff
@@ -0,0 +1,15 @@
+# Copyright 2019 C2i Change 2 improve - Eduardo Magdalena <emagdalena@c2i.es>
```
sebastienbeau: The directory name that containt the file is wrong.
Right now is "12.0.1.0.0" it should be "12.0.2.0.0" it the number of the new version

emagdalenaC2i: ok, changed

renatonlima (approved):

PR #593, opened 2020-03-20
[12.0] Add product_packaging_dimension

clementmbr:
Re-opening #531 with the preventive fix made in the relative #551 forward port, using `lngth` variable instead of `length` to avoid conflicts with JS functions. (cf https://github.com/odoo/odoo/issues/41353 )

The #531 works even so, but it might be more secure to merge this one... do you agree @grindtildeath @p-tombez ?

Thank you!
Cc @rvalyi

### DIFF
--- a/product_packaging_dimension/__init__.py
+++ b/product_packaging_dimension/__init__.py
@@ -1,0 +1,1 @@
+from . import models
--- a/product_packaging_dimension/__manifest__.py
+++ b/product_packaging_dimension/__manifest__.py
@@ -1,0 +1,18 @@
+# Copyright 2019 Camptocamp SA
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl)
+{
+    'name': 'Product Packaging Dimension',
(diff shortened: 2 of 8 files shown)


grindtildeath:
@clementmbr I'd definitely favour the usage of `lngth` with the appropriate comment as it is in #551 .
However, since #531 isn't merged yet, the change must be done there (also because your PR doesn't reflect git history correctly).
So I just did it and squashed all the commits so that it can be merged as soon as there will be enough reviews :wink:

tonygalmiche (approved):
I tested on my computer locally
I checked the code
everything seems OK for me
Thank you for this work

but why the volume is in m3 and the dimensions in mm
I think it would be better to put the volume of dm3 (liter)

tonygalmiche:
I also just tested with runbot. No problem either.

rvalyi (approved):

rvalyi:
Few reviews but the same is merged in v13 and OK for us and people who tested so far, so merging...

PR #594, opened 2020-03-21
product_secondary_unit: Unit factor decimal precision

gaikaz:
Make factor field's decimal precision optionally configurable
by creating a Decimal Precision called "Secondary Unit Factor"

**Before this PR**
* `digits` attribute is not correct according to documentation (must be tuple of `(int, int)` or `str`)
* Factor field precision is non-configurable which by default ends-up being 2 decimal places

**After this PR**
* `digits` attribute is used correctly
* Factor field is **optionally** configurable with Decimal Precision _"Secondary Unit Factor"_:
  - If not set - nothing changes (defaults to 2 decimal places)
  - If set - field will display configured number decimal places where it is visible

### DIFF
--- a/product_secondary_unit/models/product_second_unit.py
+++ b/product_secondary_unit/models/product_second_unit.py
@@ -22,7 +22,10 @@ class ProductSecondaryUnit(models.Model):
         help="Default Secondary Unit of Measure.",
     )
     factor = fields.Float(
-        string="Secondary Unit Factor", default=1.0, digits=0, required=True
+        string="Secondary Unit Factor",
+        default=1.0,
+        digits="Secondary Unit Factor",
+        required=True,
     )
 
     def name_get(self):


product_secondary_unit/models/product_second_unit.py
```diff
@@ -22,7 +22,10 @@ class ProductSecondaryUnit(models.Model):
         help="Default Secondary Unit of Measure.",
     )
     factor = fields.Float(
-        string="Secondary Unit Factor", default=1.0, digits=0, required=True
+        string="Secondary Unit Factor",
+        default=1.0,
+        digits="Secondary Unit Factor",
```
sergio-teruel, line 27:
```suggestion
        digits="Product Unit of Measure",
```

sergio-teruel: Yo do not need create a new one...

sergio-teruel (changes requested):

product_secondary_unit/models/product_second_unit.py
gaikaz: I thought about it and looked at Odoo code - arguably this is a different thing.
`'Product Unit of Measure'` is used in a lot of places for product quantity precision.
This, on the other hand, is a conversion factor - which could easily be much more precise than what is needed for quantity fields in the whole system.

pedrobaeza: But the unit of measure digits limit the conversion field resolution as well. If you have 2 decimal places for UoM, and try 0.001 as conversion factor, this is not going to be precisely expressed.

gaikaz: That's a very good point.
You made me think about how the field is saved to a database and, to my horror, after this PR it's actually being just cut off (floored, I guess) at the specified precision, while before it was actually saved in any precision (even if not visible afterward in forms).
Now all the comments `# force NUMERIC with unlimited precision` make sense, where `digits=0` is also used. Strange, that this is not mentioned in the documentation for Float field.

The original idea of this PR is to allow a user to configure the visible precision for the factor field, as it really can, in specific use cases, be something like  `0.0624`.

How about, if I reimplement the same feature in XML views instead?

pedrobaeza: I still don't see why putting a different precision from the product UoM. What do you think @sergio-teruel ?

sergio-teruel: I agree with @pedrobaeza , sorry :cry:

gaikaz:
Closing this PR as it is kind of a breaking change in disguise:
Before this PR, the precision is unlimited (even if not visible in UI)
After this PR, the Odoo starts rounding the field up to provided precision

PR #601, opened 2020-04-09
[11.0][IMP] product_multi_price: Several improvements

pedrobaeza:
* Adjusted terminology
* Better positioning of the prices list
* Adjusted README
* Remove useless constraint
* Make multi price name translatable
* Some guidelines

cc @Tecnativa TT20979

### DIFF
--- a/product_multi_price/README.rst
+++ b/product_multi_price/README.rst
@@ -39,8 +39,7 @@ Configuration
 To configure multiple prices you need to set multi prices field names first.
 To do so, you need admin permissions. Then go to:
 
-#. *Settings > General Settings > Technical > Database Structure >
-   Price Field Names*
+#. *Settings > Technical > Database Structure > Price Field Names*
 #. Create the multi price fields you need.
 
 If you have multiple companies, you can assign independet field sets for each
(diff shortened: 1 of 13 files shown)


pedrobaeza:
Ready to be reviewed

chienandalu (commented):
I wonder if there should be a translatable description field, specially to show the prices on reports

pedrobaeza:
Yes, good idea. Make name translatable.

chienandalu:
I meant another field. Translating name, could lead to issues with the proxy field, that searches by that name, don't you think?

pedrobaeza:
Good question, and after meeting both, we have removed totally the compute proxy method.

chienandalu:
Looks like romoving the proxy field makes the product compute_price method fail as it's expected to have a field with the same name of the price rule when we make the first compute. Maybe we keep it as a dummy field?

pedrobaeza:
@chienandalu check https://github.com/OCA/product-attribute/compare/def30e7616f18982803360eee43a43e80dc78f97..103784460358825ec198a0c2a13b065300e52412 for a way of avoiding the field (mimicking the technique in product_pricelist_supplierinfo)

chienandalu:
Oh, great :)

Some lint errors due to code removal: https://travis-ci.org/github/OCA/product-attribute/jobs/675338331#L312-L314

pedrobaeza:
OK, let's see now

chienandalu (approved):
Ok now :)

PR #602, opened 2020-04-17
[FIX] product_multi_price: portal and public user error

chienandalu:
When a not internal user (portal/public) with a pricelist that
contains multi price rules tries to see a page with a product affected
for such rules, an exception raises due to lack of permissions over
product.multi.price model.

cc @Tecnativa TT20979


pedrobaeza (approved):
Totally! As this method is private and only specific pricelist code is accessing it, there's no possible information leak.

pedrobaeza:
Let's fast-track it for solving ASAP the problem:

/ocabot merge patch

PR #603, opened 2020-04-20
[10.0][ADD]product_category_list

MPoch-PlanetaTIC:
Permits filter Sales Report by product categories:

Permits to assign a list of categories to a product.
Then you can filter Sales Report by products of theses categories.

cc @PlanetaTIC


rousseldenis:
@MPoch-PlanetaTIC Have you ever seen this ? 
https://github.com/OCA/product-attribute/tree/10.0/product_multi_category

I think you could extend it if it does not include categories report.

FFernandez-PlanetaTIC:
Hi @rousseldenis ,

Thank you for your comment.

We wanted to make product lists but:

- It was not a good idea to reuse categories.
- Module name and description are not appropriate.

We close it.

Apologies

PR #604, opened 2020-04-23
13.0 mig product brand

filoquin:
[IMP] product_brand: migration to 13.0


pedrobaeza:
This is already migrated, but on OCA/brand repository. You should check migration issue to see that kind of things:

https://github.com/OCA/product-attribute/issues/508

PR #607, opened 2020-04-28
[12.0][ADD]  product_logistics_uom

hparfr:
It let's you define weight uom and volume uom product per product.


product_logistics_uom/models/product_template.py
```diff
@@ -0,0 +1,64 @@
...
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+
+from odoo import api, fields, models
+
+
+class ProductTemplate(models.Model):
+    _inherit = "product.template"
+
+    volume = fields.Float(
+        help="The volume of the contents, not including any packaging, etc."
```
legalsylvain, line 12: I thinkg you should also change the text for the ``product.product`` field, don't you think ? 

I mean, overloading this value https://github.com/odoo/odoo/blob/12.0/addons/product/models/product.py#L145

legalsylvain (commented):
HI @hpar, thanks for your contribution.

I think I understand the objective of this module.

However, If you change the meaning of the volume and the weight fields. (ie = product A : ``volume = 1`` means 1 liter ; product B : ``volume = 1`` means 1 mililiter), I guess it will be incompatible with all the other OCA modules.
For exemple (for ``volume`` field) : 
- ``purchase_landed_cost``, specially [this computation.](https://github.com/OCA/purchase-workflow/blob/12.0/purchase_landed_cost/models/purchase_cost_distribution.py#L343)
- ``product_dimension``, with this volume computation [function](https://github.com/OCA/product-attribute/blob/12.0/product_dimension/models/product.py#L53)

Maybe other OCA modules are using weight field. (I didn't searched)

So, what is the road map ? Making all the OCA modules depends on your new module ? or mark your module as incompatible with all the OCA modules that are using the two fields (``weight``, ``volume``) ? 

Thanks !

pedrobaeza:
I think the solution can be to have extra additional fields for the weight/volume in other UoM, and do the inverse to the standard UoM when changing them. That way, you have the best of both worlds, and can use the "secondary weight/volume" when needed, but also keeping compatibility. The only problem can be decimal precision.

What do you think?

legalsylvain:
> [...] That way, you have the best of both worlds [...]

Nice proposal, indeed !

product_logistics_uom/models/product_template.py
hparfr: > However, If you change the meaning of the volume and the weight fields. (ie = product A : `volume = 1` means 1 liter ; product B : `volume = 1` means 1 mililiter), I guess it will be incompatible with all the other OCA modules.

Yes, it's not that simple. Some points:

1) in core v12, there is a weight_uom_id. Most of the modules don't take it in account, and it's an issue by itself. Because you can choose to switch to imperial units and these modules will provide wrong results.
2) the volume is hardcoded to m³ in v12 but not in v13. (in v13 the switch to different system will still be a problem).
3) some modules just don't care about the unit, **since it's always the same**. A sum of weight if they are both in the same unit is straightforward.

So, some other OCA modules may be patched to use weight_uom_id and be compliant with this module even without any dependency.
It will not work for volume.

But the thing is that in v13 actually, weight_uom_id has been removed.
I have suggested to odoo/odoo this PR https://github.com/odoo/odoo/pull/50252 yesterday and I hope _uom_id to be back as first class citizen but we do not know how it will end.

> So, what is the road map ? Making all the OCA modules depends on your new module ? or mark your module as incompatible with all the OCA modules that are using the two fields (`weight`, `volume`) ?

So for the future,  I see different paths:
1) odoo accept my PR, so OCA modules can be updated to depend of *_uom_id in their calculation. It will be little work for everyone but no change in dependencies. This module will be still relevant for the use case where you need different uom between products.
2) odoo reject my PR,  but within OCA we agree on the use of *_uom_id, so this module becomes a dependency. Little work for everyone.
3) no adhesion of the use of _uom_id, there can be some incompatibilities between oca modules.


To recap, this module does 3 things:
1) add [product.template]volume_uom_id  to be in pair with [product.template]weight_uom_id
2) allow to change default  weight_uom_id and volume_uom_id
3) allow to change product per product the right *_uom_id 

Note: in v12 there is no m³ in uom/data only liters.

hparfr:
> The only problem can be decimal precision.

It is. Btw, that's why I started  working on this module.

pedrobaeza:
OK, for now let's keep it then with the corresponding warnings and let's see how odoo part evolves.

mourad-ehm (approved):

PierrickBrun (approved):

alexis-via:
Nice proposal!

OCA/intrastat/intrastat_product is an example of a module using weight "directly". I wasn't aware of the native field weight_uom_id introduced in v12... we should update our intrastat code !

product_logistics_uom/readme/CONFIGURE.rst
```diff
@@ -0,0 +1,8 @@
+To change the default UoM
+
+#. Go "Genral Settings", then in "Products"
```
bealdav:
```suggestion
#. Go "General Settings", then in "Products"
```

bealdav (approved):
LGTM
I think problem raised by @legalsylvain have to be fixed in each module and shouldn't prevent to merge this module. But Ok for a grep in oca ecosystem must be achieve to fill excluded key in manifest of this one

product_logistics_uom/models/product_template.py
```diff
@@ -0,0 +1,64 @@
...
+        default=lambda x: x._get_volume_uom_id_from_ir_config_parameter(),
+    )
+    volume_uom_name = fields.Char(
+        string="Volume unit of measure label",
+        related="volume_uom_id.name",
+        readonly=True,
+    )
+
+    weight = fields.Float(
+        help="The weight of the contents, not including any packaging, etc."
```
bealdav, line 27:
```suggestion
        help="Net weight of the contents (not including any packaging, etc.)"
```

hparfr: No, I want the help messages to be coherent with odoo's default.

hparfr:
I think it's ready to be merged.
(Travis is green since we while btw)

PR #616, opened 2020-05-18
[ADD] new module product_barcode_constraint_per_company

legalsylvain:
Hi all,

Proposing a new technical module that allows to have two products with the same barcode if the products belongs to differents company.

A constraint has been introduced in the odoo 9.0 serie. (commit [here](https://github.com/odoo/odoo/commit/ede1071a09902e6eeb315cf51325a03f3991b3c3)), and it doesn't fit with some multicompany implementation.


product_barcode_constraint_per_company/readme/DESCRIPTION.rst
```diff
@@ -0,0 +1,7 @@
+This module extend Odoo functionnalities changing the constraint of the barcode
+on the ``product.product`` model.
+
+With this module, two products can have the same barcode, if there don't belong
+to the same company.
+
+This contrainst was introduced in Odoo V9.0 in [that commit](https://github.com/odoo/odoo/commit/ede1071a09902e6eeb315cf51325a03f3991b3c3), and doesn't fit with some multi companies implementation where products are not shared.
```
robinkeunen, line 7:
```suggestion
This constraint was introduced in Odoo V9.0 in [that commit](https://github.com/odoo/odoo/commit/ede1071a09902e6eeb315cf51325a03f3991b3c3), and doesn't fit with some multi companies implementation where products are not shared.
```

product_barcode_constraint_per_company/models/product_product.py
```diff
@@ -0,0 +1,31 @@
...
+            if _sql_constraint[0] == "barcode_uniq":
+                found = i
+            i += 1
+        if found != -1:
+            self._sql_constraints[found] = (
+                "barcode_uniq",
+                "unique(barcode, tmpl_company_id)", _(
+                    "A barcode can only be assigned to one"
+                    " product per company !")
+            )
```
robinkeunen, line 30: This looks like complicated looping to me, can't you do something like this?

```python
for i, constraint in enumerate(self._sql_constraints):
    if constraint[0] == "barcode_uniq":
        self._sql_constraints[i] = (
            "barcode_uniq",
            "unique(barcode, tmpl_company_id)", _(
                "A barcode can only be assigned to one"
                " product per company !")
        )
```

robinkeunen (commented):
Looks correct but I made improvement suggestions.

hparfr:
I think this module should be in stock-logistics-barcode repo.

legalsylvain:
> I think this module should be in stock-logistics-barcode repo.

Indeed ! in the same place than https://github.com/OCA/stock-logistics-barcode/tree/12.0/product_multi_ean

PR #620, opened 2020-05-24
[12.0] [ADD] Add module to define prices by product brands.

feketemihai:


pedrobaeza:
Maybe this should go to OCA/brand, or better in 13.0?

feketemihai:
@pedrobaeza Hi, i depend on this module https://github.com/OCA/product-attribute/tree/12.0/product_brand, and this one is a different implementation for Brands than the one's from OCA/brand, we use it with https://github.com/OCA/e-commerce/tree/12.0/website_sale_product_brand. 
I would leave it for like this, and for v13 (since product_brand not yet migrated) to move both of them to OCA/brand and to refactor them. What do you think?

pedrobaeza:
OK, but `product_brand` is already migrated: https://github.com/OCA/brand/tree/13.0/product_brand

feketemihai:
@pedrobaeza sorry, i just saw product_brand is in OCA/brand and it doesn't depends on https://github.com/OCA/brand/tree/12.0/brand

pedrobaeza:
Yeah, any way the refactorization can be interesting. For this PR, let's continue in this repo and put a not about moving it to OCA/brand on 13.0

feketemihai:
@pedrobaeza  Yes, it will be, but like this it's strange to have brands in 2 places, probably we cna refactor also brand module, res.brand, to have a checkbox for company owned brands, i have to check the other modules also, if we can merge into a single concept.

pedrobaeza:
Yeah, this is more an historic thing: OCA/brand didn't exists when 12.0 branch was created. It was suggested due to the family of modules, and then created, but moving `product_brand` in the middle of the version will break existing installations, so that one was kept in the same place until 13.0. That's why I ask for moving at least derivative modules.

About the refactoring, the module has been migrated without taking attention to the rest of the modules and indeed I think it's worth to have a unified brand model. Maybe it can be done on 13.0, as this version is still not very spread, or finally waiting for 14.0 and block any migration attempt without that refactoring.

feketemihai:
@pedrobaeza i will move it to oca/brand in v13, and for v14.0 to check the refactoring, since now we already have the module merged in v13, some PR in OCA/Brand for the other modules, so v14.0 seems for me more realistic without having people to change now.

PR #623, opened 2020-05-28
[13.0][ADD] product_abc_classification

MiquelRForgeFlow:
Product ABC Classification
====================

This modules includes the ABC analysis (or ABC classification), which is
used by inventory management teams to help identify the most important
products in their portfolio and ensure they prioritize managing them above
those less valuable.

Managers will create a profile with several levels (percentages) and then the
profiled products will automatically get a corresponding level using the
ABC analysis.


product_abc_classification/readme/USAGE.rst
```diff
@@ -0,0 +1,10 @@
+To use this module, you need to:
+
+#. Go to Sales or Inventory menu, then to Configuration/Products/XYZ Analysis Profile
```
emagdalenaC2i: This menu is not available in Inventory

product_abc_classification/readme/USAGE.rst
```diff
@@ -0,0 +1,10 @@
...
+and create a profile, knowing that the sum of all levels in the profile should sum 100.
```
emagdalenaC2i: Please, add a comment to explain that the user must to create several levels in that profile (maximum 12, I think)

product_abc_classification/readme/USAGE.rst
```diff
@@ -0,0 +1,10 @@
...
+
+#. Later you should go to product categories o product variants, and assign them a profile.
```
emagdalenaC2i:
```suggestion
#. Later you should go to product categories or product variants, and assign them a profile.
```

product_abc_classification/views/xyz_analysis_view.xml
```diff
@@ -0,0 +1,92 @@
...
+    </record>
+    <menuitem
+        action="product_abc_classification.xyz_analysis_profile_action"
+        id="menu_xyz_analysis_profile_config_stock"
+        parent="stock.menu_product_in_config_stock"
+        sequence="10"
+    />
+    <menuitem
+        action="product_abc_classification.xyz_analysis_profile_action"
+        id="menu_xyz_analysis_profile_config_stock"
```
emagdalenaC2i: Should have a different Id??

emagdalenaC2i (commented):
Functional review. Looks really cool. Just a few comments

emagdalenaC2i (commented):
Should the percentaje be unique?
See 
Percentaje unique

emagdalenaC2i:
And please, could you add some tests?

emagdalenaC2i:
image

emagdalenaC2i:
Why to use a selection in "Data source" and "Value" if there are just one single option to choose?

emagdalenaC2i:
Percentajes

MiquelRForgeFlow:
@emagdalenaC2i The screen about level should be unique its intended, You have repeated the number 40. And the selections are also intended, because maybe in the future it's extended.

emagdalenaC2i:
> @emagdalenaC2i The screen about level should be unique its intended, You have repeated the number 40. And the selections are also intended, because maybe in the future it's extended.

So, I can't set an ABC setting of 20%-40%-40% ?

MiquelRForgeFlow:
> So, I can't set an ABC setting of 20%-40%-40% ?

No, because it doesn't make sense. You can see explained here: https://www.eazystock.com/blog/how-to-use-abc-analysis-for-inventory-management/

emagdalenaC2i:
This module is a great idea, but you need to rethink the UI

emagdalenaC2i:
Ok, but make it sense to use a A-20% B-20% and C-60% settings?

MiquelRForgeFlow:
> Ok, but make it sense to use a A-20% B-20% and C-60% settings?

No. The percentages have to be unique.

emagdalenaC2i:
I think you could use a method like in the Payment terms
image

product_abc_classification/models/xyz_analysis_profile.py
```diff
@@ -0,0 +1,154 @@
...
+            return product_list
+
+    def _fill_data_from_stock_moves(self, date, product_list):
+        self.ensure_one()
+        moves = (
+            self.env["stock.move"]
+            .sudo()
+            .search(
+                [
+                    ("sale_line_id.state", "in", ["sale", "done"]),
```
JordiBForgeFlow: Not correct. Use simply the stock move status "done".

product_abc_classification/models/xyz_analysis_profile.py
```diff
@@ -0,0 +1,154 @@
...
+                    ("sale_line_id.order_id.date_order", ">", date),
```
JordiBForgeFlow: Not correct. Use the field "date" in the stock move.

product_abc_classification/models/xyz_analysis_profile.py
```diff
@@ -0,0 +1,154 @@
...
+            )
+        )
+        uom_id = self.env.ref("uom.product_uom_unit")
+        for product in moves.mapped("product_id"):
+            product_data = {
+                "product": product,
+                "units_sold": 0,
+            }
+            for move in moves.filtered(lambda m: m.product_id == product):
+                product_data["units_sold"] += move.product_uom._compute_quantity(
```
JordiBForgeFlow: This is not needed. The stock move already has a field for the quantity completed in the product unit of measure.

product_abc_classification/models/xyz_analysis_profile.py
```diff
@@ -0,0 +1,154 @@
...
+            return self._fill_data_from_stock_moves(date, product_list)
+        else:
+            return product_list
+
+    def _fill_data_from_stock_moves(self, date, product_list):
+        self.ensure_one()
+        moves = (
+            self.env["stock.move"]
+            .sudo()
+            .search(
```
JordiBForgeFlow: This is going to be very slow. Use search_group, and group by product_id, then accumulating on the field 'product_qty'.

JordiBForgeFlow (changes requested):

MiquelRForgeFlow:
@JordiBForgeFlow done

JordiBForgeFlow (approved):

JordiBForgeFlow:
cc @LoisRForgeFlow @NuriaMForgeFlow @didierdonze

product_abc_classification/models/product_product.py
```diff
@@ -0,0 +1,74 @@
+# Copyright 2020 ForgeFlow
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl.html).
+
+from odoo import api, fields, models
+
+
+class ProductTemplate(models.Model):
+    _inherit = "product.template"
+
+    abc_classification_profile_id = fields.Many2one(
```
legalsylvain, line 10: Hi. 

I don't understand why the abc_classification_profile_id is in the variant level, and not in the template level. Does it make sense to set a profile 1 for a variant, and a profile 2 to another variant of the same template ? 

Propose to move the profile in the template level. (and just set a related in the variant level).

What do you think ? 

Also, it could be great to have an api@depends("categ_id") to set the correct profile automatically when creating a new product, linked to a product_category with a profile defined.

legalsylvain (commented):
some remarks inline. Thanks for this new module.

product_abc_classification/models/product_product.py
MiquelRForgeFlow: @legalsylvain Sorry, I didn't have time to check this comment. Well, now if you think the code can be improved, then you can propose a PR for this version or do the refactor in the next one.

legalsylvain: Arf... this require to write migration script to change the model.

I'll see. 

But well, it doesn't answer to my question : Does it make sense to set a profile 1 for a variant, and a profile 2 to another variant of the same template ?

Thanks.

LoisRForgeFlow: @legalsylvain From my POV it makes more sense in the variant. You can have a variant that is more successful in the market than other, and therefore you classify one as A and other as B.

My 2 cents.

legalsylvain: Not sure you get it.
Of course the level a/b/c should be in the variant.
But the perfil is common for all the variant of a product template, by design. 

Concrete counter exemple welcome.

PR #625, opened 2020-06-03
[13.0][ADD] product_demand_variability

NuriaXifre:
This module allows the user to classify the products depending on its demand variability. The variability of a product is calculated as the standard deviation/total sales. ~@ForgeFlow


product_demand_variability/models/product.py
```diff
@@ -0,0 +1,82 @@
...
+class ProductCategory(models.Model):
+    _inherit = "product.category"
+
+    profile_categ_id = fields.Many2one("variability.profile")
+
+
+class ProductProduct(models.Model):
+    _inherit = "product.product"
+
+    profile_id = fields.Many2one(comodel_name="variability.profile")
```
bealdav: Hi, thanks for this nice module/features
Probably profile is a so generic name than there is colision with fields on other module like product_profile in v10/12
Variability seems more precise

rousseldenis:
@NuriaMForgeFlow Could you attend @bealdav comments ?

AdriaGForgeFlow (approved):
Local functional review, LGTM once travis and runbot are green :+1:

PR #628, opened 2020-06-14
Fix smart button alignment in brand form view

zeroheure:


rousseldenis:
@zeroheure I'm not sure this is a fix.

Base version :

image

This version :

image


Moreover, people maybe have inherited from this, so I would tend to not accept this one. @oerp-odoo @loxamir @pedrobaeza Any advice?

pedrobaeza:
Yes, I agree in not changing it and I don't see bad previously

rousseldenis:
@zeroheure IMHO The double group was done intentionally in order to override with flexibility the view. Closing this. Don't hesitate to get feedback if you want.

zeroheure:
Well, the README contains a todo about alignment. This was a fix for it.
In the base version, partner field is not aligned with other fields or widgets, plus there is a vertical bar between partner label and partner form (Odoo widget design). Working every day with this is unpleasant.
More over this was an attempt to push small patches instead of a big one with lots of changes. See my proposal #632
(Agree for double group.)

PR #630, opened 2020-06-14
update and complete french translation

zeroheure:
Update translation terms according to Odoo (Produits => Articles),  translate missing strigs.


rousseldenis:
@zeroheure Closing as you need to use https://translation.odoo-community.org/ to translate

PR #631, opened 2020-06-14
Fix product_brand wrong dependency

zeroheure:
Add sale_management as dependency because it is needed to manage brands. This fix also dependencies chaining for website_sale_product_brand (before this, website_sale didn't install automaticaly).

Curently product_brand depends on sale module, which contains only common functions between sale_management and website_sale. It doesn't allow to manage Products Brands.

### DIFF
--- a/product_brand/__manifest__.py
+++ b/product_brand/__manifest__.py
@@ -19,7 +19,7 @@
     'website': 'https://github.com/OCA/product-attribute',
     'license': 'AGPL-3',
     'depends': [
-        'sale',
+        'sale_management',
         ],
     'data': [
         'security/ir.model.access.csv',


rousseldenis:
@bguillot @ThomasBinsfeld I'm not sure about this

pedrobaeza:
Yeah, the dependency is a technical one, not a "UI" one. You can't manage brands from UI without `sale_management`, but that's not the purpose IMO.

ThomasBinsfeld:
> @bguillot @ThomasBinsfeld I'm not sure about this

Me neither. This module doesn't actually need a dependency on `sale_management` to work. Adding it can have unwanted effects on existing installations.

zeroheure:
No problem for me. I close the PR.

PR #641, opened 2020-06-26
[13.0] packaging_uom: Fix xmlid of product packaging tree view

guewen:
The erroneous xmlid is not found, so the wrong list view is shown.


TDu (approved):

rousseldenis (approved):
Indeed !

PR #642, opened 2020-07-01
[ADD][product_packaging_unece] Add module to manage UNECE packages codes

Tonow-c2c:
**Context**

* Manage UNECE codes on packaging
* Be able to express a packaging qty in smallest packaging qty
* Simplify packaging name management to ease translation and avoid people to write a name every time they add a packaging on a product

**Usage**

When I open a product form, in the inventory tab I want to see the product packaging and its related code I can quickly see the Qty per packaging to know how much of a given packaging is contain in another

 

**TODO**

* Update (Improve) the module "product_packaging_type" from product-attribute (I think ) to:
    * in the product "Inventory" tab form (Packaging list view)
        * Hide the name of the packaging
        * Move the "Packaging type" in the first position
    * When opening the packaging form view from  "Inventory" tab of the product
        * Replace the name by packaging type at the top
        * Hide name field, but populate it with on_change of packaging with the packaging type name
    * Name of the packaging type:
        * name_get should be: "NAME + ' ' + (CODE)" -> e.g. Retail box (HV)
        * name_get of the product packaging should be overridden:

```python
            if product_id is set:
                return packaging_type name (e.g. Transport box)
            else:
                return the recorded packaging name on the packaging (e.g. DHL box)
```

image

image

image


pedrobaeza:
Shouldn't this go to OCA/community-data-files the same as the rest of UNECE stuff?

product_packaging_unece/__manifest__.py
```diff
@@ -0,0 +1,14 @@
+# Copyright 2020 Camptocamp SA
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl)
+{
+    "name": "Product Packaging UNECE",
+    "summary": "Manage UNECE packaging type",
+    "version": "13.0.1.0.0",
+    "category": "Product",
+    "author": "Camptocamp, Odoo Community Association (OCA)",
+    "license": "AGPL-3",
+    "depends": ["product_packaging_dimension", "product_packaging_type"],
```
pedrobaeza: These dependencies doesn't sound very attractive for such a base module.

Tonow-c2c:
> Shouldn't this go to OCA/community-data-files the same as the rest of UNECE stuff?

thanks @pedrobaeza good question for @jgrandguillaume and @jbaudoux

jgrandguillaume:
@pedrobaeza Could be moved yes, but as product-packaging-type is there and is a dependencie, I thoug better to have them both on the same place.
Do you have a strong preference to move it ?

pedrobaeza:
Well, that's my other question, as why a foundation like UNECE codes needs so big dependencies. I don't think UNECE should be linked to that new types, but directly to packages. I think standard has enough tools for doing this is a proper without the need of that overhead.

jbaudoux (approved):
LGTM

jbaudoux (changes requested):
The product packaging type assigned to product packaging is similar to the unit of measure of a product. I agree with pedro. This should not be a separate module but be included in the product_packaging_type module.

jgrandguillaume (changes requested):
Ok, got a point.  +1 To move all this the directly into product-packaging-type module.

It was the plan since day one by the way, quote "Update (Improve) the module "product_packaging_type" from product-attribute"

So please do not add a new module for this, it make no sense.

Tonow-c2c:
I will create the PR in `product_packaging_type`

Tonow-c2c:
thanks @pedrobaeza @jbaudoux @jgrandguillaume I create new PR #644

PR #643, opened 2020-07-02
[13.0][IMP] product_packaging_unit_price_calculator

TDu:
Use the display name for the packaging in the list of packages.
Set all prices to use the Product Price digits configuration.
On load set the current unit price so package prices are displayed.
Add a button to reset the calculator.

### DIFF
--- a/product_packaging_unit_price_calculator/wizards/product_package_price.py
+++ b/product_packaging_unit_price_calculator/wizards/product_package_price.py
@@ -22,8 +22,16 @@ class ProductPackagePrice(models.TransientModel):
             return False
         return self.env["product.supplierinfo"].browse(self._context.get("active_id"))
 
-    product_id = fields.Many2one(
+    def _default_product_id(self):
+        if self._context.get("active_model") != "product.product":
+            return False
+        return self.env["product.product"].browse(self._context.get("active_id"))
+
+    product_tmpl_id = fields.Many2one(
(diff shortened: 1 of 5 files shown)


jgrandguillaume (approved):
LGTM thanks

jbaudoux (approved):
Thanks. This relates to https://github.com/OCA/product-attribute/pull/644

guewen (approved):

product_packaging_unit_price_calculator/wizards/product_package_price.py
```diff
@@ -53,35 +69,42 @@ def _default_product_supplierinfo_id(self):
...
             self.warning_message = " "
-            self._compute_package_prices()
+        self._compute_package_prices()
+
+    @api.depends(
+        "product_pricelist_item_id", "product_supplierinfo_id", "product_tmpl_id"
+    )
+    def _compute_current_unit_price(self):
+        """Compute the original unit price, the one  that the calculator will change."""
+        if self.product_pricelist_item_id:
```
simahawk, line 89: it would be better to have a loop on self here.

TDu: I understand it is required on regular models, but here the compute is for the field of a wizard, we should always have one record, no ?

guewen (approved):
/ocabot merge patch

PR #644, opened 2020-07-02
[IMP][product_packaging_type] Specific UNECE product packaging naming.

Tonow-c2c:
**Context**

* Manage UNECE codes on packaging
* Be able to express a packaging qty in smallest packaging qty
* Simplify packaging name management to ease translation and avoid people to write a name every time they add a packaging on a product

**Usage**

When I open a product form, in the inventory tab I want to see the product packaging and its related code I can quickly see the Qty per packaging to know how much of a given packaging is contain in another

 

**TODO**

* Update (Improve) the module "product_packaging_type" from product-attribute (I think ) to:
    * in the product "Inventory" tab form (Packaging list view)
        * Hide the name of the packaging
        * Move the "Packaging type" in the first position
    * When opening the packaging form view from  "Inventory" tab of the product
        * Replace the name by packaging type at the top
        * Hide name field, but populate it with on_change of packaging with the packaging type name
    * Name of the packaging type:
        * name_get should be: "NAME + ' ' + (CODE)" -> e.g. Retail box (HV)
        * name_get of the product packaging should be overridden:

```python
            if product_id is set:
                return packaging_type name (e.g. Transport box)
            else:
                return the recorded packaging name on the packaging (e.g. DHL box)
```

replace #642 

image
image
image

### DIFF
--- a/product_packaging_type/models/product_packaging_type.py
+++ b/product_packaging_type/models/product_packaging_type.py
@@ -29,6 +29,12 @@ class ProductPackagingType(models.Model):
         if msg:
             raise ValidationError(msg)
 
+    def name_get(self):
+        result = []
+        for record in self:
+            result.append((record.id, "{} ({})".format(record.name, record.code)))
+        return result
+
 
 class ProductPackaging(models.Model):
     _inherit = "product.packaging"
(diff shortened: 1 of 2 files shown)


jgrandguillaume (approved):
LGTM thx !!

jbaudoux (approved):
LGTM

PR #650, opened 2020-07-09
[13.0][FIX] product_medical: add fields

i-vyshnevska:
add fields to product.template and form view

### DIFF
--- a/product_medical/models/product_template.py
+++ b/product_medical/models/product_template.py
@@ -11,6 +11,10 @@ class ProductTemplate(models.Model):
 
     medical_certificate_url = fields.Char(string="CE Certificate medical devices")
     medical_class_id = fields.Many2one("medical.class", string="Medical Class")
+    medicine_category_id = fields.Many2one(
+        "medicine.category", string="Medical Category"
+    )
+    ppe_category_id = fields.Many2one("ppe.category", string="PPE Category")
 
     in_vitro_diagnostic = fields.Many2one(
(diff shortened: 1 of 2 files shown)


etobella:
I know now thuis comment is too late, but I think I have to tell something.

What you are proposing is not medical product, this are medical device products. A medical product can be a lot of things: A medication, a medical service... So, IMO, maybe we need to change some things here. I know that the module name, shouldn't be changed now, as it is merged, but maybe the name of the fields (`is_medical` --> `is_medical_device`). If you want, I can make the migration script. 

Also, just to clarify some things. This kind of devices are usually created by hospitals? If true, this module should need to be on OCA/vertical-medical, WDYT?

Also, If these are for devices shouldn't it be managed by `maintenance.equipment`s? Just wondering, because we are currently managing this kind of devices as equipments.

i-vyshnevska:
@etobella this can be a big discussion about implementation, the introduced addon is to handle medical products for retail that is why it is not on vertical-medical

etobella:
Ok! good for me, then might be possible to change the field `is_medical` for `is_medical_device`? At least in order to avoid confusion? 

The module name already exists so it cannot be changed, but would be interesting to propose the change for v14 on ROADMAP.

simahawk:
@i-vyshnevska ping

carlosserratoro:
This has been modified by https://github.com/OCA/product-attribute/pull/783. There, among other changes, we remove the field ```is_medical```.

PR #651, opened 2020-07-16
[13.0] Create module product_total_weight_from_packaging

grindtildeath:
Depends on:
 - [x] https://github.com/OCA/stock-logistics-warehouse/pull/939
 - [x] https://github.com/Tonow-c2c/stock-logistics-warehouse/pull/1

This module provides a weight calculation function for products based on their
packagings weight.

It uses module `stock_packaging_calculator` to get weight from packagings
having a weight defined first and fallback on product weight field if no
weight is defined on the packagings.

### DIFF
--- a/product_total_weight_from_packaging/__manifest__.py
+++ b/product_total_weight_from_packaging/__manifest__.py
@@ -1,0 +1,15 @@
+# Copyright 2020 Camptocamp SA
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl)
+{
+    "name": "Product Total Weight From Packaging",
+    "summary": "Compute estimated weight based on product's packaging weights",
+    "version": "13.0.1.3.0",
+    "development_status": "Alpha",
+    "category": "Warehouse Management",
+    "website": "https://github.com/OCA/stock-logistics-warehouse",
+    "author": "Camptocamp, Odoo Community Association (OCA)",
+    "license": "AGPL-3",
+    "application": False,
+    "installable": True,
+    "depends": ["stock_packaging_calculator", "product_packaging_dimension"],
+}
--- a/product_total_weight_from_packaging/models/product.py
+++ b/product_total_weight_from_packaging/models/product.py
@@ -12,7 +12,7 @@ class ProductProduct(models.Model):
         qty_by_packaging_with_weight = self.with_context(
             **{
                 "_packaging_filter": lambda p: p.max_weight,
-                "_with_packaging_weight": True,
+                "_packaging_values_handler": self._prepare_qty_by_packaging_values_with_weight,  # noqa
             }
         ).product_qty_by_packaging(qty)
         total_weight = sum(
@@ -23,12 +23,15 @@ class ProductProduct(models.Model):
         )
         return total_weight
 
-    def _prepare_qty_by_packaging_values(self, packaging_tuple, qty_per_pkg):
-        res = super()._prepare_qty_by_packaging_values(packaging_tuple, qty_per_pkg)
-        if self.env.context.get("_with_packaging_weight"):
-            if packaging_tuple.is_unit:
-                res["weight"] = self.weight
-            else:
-                packaging = self.env["product.packaging"].browse(packaging_tuple.id)
-                res["weight"] = packaging.max_weight
+    def _prepare_qty_by_packaging_values_with_weight(
+        self, packaging_tuple, qty_per_pkg
+    ):
+        res = {
+            "qty": qty_per_pkg,
+        }
+        if packaging_tuple.is_unit:
+            res["weight"] = self.weight
+        else:
+            packaging = self.env["product.packaging"].browse(packaging_tuple.id)
+            res["weight"] = packaging.max_weight
         return res
--- a/product_total_weight_from_packaging/models/product.py
+++ b/product_total_weight_from_packaging/models/product.py
@@ -1,0 +1,34 @@
+# Copyright 2020 Camptocamp SA
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl)
+from odoo import models
+
+
+class ProductProduct(models.Model):
+
+    _inherit = "product.product"
+
+    def get_total_weight_from_packaging(self, qty):
+        self.ensure_one()
+        qty_by_packaging_with_weight = self.with_context(
+            **{
+                "_packaging_filter": lambda p: p.max_weight,
+                "_with_packaging_weight": True,
+            }
+        ).product_qty_by_packaging(qty)
+        total_weight = sum(
+            [
+                pck.get("qty", 0) * pck.get("weight", 0)
+                for pck in qty_by_packaging_with_weight
+            ]
+        )
+        return total_weight
+
+    def _prepare_qty_by_packaging_values(self, packaging_tuple, qty_per_pkg):
+        res = super()._prepare_qty_by_packaging_values(packaging_tuple, qty_per_pkg)
+        if self.env.context.get("_with_packaging_weight"):
+            if packaging_tuple.is_unit:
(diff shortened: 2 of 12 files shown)


product_total_weight_from_packaging/models/product.py
```diff
@@ -0,0 +1,35 @@
...
+        qty_by_packaging_with_weight = self.with_context(
+            **{
+                "_packaging_filter": lambda p: p.max_weight,
+                "_with_packaging_weight": True,
+            }
+        ).product_qty_by_packaging(qty)
+        total_weight = 0
+        for packaging_dict in qty_by_packaging_with_weight:
+            if packaging_dict.get("is_unit"):
+                pack_weight = self.weight
```
simahawk: is this needed? You set the weight for units on line 30 as `self.weight` as well. :thinking:

simahawk:
the 2 pending PRs got merged

jbaudoux:
Just a small remark here : https://github.com/OCA/stock-logistics-warehouse/pull/939#issuecomment-660894458

product_total_weight_from_packaging/__manifest__.py
```diff
@@ -0,0 +1,15 @@
+# Copyright 2020 Camptocamp SA
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl)
+{
+    "name": "Product Total Weight From Packaging",
+    "summary": "Compute product quantity to pick by packaging",
```
jbaudoux: Wrong summary

Compute estimated weight based on product's packaging weights

product_total_weight_from_packaging/readme/DESCRIPTION.rst
```diff
@@ -0,0 +1,6 @@
+This module provides a weight calculation function for products based on their
```
jbaudoux:
```suggestion
This module provides a function to estimate the weight of a given quantity taking into account the product's```

**Review by jbaudoux — COMMENTED** [review]
Small description remarks

**Review by simahawk — APPROVED** [review]

**Comment by simahawk** [comment]
@grindtildeath can you squash and add the dependency on stock-logistics-wh?

File: product_total_weight_from_packaging/models/product.py [hunk 4]
```diff
@@ -0,0 +1,34 @@
...
+            [
+                pck.get("qty", 0) * pck.get("weight", 0)
+                for pck in qty_by_packaging_with_weight
+            ]
+        )
+        return total_weight
+
+    def _prepare_qty_by_packaging_values(self, packaging_tuple, qty_per_pkg):
+        res = super()._prepare_qty_by_packaging_values(packaging_tuple, qty_per_pkg)
+        if self.env.context.get("_with_packaging_weight"):
```
rousseldenis: @grindtildeath Why doing some complex things with context here as you add just an entry in the result? There is no behaviour bypassing of the original packaging caclulator.

grindtildeath: As the module can be used in different contexts, I thought we can avoid having the weight in the result if unneeded, but you're right in the sense having this value in the dict doesn't break anything from original function.
@simahawk what d'you think?

grindtildeath:
> can you squash and add the dependency on stock-logistics-wh?

@simahawk shouldn't we move the module here to avoid this dependency? :wink:  :trollface:

product_total_weight_from_packaging/models/product.py
simahawk: @grindtildeath thinking about this twice in the ctx where is needed: do we really need the weight to be in `product_qty_by_packaging` result?
Because your goal here can be achieved with `sum([x.qty * x.max_weight for x in self.packaging_ids.filtered(lambda p: p.qty and p.max_weight))])` no?

Regarding adding `weight` to the original result w/out conditions... fine if we do it in the base module. You'll have to add a dependency on delivery.
And yes, you can break things - like tests - if you add it always and it's not foreseen.

grindtildeath: @simahawk Did you check the test and usage before suggesting this? :roll_eyes:

simahawk: you have one test that only checks the result of `get_total_weight_from_packaging`.
How should I imagine something different?
In my suggestion you only miss the unit weight because I omitted it. You can add it very easily, no?
What else do I miss?

grindtildeath: @simahawk The test is as follows:

 - we have a product with a weight of 5 and following packagings
    - Pair (qty=2, max_weight=12.5)
    - Cardbox (qty=10, max_weight=55)
    - Pallet (qty=200, max_weight=0)

As stated in the module readme:
>This module provides a a function to estimate the weight
of a given quantity of a product, taking into account the product
packaging's weights.

> It uses module `stock_packaging_calculator` to get weight from product packagings
having a weight defined first and fallback on product weight field if no
weight is defined on any of the packagings.

Therefore, if we want to compute the weight for 259 units, we must do 25 * 55 + 4 * 12.5 + 5, because the pallet doesn't have a weight defined.

Anyway, I'm sorry but what does `sum([x.qty * x.max_weight for x in self.packaging_ids.filtered(lambda p: p.qty and p.max_weight))])` actually do? 
I'm still trying to figure out what you can achieve by multiplicating the quantity inside a packaging by the weight of said packaging...

If I have a 6 pack of beers and each 0.5l beer weights 0.5kg, my 6-pack should have a max weight around 3kg. And I really don't understand how your suggestion of multiplicating 3kg * 6pieces would give anything meaningful...

simahawk: Sorry, don't take my example literally. I meant "loop on them, filter them, use them as you want" but I admit that was not crystal clear :sweat_smile: 

I'm just saying that maybe you don't need that module as a base and you can rely on your own computation and/or use some bits only from that module (eg: _qty_by_pkg) and do what you want here.

Another options is, instead of overriding `_prepare_qty_by_packaging_values` and check for ctx key, you could improve `_product_qty_by_packaging` to lookup for `_packaging_values_handler` and pass your own handler.
This way the base module is even more pluggable.

Eg:

`value = self._prepare_qty_by_packaging_values(pkg, qty_per_pkg)`
becomes
```py
prepare_values = self.env.context.get('_packaging_values_handler', self._prepare_qty_by_packaging_values)
value = prepare_values(pkg, qty_per_pkg)
```
Then you use your own handler here. No overlap of features/data/whatever, no useless checks.

That being said. I don't care that much here about cosmetics and alike. That's why I approved it before.
The module is alpha: if it works and does not break anything, it's ok. Maybe just add a TODO comment there to improve it on the next round.

simahawk: > > can you squash and add the dependency on stock-logistics-wh?
> 
> @simahawk shouldn't we move the module here to avoid this dependency? wink :trollface:

go ahead if you have time. But if you change the module's name, add a migration script ;)

simahawk:
> @simahawk shouldn't we move the module here to avoid this dependency? wink :trollface:

go ahead if you have time. But if you change the module's name, add a migration script ;)

product_total_weight_from_packaging/models/product.py
simahawk: as I was touching the module, I added the custom value handler hook here https://github.com/OCA/stock-logistics-warehouse/pull/949

rousseldenis:
@grindtildeath Will you fix this ?

jgrandguillaume (approved):
LGTM, I suggest to merge this work as it is Alpha we can improve the rest on a later stage.

jgrandguillaume:
Please add the oca_dependencies @grindtildeath

grindtildeath:
@rousseldenis Can you please check last commit?

HaraldPanten (approved):

PR #652, opened 2020-07-17
[IMP] change Product packaging `Has GTIN` computation

Tonow-c2c:
* If Product packaging "Contained quantity" == 0.0 "Has GTIN" is
* In product packaging form view the "Has GTIN" is invisible

### DIFF
--- a/product_packaging_type/models/product_packaging_type.py
+++ b/product_packaging_type/models/product_packaging_type.py
@@ -57,7 +57,8 @@ class ProductPackaging(models.Model):
         related="packaging_type_id.is_default",
         store=True,
     )
-    type_has_gtin = fields.Boolean(related="packaging_type_id.has_gtin", readonly=True)
+    type_has_gtin = fields.Boolean(readonly=True, compute="_compute_type_has_gtin")
+
     type_sequence = fields.Integer(
         string="Type sequence",
         related="packaging_type_id.sequence",
@@ -82,6 +83,13 @@ class ProductPackaging(models.Model):
             "with the same type for a given product.",
         )
     ]
+
+    @api.depends("packaging_type_id", "packaging_type_id.has_gtin", "qty")
+    def _compute_type_has_gtin(self):
+        for packaging in self:
+            packaging.type_has_gtin = packaging.packaging_type_id.has_gtin
+            if not packaging.qty:
+                packaging.type_has_gtin = False
 
     @api.depends(
         "product_id",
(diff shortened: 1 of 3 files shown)


product_packaging_type/models/product_packaging_type.py
```diff
@@ -62,6 +63,13 @@ def default_packaging_type_id(self):
         compute="_compute_qty_per_type", string="Qty per package type"
     )
 
+    @api.depends("packaging_type_id", "packaging_type_id.has_gtin", "qty")
```
i-vyshnevska, line 89: should this be onchange method, as it only about UI usage?

mmequignon (approved):

product_packaging_type/models/product_packaging_type.py
```diff
@@ -51,7 +51,8 @@ def default_packaging_type_id(self):
         ondelete="restrict",
         default=lambda p: p.default_packaging_type_id(),
     )
-    type_has_gtin = fields.Boolean(related="packaging_type_id.has_gtin", readonly=True)
+    type_has_gtin = fields.Boolean(readonly=True, compute="_compute_type_has_gtin")
```
guewen: Semantically, this is very weird that this field is named `type_has_gtin`, as it *may* have a gtin according to the implementation of `_compute_type_has_gtin`. This is the kind of thing that may hurt later when someone assumes that `type_has_gtin` really has a gtin for another part of logic. Since the only goal of this field is to set the field barcode required conditionally, I would name it `barcode_required_for_gtin`.

guewen (changes requested):

guewen (approved):

product_packaging_type/models/product_packaging_type.py
```diff
@@ -57,7 +57,11 @@ def default_packaging_type_id(self):
         related="packaging_type_id.is_default",
         store=True,
     )
-    type_has_gtin = fields.Boolean(related="packaging_type_id.has_gtin", readonly=True)
+    type_has_gtin = fields.Boolean(readonly=True, compute="_compute_type_has_gtin")
```
guewen: @Tonow-c2c this one must be removed

AttributeError: 'product.packaging' object has no attribute '_compute_type_has_gtin'


```suggestion
```

Tonow-c2c: thanks :pray:  sorry for the issue in the rebase :grimacing:

PR #653, opened 2020-07-20
[12.0][FIX] packaging_uom: use factor 1.0 if packaging quantity equals 0

AdriaGForgeFlow:
This PR is intended to fix a small issue that appeared when running tests here [stock-logistics-warehouse/pull/938](https://github.com/OCA/stock-logistics-warehouse/pull/938).

When a packaging has qty equal to 0, there was a division by zero which raised the following error:
image

The fix consists on default the value of factor to 1.0 if the quantity of the package is 0.

cc ~ @ForgeFlow

### DIFF
--- a/packaging_uom/models/product_packaging.py
+++ b/packaging_uom/models/product_packaging.py
@@ -67,8 +67,9 @@ class ProductPackaging(models.Model):
         """
         for packaging in self:
             category_id = packaging.product_id.uom_id.category_id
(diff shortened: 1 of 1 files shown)


LoisRForgeFlow (approved):
LGTM

MiquelRForgeFlow (approved):

AaronHForgeFlow (approved):

PR #656, opened 2020-07-24
[13.0] Replace SQL constraint for unique packaging type by Python constrains

guewen:
In  product_packaging_type.

The SQL constraint I [added previously](https://github.com/OCA/product-attribute/pull/654) prevent some demo data to
be installed (e.g. in module OCA/wms::stock_storage_type).

The issue is that the packaging type is required, is part of the unique
constraint, and has a default packaging type.
When the module is installed and we have existing data or demo data,
the default type will be set on all the packagings, meaning we have
to ignore the unique constraint on the default type.
This was already handled using an "EXCLUDE" constraint, but it does
not solve the issue with demo data, the related field
"packaging_type_is_default", used as exclusion for the unique
constraint, is first inserted with a NULL value before being set
to True.

### DIFF
--- a/product_packaging_type/models/product_packaging_type.py
+++ b/product_packaging_type/models/product_packaging_type.py
@@ -51,16 +51,9 @@ class ProductPackaging(models.Model):
         ondelete="restrict",
         default=lambda p: p.default_packaging_type_id(),
     )
-    # Used for the unique constraint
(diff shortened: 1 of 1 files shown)


grindtildeath (approved):
:+1:

simahawk (approved):

PR #658, opened 2020-07-30
[13.0] Add product_template_tags_code module

Tonow-c2c:
This module allow to reference tags record a code is computed and writeable.

image

Peek 2020-07-30 18-10


product_template_tags_code/README.rst
```diff
@@ -0,0 +1,82 @@
...
+.. |badge4| image:: https://img.shields.io/badge/weblate-Translate%20me-F47D42.png
+    :target: https://translation.odoo-community.org/projects/product-attribute-13-0/product-attribute-13-0-product_template_tags_code
+    :alt: Translate me on Weblate
+.. |badge5| image:: https://img.shields.io/badge/runbot-Try%20me-875A7B.png
+    :target: https://runbot.odoo-community.org/runbot/135/13.0
+    :alt: Try me on Runbot
+
+|badge1| |badge2| |badge3| |badge4| |badge5| 
+
+To allow to reference tags record a code is computed and writable.
```
simahawk:
```suggestion
Allows to reference tag records trough a unique code. 
The code is computed automatically if not provided.
```

product_template_tags_code/models/product_template_tag.py
```diff
@@ -0,0 +1,38 @@
...
+from odoo import api, fields, models
+
+from odoo.addons.http_routing.models.ir_http import slugify
+
+
+class ProductTemplateTag(models.Model):
+
+    _inherit = "product.template.tag"
+
+    code = fields.Char(string="Code")
```
simahawk: what happened to the `compute` version?

product_template_tags_code/models/product_template_tag.py
```diff
@@ -0,0 +1,38 @@
...
+
+    def _update_code(self):
+        for rec in self:
+            code = False
+            if rec.code and rec.code != slugify(rec.code):
+                # make sure the code is always clean
+                code = slugify(rec.code)
+            elif rec.name and rec.name.strip():
+                code = slugify(rec.name)
+            super(ProductTemplateTag, rec).write({"code": code})
```
simahawk: this writes twice every time

product_template_tags_code/tests/test_product_template_tags.py
```diff
@@ -0,0 +1,34 @@
+# Copyright 2020 Camptocamp
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+from odoo.tests.common import SavepointCase
+
+
+class TestProductTemplateTagBase(SavepointCase):
```
simahawk: remove this, is in the base module already

simahawk (changes requested):

product_template_tags_code/__manifest__.py
```diff
@@ -0,0 +1,13 @@
...
+{
+    "name": "Product Template Tags",
+    "summary": "This addon allow to add code on products tags",
+    "version": "13.0.1.0.0",
+    "license": "AGPL-3",
+    "author": "Camptocamp SA, Odoo Community Association (OCA)",
+    "website": "https://github.com/OCA/product-attribute",
+    "depends": ["product", "product_template_tags"],
+    "data": ["views/product_template_tag.xml"],
+}
```
simahawk, line 13: where is the migration step? Or the post install hook? You should compute codes for existing records.

product_template_tags_code/models/product_template_tag.py
Tonow-c2c: the compute finally don't work with the @sebalix advice I try this solution

the compute was not use in the record creation

product_template_tags_code/models/product_template_tag.py
Tonow-c2c: @sebalix propose me this solution to avoid to fall in infinite writing loop with the name and the code

product_template_tags_code/models/product_template_tag.py
simahawk: but then why not using an inverse func?

Tonow-c2c: cause @sebalix tell me it's easier like this, but I can change :wink:

simahawk: weird, it works for me for both create and write :thinking:

product_template_tags_code/models/product_template_tag.py
rousseldenis: I suggest you to replace write() and create() by an inverse on code field. This is neater and more comprehensive. Moreover, I think that will avoid infinite loop.

product_template_tags_code/models/product_template_tag.py
```diff
@@ -0,0 +1,38 @@
...
+    )
+
+    _sql_constraints = [
+        ("code_uniq", "unique(code)", "Product template code already exists",)
+    ]
+
+    @api.depends("name", "code")
+    def _compute_code(self):
+        for rec in self:
+            if rec.code and not rec.code == slugify(rec.code):
```
simahawk: you can remove this part as the slug-ification is done in the inverse method anyway

product_template_tags_code/models/product_template_tag.py
```diff
@@ -0,0 +1,35 @@
...
+
+
+class ProductTemplateTag(models.Model):
+
+    _inherit = "product.template.tag"
+
+    code = fields.Char(
+        string="Code",
+        compute="_compute_code",
+        readonly=False,
```
sebalix, line 16: `readonly=False` is not mandatory as soon as there is an `inverse`

sebalix (approved):
Small comment :+1:

product_template_tags_code/models/product_template_tag.py
sebalix: We didn't test the inverse actually (there was only the compute), and I was tired, and newbie, sooo... :p

product_template_tags_code/models/product_template_tag.py
```diff
@@ -0,0 +1,35 @@
...
+        inverse="_inverse_code",
+        store=True,
+    )
+
+    _sql_constraints = [
+        ("code_uniq", "unique(code)", "Product template code already exists",)
```
simahawk:
```suggestion
        ("code_uniq", "unique(code)", "Product template tag code already exists",)
```

product_template_tags_code/tests/test_product_template_tags.py
```diff
@@ -0,0 +1,40 @@
+# Copyright 2020 Camptocamp
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+from ...product_template_tags.tests.test_product_template_tags import (
```
simahawk:
```suggestion
from odoo.addons.product_template_tags.tests.test_product_template_tags import (
```

simahawk (approved):

simahawk:
@Tonow-c2c 2 small glitches

PR #660, opened 2020-07-31
[12.0] ADD product_model_viewer

eLBati:
See https://github.com/OCA/web/tree/12.0/web_widget_model_viewer


pedrobaeza:
Can you provide demo data for showcasing both modules?

eLBati:
@pedrobaeza right, done

mamcode (approved):
LGTM, very good contribution.

marcelsavegnago (approved):

pedrobaeza:
Several thoughts:

- You are linking the 3D model to the template, but there can be the option of having a different 3D model for each variant. I think you should handle this the same way as product images, defaulting to the template image, but having the option to put a different one in the variant.
- Does this worth to be split from `website_sale_product_model_viewer`? If you plan to do another modules for any feature in backend or without `website_sale`, then it can be, but if the only idea is to put it on the e-commerce, then not sure. If not split, I would move the field to the e-commerce tab.
- It's worth to mention in the README in the usage section that with demo data, the product "Office Chair Black" contains a sample 3D model.

mamcode:
Agree with @pedrobaeza  in all but I think that split with `website_sale_product_model_viewer` is good because there may be projects without e-commerce that only need to have the 3D model in the product in the backend.

andreampiovesana:
I think that we have to emulate the image, so is better to link the 3d to the product.product model?

eLBati:
@pedrobaeza @mamcode @andreampiovesana implemented variant handling, similarly to images.
Added `USAGE.rst`.

I kept `product_model_viewer`splitted from `website_sale_product_model_viewer` for the reasons explained by @mamcode . The 3D model of products could be also used by other modules, not related to e-commerce.

pedrobaeza:
OK with the split for future uses.

product_model_viewer/views/product_views.xml
```diff
@@ -0,0 +1,15 @@
+<?xml version="1.0" encoding="utf-8"?>
+<odoo>
+    <record id="product_template_form_view_3d" model="ir.ui.view">
+        <field name="name">product_template_form_view_3d</field>
+        <field name="model">product.template</field>
+        <field name="inherit_id" ref="product.product_template_form_view"/>
+        <field name="arch" type="xml">
+            <xpath expr="//notebook" position="inside">
+                <page string="3D model">
```
bealdav: Congratulations for this really nice module

Completely expressive like that, I think (only my opinion)
```suggestion
                <page string="3D">
```
When you have many tabs, shorter ones are good ;-)

eLBati: Thanks, done

andreampiovesana (approved):

bealdav (approved):

rousseldenis:
@eLBati Ready ?

eLBati:
@rousseldenis this is ready for me, thanks

PR #661, opened 2020-07-31
[11.0][IMP] product_multi_price: add show multi-price access group

ernestotejeda:
Add a new access group to show/hide multi-prices list in product form view.

cc @Tecnativa TT24988

### DIFF
--- a/product_multi_price/hooks.py
+++ b/product_multi_price/hooks.py
@@ -1,0 +1,13 @@
+# Copyright 2020 Tecnativa - Ernesto Tejeda
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+
+from odoo import api, SUPERUSER_ID
+
+
+def post_init_hook(cr, registry):
+    env = api.Environment(cr, SUPERUSER_ID, {})
+    group_id = env.ref('product_multi_price.group_show_multi_prices').id
+    default_user = env.ref('base.default_user')
+    user = env['res.users'].with_context(active_test=False).search(
+        [("share", "=", False)])
+    (user - default_user).write({'groups_id': [(4, group_id, None)]})
--- a/product_multi_price/views/product_views.xml
+++ b/product_multi_price/views/product_views.xml
@@ -4,6 +4,7 @@
     <record id="product_template_form_view" model="ir.ui.view">
         <field name="model">product.template</field>
         <field name="inherit_id" ref="product.product_template_form_view"/>
+        <field name="groups_id" eval="[(4, ref('product_multi_price.group_show_multi_prices'))]"/>
         <field name="arch" type="xml">
             <xpath expr="//div[@name='standard_price_uom']" position="after">
                 <field name="price_ids"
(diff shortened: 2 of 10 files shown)


product_multi_price/hooks.py
```diff
@@ -0,0 +1,10 @@
+# Copyright 2020 Tecnativa - Ernesto Tejeda
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+
+from odoo import api, SUPERUSER_ID
+
+
+def post_init_hook(cr, registry):
+    env = api.Environment(cr, SUPERUSER_ID, {})
+    group_id = env.ref('product_multi_price.group_show_multi_prices').id
+    env['res.users'].search([]).write({'groups_id': [(4, group_id, None)]})
```
pedrobaeza: Write also on inactive users, but only internal ones (`[("share", "=", False)]`).

product_multi_price/migrations/11.0.1.2.0/post-migration.py
```diff
@@ -0,0 +1,9 @@
+#  Copyright 2020 Tecnativa - Ernesto Tejeda
```
pedrobaeza: You have to put a higher version than the existing one, and bump manually the module version, so there's no possibility of this not being triggered, even if ocabot is launched with other bump. Use `11.0.2.0.0`, as this is a major change in the module.

product_multi_price/migrations/11.0.1.2.0/post-migration.py
```diff
@@ -0,0 +1,9 @@
...
+#  License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl
+from openupgradelib import openupgrade
+
+
+@openupgrade.migrate(use_env=True)
+def migrate(env, version):
+    group_id = env.ref('product_multi_price.group_show_multi_prices').id
```
pedrobaeza: The idea here is not to duplicate code, but call the post_init_hook, and don't use openupgradelib, as it's not needed:
```python
from odoo.addons.product_multi_price.hooks import post_init_hook

def migrate(cr, version):
    post_init_hook(cr, False)
```

product_multi_price/views/product_views.xml
```diff
@@ -10,6 +10,7 @@
                     widget="one2many_list"
                     context="{'default_product_id': product_variant_id,'tree_view_ref':'product_multi_price.product_multi_price_tree_view2', 'form_view_ref':'product_multi_price.product_multi_price_form_view'}"
                     attrs="{'invisible': [('product_variant_count', '&gt;', 1), ('is_product_variant', '=', False)]}"
+                    groups="product_multi_price.group_show_multi_prices"
```
pedrobaeza: Put the group in the view, not in the field.

pedrobaeza (changes requested):

ernestotejeda:
@pedrobaeza Changes done

pedrobaeza (approved):

rousseldenis (approved):
LGTM

pedrobaeza:
Cherry-picked to v12 in 84166cdf1676e491397bdaeb91d64416e956fa66

PR #662, opened 2020-08-17
[12.0][ADD] product_import_autocomplete_attribute

kevinkhao:
Trigger variant attribute sync when importing variants.

Although product template attribute choices are synced from template to product variant, the reverse is not true.

In other words,

* Base function product_template.create_variant_ids() uses product template's product.attribute to create variants (tmpl->variant)
* In reverse, this module uses the product.attribute on variants, to update product.attribute on the product template (variant->tmpl)

This product.attribute sync is triggered when you do an import of product variants.


kevinkhao:
Following feedback that the module was not easily understandable, here is some additional information.
Step 1: observe the template attributes for T-shirt
step_1

Step 2: import variant with some attributes
step_2_variant
Observe the variant with new attributes. It is reflected on the product template.
step_2_tmpl
Step 3: import some other variants.
step_3_variant
Observe it is reflected on product template.
step_3_tmpl

sebastienbeau:
replaced / refactored / migrated in v14 here : https://github.com/OCA/product-attribute/pull/830
We will not plan to work any more on V12 for this module so we close it

PR #663, opened 2020-08-20
[IMP] add sale_price basond on sale_margin to supplierinfo

mourad-ehm:
Improve pricelist_supplierinfo. Add sale price based on sale_margin on product supplier info.
Its allows to simply manage several price list without redefine  sale price on each pricelist (only add discount for example).

### DIFF
--- a/product_pricelist_supplierinfo/models/product_supplierinfo.py
+++ b/product_pricelist_supplierinfo/models/product_supplierinfo.py
@@ -1,0 +1,22 @@
+# Copyright 2020 Akretion - Mourad EL HADJ MIMOUNE
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+
+from odoo import api, fields, models
+
+
+class ProductSupplierinfo(models.Model):
+    _inherit = 'product.supplierinfo'
+
+    sale_margin = fields.Float(
+        'Sale Margin', default=0, digits=(16, 2),
+        help="Margin to apply on price to obtain sale price")
+
+    @api.multi
+    def _get_supplierinfo_pricelist_price(self):
+        self.ensure_one()
+        sale_price = self.price
+        if self.sale_margin:
+            sale_price = (
+                (self.price + (self.price * (self.sale_margin / 100)))
+                or 0.0)
+        return sale_price
--- a/product_pricelist_supplierinfo/readme/DESCRIPTION.rst
+++ b/product_pricelist_supplierinfo/readme/DESCRIPTION.rst
@@ -1,3 +1,6 @@
 This module allows you to create a sales pricelist based on product
 supplierinfo prices. If you want, you can bypass minimum quantity in pricelist
 item.
+
+We can also define sale marging applied on purchase price directly on supplier info.
+For this, you must add users to "Show sale margin on Product Supplierinfo" group.
--- a/product_pricelist_supplierinfo/views/product_supplierinfo_view.xml
+++ b/product_pricelist_supplierinfo/views/product_supplierinfo_view.xml
@@ -1,0 +1,31 @@
+<?xml version="1.0" encoding="utf-8"?>
+<odoo>
+
+    <record id="product_supplierinfo_tree_view" model="ir.ui.view">
+        <field name="name">product.supplierinfo.tree.view.pricelist</field>
+        <field name="model">product.supplierinfo</field>
+        <field name="inherit_id" ref="product.product_supplierinfo_tree_view"/>
+        <field name="groups_id" eval="[(4,ref('product_pricelist_supplierinfo.group_supplierinfo_pricelist_sale_margin'))]"/>
+        <field name="arch" type="xml">
+            <field name="price" position="after">
+                <field name="sale_margin" string="Sale margin"/>
+            </field>
+        </field>
+    </record>
+
+    <record id="product_supplierinfo_form_view" model="ir.ui.view">
+        <field name="name">product.supplierinfo.form.view.pricelist</field>
+        <field name="model">product.supplierinfo</field>
+        <field name="inherit_id" ref="product.product_supplierinfo_form_view"/>
+        <field name="groups_id" eval="[(4,ref('product_pricelist_supplierinfo.group_supplierinfo_pricelist_sale_margin'))]"/>
+        <field name="arch" type="xml">
+            <xpath expr="//label[@for='date_start']" position="before">
+                <label for="sale_margin"/>
+                <div>
+                    <field name="sale_margin" class="oe_inline"/>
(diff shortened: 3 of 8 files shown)


mourad-ehm:
Hi @eLBati, @pedrobaeza, 
New PR

pedrobaeza:
I don't get why you need this if the standard formula feature allows you to do it...

mourad-ehm:
Standard formula don't distinguish between discount applied on sale price (which can appear on sale order) and margin applied on purchase price (in several case customer must not see the margin we apply but only discount).
With this feature we can define a public price base on on sale margin in public price list and discount applied for a customer by a specific price list (to cleary separate disscount and margin).
For some Client Seller can negotiates price with customers and apply discount but he must not see sale margin (saler take on count only discount not margin).

pedrobaeza:
Sorry, but I don't understand the difference. Are you talking about the visibility of the discount?

mourad-ehm:
Example:
Purchase price = 10 
If we want to apply a margin by using list price formula (with out this functionality) we must define a negative discount  
price = (price - (price * (rule.**price_discount** / 100))) or 0.0  price = 10 - (10 * -30/100) = 13
If we display price list discount on sale line we take -30 (which is the margin applied on purchase price but not a discount).
If I want applied a discount on this price I must recompute de discount by taking in account margin or define a new price list base on the first one.

mourad-ehm:
ping @sebastienbeau

sebastienbeau:
@pedrobaeza some extra information regarding our case.

Our customer have a lot of product_supplier_info (around 5 per product same supplier but with different price depending of the qty), the margin (and so the sale price) will depend of the qty

example
qty, price, marging
1,    10,     40%
10,   8,      35%
100, 6,      35%
...

Using pricelist item to ajust the marge mean creating 5 pricelist.item per product (not super fast in the UI, and make import file more complex).

So for making it simpler we directly add the margin on the product.supplier.info.

If you think it make too complex the UI when you do not need it, we can add a group to hide the margin ? (if you want to set the margin on the product.supplier.info you just need to activate the group)

What do you think ?

pedrobaeza (changes requested):
Thanks for the extra information. Now I get the need and indeed it wouldn't be comfortable to have the 5 product.supplierinfo records + 5 product.pricelist.item with the minimum quantity and the margin.

It's OK for me to add it then, but please:

- Don't use the extra compute field, but apply it on the method `_get_supplierinfo_pricelist_price`. It doesn't add value to spread the computation code across files and overload registry with fields for only one use. If you plan to override that computation, then it's OK for me to add a hook method to get the price. It's also important to not apply always the margin if it's 0 or 100.
- Add the security group as you propose, for not complicating the UI for regular users.
- Add instructions on README about the new feature.

Thanks also for the extra test exercising the new feature!

product_pricelist_supplierinfo/models/product_supplierinfo.py
```diff
@@ -11,15 +10,11 @@ class ProductSupplierinfo(models.Model):
...
-    @api.depends('price', 'sale_margin')
-    def _compute_product_price(self):
-        for product in self:
-            product.sale_price = (
-                (product.price + (product.price * (product.sale_margin / 100)))
-                or 0.0)
+    @api.multi
+    def _get_supplierinfo_pricelist_price(self):
+        self.ensure_one()
+        sale_price = (
```
pedrobaeza: Please don't do the math if sale_margin == 100 or 0

pedrobaeza: ...or 0

mourad-ehm: Why you ask to exclude sale_margin == 100 or 0 ? I think that the expression is correct for all values.

pedrobaeza: Yes, but it overheads the computation without reason. If 0, result = price. If 100, result = 0, isn't it?

pedrobaeza: Also I think you can have some rounding problem doing the math

mourad-ehm: Here we apply sale margin but not discount. If 0 result = price (OK) But if 100 result = 2 * price (price + price).

pedrobaeza: OK, then just `if margin: <operation> else: price`

mourad-ehm: Ok

product_pricelist_supplierinfo/views/product_supplierinfo_view.xml
```diff
@@ -0,0 +1,28 @@
...
+<odoo>
+
+    <record id="product_supplierinfo_tree_view" model="ir.ui.view">
+        <field name="name">product.supplierinfo.tree.view.pricelist</field>
+        <field name="model">product.supplierinfo</field>
+        <field name="inherit_id" ref="product.product_supplierinfo_tree_view"/>
+        <field name="arch" type="xml">
+            <field name="price" position="after">
+                <field name="sale_margin"
+                   groups="product_pricelist_supplierinfo.group_supplierinfo_pricelist_sale_margin"/>
```
pedrobaeza: You can put the group at view level, so that is not loaded at all and you optimize a bit that part.

sebastienbeau:
The label for the marge is missing in the view form
image

sebastienbeau (approved):
Code review and test

sebastienbeau:
@pedrobaeza can you update your review? thanks

sebastienbeau:
Wait I am going to fix the warning

sebastienbeau:
I will squash the commit

sebastienbeau:
Commit squashed, should be ready for merging

product_pricelist_supplierinfo/readme/DESCRIPTION.rst
```diff
@@ -1,3 +1,5 @@
 This module allows you to create a sales pricelist based on product
 supplierinfo prices. If you want, you can bypass minimum quantity in pricelist
 item.
+We can also define sale marging applied on purchase price directly on supplier info.
```
pedrobaeza, line 5: An empty line in between is needed here for proper RST syntax

pedrobaeza (approved):
Little comment, but OK for me

product_pricelist_supplierinfo/readme/DESCRIPTION.rst
sebastienbeau: fixed, thanks for the catch

PR #665, opened 2020-08-21
[12.0][ADD] product_supplierinfo_group

kevinkhao:
Product price values can be tedious and error-prone to enter one by one when you need to repeat the same values:

* vendor,
* product name,
* variant
* product code

Whereas most of the time you just want to enter the minimum quantities, price, dates, delay.

With this module, price lines look to their parent (product.supplierinfo.group) in order to get the values from these repetitive fields.

In addition, it adds a computed field that summarizes all price/quantities deals.

Here are some screenshots for before/after comparison.

Before:
before_1
Boring !
before_2

After:
after_1
Much better
after_2


sebastienbeau:
@legalsylvain maybe this PR can interest you ;)

bealdav:
Ergonomy ideas: 

1/ make it compatible with this module https://github.com/OCA/web/tree/12.0/web_tree_duplicate#configuration : set duplicate=True
2/ Use double column per row for fields in supplier info group

To discuss

kevinkhao:
> Ergonomy ideas:
> 
> 1/ make it compatible with this module https://github.com/OCA/web/tree/12.0/web_tree_duplicate#configuration : set duplicate=True
> 2/ Use double column per row for fields in supplier info group
> 
> To discuss

Thanks, view change is applied

product_supplierinfo_group/models/product_supplierinfo_group.py
```diff
@@ -0,0 +1,66 @@
...
+
+    product_tmpl_id = fields.Many2one("product.template", required=True)
+    supplierinfo_ids = fields.One2many("product.supplierinfo", "supplierinfo_group_id")
+    product_id = fields.Many2one(
+        "product.product",
+        "Product Variant",
+        help="If not set, the vendor price will apply to all "
+        "variants of this product.",
+    )
+    name = fields.Many2one(
```
sebastienbeau: I know that odoo use "name" for the link to the partner but as it's an new object I prefers to use "partner_id" here.

rousseldenis: Agree

kevinkhao: done

product_supplierinfo_group/models/product_supplierinfo.py
```diff
@@ -0,0 +1,69 @@
...
+    "product_id": "product_id",
+    "product_name": "product_name",
+    "product_code": "product_code",
+}
+
+
+class ProductSupplierinfo(models.Model):
+    _inherit = "product.supplierinfo"
+
+    supplierinfo_group_id = fields.Many2one("product.supplierinfo.group", required=True)
```
rousseldenis: I think you maybe needs to set a default (see runbot warning)

kevinkhao: @rousseldenis thanks for the remark
This meant a bit more head scratching than expected :)
see: https://github.com/OCA/product-attribute/pull/665/commits/46a04187fbd6595c6db4626eb2ef0a4ad0561f94

product_supplierinfo_group/views/product.xml
```diff
@@ -0,0 +1,21 @@
+<?xml version="1.0" encoding="utf-8"?>
+<odoo>
+    <record id="product_template_tree_view" model="ir.ui.view">
+        <field name="model">product.template</field>
+        <field name="inherit_id" ref="product.product_template_only_form_view"/>
+        <field name="arch" type="xml">
+            <field name="seller_ids" position="replace">
```
sebastienbeau: Hi, you should avoid using replace as it can break other module set invisible instead

Moreover there is two field seller_ids and variant_seller_ids we should hide both see here

https://github.com/OCA/OCB/blob/12.0/addons/purchase/views/product_views.xml#L14

product_supplierinfo_group/models/product_supplierinfo.py
```diff
@@ -0,0 +1,74 @@
...
+    product_tmpl_id = fields.Many2one(
+        related="supplierinfo_group_id.product_tmpl_id", store=True
+    )
+    name = fields.Many2one(
+        related="supplierinfo_group_id.partner_id", required=False, store=True
```
nikul-serpentcs: @kevinkhao Here no need to add `required=False`, Because by default `required False`.

product_supplierinfo_group/models/product_supplierinfo_group.py
```diff
@@ -0,0 +1,59 @@
+# Copyright (C) 2012 Savoir-faire Linux (<http://www.savoirfairelinux.com>).
+# Copyright (C) 2015-Today GRAP (http://www.grap.coop)
+# Copyright (C) 2020-Today Akretion (http://www.akretion.com)
+# @author Sylvain LE GAL (https://twitter.com/legalsylvain)
+# License GPL-3.0 or later (http://www.gnu.org/licenses/agpl).
```
nikul-serpentcs:
```suggestion
# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
```

product_supplierinfo_group/models/product_supplierinfo.py
kevinkhao: @nikul-serpentcs thanks for the catch

product_supplierinfo_group/models/product_supplierinfo_group.py
kevinkhao: done

product_supplierinfo_group/models/product_supplierinfo.py
kevinkhao: actually we do need to set it to required=False https://github.com/odoo/odoo/blob/d87111ae3e3cf3d654180c986745623ebe731667/addons/product/models/product.py#L743 if we don't then we have problems when trying to create a new record

sebastienbeau:
@kevinkhao I had pushed a fix when installing on an empty db.
Also never ever use cr.commit() do it only if you have a really really good reason do to it (really really rare case maybe 1 time every 10 years)

kevinkhao:
@sebastienbeau noted with thanks
it still gets stuck on the runbot, adding a temporary trivial commit to regenerate logs (they are currently inaccessible)
I also just realized I didn't yet do a PR for the product_supplierinfo_group_margin yet because of the red status

sebastienbeau (approved):
Code review and test.
Everything is green now

bealdav (changes requested):
Thanks a lot for this module. Just tested, it works very well.
Just a point

You should define a domain for product_id because here it's possible to mix product_id with a product.template which don't have a matching product_tmpl_id.

Selection_080
bolt != desk

kevinkhao:
> Thanks a lot for this module. Just tested, it works very well.
> Just a point
> 
> You should define a domain for product_id because here it's possible to mix product_id with a product.template which don't have a matching product_tmpl_id.
> 
> Selection_080
> bolt != desk

Thank you for the review.

What you are saying makes sense. In the last commit I tried to do this restriction. It works great on product.product, but not on product.template. I am suspecting it is a limitation of the JS side of Odoo, more specifically the popup of when you click the One2many widget on "Add a line..."

Something funky happens when you reference the parent. In our case it is product_tmpl_id on product.supplierinfo.group. Because product_tmpl_id is on our popup and refers to the active_id of the screen that is behind the popup. Then the onchange seems broken.

From the product template screen:

Screenshot from 2020-09-23 21-41-37

Nothing happens.

From the product.product screen:

Screenshot from 2020-09-23 21-41-51

Works great.


I will need to spend a bit more time thinking about it. Do you have some idea ?

kevinkhao:
Fixed the issue.

The problem was due to the fact that in a One2many field you need to specify in context the default value of the parent (here the child is supplierinfo.group and the parent is product_tmpl_id) if you want to use it in domains on onchanges. Or else, it will only be set on write (so, only when you click "Save", before that you won't have access to it for onchanges/domains...).

bealdav (approved):
LGTM, thanks fot the fix

PR #667, opened 2020-08-27
[FIX] product_pricelist_supplierinfo: variant domain

ruuter:
Before this commit, all supplierinfo records with product_tmpl_id set,
were considered as possible price for all product variants, even if
product_id was set also. That resulted in same price for all variants
even if supplierinfo records had different price for all variants.

Now, if product variant is set on a supplier record, only that is taken
into consideration for respective variant. If only template is set, it
will still be considered.

Maybe worth to mention, that it removing template relation from supplierinfo record sort of fixed the issue, but Odoo UI does not even allow selecting variant relation without template set, because of the filter domain on variant field. Also, doing that introduced permission issues down the line, resulting in all prices being 0 for non internal users.

### DIFF
--- a/product_pricelist_supplierinfo/models/product_template.py
+++ b/product_pricelist_supplierinfo/models/product_template.py
@@ -13,41 +13,15 @@ class ProductTemplate(models.Model):
             self, rule, date=None, quantity=None, product_id=None):
         """Method for getting the price from supplier info."""
         self.ensure_one()
+        price = 0.0
+        product = self.product_variant_id
         if product_id:
-            domain = [
-                '|',
-                ('product_id', '=', product_id),
-                ('product_tmpl_id', '=', self.id),
-            ]
-        else:
-            domain = [
-                ('product_tmpl_id', '=', self.id),
-            ]
-        if not rule.no_supplierinfo_min_quantity and quantity:
-            domain += [
-                '|',
-                ('min_qty', '=', False),
-                ('min_qty', '<=', quantity),
-            ]
-        if date:
-            domain += [
-                '|',
-                ('date_start', '=', False),
-                ('date_start', '<=', date),
-                '|',
-                ('date_end', '=', False),
-                ('date_end', '>=', date),
-            ]
-        # We use a different default order because we are interested in getting
-        # the price for lowest minimum quantity if no_supplierinfo_min_quantity
-        supplierinfos = self.env['product.supplierinfo'].search(
-            domain, order='min_qty,sequence,price',
-        )
+            product = product.browse(product_id)
         if rule.no_supplierinfo_min_quantity:
-            selected_supplierinfo = supplierinfos[:1]
-        else:
-            selected_supplierinfo = supplierinfos[-1:]
-        price = selected_supplierinfo._get_supplierinfo_pricelist_price()
+            quantity = 1.0
+        seller = product._select_seller(quantity=quantity, date=date)
+        if seller:
+            price = seller._get_supplierinfo_pricelist_price()
         if price:
             # We have to replicate this logic in this method as pricelist
             # method are atomic and we can't hack inside.
--- a/product_pricelist_supplierinfo/tests/test_product_supplierinfo.py
+++ b/product_pricelist_supplierinfo/tests/test_product_supplierinfo.py
@@ -2,6 +2,8 @@
 # Copyright 2018 Tecnativa - Pedro M. Baeza
 # License AGPL-3 - See http://www.gnu.org/licenses/agpl-3.0.html
 
+from datetime import date
+
 from odoo.tests import common
 
 
@@ -34,18 +36,6 @@ class TestProductSupplierinfo(common.SavepointCase):
                 }),
             ],
         })
-        cls.product.write({'seller_ids': [
-            (0, 0, {
-                'name': cls.supplier1.id,
-                'min_qty': 5,
-                'price': 50,
-            }),
-            (0, 0, {
-                'name': cls.supplier2.id,
-                'min_qty': 1,
-                'price': 10,
-            }),
-        ]})
         cls.pricelist = cls.env['product.pricelist'].create({
             'name': 'Supplierinfo Pricelist',
             'discount_policy': 'without_discount',
@@ -117,7 +107,7 @@ class TestProductSupplierinfo(common.SavepointCase):
         )[0].date_start = '2018-12-31'
         self.assertAlmostEqual(
             self.pricelist.get_product_price(
-                self.product, 5, False, date='2019-01-01',
+                self.product, 5, False, date=date(2019, 1, 1),
             ), 50,
         )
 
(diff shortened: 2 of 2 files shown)


pedrobaeza:
I don't really get the problem. Please include a test exercising the problem (failing before this fix).

ruuter:
@pedrobaeza Sorry if I'm not clear. Problem is that variant based prices are not considered. The domain is wrong. It selects all records for a template, not for specific variant. But it should select supplierinfo records that match specific variant, not all for that template. As a result end price is same for all variants.

Okay, I will add test case.

ruuter:
@pedrobaeza Test case added. Also, fixed a mistake a made in my original fix.

Current domain now selects correct records, except in case template id is not set on supplierinfo record, but I think that is fine? It does not make sense to have variant set and not template anyway and Odoo UI does not let you do it easily also.

pedrobaeza:
I'm not sure with this about the priority, as if you have a supplierinfo for the template and another for the variant, the first one may be used. I'm also thinking why I didn't use `_select_seller` method that I remember it exists...

ruuter:
@pedrobaeza Right, that is true, but its same now, isn't it? Current domain also selects both, for template or variant and later selects first or last. That does not change. But currently, it selects also records of other variants, which makes it unusable if you have different prices per variant.

ruuter:
@pedrobaeza Is there anything else I could do to get this merged or a reason why not?

pedrobaeza:
What about my comment about `_select_seller`? Anyway, this requires at least 2 approvals.

ruuter:
I am not sure what you meant by that comment. Do you mean you would prefer a re-write using `_select_seller` instead my quick fix? This is more work. Hoped my fix is ok for now, as it should not break any existing code.

pedrobaeza:
Yes, I try to standardize the way the supplierinfo is selected, as core code doesn't have this bug.

ruuter:
Okay, I try to do that when I have time.

ruuter:
@pedrobaeza How does it look now? Also, why is Codecov failing? I'm not sure how to fix that.

pedrobaeza:
Seems better, but several remarks:

- Don't add empty lines in the middle of the method. They are discouraged by PEP8 itself.
- Why doing that weird date operation? Which backward  compatibility is that? In any case, you can convert directly to date, but I don't see it needed.

ruuter:
Okay, will remove empty lines.

Date operation is there to comply with current tests which all call the method with string, while _select_seller expects date object.

pedrobaeza:
Well, I think tests should be adapted, as Odoo is handling everything already as dates, but if you don't want that extra charge, please convert at least directly to date: `from datetime import date; date.strptime(...)`

ruuter:
Ah okay, that's what you meant by directly, but sure I agree, better change the tests. I was just not sure if it's okay to do that.

pedrobaeza:
@eLBati can you check if this change also satisfies your problem?

ruuter:
> from datetime import date; date.strptime(...)

@pedrobaeza date object do not have strptime method. See: https://docs.python.org/3/library/datetime.html#strftime-strptime-behavior

pedrobaeza:
OK, I didn't know that. Maybe last time I did that was in P2.7 (or I'm making up that feature...).

ruuter:
So I removed all empty lines now and moved the date conversion to test instead.

product_pricelist_supplierinfo/tests/test_product_supplierinfo.py
```diff
@@ -117,7 +119,8 @@ def test_pricelist_dates(self):
         )[0].date_start = '2018-12-31'
         self.assertAlmostEqual(
             self.pricelist.get_product_price(
-                self.product, 5, False, date='2019-01-01',
+                self.product, 5, False,
+                date=datetime.strptime('2019-01-01', '%Y-%m-%d').date(),
```
pedrobaeza: If you want to do it simpler:
```python
from datetime import date
...
date=date(2019, 01, 01)
```

ruuter: Ah yes, of course, much better :) Tnx!

pedrobaeza:
@sergio-teruel you studied also the method `_select_seller`. Do you see any problem?

product_pricelist_supplierinfo/models/product_template.py
```diff
@@ -13,40 +13,12 @@ def _get_supplierinfo_pricelist_price(
...
         if product_id:
-            domain = [
-                '|',
-                ('product_id', '=', product_id),
-                ('product_tmpl_id', '=', self.id),
-            ]
-        else:
-            domain = [
-                ('product_tmpl_id', '=', self.id),
-            ]
```
sergio-teruel, line 25: I think that this part of domain can be replaced by 
                "|",
                ("product_id", "=", self.id),
                "&",
                ("product_tmpl_id", "=", self.product_tmpl_id.id),
                ("product_id", "=", False),

pedrobaeza: @sergio-teruel we are removing all that domain stuff and using `_select_seller` instead.

sergio-teruel:
I think that the problem is when for a product that has more than one variants and you defined a seller record for a template and other for only one variant.

pedrobaeza:
`_select_seller` is not doing that correctly?

product_pricelist_supplierinfo/models/product_template.py
sergio-teruel: yeahhh I saw!! I have done a comment

product_pricelist_supplierinfo/models/product_template.py
```diff
@@ -13,40 +13,12 @@ def _get_supplierinfo_pricelist_price(
             self, rule, date=None, quantity=None, product_id=None):
         """Method for getting the price from supplier info."""
         self.ensure_one()
+        product = self.product_variant_id
```
sergio-teruel, line 17: But... If i have more than one...???

product_pricelist_supplierinfo/tests/test_product_supplierinfo.py
```diff
@@ -133,3 +135,37 @@ def test_pricelist_based_price_round(self):
...
+        })
+        attr = self.env.ref('product.product_attribute_1')
+        attr_val1 = self.env.ref('product.product_attribute_value_1')
+        attr_val2 = self.env.ref('product.product_attribute_value_2')
+        self.env['product.template.attribute.line'].create({
+            'product_tmpl_id': tmpl.id,
+            'attribute_id': attr.id,
+            'value_ids': [(6, 0, [attr_val1.id, attr_val2.id])],
+        })
+        tmpl.create_variant_ids()
```
kos94ok-3D: Why not use?
<pre>
tmpl = self.env['product.template'].create({
    'name': 'Test Product',
    'attribute_line_ids': [[0, 0, {
        'attribute_id': self.env.ref('product.product_attribute_1').id,
        'value_ids': [
            (4, self.env.ref('product.product_attribute_value_1').id),
            (4, self.env.ref('product.product_attribute_value_2').id),
        ],
    ],
})
</pre>

product_pricelist_supplierinfo/models/product_template.py
ruuter: hmm... then you should not use template to get the variant price? I mean, if you have variant based prices, what price are you expecting to get when asking for template price?

product_pricelist_supplierinfo/tests/test_product_supplierinfo.py
ruuter: Can't find any reason why not. Yours is indeed more compact. I will change it. Thanks.

eLBati (changes requested):
Testing the use case described at https://github.com/OCA/product-attribute/pull/580 I get
```
  File "/home/elbati/workspace/odoo/instances/odoo12/parts/product-attribute/product_pricelist_supplierinfo/models/product_supplierinfo.py", line 16, in _get_supplierinfo_pricelist_price
    self.ensure_one()
  File "/home/elbati/workspace/odoo/instances/odoo12/parts/odoo/odoo/models.py", line 4749, in ensure_one
    raise ValueError("Expected singleton: %s" % self)
ValueError: Expected singleton: product.supplierinfo()
```

I suggest to proceed with https://github.com/OCA/product-attribute/pull/580 and evaluate this work about variant domain afterwards, if needed

pedrobaeza:
@eLBati it's a nonsense to change something that will be potentially removed, so it's better to proceed with this.

ruuter:
@eLBati I just discovered this error myself also. Should be fixed now. Please test again. At least b4cb1365383b4c76d9ac27d259d74685f9709268 should be solved by using _select_seller, if I understand your issue correctly.

The error came from the commit c7c838d7 merged meanwhile. It expects singleton, but _select_seller is not guaranteed to return any.

product_pricelist_supplierinfo/models/product_template.py
```diff
@@ -13,41 +13,16 @@ def _get_supplierinfo_pricelist_price(
...
         if rule.no_supplierinfo_min_quantity:
-            selected_supplierinfo = supplierinfos[:1]
+            quantity = 1.0
+        seller = product._select_seller(quantity=quantity, date=date)
+        if seller:
+            price = seller._get_supplierinfo_pricelist_price()
         else:
-            selected_supplierinfo = supplierinfos[-1:]
-        price = selected_supplierinfo._get_supplierinfo_pricelist_price()
+            price = 0.0
```
CarmenMiranda: Hi @ruuter I think you can delete this `else` and set the price before the seller, something like this:

```python
price = 0.0
seller = product._select_seller(quantity=quantity, date=date)
if seller:
    price = seller._get_supplierinfo_pricelist_price()
```
This way you avoid the need of doing another test to cover it and the warning of the coverage will be discarded

CarmenMiranda (changes requested):

product_pricelist_supplierinfo/models/product_template.py
ruuter: Hmm okay thanks! Did not know it affects coverage.

CarmenMiranda: You're welcome :smiley: It was affecting because there wasn't a test to cover the else so the line was not covered and a warning appeared but now it is covered and no warning is shown

CarmenMiranda:
Hi @pedrobaeza I checked the PR days ago, I only did one comment and it was resolved, other than that it's LGTM :+1:

eLBati (approved):
Case described at https://github.com/OCA/product-attribute/pull/580 is resolved, thanks.

Extracting "Allow to only match prices from selected supplier" from https://github.com/OCA/product-attribute/pull/580 with https://github.com/ruuter/product-attribute/pull/1 , in case you wanted to add it here

pedrobaeza:
Let's merge it and then we continue the PRs over this base. Thanks all for your involvement.

/ocabot merge major

eLBati:
:+1: created https://github.com/OCA/product-attribute/pull/694

PR #675, opened 2020-09-03
[FIX] [13.0] product_dimension Contributors website

emagdalenaC2i:


emagdalenaC2i:
@OCA/product-maintainers Please could you merge this PR?

rousseldenis:
@emagdalenaC2i Is that wanted to change to http ? The actual trends is to force https as standard browsers are going to unsupport http urls.

If you have issues with certificates, you can have a look at letsencrypt. Free one !

emagdalenaC2i:
I know, but due to some internal developments (that actually don't work with https) we will keep the http and don't change to https for now.

And as you can see the link https://www.c2i.es doesn't work

PR #676, opened 2020-09-14
[FIX] product_brand: fix product form view

len-foss:


marcelsavegnago (approved):

rousseldenis (approved):

pedrobaeza:
Can you put the before and after for showcasing the problem?

len-foss:
@pedrobaeza I think this can be closed; the problem I tried to solve was probably caused by another module. 
That morning I checked on the runbot and thought it was also broken there but I was probably not completely awake, because it seems completely fine.
I'll recheck tomorrow and close it if I don't find a case where it breaks.

pedrobaeza:
OK, that was the reflection I as looking for, as I haven't seen such problem :smile: 

It can be a compatibility problem between several OCA modules though, and if the case, we can check both and find best solution, although probably is in the other part (not in this module) the non-compatible modification. If it's an incompatibility with a custom module, the same applies, but if the non-compatible modification is on the custom, then it should be fixed there.

len-foss:
Yeah of course. 
Since the view definition was weird I assumed it came from that and stopped my investigation, but I was completely off course. Sorry for the noise...

pedrobaeza:
OK, no worry. Waiting your confirmation for closing this.

PR #677, opened 2020-09-14
[13.0] Add product_barcode_required

simahawk:
Make product barcode required when enabled.


product_barcode_required/__manifest__.py
```diff
@@ -0,0 +1,15 @@
+# Copyright 2020 Camptocamp SA
+# License LGPL-3.0 or later (https://www.gnu.org/licenses/lgpl)
+{
+    "name": "Product barcode required",
+    "summary": "Make product barcode required when enabled",
+    "version": "13.0.1.0.0",
+    "development_status": "Alpha",
```
guewen: I don't think it should be alpha, this module shouldn't change too much.

guewen (approved):

product_barcode_required/models/product.py
```diff
@@ -0,0 +1,57 @@
...
+        res = super().write(vals)
+        self._check_barcode_required()
+        return res
+
+    def _check_barcode_required(self):
+        for rec in self:
+            if rec._is_barcode_required_enabled() and rec._is_barcode_required():
+                raise exceptions.ValidationError(
+                    _("Product '{0.name}' has no barcode!").format(rec)
+                )
```
len-foss: The suggested change is a bit more verbose, but in general it seems an anti-pattern to only have the first error (it's like running test with the failfast option always enabled...). 
```suggestion                        
bad_products = self.filtered(
    lambda p: ._is_barcode_required_enabled() and rec._is_barcode_required()
)
if bad_products:
    names = ", ".join(n for n in bad.products.mapped("name") if n)
    names_str = names if len(names) < 300 else names[:300] + "..."
    message = _("The following products have no barcode:")
    raise exceptions.ValidationError(message + "\n" + names_str)
```

product_barcode_required/models/product.py
```diff
@@ -0,0 +1,57 @@
+# Copyright 2020 Camptocamp SA
+# License LGPL-3.0 or later (https://www.gnu.org/licenses/lgpl)
+
+from odoo import _, api, exceptions, models
+
+
+class BarcodeRequiredMixin(models.AbstractModel):
```
len-foss: I tend to prefer having very small models in the same file, but I've been told a few times the OCA code should respect the 1model/1class/1file.

product_barcode_required/tests/test_barcode.py
```diff
@@ -0,0 +1,91 @@
...
+        """Requirement enabled, default barcode to default_code."""
+        form = Form(self.env["product.product"])
+        form.name = "Prod A"
+        form.default_code = "PROD-A"
+        self.assertEqual(form.barcode, "PROD-A")
+        record = form.save()
+        self.assertEqual(record.barcode, "PROD-A")
+
+    def test_validation_create(self):
+        """Can create a record w/out barcode since the constraint is not enabled."""
```
len-foss: Can not... is enabled

product_barcode_required/tests/test_barcode.py
```diff
@@ -0,0 +1,91 @@
...
+        with self.assertRaises(exceptions.ValidationError) as err:
+            self.env["product.product"].create(
+                {"name": "Variant A", "default_code": "VAR-A"}
+            )
+
+        self.assertEqual(err.exception.name, "Product 'Variant A' has no barcode!")
+
+    def test_validation_write(self):
+        """Can create a record w/out barcode since the constraint is not enabled."""
```
len-foss: Can not remove the barcode of a product since the constraint is enabled

len-foss (approved):
Code review: LGTM (with a few remarks)

simahawk:
@guewen @len-foss I've refactored and improved/fixed some parts (eg: no fail on automatic creation from template, make field required in views, etc). Could you review again?

len-foss (approved):
:+1:

guewen (approved):

product_barcode_required/models/product_barcode_mixin.py
```diff
@@ -0,0 +1,40 @@
...
+                rec.barcode = rec.default_code
+
+    @api.depends("type", "barcode")
+    def _compute_is_barcode_required(self):
+        for rec in self:
+            rec.is_barcode_required = (
+                rec._is_barcode_required_enabled() and rec._is_barcode_required()
+            )
+
+    def _is_barcode_required(self):
```
lmignon: I would prefer an other name for this method.... `_is_barcode_missing`

product_barcode_required/models/product_barcode_mixin.py
```diff
@@ -0,0 +1,40 @@
...
+        self.ensure_one()
+        return self.type != "service" and not self.barcode
+
+    def _is_barcode_required_enabled(self):
+        return self.env.company.product_variant_barcode_required
+
+    def _check_barcode_required(self):
```
lmignon, line 34: @simahawk Why not `api.constraints("barcode")`

product_barcode_required/models/product_barcode_mixin.py
simahawk: renamed

product_barcode_required/models/product_barcode_mixin.py
simahawk: because I don't want to enable it by default. Also, on template is useless since it comes from the 1st variant.

lmignon: In such a case you could add the constrains only into 'product.product'. That will avoid to override `write`and `create` in `product.template` and `product.product`, will remove a lot of code and avoid using the context to pass arguments (IMO it's always a bad practice to use the context in this way even if in > 13, there is no more performance impact when using this kind of trick) 

```python
    _name = "product.product"
    _inherit = ["product.product", "product.barcode.required.mixin"]

    @api.constrains("is_barcode_required")
    def _chek_is_barcode_required(self):
        return super()._chek_is_barcode_required(self):

    _name = "product.template"
    _inherit = ["product.template", "product.barcode.required.mixin"]
```

simahawk: you cannot avoid the ctx key in the template because the template triggers the create/write of product.product when you update variant attributes. BUT at that stage you might have no barcode because the one that you set on the template is applied only to the 1st variant.
Also, to limit the possibility to have errors, I added a call to the onchange in the create/write to make sure we can have a default barcode when possible (maybe here I could make barcode a computed/writeable field).

simahawk: ah, what I could change is to set the key only in `_create_variant_ids`.
Also, AFAIS your constrain won't work if is_barcode_required does not change -> the check must run always.

Regarding bad practice: I do not agree. It's the standard way since the env was introduced in Odoo to propagate flags/vars across models over the same env. If you want to avoid it for perf reasons and you can, fine.
But here I don't see how you can control that particular behavior.

In any case, adding the flag in `_create_variant_ids` reduces the impact. Fine for you?

lmignon: In the same time `barcode` could become a computed field `readonly=False`. That would avoid to have to call to `_onchange_code` into the `create` and `write`method 

```python
    _name = "product.product"
    _inherit = ["product.product", "product.barcode.required.mixin"]

    barcode = fields.Char(compute="_compute_barcode", readonly=False)
 
    @api.depends("default_code", "is_barcode_required")
    def _compute_barcode(self):
        for rec in self:
            if rec.is_barcode_required and rec.default_code:
                rec.barcode = rec.default_code
```
Do we even still need the mixin? `

The code for `is_barcode_required` field could be put into `product.product` and the field makes available into `product.tempate` as related

```python
    _inherit = "product.product"
    
    barcode = fields.Char(compute="_compute_barcode", readonly=False)
    is_barcode_required = fields.Boolean(compute="_compute_is_barcode_required")
 
    @api.depends("default_code", "is_barcode_required")
    def _compute_barcode(self):
        for rec in self:
            if rec.is_barcode_required and rec.default_code:
                rec.barcode = rec.default_code

    @api.depends("type", "barcode")
    def _compute_is_barcode_required(self):
        for rec in self:
            rec.is_barcode_required = (
                rec._is_barcode_required_enabled() and rec._is_barcode_missing()
            )

    def _is_barcode_missing(self):
        self.ensure_one()
        return self.type != "service" and not self.barcode

    def _is_barcode_required_enabled(self):
        return self.env.company.product_variant_barcode_required
    
    @api.constrains("is_barcode_required")
    def _chek_is_barcode_required(self):
        """Check if barcode required, to be used in your own contraint."""
        # Make error nicer up to 30 records.
        failing = [x.display_name for x in self[:30] if x.is_barcode_required]
        if failing:
            failed_list = "\n  * " + "\n  * ".join(failing)
            raise exceptions.ValidationError(
                _("These products have no barcode:\n{}").format(failed_list)
            )

    _inherit = "product.template"

    is_barcode_required = fields.Boolean(related="product_variant_id.is_barcode_required")
    
```

simahawk: barcode computed: read my https://github.com/OCA/product-attribute/pull/677#discussion_r510658087 ;)
mixin: yes, because you want to check on the template as well. Also, you could reuse it for other models having barcodes to make required.
constrains: it does not work as long as is_barcode_required does not change.

lmignon (approved):
Thank you for all your changes @simahawk LGTM (Code review)

lmignon:
@simahawk AFAIK if we want to publish a release on pypi (not a .dev release) we must always bump the version. ( a limitation into runbot odoo...)

rousseldenis:
> @simahawk AFAIK if we want to publish a release on pypi (not a .dev release) we must always bump the version. ( a limitation into runbot odoo...)

Indeed, for migration ones, I personally bump patch for migration PR's.

PR #678, opened 2020-09-16
[12.0][IMP] product_code_unique: Allow to use inactive products

rousseldenis:
Module limitation was it was impossible to create a new product with the same default code as an archived one.

@max3903 @rven @agyamuta Can you confirm this use case is correct ?

### DIFF
--- a/product_code_unique/models/product.py
+++ b/product_code_unique/models/product.py
@@ -8,5 +8,5 @@ class ProductProduct(models.Model):
     _inherit = 'product.product'
 
     _sql_constraints = [
-        ('default_code_uniq', 'unique(default_code)',
+        ('default_code_uniq', 'EXCLUDE (default_code WITH =) WHERE (active = True)',
             'Internal Reference must be unique across the database!'), ]
--- a/product_code_unique/tests/__init__.py
+++ b/product_code_unique/tests/__init__.py
@@ -1,0 +1,1 @@
+from . import test_code_unique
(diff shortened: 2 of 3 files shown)


pedrobaeza:
I prefer to prevent such code duplication, encouraging to re-activate the previous product or empty the code before creating the duplicated one.

rousseldenis:
> I prefer to prevent such code duplication, encouraging to re-activate the previous product or empty the code before creating the duplicated one.

Deactivating product is safe enough and is easy for user (that's the process encouraged by Odoo - see some exceptions on uom or lots for instance). As if you try to reactivate the archived one, constraint will raise.

pedrobaeza:
I don't get you. I try to put the following user case:

- A product with code XXX is deactivated (by mistake or because they think it's not going to be used anymore).
- Then it's needed again and you try to create a new product with same code.
- Currently, it's not allowed, so you have to check inactive products and decide to activate again the product, or to empty that code because is a code reuse.
- With your change, this will be allowed and I think it's not correct.

rousseldenis:
> * Currently, it's not allowed, so you have to check inactive products and decide to activate again the product, or to empty that code because is a code reuse.

With current code, you have to filter on inactive products which is not easy for some users.

pedrobaeza:
You can improve the constraint message for indicating that (something like "Product with duplicated code may be archived").

len-foss:
If I can add some motivation, it seems that this changeset actually makes the module compatible with Odoo's variant management.
If you change attributes on a product (it's a bad idea, granted, but it's an intended use-case), Odoo will archive variants that are already sold and create new ones (in most cases). So before change it could fail; after change it would always work as expected.

However, in the manual case I think Pedro is right, a change in the error message would be better.

pedrobaeza:
Then maybe add some code for removing product's reference of the deactivated records obtained from the Odoo's variant management?

PR #680, opened 2020-09-17
[13.0][FIX] loop on product.pricelist.item fields to avoid errors when updating assortment items

benwillig:


rousseldenis:
@Cedric-Pigeon What do you think of this ? Why removing blacklist.extend(...)?

PR #682, opened 2020-09-22
[12.0][ADD] product_pricelist_show_product_ref

joao-p-marques:
Show Product Internal Code on Product Pricelist

TT25552

ping @pedrobaeza


joao-p-marques:
@victoralmau and @ernestotejeda can you please review this?

product_pricelist_show_product_ref/__manifest__.py
```diff
@@ -0,0 +1,14 @@
...
+    "name": "Show Product Ref on Product Pricelist",
+    "summary": "Show Product Ref on Product Pricelist",
+    "version": "12.0.1.0.0",
+    "development_status": "Beta",
+    "category": "Product",
+    "website": "https://www.github.com/OCA/product-attribute",
+    "author": "Tecnativa, Odoo Community Association (OCA)",
+    "license": "AGPL-3",
+    "installable": True,
+    "depends": ["sale_management"],
```
victoralmau, line 13:
```suggestion
    "depends": ["sale_management"],
    "maintainers": ["joao-p-marques"],
```
Maybe add a maintainer in manifest it's a good idea in future.

joao-p-marques: Ok, thanks for the tip! Already made the changes :+1:

victoralmau (approved):

product_pricelist_show_product_ref/models/product_pricelist.py
```diff
@@ -0,0 +1,20 @@
...
+        if product.product_id:
+            product.name = product.product_id.display_name
+
+    def _get_pricelist_item_name_price(self):
+        super(). _get_pricelist_item_name_price()
+        if len(self) > 1:
+            for product in self:
+                self._set_pricelist_item_name(product)
+        else:
+            self._set_pricelist_item_name(self)
```
ernestotejeda:
```suggestion
    def _get_pricelist_item_name_price(self):
        res = super(). _get_pricelist_item_name_price()
        for record in self.filtered("product_id"):
            record.name = product.product_id.display_name
        return res
```
I think this could be simplified to this, what do you think?
I think also that it's a good practice to get the value of super () ... and then return it even though it returns None (ping @pedrobaeza )

product_pricelist_show_product_ref/tests/test_product_pricelist_show_product_ref.py
```diff
@@ -0,0 +1,85 @@
...
+            'applied_on': '0_product_variant',
+            'product_id': cls.product.id,
+            'compute_price': 'fixed',
+            'fixed_price': 100,
+        })
+
+    def test_code_in_display_name_in_product(self):
+        """Test product item has code in display_name
+        """
+        self.pricelist_item_product_product._get_pricelist_item_name_price()
```
ernestotejeda: Is it neccessary? perhaps you can check diretly test the display_name because 'name' is the computed field that has linked the _get_pricelist_item_name_price method.

ernestotejeda (commented):
Fuctional review: :+1: 
Code review: Ok, but I have a few comments.

product_pricelist_show_product_ref/models/product_pricelist.py
joao-p-marques: In this case, the super() function doesn't return anything indeed, but we can add the call to increase resilience. 
Your other suggestion seems like a good idea. I've tested and it works, so I'm committing it

product_pricelist_show_product_ref/tests/test_product_pricelist_show_product_ref.py
joao-p-marques: Yes, indeed it shouldn't be necessary. However, maybe we can leave it as is to make it explicit that that method is being called, as we are in a test... What do you think @pedrobaeza ?

pedrobaeza: Tests mustn't call directly to internal methods, as this tricks the test itself. The idea of the test is to reproduce the expected behavior, so yes, it's better to call `name_get` than to call the method directly.

joao-p-marques: Ok. Making the changes. Thank you!

joao-p-marques: Does the latest version look better? Or do I still need to make changes?

ernestotejeda: LGTM :)

ernestotejeda (approved):

PR #686, opened 2020-09-28
[IMP] product_template_tags: speed up tag.products_count compute

moylop260:
The tag.products_count field was store=True
If more than one user is adding/removing the same tag to different products
at the same time it will raise concurrencies issues

In fact, if you are configuring/importing your products
this compute could be so slow
Since that each tag will be updated for each product

It was changed to store=False in order to avoid these issues.

The tag.products_count field is displayed in tree view
So it needs to be so faster
Then the compute was changed using a query directly with "group by"
in order to process it faster

Note: Only this compute spend 3 hours in our importation/configuration process

image

### DIFF
--- a/product_template_tags/models/product_template_tag.py
+++ b/product_template_tags/models/product_template_tag.py
@@ -16,7 +16,7 @@ class ProductTemplateTag(models.Model):
         relation='product_template_product_tag_rel',
         column1='tag_id', column2='product_tmpl_id')
     products_count = fields.Integer(
-        string="# of Products", compute='_compute_products_count', store=True)
+        string="# of Products", compute='_compute_products_count')
     company_id = fields.Many2one(
         comodel_name='res.company', string="Company",
         default=lambda self: self._default_company())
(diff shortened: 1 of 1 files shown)


product_template_tags/models/product_template_tag.py
```diff
@@ -28,5 +28,12 @@ def _default_company(self):
     @api.multi
     @api.depends('product_tmpl_ids')
     def _compute_products_count(self):
+        if not self.ids:
+            return
+        self.env.cr.execute("""SELECT tag_id, COUNT(DISTINCT(product_tmpl_id))
```
luisg123v: Could product templates be duplicated for a tag in this table?

product_template_tags/models/product_template_tag.py
```diff
@@ -28,5 +28,12 @@ def _default_company(self):
...
+        FROM product_template_product_tag_rel
```
luisg123v: Would you consider indenting/prettifying the query?

I consider that, when dealing with queries, readability is more relevant than ever.

luisg123v (changes requested):

product_template_tags/models/product_template_tag.py
moylop260: What did you mean?

product_template_tags/models/product_template_tag.py
moylop260: No, they could not. Since that m2m fields have complex unique constraint and not null for each column
It was changed

luisg123v (approved):

patrickrwilson (approved):
Works as intended, LGTM

PR #688, opened 2020-09-29
[12.0] FIX product_multi_price: do not force price names to have a company, as already expected by record rules

eLBati:


pedrobaeza:
The unique constraint will arise for company_id = null?

pedrobaeza:
@chienandalu I remember this was forced for avoiding problems in product table, isn't it?

chienandalu:
Yes, otherwise we couldn't have the same name for different companies

eLBati:
@pedrobaeza @chienandalu Hi, I updated the PR, avoiding to use `company` in `_get_multiprice_pricelist_price`: `price` will be returned according to record rules.

> The unique constraint will arise for company_id = null?

If company_id = null, that price will be seen by every company, so it is correct to avoid to create 2 prices with the same name and company_id = null.

mrcast (approved):
Functional test ok
thanks

pedrobaeza:
Yeah, but I remember the problem: you have a price name called "X" without company, and another with the same name assigned to the current company, and both will be visible, being an error. Why do you need to use same price name record?

mrcast:
> Yeah, but I remember the problem: you have a price name called "X" without company, and another with the same name assigned to the current company, and both will be visible, being an error. Why do you need to use same price name record?

@pedrobaeza : If I'm not mistaken this use case it's controlled
image

eLBati:
I think @pedrobaeza was describing this case:

image

image


but it is not giving error

pedrobaeza:
That's it!

product_multi_price/models/product_product.py
```diff
@@ -15,9 +15,7 @@ class ProductProduct(models.Model):
     def _get_multiprice_pricelist_price(self, rule):
         """Method for getting the price from multi price."""
         self.ensure_one()
-        company = rule.company_id or self.env.user.company_id
-        price = self.env['product.multi.price'].sudo().search([
-            ('company_id', '=', company.id),
```
pedrobaeza, line 20: This was needed for calls to this in sudo mode (which happens in some places, for example, in e-commerce).

eLBati:
> Why do you need to use same price name record?

In our customer's multi company environment, products are shared between companies, so are the prices.
The user should not be forced to write the same price N times, switching between companies in the same product form.

pedrobaeza (changes requested):
Your custom case will break standard behavior, so I'm afraid that this can't be added here by the reasons stated in previous comments. You can add an extra module with that behavior and add in this one the hooks you need.

eLBati:
OkayGuy

eLBati:
:grin:

PR #689, opened 2020-09-29
[ADD] product_lifecycle: add module

GavyMG:
This module applies a part of the product lifecycle management concept by 
adding a status field and replacement product fields that will be used as 
alternatives in sales operations when a product is discontinued.


product_lifecycle/__manifest__.py
```diff
@@ -0,0 +1,24 @@
...
+        "views/sale_order_views.xml",
+        "views/assets.xml",
+    ],
+    "demo": [
+        "demo/replacement_product_demo.xml",
+    ],
+    "qweb": [
+        "static/src/xml/replacement_template.xml",
+    ],
+    "installable": True,
```
luisg123v, line 26: Could you add my user as maintainer, please?

luisg123v (changes requested):
Since this is not a migrated module but a new one, squashing will be required.

product_lifecycle/__manifest__.py
GavyMG: Done. Could you check this MR?

lmignon:
@GavyMG  Part of the proposed logic is already provided  by https://github.com/OCA/product-attribute/tree/12.0/product_state Would it be possible to base your new addon on the first one?

PR #690, opened 2020-09-29
[IMP] product_template_tags: speed up tag.products_count compute

GavyMG:
The tag.products_count field was store=True
If more than one user is adding/removing the same tag to different products
at the same time it will raise concurrencies issues

In fact, if you are configuring/importing your products
this compute could be so slow
Since that each tag will be updated for each product

It was changed to store=False in order to avoid these issues.

The tag.products_count field is displayed in tree view
So it needs to be so faster
Then the compute was changed using a query directly with "group by"
in order to process it faster

FWP of https://github.com/OCA/product-attribute/pull/686/

### DIFF
--- a/product_template_tags/models/product_template_tag.py
+++ b/product_template_tags/models/product_template_tag.py
@@ -19,7 +19,7 @@ class ProductTemplateTag(models.Model):
         column2="product_tmpl_id",
     )
     products_count = fields.Integer(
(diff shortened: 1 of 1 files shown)


moylop260:
It is a approved FowardPort from: https://github.com/OCA/product-attribute/pull/686

PR #668, opened 2020-08-27
[12.0][ADD] product_pricelist_supplierinfo: Partner to get pricelist price

CarmenMiranda:
Adding the option to send a `partner_id` to get the pricelist price of a product and obtaining the `partner_id` of the customer to whom the price is being computed.

PR #694, opened 2020-10-02
[12.0] IMP product_pricelist_supplierinfo allowing to filter suppliers to get prices from

eLBati:


product_pricelist_supplierinfo/tests/test_product_supplierinfo.py
```diff
@@ -38,7 +38,6 @@ def setUpClass(cls):
         })
         cls.pricelist = cls.env['product.pricelist'].create({
             'name': 'Supplierinfo Pricelist',
-            'discount_policy': 'without_discount',
```
pedrobaeza, line 41: Why removing this?

eLBati: Because it is a field added by `sale` and `product_pricelist_supplierinfo` does not depend on `sale`

pedrobaeza: OK, thanks for explaining. I suppose this doesn't have side effects.

product_pricelist_supplierinfo/models/product_pricelist.py
```diff
@@ -44,3 +44,6 @@ class ProductPricelistItem(models.Model):
     no_supplierinfo_min_quantity = fields.Boolean(
         string='Ignore Supplier Info Min. Quantity',
     )
+    filter_supplier_id = fields.Many2one(
```
pedrobaeza, line 47: Does it makes sense to make this a many2many for being more versatile?

eLBati: Hi @pedrobaeza , you can achieve the same goal by creating one rule for every supplier. I think it's more clear for end user.

Our use case is: for all the products supplied by a vendor, apply markup.
With a m2m, what's the meaning of the filter when you have more vendors for the same product?

pedrobaeza: OK

pedrobaeza (approved):

moylop260:
@eLBati 

What do you think about the following comment?
 - https://github.com/OCA/product-attribute/pull/668#issuecomment-707818782

What is the best option for this feature or both could be merged it?

eLBati:
Hi @moylop260 
I see #668 was later closed https://github.com/OCA/product-attribute/pull/668#issuecomment-708868902 
so I think we can proceed with this and evaluate other use cases successively

PR #698, opened 2020-10-04
Update ar.po

Emadbox:
Adding the missed translates


pedrobaeza:
Please use Weblate for this: odoo-community.org/page/translate

PR #701, opened 2020-10-19
[12.0][IMP] product_pricelist_show_product_ref: display product code in pricelists for templates

joao-p-marques:
In the preovious version, the product code was only added for product
variants.
It is now added to product templates as well.

@Tecnativa 
TT26296


joao-p-marques:
@pedrobaeza @sergio-teruel pls review

pedrobaeza (approved):

sergio-teruel (approved):

PR #706, opened 2020-10-27
product_barcode_required: fix required on template form

simahawk:
Barcode on product template should not be required
if you have more than 1 variant.

When more than 1 variant is generated the field is hidden
and you cannot populate it anyway.

This, for instance, prevents duplicating products with more than one variant.


sebalix (approved):

lmignon (commented):
@simahawk Why not an override of the ` _compute_is_barcode_required(self)`method in product_template?

simahawk:
@lmignon thanks for challenging :+1: 
Can you pass by again? Computed method overridden and test added.

simahawk:
@lmignon ping :)

lmignon (approved):
LGTM (code review only)

PR #708, opened 2020-11-04
Add product_uom_updatable module

ajaniszewska-dev:
Add product_uom_updatable module.
Aim of this module is to allow change the product uom if the new uom is in the same category and has the same factor.


product_uom_updatable/models/product_template.py
```diff
@@ -0,0 +1,75 @@
...
+
+
+class ProductTemplate(models.Model):
+
+    _inherit = "product.template"
+
+    def write(self, vals):
+        uom_id = vals.pop("uom_id", False)
+        uom_po_id = vals.pop("uom_po_id", False)
+        res = super(ProductTemplate, self).write(vals)
```
sebalix:
```suggestion
        res = super().write(vals)
```

sebalix (commented):
You are missing the `readme/` template directory with relevant info (description, contributors etc), see https://github.com/OCA/maintainer-tools/tree/master/template/module/readme.

product_uom_updatable/models/product_template.py
```diff
@@ -0,0 +1,75 @@
...
+            new_uom = uom_obj.browse(uom_id)
+            if (
+                key.category_id == new_uom.category_id
+                and key.factor_inv == new_uom.factor_inv
+            ):
+                self.env.cr.execute(
+                    "UPDATE product_template SET uom_id = %(uom)s WHERE id in \
+                     %(product_id)s",
+                    {"uom": new_uom.id, "product_id": tuple(product_id_list)},
+                )
```
guewen, line 41: The cache needs to be invalidated, because it would still contain the previous value.

```
self.invalidate_cache(fnames=["uom_id"], ids=product_id_list)
```

product_uom_updatable/models/product_template.py
```diff
@@ -0,0 +1,75 @@
...
+            new_uom_po = uom_obj.browse(uom_po_id)
+            if (
+                key.category_id == new_uom_po.category_id
+                and key.factor_inv == new_uom_po.factor_inv
+            ):
+                self.env.cr.execute(
+                    "UPDATE product_template SET uom_po_id = %(uom)s WHERE id \
+                    in %(product_id)s",
+                    {"uom": new_uom_po.id, "product_id": tuple(product_id_list)},
+                )
```
guewen: need cache invalidation as well (see my other comment)

product_uom_updatable/models/product_template.py
```diff
@@ -0,0 +1,75 @@
...
+            self._update_uom_po(uom_po_id)
+        return res
+
+    def _update_uom(self, uom_id):
+        uom_obj = self.env["uom.uom"]
+        for key, products_group in groupby(self, key=lambda r: r.uom_id):
+            product_list = list(products_group)
+            product_id_list = []
+            for product in product_list:
+                product_id_list.append(product.id)
```
guewen:
```suggestion
            product_ids = [product.id for product in product_list]
```

On the variable naming: `product_ids` is widely used for a list of ids

product_uom_updatable/__manifest__.py
```diff
@@ -0,0 +1,17 @@
+# Copyright 2020 Acsone SA/NV
+# Copyright 2020 Camptocamp
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+{
+    "name": "Alc Product Uom Updatable",
```
guewen: Alc?

product_uom_updatable/models/product_template.py
```diff
@@ -0,0 +1,75 @@
...
+        res = super(ProductTemplate, self).write(vals)
+        if uom_id:
+            self._update_uom(uom_id)
+        if uom_po_id:
+            self._update_uom_po(uom_po_id)
+        return res
+
+    def _update_uom(self, uom_id):
+        uom_obj = self.env["uom.uom"]
+        for key, products_group in groupby(self, key=lambda r: r.uom_id):
```
guewen: When using `groupby`, records have to be sorted using the same key (unless you actually want to have several groups for the same key if they are not consecutive, but not the case here).
From https://docs.python.org/3.7/library/itertools.html#itertools.groupby:

> The operation of groupby() is similar to the uniq filter in Unix. It generates a break or new group every time the value of the key function changes (which is why it is usually necessary to have sorted the data using the same key function). That behavior differs from SQL’s GROUP BY which aggregates common elements regardless of their input order.

product_uom_updatable/models/product_template.py
```diff
@@ -0,0 +1,75 @@
...
+    _inherit = "product.template"
+
+    def write(self, vals):
+        uom_id = vals.pop("uom_id", False)
+        uom_po_id = vals.pop("uom_po_id", False)
+        res = super(ProductTemplate, self).write(vals)
+        if uom_id:
+            self._update_uom(uom_id)
+        if uom_po_id:
+            self._update_uom_po(uom_po_id)
```
guewen: Is there a reason for updating the uom *after* the call to super? Intuitively, I would have done it before, so if something in `write` is based on the uom (any computation based on it), it would be correctly using the new one.

product_uom_updatable/models/product_template.py
```diff
@@ -0,0 +1,75 @@
...
+                    "UPDATE product_template SET uom_id = %(uom)s WHERE id in \
+                     %(product_id)s",
+                    {"uom": new_uom.id, "product_id": tuple(product_id_list)},
+                )
+            else:
+                raise UserError(
+                    _(
+                        "You can not change the unit of measure of a product \
+                        to a new unit that doesn't have the same category \
+                        and factor"
```
guewen:
```suggestion
                        "You can not change the unit of measure of a product "
                        "to a new unit that doesn't have the same category "
                        "and factor"
```

Prevents inserting blanks

product_uom_updatable/models/product_template.py
```diff
@@ -0,0 +1,75 @@
...
+                    )
+                )
+
+    def _update_uom_po(self, uom_po_id):
```
guewen: couldn't it be the same method (`_update_uom`) with a field name parameter?

product_uom_updatable/__manifest__.py
```diff
@@ -0,0 +1,17 @@
...
+    "summary": """
+        allows products uom to be modified after be used in a stock picking if \
+         the product uom is of the same category""",
+    "version": "13.0.1.0.0",
+    "license": "AGPL-3",
+    "author": "Acsone SA/NV, Odoo Community Association (OCA)",
```
simahawk:
```suggestion
    "author": "ACSONE SA/NV, Camptocamp, Odoo Community Association (OCA)",
```

product_uom_updatable/__manifest__.py
```diff
@@ -0,0 +1,17 @@
...
+    "website": "www.acsone.eu",
```
simahawk:
```suggestion
    "website": "https://github.com/OCA/product-attribute",
```

product_uom_updatable/__manifest__.py
```diff
@@ -0,0 +1,17 @@
...
+    "depends": ["stock"],
+    "data": [],
+    "demo": [],
```
simahawk:
```suggestion
```

product_uom_updatable/tests/test_product_uom_update.py
```diff
@@ -0,0 +1,58 @@
+# Copyright 2020 Acsone SA/NV
+# Copyright 2020 Camptocamp
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+from odoo.tests.common import TransactionCase
```
simahawk: use SavepointCase and disable tracking pls. Like:
```py
cls.env = cls.env(context=dict(cls.env.context, tracking_disable=True))
```

simahawk (changes requested):
same remarks as Guewen

product_uom_updatable/models/product_template.py
```diff
@@ -0,0 +1,50 @@
...
+        if uom_id:
+            self._update_uom(uom_id,"uom_id")
+        if uom_po_id:
+            self._update_uom(uom_po_id,"uom_po_id")
+        res = super().write(vals)
+        return res
+
+    def _update_uom(self, uom_id, field_name):
+        uom_obj = self.env["uom.uom"]
+        sorted_items = sorted(self, key=lambda r: r.uom_id)
```
guewen:
```suggestion
        sorted_items = sorted(self, key=lambda r: r[field_name])
```

product_uom_updatable/models/product_template.py
```diff
@@ -0,0 +1,50 @@
...
+        for key, products_group in groupby(sorted_items, key=lambda r: r.uom_id):
```
guewen:
```suggestion
        for key, products_group in groupby(sorted_items, key=lambda r: r[field_name]):
```

product_uom_updatable/models/product_template.py
```diff
@@ -0,0 +1,53 @@
...
+        if uom_po_id:
+            self._update_uom(uom_po_id, "uom_po_id")
+        res = super().write(vals)
+        return res
+
+    def _update_uom(self, uom_id, field_name):
+        uom_obj = self.env["uom.uom"]
+        sorted_items = sorted(self, key=lambda r: r[field_name])
+        for key, products_group in groupby(sorted_items, key=lambda r: r[field_name]):
+            product_list = list(products_group)
```
simahawk: you don't need to call `list` here

product_uom_updatable/tests/test_product_uom_update.py
```diff
@@ -0,0 +1,59 @@
...
+        self.assertEqual(self.product.uom_id, self.uom_unit)
+        self.assertEqual(self.uom_unit.category_id, self.new_uom.category_id)
+        self.assertEqual(self.uom_unit.factor_inv, self.new_uom.factor_inv)
+        # uom is changed with another uom with the same category
+        self.product.update({"uom_id": self.new_uom.id})
+        self.assertEqual(self.product.uom_id, self.new_uom)
+        # uom_po can also be changed with another uom with the same category
+        self.assertEqual(self.product.uom_po_id, self.uom_unit)
+        self.product.update({"uom_po_id": self.new_uom.id})
+        self.assertEqual(self.product.uom_po_id, self.new_uom)
```
simahawk: you should test the error as well with `self.assertRaises`

product_uom_updatable/models/product_template.py
```diff
@@ -0,0 +1,53 @@
...
+            product_ids = [product.id for product in product_list]
+            new_uom = uom_obj.browse(uom_id)
+            if (
+                key.category_id == new_uom.category_id
+                and key.factor_inv == new_uom.factor_inv
+            ):
+                self.env.cr.execute(
```
simahawk: TBH I don't get why you pop the value and the write it via SQL instead of leaving it in `values`. :thinking: 
Also, couldn't you do this check in a constraint?

ajaniszewska-dev: before writing anything we want to check which parameter (uom_id or uom_po_id) we have in vals so we know what we want to update. We can leave it in values and pass it to this method, but still we will have to check what's inside if we want to use it in query.

product_uom_updatable/i18n/product_uom_updatable.pot
```diff
@@ -0,0 +1,29 @@
+# Translation of Odoo Server.
```
i-vyshnevska: pot file is useless remove it

product_uom_updatable/tests/test_product_uom_update.py
```diff
@@ -0,0 +1,75 @@
...
+                "uom_type": "smaller",
+            }
+        )
+        # verify that the product has stock_moves
+        self.assertTrue(self.product.stock_move_ids)
+        self.assertEqual(self.product.uom_id, self.uom_unit)
+        self.assertEqual(self.uom_unit.category_id, self.new_uom.category_id)
+        self.assertEqual(self.uom_unit.factor_inv, self.new_uom.factor_inv)
+        # uom is changed with another uom with the same category
+        self.product.update({"uom_id": self.new_uom.id})
```
i-vyshnevska: you need to test that updates on product_tmpl_id uom_id will impact product_template by itself and related product_product

product_uom_updatable/models/product_template.py
```diff
@@ -0,0 +1,51 @@
...
+            self._update_uom(uom_id, "uom_id")
+        if uom_po_id:
+            self._update_uom(uom_po_id, "uom_po_id")
+        res = super().write(vals)
+        return res
+
+    def _update_uom(self, uom_id, field_name):
+        uom_obj = self.env["uom.uom"]
+        sorted_items = sorted(self, key=lambda r: r[field_name])
+        for key, products_group in groupby(sorted_items, key=lambda r: r[field_name]):
```
i-vyshnevska, line 28: the  new uom will be the same for all products so you can search for all products,  map their uom at once 
if all products have one category are the same than you can do an update
in this case you don't need to iterate on grouped records, and you will have only one sql request

PR #710, opened 2020-11-06
[IMP] product_brand: faster BI

yajo:
- Index one field that is used in the BI report for grouping, so it goes faster.
- Fix a linter instead of silenting it.

@Tecnativa TT26399


pedrobaeza:
Travis fixed in 9fae87a62ade6ad7496b9264fa0e1139bdae36e1 and fast-tracking as very straight-forward:

/ocabot merge patch

Please FW-port it to v13

PR #714, opened 2020-11-10
[13.0] [MIG] product_lot_sequence

Du-ma:
@AdriaGForgeFlow 

Yo, looks :grapes: ?


rousseldenis (changes requested):
@Du-ma Thanks for your contribution. IMHO as this process is quite critical for lot generation, we should add tests in this migration process. Could you update ? @AdriaGForgeFlow

AdriaGForgeFlow:
I agree with @rousseldenis, tests would help a lot

Du-ma:
@AdriaGForgeFlow @rousseldenis 
Lookin' good?

AdriaGForgeFlow (approved):
Functional and code review :+1:

Du-ma:
I just noticed that if you change the sequence, the fields that should change on product, don't, they stay the same.
Should I fix that?
On the top of that I could probably make fields editable not only on sequence but on the product too (i think that adding a relation would do the trick).
When there is no sequence selected, all sequence related fields could be hidden and a button appear next to the sequence field letting you to auto-generate a new sequence (a button with an icon "+" would do the trick I think).
If I should fix that, should I be doing these changes on `MIG` PR (this one), or should I do them in a new PR (`FIX` or `IMP` not sure) after this one?

gaikaz:
@rousseldenis Could you do another review, please?
Also, Du-ma here is suggesting to make all those fields related to the actual sequence fields, which personally I don't understand why it wasn't done in the first place.
@AdriaGForgeFlow Any insight on this choice?

AdriaGForgeFlow:
I think this might be an interesting approach but I wouldn't do it in the migration PR. IMO I would merge this PR as is and then probably open another PR with the improvements.
Does that make sense to you @rousseldenis @Du-ma

Du-ma:
Yea, that sounds fair to me, hbu @rousseldenis ?

rousseldenis (approved):

PR #723, opened 2020-11-16
[14.0][MIG] product_multi_category

watthanun:


tonygalmiche (approved):
LGTM : Functional review only with runbot

image

product_multi_category/__manifest__.py
```diff
@@ -0,0 +1,15 @@
+# Copyright 2009 Akretion,Guewen Baconnier,Camptocamp,Avanzosc,Sharoon Thomas,Sodexis
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+
+{
+    "name": "Product - Many Categories",
+    "version": "14.0.1.0.0",
```
MeetKD, line 6: @watthanun Summary key is missing.

product_multi_category/models/product.py
```diff
@@ -0,0 +1,16 @@
...
+
+class ProductTemplate(models.Model):
+    _inherit = "product.template"
+
+    categ_ids = fields.Many2many(
+        comodel_name="product.category",
+        relation="product_categ_rel",
+        column1="product_id",
+        column2="categ_id",
+        string="Extra Categories",
```
MeetKD, line 15: Instead of extra categories my suggestion is to change the string to multi categories.

MeetKD (commented):
Minor Change + Suggestion.

joao-p-marques (approved):

product_multi_category/models/product.py
joao-p-marques: IMHO that would make sense if we just had a single field for all the categories. As they are separated between `Product Category` and  `Extra Categories`, I'd say it makes sense to keep it this way.

product_multi_category/__manifest__.py
pedrobaeza: Not really needed.

PR #726, opened 2020-11-19
[13.0] Add product_uom_updatable

ajaniszewska-dev:
Add product_uom_updatable module.
Aim of this module is to allow change the product uom if the new uom is in the same category and has the same factor.


simahawk:
`alc_product_uom_updatable`?

ajaniszewska-dev:
> `alc_product_uom_updatable`?

@simahawk renamed module in both commits.

product_uom_updatable/models/product_template.py
```diff
@@ -0,0 +1,51 @@
...
+                key.category_id == new_uom.category_id
+                and key.factor_inv == new_uom.factor_inv
+            ):
+                # pylint: disable=sql-injection
+                query = (
+                    "UPDATE product_template SET "
+                    + field_name.strip("'")
+                    + " = %s WHERE id in \
+                     %s"
+                )
```
guewen: I thought I had pointed you to https://www.psycopg.org/docs/sql.html#module-psycopg2.sql which is the psycopg2's way to compose queries and deal with sql injections safely.

That would be something like:

```suggestion
                query = sql.SQL(
                    "UPDATE product_template SET {field} "
                    " = %s WHERE id in %s"
                ).format(field=sql.Identifier(field_name))
```

BTW, you don't need `+` to append several strings together, this is automatically concatenated:

```
>>> abc = (
...     "a"
...     "b"
...     "c"
... )
>>>
>>> abc
'abc'
>>>
```

product_uom_updatable/models/product_template.py
```diff
@@ -0,0 +1,51 @@
...
+            self._update_uom(uom_po_id, "uom_po_id")
+        res = super().write(vals)
+        return res
+
+    def _update_uom(self, uom_id, field_name):
+        uom_obj = self.env["uom.uom"]
+        sorted_items = sorted(self, key=lambda r: r[field_name])
+        for key, products_group in groupby(sorted_items, key=lambda r: r[field_name]):
+            product_ids = [product.id for product in products_group]
+            new_uom = uom_obj.browse(uom_id)
```
guewen, line 30: you don't need to instanciate a recordset with browse if you use only the id afterwards

guewen (approved):

product_uom_updatable/tests/test_product_uom_update.py
```diff
@@ -0,0 +1,83 @@
+# Copyright 2020 Acsone SA/NV
+# Copyright 2020 Camptocamp
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+from odoo.exceptions import UserError
+from odoo.tests.common import SavepointCase
+
+
+class TestProductUomUpdate(SavepointCase):
+    def setUp(self):
```
simahawk, line 10: next time replace w/ setUpClass pls :wink:

simahawk (approved):

PR #730, opened 2020-11-27
[14.0][MIG]product_barcode_required

watthanun:


rousseldenis:
@watthanun Thanks for your contribution. Could you check https://github.com/OCA/maintainer-tools/wiki/Migration-to-version-14.0 ?

As it seems that you included other modules.

PR #737, opened 2020-12-08
[12.0][IMP] product_sequence: Use parent categories to determine the prefix …

khaeusler:
… if the category has no settings for the prefix.

We have implemented a small improvement for determining the prefix.

Currently only the prefix from the selected category or a standard prefix is used.
With this change you can set in the settings whether the prefix of a parent category should be used if no prefix is defined for the selected category.

### DIFF
--- a/product_sequence/models/product_product.py
+++ b/product_sequence/models/product_product.py
@@ -23,18 +23,15 @@ class ProductProduct(models.Model):
         if 'default_code' not in vals or vals['default_code'] == '/':
             categ_id = vals.get("categ_id")
             template_id = vals.get("product_tmpl_id")
-            categ = sequence = False
+            category = self.env['product.category']
             if categ_id:
                 # Created as a product.product
-                categ = self.env['product.category'].browse(categ_id)
+                category = category.browse(categ_id)
             elif template_id:
                 # Created from a product.template
                 template = self.env["product.template"].browse(template_id)
-                categ = template.categ_id
-            if categ:
-                sequence = categ.sequence_id
-            if not sequence:
-                sequence = self.env.ref('product_sequence.seq_product_auto')
+                category = template.categ_id
+            sequence = self.env["ir.sequence"].get_category_sequence_id(category)
             vals['default_code'] = sequence.next_by_id()
         return super().create(vals)
 
@@ -49,10 +46,7 @@ class ProductProduct(models.Model):
             for product in self:
                 category_id = vals.get('categ_id', product.categ_id.id)
                 category = product_category_obj.browse(category_id)
-                sequence = category.exists() and category.sequence_id
-                if not sequence:
-                    sequence = self.env.ref(
-                        'product_sequence.seq_product_auto')
+                sequence = self.env["ir.sequence"].get_category_sequence_id(category)
                 ref = sequence.next_by_id()
                 vals['default_code'] = ref
                 if len(product.product_tmpl_id.product_variant_ids) == 1:
--- a/product_sequence/__manifest__.py
+++ b/product_sequence/__manifest__.py
@@ -6,7 +6,7 @@
 
 {
     'name': 'Product Sequence',
-    'version': '12.0.1.0.1',
+    'version': '12.0.2.0.1',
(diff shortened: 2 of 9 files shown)


marylla (approved):
I made a functional review. LGTM :+1: 

Before activation of new setting:

- Go to Inventory > Configuration > Product Categories
- Set a prefix for categorie "All / Saleable" like "Prod-"
- Create a new product with this category --> the default code was "Prod-0001"
- Create a new product with the child category "All / Saleable / Office Furniture" --> the default code was "PR/0005" (standard sequence)

And then:

- Go to Settings > General Settings 
- Activate "Use parent categories to determine the prefix"
- Create a new product with the child category "All / Saleable / Office Furniture" --> the default code was "Prod-0002".

chandni299 (approved):

product_sequence/models/product_category.py
```diff
@@ -21,6 +21,15 @@ class ProductCategory(models.Model):
         copy=False, readonly=True,
     )
 
+    @api.model
+    def get_sequence_id(self):
```
yweng8111: I think it is better to use the decorator @api.multi with self.ensure_one() or @api.one

yweng8111 (approved):

product_sequence/models/product_category.py
khaeusler: @api.one is deprecated and with ensure_one the following does not work:
```python
self.env['product.category'].get_sequence_id()
```

Or do you mean it like that?
```python
@api.multi
def get_sequence_id(self):
    default_sequence = self.env.ref('product_sequence.seq_product_auto')
    if self:
        self.ensure_one()
        current_category = self
        if self.env.user.company_id.use_parent_categories_to_determine_prefix:
            while not current_category.sequence_id and current_category.parent_id:
                current_category = current_category.parent_id
        return current_category.sequence_id or default_sequence
    else:
        return default_sequence
```

yweng8111: I know what you want, But if you define the function with "@api.model". It confuse me that It can only be called from self.env['product.category']

product_sequence/models/product_product.py
```diff
@@ -23,18 +23,15 @@ def create(self, vals):
...
                 categ = self.env['product.category'].browse(categ_id)
             elif template_id:
                 # Created from a product.template
                 template = self.env["product.template"].browse(template_id)
                 categ = template.categ_id
-            if categ:
-                sequence = categ.sequence_id
-            if not sequence:
-                sequence = self.env.ref('product_sequence.seq_product_auto')
+            sequence = categ.get_sequence_id()
```
yweng8111, line 34: Maybe you can keep the If clause, and call you function there.

yweng8111, line 34: Maybe you can keep the If clause, and call you function there.

yweng8111: Wait. Actually "categ_id" of product is required field. Can you create a product without category?

product_sequence/models/product_product.py
```diff
@@ -49,10 +46,7 @@ def write(self, vals):
             for product in self:
                 category_id = vals.get('categ_id', product.categ_id.id)
                 category = product_category_obj.browse(category_id)
-                sequence = category.exists() and category.sequence_id
-                if not sequence:
-                    sequence = self.env.ref(
-                        'product_sequence.seq_product_auto')
+                sequence = category.get_sequence_id()
```
yweng8111: Maybe you can keep the if clause and call you function there.

benbrich:
Hi @rousseldenis what's preventing us from merging this one? Can I be of any help? Thanks for the work!

product_sequence/models/product_product.py
yweng8111: Or you can define you function like this. It could not be a function of "product,category" but for 'ir.sequence'
 
`@api.model`
`def get_categ_sequence_id(self, category=False):`

khaeusler: done

yweng8111:
Well done. Thanks for the work!

yweng8111:
Hi @rousseldenis, would you like to review the code again? I think it can be merged now.

PR #732, opened 2020-11-27
[ADD] product_order_noname: Speedup product retrieve

WR-96:
This module improves retrieving products when a language is installed
and configured. The default behavior is to order products by name,
which in this case would be slow because Odoo will process the
product's name to translate it.

When ordering products by name and having a language configured the
Odoo's ORM makes a join to `product_template` and then another join but
this time with a nested SELECT to retrieve the translation. This is very
slow because of the lack of index on the JOIN, making a query like
`env['product.product'].search([], limit=10)` taking up to 1027.282 ms.

With the changes introduced here the query is considerably simpler, just
as the following:

```sql
SELECT "product_product".id
FROM   "product_product"
WHERE  ( "product_product"."active" = true )
ORDER  BY "product_product"."default_code"
LIMIT  10
```
Also, the execution time dramatically decrease, from 1027.282 ms to
0.624 ms which is about 1.65k times faster.

Related to:
 - https://github.com/odoo/odoo/pull/61618

PR #738, opened 2020-12-09
[13.0][ADD] product_order_noname: Speedup product retrieve

fernandahf:
This module improves retrieving products when a language is installed
and configured. The default behaviour is to order products by name,
which in this case would be slow because Odoo will process the
product's name to translate it.

When ordering products by name and having a language configured the
Odoo's ORM makes a join to `product_template` an then another join but
this time with a nested SELECT to retrieve the translation. This is very
slow because the lack of index on the JOIN, making a query like
`env['product.product'].search([], limit=10)` taking up to 1027.282 ms.

With the changes introduced here the query is considerably simpler, just
as the following:

    SELECT "product_product".id
    FROM   "product_product"
    WHERE  ( "product_product"."active" = true )
    ORDER  BY "product_product"."default_code"
    LIMIT  10

Also the execution time dramatically decrease, from 1027.282 ms to
0.624 ms which is about 1.65k times faster.


product_order_noname/README.rst
```diff
@@ -0,0 +1,181 @@
...
+   !!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!
+
+.. |badge1| image:: https://img.shields.io/badge/maturity-Production%2FStable-green.png
+    :target: https://odoo-community.org/page/development-status
+    :alt: Production/Stable
+.. |badge2| image:: https://img.shields.io/badge/licence-LGPL--3-blue.png
+    :target: http://www.gnu.org/licenses/lgpl-3.0-standalone.html
+    :alt: License: LGPL-3
+.. |badge3| image:: https://img.shields.io/badge/github-OCA%2Fproduct--attribute-lightgray.png?logo=github
+    :target: https://github.com/OCA/product-attribute/tree/12.0/product_order_noname
```
moylop260: Could you re-generate this file using 13.0 in the parameters,
please?

moylop260:
Fast-tracking since that it is a fowarding from an approve module:
 - https://github.com/OCA/product-attribute/pull/732

PR #739, opened 2020-12-09
[14.0][ADD] product_order_noname: Speedup product retrieve

fernandahf:
This module improves retrieving products when a language is installed
and configured. The default behaviour is to order products by name,
which in this case would be slow because Odoo will process the
product's name to translate it.

When ordering products by name and having a language configured the
Odoo's ORM makes a join to `product_template` an then another join but
this time with a nested SELECT to retrieve the translation. This is very
slow because the lack of index on the JOIN, making a query like
`env['product.product'].search([], limit=10)` taking up to 1027.282 ms.

With the changes introduced here the query is considerably simpler, just
as the following:

    SELECT "product_product".id
    FROM   "product_product"
    WHERE  ( "product_product"."active" = true )
    ORDER  BY "product_product"."default_code"
    LIMIT  10

Also the execution time dramatically decrease, from 1027.282 ms to
0.624 ms which is about 1.65k times faster.


luisg123v:
I think the module should be migrated only and after it's already migrated to 13.0, as the branch 14.0 must contain the full history from previous versions.

moylop260:
> @luisg123v I think the module should be migrated only and after it's already migrated to 13.0, as the branch 14.0 must contain the full history from previous versions.

You are right!
But for this case it is a new module using just one commit
So, I think we can merge it as it

FYI I just re-generated the README

moylop260:
I just fixed the pre-commit error using the latest sha of maintainer-tools

moylop260:
@sbidoul 

Do you know if the bot is tired or sleeping?

Rergads!

luisg123v:
> Do you know if the bot is tired or sleeping?

:laughing: :laughing: :laughing:

sbidoul:
https://github.com/OCA/hr/pull/946#issuecomment-742610601 It has been stuck for a while and now has a long queue to process.

PR #744, opened 2020-12-09
[13.0][REF] product_order_noname: Change product.template order

fernandahf:
Product.template has `_order='name'`
even if the `name` field has an `index=True`
This index is not used because of this field has a
`translate=True` parameter

So, it is in the same situation like product.product

So, it is changed using `_order='default_code, id'` instead.
For a faster sort it requires an index so the field `default_code`
is redefined to `index=True`


fernandahf:
@moylop260

Could you review this, please?

PR #752, opened 2020-12-13
[ADD] 12.0 : product_active_propagate : new module to avoid having inconsistent states between the active field of the variant and the active field of the template

### LINKED ISSUE #750: [RFC] module to propagate "active" value between product variant and template
Hi all,

here is the need : 
in odoo, for products, we have two fields ``active``, one on the ``product.product`` model, one on the ``product.template`` model. There are not link between both fields.
so, it can occures that an active template has only disabled variant but is still visible. So users looking for products will see a product in a view, but not in the other view that is confusing. It forces the user to disable two times the products to archive it really.

i'm looking for a module ``product_active_propagate`` with the following features : 

- if a user disables a template, it disables all the variants ;
- if a user ensables a template, in enables all the variants ;  
- if a user disables a variant, and if it was the last active variant, it disable also the template ;
- the a user enable a variant, it enables the templates ; 

Do you know an existing module for that features ? 

thanks

legalsylvain:
This module extends the functionnaly of Odoo Product module.

By default, in Odoo,
- when a user disable a template, it will disable the related variants.
- when a user enable a template, it will enable the related variants.

But if a user disables all the variants of a template, the related template
will still remain active, what doesn't make sense in many case,
and which will force the user to disable the template as well,
to avoid seeing the obsolete template.

This module avoids having inconsistent states between the active field of the varinats
and the active field of the template.

Once installed:

- if a user disables a variant, it will disable the template, if all the variants
  are disabled.

- if a user enables a variant, it will enable the template.

Fixes: #750


product_active_propagate/__init__.py
```diff
@@ -0,0 +1,2 @@
+from . import models
+from .hooks import post_init_hook
```
ivantodorovich: It seems this file is missing

ivantodorovich (changes requested):

product_active_propagate/__init__.py
legalsylvain: thanks. Fixed.

ivantodorovich (approved):
Travis is complaining about the missing OCA in authors manifest key, but otherwise looks ok 👍🏻 code review

PierrickBrun (approved):

product_active_propagate/tests/test_module.py
```diff
@@ -0,0 +1,51 @@
...
+                "Disabling %d of the %d variants should not disable"
+                " the template." % (count, len(self.all_variants)))
+
+        # Disable the last variant
+        self.all_variants[-1:].active = False
+        self.assertEqual(
+            self.template.active, False,
+            "Disabling all variants should disable the template.")
+
+        # Enable againt the last variant
```
mamcode:
```suggestion
        # Enable against the last variant
```

product_active_propagate/readme/INSTALL.rst
```diff
@@ -0,0 +1,34 @@
...
+    UPDATE product_template pt
+    SET active = false
+    AND id in (
+        SELECT pt.id
+        FROM product_template pt
+        INNER JOIN product_product pp on pp.product_tmpl_id = pt.id
+        WHERE pt.active = true
+        GROUP BY pt.id
+        HAVING sum((pp.active)::int) = 0
+    );
```
mamcode:
```suggestion
    UPDATE product_template pt
    SET active = false
    WHERE id in (
        SELECT pt.id
        FROM product_template pt
        INNER JOIN product_product pp on pp.product_tmpl_id = pt.id
        WHERE pt.active = true
        GROUP BY pt.id
        HAVING sum((pp.active)::int) = 0
    );
```

product_active_propagate/readme/INSTALL.rst
```diff
@@ -0,0 +1,34 @@
...
+    UPDATE product_template pt
+    SET active = true
+    AND id in (
+        SELECT pt.id
+        FROM product_template pt
+        INNER JOIN product_product pp on pp.product_tmpl_id = pt.id
+        WHERE pt.active = false
+        GROUP BY pt.id
+        HAVING sum((pp.active)::int) != 0
+    );
```
mamcode:
```suggestion
    UPDATE product_template pt
    SET active = true
    WHERE id in (
        SELECT pt.id
        FROM product_template pt
        INNER JOIN product_product pp on pp.product_tmpl_id = pt.id
        WHERE pt.active = false
        GROUP BY pt.id
        HAVING sum((pp.active)::int) != 0
    );
```

product_active_propagate/readme/INSTALL.rst
mamcode: I don't know if this depends on version of postgresql but in my local installation the current query disables all the templates, with this change it works, disabling only the templates that do not have a single variant active.

mamcode (changes requested):
Good job, thanks for the work! please see my comments.

product_active_propagate/tests/test_module.py
legalsylvain: done. (I wrote again, not against ;-))

product_active_propagate/readme/INSTALL.rst
legalsylvain: how ! thanks a lot ! my bad. I fixed the request.

legalsylvain:
Hi @mamcode, thanks a lot for your review. I have applied your remarks

mamcode (approved):
LGTM :+1:

GSLabIt:
is this ready to merge?

legalsylvain:
@OCA/product-maintainers  : I think this one is ready.

bodedra:
This is working fine in v13.0. So it seems module is only useful for v12.0. :thinking: 

I have tested it with following scenario in runbot https://8423963-13-0-all.runbot31.odoo.com/web#id=97&action=356&model=product.template&view_type=form&cids=&menu_id=246 :

- Create a template and 2 variants.
- Disable 2 variants and verify template. It disable automatically.
- Active template, it activated 2 variants. OR active 1 or more variants, it activated template. 
- If I disable template, it disable 2 variants.

LMK am I missing anything here?

legalsylvain:
> This is working fine in v13.0. So it seems module is only useful for v12.0. 

Good to know.

PR #755, opened 2020-12-17
[MIG] 14.0 base_product_mass_addition

PierrickBrun:
based on the 13.0 PR https://github.com/OCA/product-attribute/pull/578 

I experienced slow inputs while testing so I decided to go back to using `onchange_helper`.

Since not everyone wants this dependency I made it optional, falling back to `odoo.test.Form` if the module is not installed.

I don't know how to test both methods in travis but I ran `purchase_quick` tests with both solutions locally without problems


base_product_mass_addition/models/product_mass_addition.py
```diff
@@ -0,0 +1,96 @@
...
+        raise NotImplementedError
+
+    def _complete_quick_line_vals(self, vals, lines_key=""):
+        if not lines_key:
+            raise NotImplementedError
+        init_keys = ["product_id"]
+        init_vals = {key: val for key, val in vals.items() if key in init_keys}
+        update_vals = {key: val for key, val in vals.items() if key not in init_keys}
+        # Try to use onchange_helper if installed
+        try:
```
legalsylvain: I think you should better test if the module ``onchange_helper`` is installed. So, you avoid a try catch that could hide other errors.

legalsylvain (approved):
:+1: thanks for making the usage of Form optional.
I use this module in V12, and had to patch it to avoid slow performance.

Just one question : there is written ``Once onchange_helper is stabilized,`` : could you mention witch problem we can face, using ``onchange_helper`` ?  Not sure to understand.

thanks !

PierrickBrun:
> Just one question : there is written `Once onchange_helper is stabilized,` : could you mention witch problem we can face, using `onchange_helper` ? Not sure to understand.

This is obsolete (I think in 12.0 onchange_helper needed some refactor), I changed the text to be coherent with the current state

base_product_mass_addition/models/product_mass_addition.py
PierrickBrun: That would require to do a `.search()` everytime this function is called, I'm not sure this is a good idea to add that (or there is a less costly way to do it)

Also, I only catch AttributeError, which shouldn't happen often outside of development work and I logged the error message to make it more obvious

bealdav (approved):

legalsylvain:
Hi Pierrick. If onchange_helper works correctly, 
- we should depend on it. 

Otherwise, test will not be possible without reason, that's not good.

base_product_mass_addition/README.rst
```diff
@@ -0,0 +1,98 @@
...
+
+It is useful when you don't want to add every product line possible to your object.
+From the product grid you only have to set the quantity for each product.
+
+Example implementations:
+ - purchase_quick
+ - stock_picking_quick
+
+This module does not depend on onchange_helper but it is faster when installed.
+If not, it will us Form emulation from backend.
```
kittiu:
```suggestion
If not, it will use Form emulation from backend.
```

kittiu:
First of all, I find this module and the purchase_quick very creative and useful :+1: 

I am not the user of onchange_helper. But as the end functional user, I would prefer this to work fast without digging, even has to depend on onchange_helper. Otherwise, how could I know that, if it is slow and I need to patch it.

Why not just go on with onchange_helper now, it already works well without refactor, isn't it? And then another PR to refactor it later.

Just my small opinion.

PierrickBrun:
To the general demand, I made onchange_helper a clear dependency !

hparfr (approved):

sebastienbeau:
@PierrickBrun can you check with runbot is red ?

PierrickBrun:
> @PierrickBrun can you check with runbot is red ?

It's green now

PR #758, opened 2020-12-21
[FIX] product_pricelist_supplierinfo: ensure product exists

ruuter:
It is possible that this method is called on an inactive template, in which case product_variant_id returns empty recordset and
_select_seller will raise expected singleton error. This can happen for example when archiving a product.

### DIFF
--- a/product_pricelist_supplierinfo/models/product_template.py
+++ b/product_pricelist_supplierinfo/models/product_template.py
@@ -19,10 +19,15 @@ class ProductTemplate(models.Model):
             product = product.browse(product_id)
         if rule.no_supplierinfo_min_quantity:
             quantity = 1.0
-        seller = product._select_seller(
-            partner_id=rule.filter_supplier_id, quantity=quantity, date=date)
-        if seller:
-            price = seller._get_supplierinfo_pricelist_price()
+        # The product_variant_id returns empty recordset if template is not
+        # active, so we must ensure variant exists or _select_seller fails.
+        if product:
+            seller = product._select_seller(
+                partner_id=rule.filter_supplier_id,
+                quantity=quantity,
+                date=date)
+            if seller:
+                price = seller._get_supplierinfo_pricelist_price()
         if price:
             # We need to convert the price if the pricelist and seller have
             # different currencies so the price have the pricelist currency


pedrobaeza:
There's no possibility of such thing unless you are calling the method from custom code, and thus, you need to check that outside this module.

ruuter:
Hmm, I'm not so sure about that. Maybe I'm wrong, but only involved custom module seems to be OCA auditlog, which calls read
on deactivated product, which in turn triggers recompute of website_price. Finally it all fails here:
https://github.com/odoo/odoo/blob/12.0/addons/sale/models/product_template.py#L270

pedrobaeza:
Please provide then steps to reproduce the problem in a runbot.

ruuter:
Sorry, don't have time for that now. Maybe later. But I can confirm, that when I turn off Auditlog rule on Product Template, it does not happen anymore.

pedrobaeza:
Something weird is happening then on that place.

ruuter:
@pedrobaeza Finally had time to look into it again. Unfortunately I am not able to login to any of the OCA runbot instances. Is this general issue or am I doing something wrong?

However, debugging the issue locally, I do not see anything wrong with the auditlog module. It simply reads all fields from product, before and after: https://github.com/OCA/server-tools/blob/12.0/auditlog/models/rule.py#L308-L314
This triggers compute on all fields, including `website_price`, which eventually fails here: https://github.com/OCA/product-attribute/blob/12.0/product_pricelist_supplierinfo/models/product_template.py#L22, because `self.product_variant_id` returns empty recordset, since at this point all variants are deactivated.

All this seems to make sense to me. Computing price for template is not possible if all variants are archived, but the `_get_supplierinfo_pricelist_price` does not take that into account. It blindly trusts that `product_variant_id` always returns a variant. Are you sure it should be like this? If so, where is the issue? I can not see it.

pedrobaeza:
OK, thanks for the explanation and the debugging. It seems correct to me, except that you are not adding such explanation as comment in code for avoiding to question this again when migrating or refactoring the module, so this can disappear in that cases not being seen as needed.

ruuter:
Okay, I will add the comment.

One more question. Do you think maybe better solution is to use:
`self.with_context(active_test=False).product_variant_id` instead if condition?

pedrobaeza:
That can alter other results, so no.

ruuter:
Ok tnx, good to know. Comment added.

pedrobaeza (approved):

pedrobaeza:
Let's fast track it:

/ocabot merge patch

PR #760, opened 2020-12-26
[MIG] product_packaging_type: Migration to 14.0

phuctranfxvn:


hparfr (approved):

rvalyi (approved):
I approve the migration through basically it only consists in changing the version number to 14, so I'm not too sure if it is fair to make you a listed module contributor (this would not scale). I guess this is to be decided with CampToCamp authors. cc @damdam-s  @jgrandguillaume

PR #775, opened 2021-01-20
[12.0][FIX] product_dimension : compatibility with product_logistics_uom's volume_uom_id

clementmbr:
Both modules **[product_dimension](https://github.com/OCA/product-attribute/tree/12.0/product_dimension)** and **[product_logistics_uom](https://github.com/OCA/product-attribute/tree/12.0/product_logistics_uom)** are merged in 12.0 but could not be used together.

**product_logistics_uom** introduce a field `volume_uom_id` allowing to register product's volume in any choosen Unit of Measure, but **product_dimension** always calculated the product's volume in cubic metre.

This PR makes **product_dimension** taking `volume_uom_id` into account when calculating the product's volume.

Cf the debate on https://github.com/OCA/product-attribute/pull/607#discussion_r416687807
Cc @hparfr @legalsylvain @bealdav @rvalyi

### DIFF
--- a/product_dimension/models/product.py
+++ b/product_dimension/models/product.py
@@ -6,76 +6,82 @@ from odoo import models, fields, api
 
 
 class Product(models.Model):
-    _inherit = 'product.product'
+    _inherit = "product.product"
 
-    @api.onchange(
-        'product_length', 'product_height', 'product_width', 'dimensional_uom_id')
-    def onchange_calculate_volume(self):
-        self.volume = self.env['product.template']._calc_volume(
-            self.product_length, self.product_height,
-            self.product_width, self.dimensional_uom_id)
+    product_length = fields.Float("length")
+    product_height = fields.Float("height")
+    product_width = fields.Float("width")
+    dimensional_uom_id = fields.Many2one(
+        "uom.uom",
+        "Dimensional UoM",
+        domain=lambda self: self._get_dimension_uom_domain(),
+        help="UoM for length, height, width",
+        default=lambda self: self.env.ref("uom.product_uom_meter"),
+    )
 
     @api.model
     def _get_dimension_uom_domain(self):
-        return [
-            ('category_id', '=', self.env.ref('uom.uom_categ_length').id)
-        ]
+        return [("category_id", "=", self.env.ref("uom.uom_categ_length").id)]
 
-    product_length = fields.Float('length')
-    product_height = fields.Float('height')
-    product_width = fields.Float('width')
-    dimensional_uom_id = fields.Many2one(
-        'uom.uom',
-        'Dimensional UoM',
-        domain=lambda self: self._get_dimension_uom_domain(),
-        help='UoM for length, height, width',
-        default=lambda self: self.env.ref('uom.product_uom_meter'),
+    @api.onchange(
+        "product_length", "product_height", "product_width", "dimensional_uom_id"
+    )
+    def onchange_calculate_volume(self):
+        self.volume = self.env["product.template"]._calc_volume(
+            self.product_length,
+            self.product_height,
+            self.product_width,
+            self.dimensional_uom_id,
         )
 
 
 class ProductTemplate(models.Model):
-    _inherit = 'product.template'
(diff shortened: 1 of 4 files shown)


product_dimension/models/product.py
```diff
@@ -6,76 +6,98 @@
...
+        uom_litre = self.env.ref("uom.product_uom_litre")
+        volume_litre = 0
 
-    def convert_to_meters(self, measure, dimensional_uom):
-        uom_meters = self.env.ref('uom.product_uom_meter')
+        if product_length and product_height and product_width and uom_id:
+            length_m = self._convert_to_meters(product_length, uom_id)
+            height_m = self._convert_to_meters(product_height, uom_id)
+            width_m = self._convert_to_meters(product_width, uom_id)
+            volume_litre = (length_m * height_m * width_m) * 1000
```
hparfr:
```suggestion
            volume_m³ = (length_m * height_m * width_m) 
```

product_dimension/models/product.py
```diff
@@ -6,76 +6,98 @@
...
-    def onchange_calculate_volume(self):
-        self.volume = self._calc_volume(
-            self.product_length, self.product_height, self.product_width,
-            self.dimensional_uom_id)
+    @api.model
+    def _calc_volume(
+        self, product_length, product_height, product_width, uom_id, volume_uom_id
+    ):
+        uom_litre = self.env.ref("uom.product_uom_litre")
+        volume_litre = 0
```
hparfr:
```suggestion
        volume_m³ = 0
```

product_dimension/models/product.py
```diff
@@ -6,76 +6,98 @@
...
+            height_m = self._convert_to_meters(product_height, uom_id)
+            width_m = self._convert_to_meters(product_width, uom_id)
+            volume_litre = (length_m * height_m * width_m) * 1000
 
-        return dimensional_uom._compute_quantity(
-            qty=measure,
-            to_unit=uom_meters,
-            round=False,
+        return uom_litre._compute_quantity(
+            qty=volume_litre, to_unit=volume_uom_id, round=False,
```
hparfr, line 69:
```suggestion
        volume_litre = volume_m³ / 1000
        return uom_litre._compute_quantity(
            qty=volume_litre, to_unit=volume_uom_id, round=False,
```

hparfr (changes requested):
Thanks for addressing this problem.

I think it's better to express volumes in m³ than liters.

product_dimension/models/product.py
clementmbr: Hi @hparfr ! Thanks for the review !
In fact I used litres because it's the easiest UoM I found registered in native uom module datas :
https://github.com/odoo/odoo/blob/12.0/addons/uom/data/uom_data.xml#L90

In your suggestion I didn't catch where does your `uom_m³` comes from...

hparfr: It's not available in 12.0 as UoM by default in v12 but it's hardcoded in product https://github.com/odoo/odoo/blob/12.0/addons/product/models/product_template.py#L83
 
By the way it's available in 14.0 https://github.com/odoo/odoo/blob/14.0/addons/uom/data/uom_data.xml#L95

I have updated my suggestion. In that way, for the migration in v14 there will be less changes.

clementmbr: Hmm... As I understand `volume_litre` and `volume_m³` as purely temporary variables to be able to make the conversion based on a known UoM, it looks like it doesn't really matter what is this temporary variable UoM...

But ok to use both `volume_litre` and `volume_m³` instead of only `volume_litre` if it helps for the 14.0 migration.

clementmbr:
> Thanks for addressing this problem.
> 
> I think it's better to express volumes in m³ than liters.

Done ! @hparfr let me know if it's ok for you now

hparfr (approved):

rvalyi (approved):
great work @clementmbr !

hparfr (approved):

marcelsavegnago (approved):

PR #776, opened 2021-01-21
[12.0][FIX] product_pricelist_supplierinfo: Convert price to UOM on the Sale

CarmenMiranda:
If we have a product with a supplier with a UOM "A" but the product is sold with a UOM "B" the price we are getting is the same no matter the UOM. For example:
Product with the price 1200 per Dozen and we sell a 1 unit we are getting the price of 1200 for that unit when we should be obtaining 100.

With these changes, we are converting the price to the UOM that is being used on the sale or the default on the product.
Also, a test was added to prove the previous behaviour.

### DIFF
--- a/product_pricelist_supplierinfo/models/product_template.py
+++ b/product_pricelist_supplierinfo/models/product_template.py
@@ -37,6 +37,12 @@ class ProductTemplate(models.Model):
             # Verbatim copy of part of product.pricelist._compute_price_rule.
             qty_uom_id = self._context.get('uom') or self.uom_id.id
             price_uom = self.env['uom.uom'].browse([qty_uom_id])
+
+            # We need to convert the price to the uom used on the sale, if the
+            # uom on the seller is a different one that the one used there.
+            if seller and seller.product_uom != price_uom:
+                price = seller.product_uom._compute_price(price, price_uom)
+
             convert_to_price_uom = (
                 lambda price: self.uom_id._compute_price(
                     price, price_uom))
--- a/product_pricelist_supplierinfo/tests/test_product_supplierinfo.py
+++ b/product_pricelist_supplierinfo/tests/test_product_supplierinfo.py
@@ -36,6 +36,18 @@ class TestProductSupplierinfo(common.SavepointCase):
                 }),
             ],
         })
+        cls.product_with_diff_uom = cls.env['product.product'].create({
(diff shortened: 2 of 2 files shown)


CarmenMiranda:
Hi @JulioSerna This is the PR for the issue with the difference between supplierinfo uom and sale line uom, could you review it? please :smile:

JulioSerna:
@moylop260 

LGTM 👍

moylop260 (approved):
LGTM

moylop260:
@pedrobaeza 
What do you think?

product_pricelist_supplierinfo/models/product_template.py
```diff
@@ -37,6 +37,12 @@ def _get_supplierinfo_pricelist_price(
             # Verbatim copy of part of product.pricelist._compute_price_rule.
             qty_uom_id = self._context.get('uom') or self.uom_id.id
             price_uom = self.env['uom.uom'].browse([qty_uom_id])
+
```
pedrobaeza, line 40: Please don't add empty lines inside a method. It's discouraged in PEP8.

pedrobaeza (commented):
The same in all places

pedrobaeza:
The codes changes themselves are OK.

product_pricelist_supplierinfo/models/product_template.py
moylop260: If it is a PEP and the OCA likes to follow it you should make a issue in pylint-odoo in order to enable it

if not, maybe OCA doesn't like to follow this particular PEP

So, you can avoid checking this again and again for each PR and delegate it to the CI

pedrobaeza: Well, as I'm the reviewer and original coder of this module, don't want to introduce these ugly lines in the module, so please remove them, independently from the pylint thing.

moylop260: Well, as we are the creator and contributor of this change of this module and it is under umbrella guidelines and CI we don't want fix Pedro's guidelines that are not discussed in the OCA and the CI PEP style is green

pedrobaeza: OK, as you prefer, then don't ping me again on any PR.

marcelsavegnago (approved):

PR #777, opened 2021-01-21
[12.0][FIX] product_supplierinfo_for_customer: Crash when the price is computed for a template instead of variant

sergio-teruel:
cc @Tecnativa TT27887
ping @chienandalu @CarlosRoca13

### DIFF
--- a/product_supplierinfo_for_customer/models/product_customerinfo_mixin.py
+++ b/product_supplierinfo_for_customer/models/product_customerinfo_mixin.py
@@ -1,0 +1,58 @@
+# Copyright 2021 Tecnativa - Sergio Teruel
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl.html).
+from odoo import fields, models
+
+
+class ProductSupplierInfoMixin(models.AbstractModel):
+    _name = 'product.customerinfo.mixin'
+    _description = 'Common methods to compute prices based on partners'
+
+    def _get_price_from_customerinfo(self, partner_id):
+        self.ensure_one()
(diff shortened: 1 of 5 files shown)


CarlosRoca13 (approved):
LGTM :+1:

pedrobaeza:
Can you try to avoid code duplication making synergies between existing variant code and current one?

sergio-teruel:
Ok.. Ready!!

pedrobaeza:
Good job extracting the common parts. I think that you don't need a mixin as `product.product` already inherits from `product.template`. You can:

- Put all the mixin logic in `product.template`.
- Add a method there for obtaining the domain to use.
- On `product.product`, override such method for returning the changed domain.
- And that's all!

product_supplierinfo_for_customer/models/product_customerinfo_mixin.py
```diff
@@ -0,0 +1,58 @@
...
+        else:
+            domain = [
+                ("name", "=", partner_id),
+                "|",
+                ("product_id", "=", self.id),
+                "&",
+                ("product_tmpl_id", "=", self.product_tmpl_id.id),
+                ("product_id", "=", False),
+            ]
+        customerinfo = self.env["product.customerinfo"].search(
```
chienandalu:
```suggestion
        customerinfo = self.env["product.customerinfo"].sudo().search(
```

chienandalu (approved):

pedrobaeza:
Is this needed for v13?

/ocabot merge patch

PR #781, opened 2021-01-26
[10.0][ADD] product_abc_classification

lmignon:
This module is a rework of #623.  
> This modules includes the ABC analysis (or ABC classification), which is
used by inventory management teams to help identify the most important
products in their portfolio and ensure they prioritize managing them above
those less valuable.

> Managers will create a profile with several levels (percentages) and then the
profiled products will automatically get a corresponding level using the
ABC analysis.

The original code has been split into 2 addons.
* product_abc_classification_base: provides the bases models to use to configure a classification profile and assign a classification level to the products
* product_abc_classification_sale_stock: provides a  concrete profile that compute the classification from the number of delivered sale order line by product.

TODO
* [x] add tests


product_abc_classification_base/models/abc_classification_profile.py
```diff
@@ -51,6 +51,17 @@ def _check_levels(self):
                 raise ValidationError(
                     _("The percentages of the levels must be unique.")
                 )
+            percentage_productss = profile.level_ids.mapped(
+                "percentage_products"
+            )
+            total = sum(percentage_productss)
+            if profile.level_ids and total != 100.0:
```
jbaudoux, line 61: float_compare?

product_abc_classification_base/models/abc_classification_level.py
```diff
@@ -0,0 +1,58 @@
...
+                )
+            if level.percentage <= 0.0:
+                raise ValidationError(
+                    _("The percentage should be a positive number.")
+                )
+
+    @api.constrains("percentage_products")
+    def _check_percentage_products(self):
+        for level in self:
+            if level.percentage_products > 100.0:
```
jbaudoux, line 49: float_compare?

product_abc_classification_base/models/abc_classification_level.py
```diff
@@ -0,0 +1,58 @@
...
+                raise ValidationError(
+                    _("The percentage of products cannot be greater than 100.")
+                )
+            if level.percentage_products <= 0.0:
```
jbaudoux, line 53: float_compare?

MiquelRForgeFlow:
I know this PR is a rework, but I would like to point you that meantime the one in v13 has had some fixes and improvements, see https://github.com/OCA/product-attribute/pull/1006 and https://github.com/OCA/product-attribute/pull/1010 😉

lmignon:
@MiquelRForgeFlow In this version we support multiple classification profiles by product, computation rules by warehouse, computed values history ... It's also very important to us to put the computation rules in dedicated addon to ease maintainability and lower the complexity. For now we are locked on odoo 10.0. We could start the migration to odoo 14/16? by end of this year. At this time we'll try to makes all the changes converging.

pedrobaeza:
Yeah, I see the splitting also as something needed for following Odoo's modularity approach. @MiquelRForgeFlow you may have it into account on the migration to v15.

MiquelRForgeFlow:
Well, for now I don't have plans to migrate that module.

pedrobaeza:
OK, we are enrolling to it on v13, converging from exiting v12 `sale_product_classification`, and probably doing already an extension `product_abc_classification_sale` on v13 for such specific metric. I would prefer to let the base module called `product_abc_classification` (without `_base`), and put suffixes for the rest of extensions, as this is more auto-explanatory IMO.

PR #785, opened 2021-01-28
14.0 mig product supplierinfo group

mourad-ehm:


ferranmateu (commented):
I've couldn't install the module in a local database. 

image

sebastienbeau:
@sbidoul runboat seem to have an issue

sebastienbeau (approved):
LGTM, code review, fix, tested, running in prod

product_supplierinfo_group/models/product_supplierinfo.py
```diff
@@ -0,0 +1,79 @@
...
+            )
+        return group
+
+    @api.model_create_multi
+    def create(self, list_vals):
+        for vals in list_vals:
+            if not vals.get("group_id"):
+                vals["group_id"] = self._get_or_create_group(vals).id
+                # remove useless key
+                for key in MAPPING_RELATED.keys():
```
florian-dacosta: Would it be possible to call a method just as `_fields_for_group_match` but to return `MAPPING_RELATED`?

From what I understand, MAPPING_MATCH_GROUP need to be outside the class so it can be used in the pre init hook, but it is overiddable because of `_fields_for_group_match`

I guess we should have the same thing for `MAPPING_RELATED`. So the module will be fully overidable in case of additional custom fields on supplierinfo, that should be in the group part.

florian-dacosta (approved):
One comment, not mandatory so I approve.
Code review + tests LGTM
Interesting/clever module!

PR #787, opened 2021-01-28
[12.0] product_price_history

eLBati:
This module allows to manage product history prices.


product_price_history/models/product.py
```diff
@@ -0,0 +1,32 @@
...
+
+
+class ProductPriceHistory(models.Model):
+    _inherit = 'product.price.history'
+
+    product_tmpl_id = fields.Many2one(
+        string='Product Template', store=True, related='product_id.product_tmpl_id')
+
+
+class ProductTemplate(models.Model):
```
rousseldenis: @eLBati Could you split => one file per model ?

product_price_history/models/product.py
```diff
@@ -0,0 +1,32 @@
...
+    _inherit = 'product.template'
+
+    price_history_count = fields.Integer(
+        "Price History Count", compute='compute_price_history_count')
+
+    @api.one
```
rousseldenis: Don't use api.one

product_price_history/models/product.py
eLBati: thanks, done

product_price_history/models/product.py
eLBati: done

product_price_history/models/product_product.py
```diff
@@ -0,0 +1,14 @@
...
+class ProductProduct(models.Model):
+    _inherit = 'product.product'
+
+    price_history_count = fields.Integer(
+        "Price History Count", compute='compute_price_history_count')
+
+    @api.multi
+    def compute_price_history_count(self):
+        for p in self:
+            p.price_history_count = self.env['product.price.history'].search_count(
```
rousseldenis, line 13: Maybe you should use a computed field to get price_history_ids.
Then use len() in loop. That should avoid a query per loop if multi recordset.

product_price_history/readme/USAGE.rst
```diff
@@ -0,0 +1 @@
+Go to ``Inventory > Master data > Product price history`` and handle prices
```
legalsylvain: In a UX PoV I'll expect this menu in the sale menu and not in a stock menu. Don't you think ?

andreampiovesana: better stock, it refers to costs

andreampiovesana (approved):
go

product_price_history/readme/USAGE.rst
legalsylvain: Oh ! I didnt saw that. In that case, I think name and menu are error prone. Should be product_standard_price_history and standard price history. (Or "cost", that is also used)
Price, in Odoo refers more to sale price.

eLBati: I kept the module name, because it is related to model `product.price.history`, and modified manifest name and summary

product_price_history/models/product_product.py
eLBati: `price_history_count` only appears in form views, I don't think we have performance issues here

product_price_history/readme/USAGE.rst
legalsylvain: Thanks !
I understand your PoV regarding the name of the module.

Could you also change the name of the menu ? ``Product Price History`` > ``Product Cost History``.
(the translation is OK (``Storia costo medio prodotto``, but the original name is not)

eLBati: @legalsylvain done, thanks

legalsylvain: thanks !

ferranmateu (approved):
Functional review OK!

astirpe (approved):
LGTM!

hbrunn (approved):

PR #792, opened 2021-02-07
[MIG] product_packaging_type_required: Migration to 14.0

phuctranfxvn:
Standard migration, requires:
- `product_packaging_type_required` migrated in https://github.com/OCA/product-attribute/pull/760


test-requirements.txt
```diff
@@ -0,0 +1 @@
+odoo14-addon-product_packaging_type @ git+https://github.com/trobz/product-attribute.git@14.0-mig-product_packaging_type#subdirectory=setup/product_packaging_type
```
sebalix: Depdendency merged: https://github.com/OCA/product-attribute/pull/760, to remove

sebalix (commented):
LG, dependencies to clean up

test-requirements.txt
simahawk: @phuctranerp ping

hailangvn: @simahawk , @sebalix, Thanks for your review. I will work on this.

hailangvn:
@simahawk , @sebalix, the PR has been updated. Please let us know if anything else is needed. Thanks.

product_packaging_type_required/models/product_packaging_type.py
```diff
@@ -0,0 +1,45 @@
...
+        packages.
+        """
+        existing_products = self.env["product.product"].search(
+            [("type", "in", ("product", "consu"))]
+        )
+        i = 0
+        required_packaging_types = self.search([("required", "=", True)])
+        for product in existing_products:
+            if limit and i == limit:
+                break
```
simahawk: this is wrong in the original implementation. The limit is never used in fact.
The minimum correction would be:

```suggestion
        i = 0
        required_packaging_types = self.search([("required", "=", True)])
        for i, product in enumerate(existing_products):
            if limit and i == limit:
                break
```
but to be honest I don't find this very effective because you will  always process the same products and packaging.
IMO the only reason why it worked so far is that the limit was never used.

**Hence, please do not fix it.** 
Add a `FIXME: limit is never used and the whole method should be refactored`.

There are no tests too, so it's better to refactor it without them.
I'll have a look at this.

simahawk (changes requested):

product_packaging_type_required/models/product_packaging_type.py
hailangvn: Thanks @simahawk for spotting this. PR has been updated per your request.

simahawk (approved):

mmequignon (approved):

simahawk:
FTR I did some change on v13 here https://github.com/OCA/product-attribute/pull/903 and we'll fwd port them to v14 later.

/ocabot merge nobump

Issue #798, opened 2021-02-13
[RFC] product.category "type" field


legalsylvain:
Hi all,

I recently migrated from 8.0 to 12.0. 
regarding the ``product.category``, the field ``type`` disappeared. 

it was present in V10 : https://github.com/odoo/odoo/blob/10.0/addons/product/models/product.py#L24
and not in V11 : https://github.com/odoo/odoo/blob/11.0/addons/product/models/product.py#L16

Do you know if there are any module to restore this usefull feature ?

thanks !

PR #804, opened 2021-02-13
12.0 add product category type

legalsylvain:
- Very simple module that adds 'Type' field on Product Categories to distinguish between parent and final categories.
- also fix ``product_category_active`` module, that use incorrect ``decoration-info`` for disabled categories (instead of ``decoration-muted``, that conflicts with that module.

product_category_tree

* Categories (type view) can contain only categories.

* Categories (type normal) can contain only products.

It is so impossible to select a category (type view) in the product template form view.

product_template_form

### Note

This module restores a feature that was present in Odoo Community
Edition until the V10 revision.

Ref: https://github.com/odoo/odoo/blob/10.0/addons/product/models/product.py#L24

Fix : #798 

CC : @quentinDupont, @grap #GRAPOCA


bealdav (approved):

hparfr (approved):

legalsylvain:
Thanks for your review ! 
i'll made a change to rename ``type`` field to ``category_type`` field, to avoid to use reserved keyword. (with according migration script).

Issue #220, opened 2017-01-04
product dimension not saved in record creation


mehdi-ghezal:
**_module:_** product_dimension
**_version:_** 9.0

### Steps to reproduce

- Install module
- Create a new product template with dimension
- Save the product
- Check the dimension

### Current behavior

Product dimension are **not** saved for product template

### Expected behavior

Product dimension **should be** saved for product template

### Reason

Product template use related fields for dimension (width, height, ...) that are bind to product.product ; during record creation related field are not persist.

When update a record, everything work as expected.

pedrobaeza:
@leemannd, can you check this issue?

leemannd:
@pedrobaeza Yep. I'm going to check this in the end of the week, beginning of the next one.
EDIT: @pedrobaeza @mehdi-ghezal This [pr](https://github.com/OCA/product-attribute/pull/221/files) from @vrenaville should correct the behavior.

PR #807, opened 2021-02-18
[14.0][FIX] product_dimension: dimension not saved at record creation

petrus-v:
... while creating product template with dimension
fields related are not write to the default product_product
I guess the instance is not set created so we get default value
from the product product

### DIFF
--- a/product_dimension/models/product.py
+++ b/product_dimension/models/product.py
@@ -69,6 +69,19 @@ class ProductTemplate(models.Model):
             round=False,
         )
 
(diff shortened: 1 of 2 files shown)


petrus-v:
This is a purpose to fix #220 !

I wonder if we could find a better way ?

petrus-v:
Dear @OCA/product-maintainers, this PR needs review, is it possible to replace stale tag  by needs reviews ?

regards !

PR #809, opened 2021-02-23
[ADD] pos_product_cost_security: glue module

chienandalu:
Glue module between Point of Sale and Product Cost Security

cc @Tecnativa TT28388


chienandalu:
This is ready

pos_product_cost_security/models/ir_http.py
```diff
@@ -0,0 +1,16 @@
...
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl.html).
+from odoo import models
+
+
+class Http(models.AbstractModel):
+    _inherit = "ir.http"
+
+    def session_info(self):
+        """This context comes later from UI to the rpc call"""
+        result = super(Http, self).session_info()
```
pedrobaeza: Can we avoid this override in so such high level?

chienandalu: I saw no other way to inject into the user_context to get it with the rpc call: https://github.com/odoo/odoo/blob/14.0/addons/point_of_sale/static/src/js/models.js#L655

pedrobaeza: But you talk about having already an existing context when on PoS. I think we can re-use it without fear of not having side effects (just bypass that security, not doing `sudo`).

@Yajo do you know anyway other way to inject variables in session context?

chienandalu: Ok, I got a cleaner workaround to avoid injecting contexts all around

pos_product_cost_security/readme/DESCRIPTION.rst
```diff
@@ -0,0 +1,2 @@
+This is a glue module to avoid permission errors when loading data in the Point of Sale
+if the user has no `standard_prices` (cost) field permissions.
```
yajo:
```suggestion
if the user has no ``standard_prices`` (cost) field permissions.
```

pos_product_cost_security/tests/test_pos_product_cost.py
```diff
@@ -0,0 +1,41 @@
+# Copyright 2021 Tecnativa - David Vidal
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+from odoo.tests import HttpCase, tagged
+
+
+@tagged("post_install", "-at_install")
```
yajo, line 6: Can this be avoided? It usually produces random failures on integration environments.

```suggestion
```

pos_product_cost_security/models/product_product.py
```diff
@@ -0,0 +1,18 @@
...
+
+    def read(self, fields=None, load="_classic_read"):
+        """Make an exception when loading data to PoS"""
+        if self.env.context.get(
+            "pos_override_cost_security"
+        ) and self.env.user.has_group("point_of_sale.group_pos_user"):
+            fields = fields if fields else []
+            other_fields = [f for f in fields if f != "standard_price"]
+            self.check_field_access_rights("read", other_fields)
+            return super(ProductProduct, self.sudo()).read(fields=fields, load=load)
```
yajo, line 17: Hummm 🤔 OK but... if the request is asking for that field, couldn't it be because it's needed? Or is it just that the PoS just downloads all fields always?

Also, wouldn't it be a cleaner fix would be to add `groups` to the `standard_price` field?

yajo (approved):

pos_product_cost_security/models/product_product.py
chienandalu: Well, the PoS loads very specific fields. The only one in conflict is `standard_price`, but is a field needed to compute prices if a cost based pricelist is there.

We can't give access to the PoS users, as they don't necessarily have to own such permission.

So the thing is that we need to load the data for the PoS to work (remember that all the logic is reimplemented in the frontend)

pos_product_cost_security/tests/test_pos_product_cost.py
chienandalu: Well, precisely the PoS tests are a special kind: https://github.com/odoo/odoo/blob/14.0/addons/point_of_sale/tests/test_frontend.py#L474

pos_product_cost_security/readme/DESCRIPTION.rst
chienandalu: fixed

pedrobaeza (approved):

PR #811, opened 2021-02-25
[14.0] add base_uom_converter

petrus-v:
This a propose to convert quantities from an UoM category to an other one. This module does nothing special by itself for end users.

For instance working in building industry to install tile flooring for a given
meter square you would be able to extract a formula in the form of a * x + b to
define the expected time in hours to process it. (ie: 30 m2 * 0h10/m2 + 1h = 6h) 
the coefficient and the constant are saved in a scale that is selected by then given quantity (here 30m2).


marcelsavegnago:
Hello @petrus-v, with this solution do you think it is possible to handle any specific conversion for a product?

petrus-v:
@marcelsavegnago thanks for you interest :)

> Hello @petrus-v, with this solution do you think it is possible to handle any specific conversion for a product?

I guess we can use this module as base to handle different scenarios to convert product quantity. For instance let's imagine you are salling tiles (in m2) with it's related service product to install it where cost use hours UoM.

We could link an uom.converter to the service product to convert from m2 to hours, then on sale order it would be possible to compute unit price per m2 for this service. From my opinion this behavior would be implement in a dedicated module.

petrus-v:
@marcelsavegnago for the story I've managed this conversion over the pricelist in the meantime but this it's quite hacky and makes bugs in margin computations.

petrus-v:
for fun I've add matplotlib to extend this module and display a graph of what would be the expect result to avoid step while configuration a converter scale

image

I've avoid to add it in the module to avoid extra dependency here but could move it if you think it's fare enough ?

marcelsavegnago:
> @marcelsavegnago thanks for you interest :)
> 
> > Hello @petrus-v, with this solution do you think it is possible to handle any specific conversion for a product?
> 
> I guess we can use this module as base to handle different scenarios to convert product quantity. For instance let's imagine you are salling tiles (in m2) with it's related service product to install it where cost use hours UoM.
> 
> We could link an uom.converter to the service product to convert from m2 to hours, then on sale order it would be possible to compute unit price per m2 for this service. From my opinion this behavior would be implement in a dedicated module.

Got it. It is a very useful feature. I believe it will help a lot.

marcelsavegnago:
> I've avoid to add it in the module to avoid extra dependency here but could move it if you think it's fare enough ?

Very cool.

I've avoid to add it in the module to avoid extra dependency here but could move it if you think it's fare enough ?

base_uom_converter/models/uom_converter.py
```diff
@@ -0,0 +1,139 @@
...
+        self.ensure_one()
+        if not uom_qty:
+            uom_qty = self.from_uom_id
+        else:
+            if uom_qty.category_id.id != self.from_uom_id.category_id.id:
+                raise ValidationError(
+                    _(
+                        "You can't convert {uom} (expected {category} uom "
+                        "category) to {to_uom} using this converter {converter}."
+                    ).format(
```
hparfr: please use fstrings.

base_uom_converter/README.rst
```diff
@@ -0,0 +1,108 @@
...
+~~~~~~~
+
+* C2i Change 2 improve
+
+Contributors
+~~~~~~~~~~~~
+
+* `C2i Change 2 improve <https://www.c2i.es>`_:
+
+  * Eduardo Magdalena <emagdalena@c2i.es>
```
hparfr: Not @petrus-v ?

hparfr (changes requested):
It's an interesting module.

Some minor remark about strings.

base_uom_converter/models/uom_converter.py
petrus-v: hum... I thought that the fstring were interpreted before the translation of the string, which would not be appropriate, I will test

base_uom_converter/README.rst
petrus-v: oups well seen :eyes: thanks

base_uom_converter/models/uom_converter.py
petrus-v: as expected:
 if I use that code
```diff
                 raise ValidationError(
                     _(
-                        "You can't convert {uom} to {to_uom} (expect "
-                        "{expected_category} unit category) using this "
-                        "converter {converter}."
-                    ).format(
-                        uom=uom_qty.name,
-                        to_uom=result_uom.name,
-                        expected_category=self.to_uom_id.category_id.name,
-                        converter=self.name,
+                        f"You can't convert {uom_qty.name} to {result_uom.name} (expect "
+                        f"{self.to_uom_id.category_id.name} unit category) using this "
+                        f"converter {self.name}."
                     )
                 )
```
I'm getting this error will translating the module

```
2021-04-09 10:07:04,450 12761 ERROR enrj_com_uom_base_converter odoo.tools.translate: Failed to extract terms from somewhere/src/product-attribute/base_uom_converter/models/uom_converter.py 
  File "somewhere/ocb/odoo/tools/translate.py", line 1118, in _babel_extract_terms
    for extracted in extract.extract(extract_method, src_file, keywords=extract_keywords, options=options):
  File "somewhere/lib/python3.7/site-packages/babel/messages/extract.py", line 322, in extract
    for lineno, funcname, messages, comments in results:
  File "somewhere/lib/python3.7/site-packages/babel/messages/extract.py", line 484, in extract_python
    value = eval(code, {'__builtins__': {}}, {})
  File "<string>", line 2, in <module>
NameError: name 'uom_qty' is not defined
```

base_uom_converter/README.rst
petrus-v: done

base_uom_converter/models/uom_converter.py
hparfr: Strange, I did the following test: 
```py
$ odoo shell
from odoo import _
_(f"Hello { self.name } ") 
> 'Hello OdooBot '
```

petrus-v: My traceback wasn't raising at runtime but while generating po/pot files.

Also, in your example `"Hello OdooBot "` will be interpreted before calling translation function `_()`  we don't want to translate each case for each users only the template string "Hello {something}" which will translate for any users. That's why I use `.format(` after the result of the translated string (`_()`).

Note that [odoo offer some facitilies](https://github.com/OCA/OCB/blob/14.0/odoo/tools/translate.py#L451) using the old deprecated `%-formatting` that we can call like this `_("Hello %", self.name)` but I avoided to depends on that assumption if odoo change that behavior in future versions.

hparfr: I was not aware of these issues with f-strings.

.format() have some security issues and should be avoided. 

% is an acceptable solution then

petrus-v: I didn't know about security issue using `.format()` could you point me some references or tell me more about which attack to mitigate? the templated string is not set by end users I can't see why there are more risk than %-string ? :thinking:

hparfr: https://lucumr.pocoo.org/2016/12/29/careful-with-str-format/

petrus-v: Thanks, oh yes of course users can access to translation string so I supposed we would get the same kind of issue with f-string by the way.

[string.Template](https://docs.python.org/3/library/string.html?highlight=template#string.Template) could be an alternative. I'll change to `%-string` !

hparfr: I suppose f-string are safe according to:
https://security.stackexchange.com/questions/238338/are-there-any-security-concerns-to-using-python-f-strings-with-user-input

petrus-v: In fact I can't see an easy way to set a variable with an f-string template so that kind of security issue  (where end users are able to set the template string) is not possible !

To do something similar as funcionnality (let enduser set the template string) like in the last stackexchange would require an insecure eval likes:
```python
mavar = "{obj.name}"

class A:
  name="hello"
 
obj = A()

eval(f"f'{mavar}'")
'hello'

# or

mavar = "f'{obj.name}'"

eval(mavar)
'hello'
```
I was curious to make sure the odoo `safe_eval` mitigate f-string usage, it does:

```python
>>> from odoo.tools.safe_eval import safe_eval
>>> safe_eval('test', locals_dict={'test': "hello"})
'hello'
>>> safe_eval('f"{test}"', locals_dict={'test': "hello"})
  File "<stdin>", line 1, in <module>
  File "/home/pverkest/clients/cgi/cgi-erp/ocb/odoo/tools/safe_eval.py", line 328, in safe_eval
    c = test_expr(expr, _SAFE_OPCODES, mode=mode)
  File "/home/pverkest/clients/cgi/cgi-erp/ocb/odoo/tools/safe_eval.py", line 189, in test_expr
    assert_valid_codeobj(allowed_codes, code_obj, expr)
  File "/home/pverkest/clients/cgi/cgi-erp/ocb/odoo/tools/safe_eval.py", line 166, in assert_valid_codeobj
    raise ValueError("forbidden opcode(s) in %r: %s" % (expr, ', '.join(opname[x] for x in (code_codes - allowed_codes))))
ValueError: forbidden opcode(s) in 'f"{test}"': FORMAT_VALUE
```

petrus-v: @hparfr thanks for this review, I've push 7ebb445 to use %-string instead format-string.

hparfr (approved):

petrus-v:
Hi dear @OCA/product-maintainers, I'm getting a bit frustrated that one PRs with one positive review get closed automatically. I should probably advertise and request more to get my PRs reviews and do more review as well (I try to review at least 2 PRs when I a create a new one and will try to organize myself give more reviews in coming months).

Is there any guideline to re-open stale PR ? 

Or should consider there is no interest by the community to this work (which I would feel un-fair where large company can do the work and review their PRs while as freelance I can't) ? 

Is there any guideline to kindly request review ?

Regards,
Pierre

PR #813, opened 2021-02-25
[12.0] FIX product_pricelist_supplierinfo: _get_supplierinfo_pricelist_price when date is datetime

eLBati:
date may come from context


eLBati:
I can't reproduce the travis error locally, running exactly the same commands.

It seems related to `product_weight` module.
@tbaden @florian-dacosta any hint?

pedrobaeza:
Main branch should be gone, so let's try to merge (following the same patch proposed at #820):

/ocabot merge patch

pedrobaeza:
Is this needed as well in v13?

eLBati:
@pedrobaeza thanks.

Module is not available for v13

PR #814, opened 2021-02-26
[11.0][mig] product_gtin

dsolanki-initos:


dsolanki-initos:
We drop this PR as we currently focus on the Latest version

PR #820, opened 2021-03-05
[FIX] product_pricelist_supplierinfo: Convert to date the date in context 

sergio-teruel:
cc @Tecnativa  TT28603
ping @carlosdauden @chienandalu


sergio-teruel:
Ups!! similar fix here https://github.com/OCA/product-attribute/pull/813

pedrobaeza:
Duplicated of #813

PR #821, opened 2021-03-06
[FIX] product_pricelist_supplierinfo: Fix broken get price with False date parameter

sergio-teruel:
cc @Tecnativa TT28603

### DIFF
--- a/product_pricelist_supplierinfo/models/product_template.py
+++ b/product_pricelist_supplierinfo/models/product_template.py
@@ -25,10 +25,11 @@ class ProductTemplate(models.Model):
         if product:
             if type(date) == datetime:
                 date = date.date()
+            # Fails is date is False so pass None
             seller = product._select_seller(
                 partner_id=rule.filter_supplier_id,
                 quantity=quantity,
-                date=date)
+                date=date or None)
             if seller:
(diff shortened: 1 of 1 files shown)


product_pricelist_supplierinfo/models/product_template.py
```diff
@@ -25,10 +25,11 @@ def _get_supplierinfo_pricelist_price(
         if product:
             if type(date) == datetime:
                 date = date.date()
+            # Fails is date is False so pass None
```
pedrobaeza, line 28:
```suggestion
            # Fails if date is False, so pass None
```

pedrobaeza:
Isn't this an error in the module that passes False instead of None?

carlosdauden:
Maybe we should change this:
https://github.com/OCA/product-attribute/blob/95d0805600cbfa21b32550f2fa760ffd4e5b53b3/product_pricelist_supplierinfo/models/product_pricelist.py#L29
to this other:
`date=date or context.get('date') or fields.Date.today(),`

pedrobaeza:
Yeah, if that is the problem, there's a chance of having `date` key in the context, but containing `False`/`None`.

chienandalu (commented):
I'd vote for @carlosdauden suggestion

PR #826, opened 2021-03-10
12.0 add product brand supplier discount

MPoch-PlanetaTIC:
product_brand_supplier_discount module permits to assign a purchase discount
to all products of a product.brand for specific suppliers


rousseldenis:
@MPoch-PlanetaTIC Could you check travis?

But I think brand related things have been moved to https://github.com/OCA/brand

PR #828, opened 2021-03-11
[13.0] [MIG] product_pricelist_supplierinfo: MIgration to v13.0

CarlosRoca13:
cc @Tecnativa TT28618

please @sergio-teruel @carlosdauden review this


product_pricelist_supplierinfo/models/product_template.py
```diff
@@ -48,7 +48,7 @@ def _get_supplierinfo_pricelist_price(
             if seller and seller.product_uom != price_uom:
                 price = seller.product_uom._compute_price(price, price_uom)
 
-            convert_to_price_uom = lambda price: self.uom_id._compute_price(
+            convert_to_price_uom = lambda price: self.uom_id._compute_price(  # noqa
```
joao-p-marques: Why this `noqa` here? Can you explain a little what is the concrete problem?
IMHO, if the problem is not fixable, it is better to specify the error you are ignoring, rather than ignoring everything.

product_pricelist_supplierinfo/models/product_pricelist.py
```diff
@@ -0,0 +1,49 @@
...
+
+class ProductPricelist(models.Model):
+    _inherit = "product.pricelist"
+
+    def _compute_price_rule(self, products_qty_partner, date=False, uom_id=False):
+        """Recompute price after calling the atomic super method for
+        getting proper prices when based on supplier info.
+        """
+        rule_obj = self.env["product.pricelist.item"]
+        result = super(ProductPricelist, self)._compute_price_rule(
```
joao-p-marques:
```suggestion
        result = super()._compute_price_rule(
```

product_pricelist_supplierinfo/tests/test_product_supplierinfo.py
```diff
@@ -0,0 +1,300 @@
...
+
+from datetime import date
+
+from odoo.tests import common
+
+
+class TestProductSupplierinfo(common.SavepointCase):
+    @classmethod
+    def setUpClass(cls):
+        super(TestProductSupplierinfo, cls).setUpClass()
```
joao-p-marques:
```suggestion
        super().setUpClass()
```

joao-p-marques (commented):
Code review

CarlosRoca13:
Changes done @joao-p-marques 

That `noqa` was for avoid the precommit error `E731 do not assign a lambda expression, use a def`. Now it is specified

joao-p-marques (approved):

product_pricelist_supplierinfo/models/product_template.py
```diff
@@ -48,7 +48,7 @@ def _get_supplierinfo_pricelist_price(
             if seller and seller.product_uom != price_uom:
                 price = seller.product_uom._compute_price(price, price_uom)
 
-            convert_to_price_uom = lambda price: self.uom_id._compute_price(
+            convert_to_price_uom = lambda price: self.uom_id._compute_price(  # noqa: E731
                 price, price_uom
             )
```
chienandalu: Don't sweep under the carpet :wink: You can make an internal method to achieve the same.

```suggestion
            def convert_to_price_uom(price, price_uom=price_uom):
                return self.uom_id._compute_price(price, price_uom)
```

Anyway, maybe it's pointless to don't call it directly anyway

CarlosRoca13:
@chienandalu changed!! I didn't know that I can define a method inside other :) Thanks for the review

chienandalu:
Yeah, ok. But my question remains, why not to call the original method directly with the proper arguments?

CarlosRoca13:
True, I didn't saw that comment 😅😅 

BTW I don't know why it was make like that... I have done what you say

chienandalu:
> BTW I don't know why it was make like that... I have done what you say

My guess is saving lines :slightly_smiling_face:

victoralmau (approved):

chienandalu (approved):

PR #829, opened 2021-03-12
[14.0][ADD] product_pricelist_from_commitment_date

jcoux:
When the sale order commitment date is set,
this date is used by pricelist to compute price unit instead of using order date.


jcoux:
Bad place: to move into sale-workflow ;)

PR #831, opened 2021-03-17
[FIX] product_standard_price_tax_included: handle case where pricelist items has discount (and surcharge, ...)

legalsylvain:
Context : without that patch, if a pricelist has a discount / surcharge and the ``base`` is ``standard_price_tax_included``. The computation is bad. 
The problem comes from a useless overload of ``_compute_price_rule`` that I removed. (my bad during the port of this module).

I also improved tests that was failing if the currency of the demo pricelist is not the same as the currency of the company. (demo data doesn't seems to be predictive)

CC : @grap, @quentinDupont (deck/#/board/144/card/1326)


legalsylvain:
/ocabot merge patch

(Patch used un production since 15 monthes)

PR #832, opened 2021-03-17
[14.0][ADD] product_supplierinfo_code

rousseldenis:


hparfr (approved):

product_supplierinfo_code/views/product_template.xml
```diff
@@ -0,0 +1,63 @@
...
+     License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl). -->
+<odoo>
+
+    <record model="ir.ui.view" id="product_template_form_view">
+        <field name="name">product.template.form (in product_supplierinfo_code)</field>
+        <field name="model">product.template</field>
+        <field name="inherit_id" ref="product.product_template_form_view" />
+        <field name="arch" type="xml">
+            <group name="purchase" position="inside">
+                <field name="supplier_product_code" />
```
florian-dacosta, line 12: @rousseldenis 
It seems really strange in the form view, because the code is displayed with no label or information with the meaning.
Also, I don't see the point to display it here since we can see it in the supplierinfo list, in the same page.
supplier_code

Shouldn't we remove this view ?
Other wise, works well

florian-dacosta:
@rousseldenis 
Also, we had a similar module, with just one more feature : the possibility to search a picking by supplier code : https://github.com/akretion/odoo-usability/blob/10.0/product_search_supplier_code/views/picking_view.xml

Do you think it could be included here ? Or should we have a separated module just for this ?
IMO, if we want to be able to search by supplier code, it is usually because we use the stock module, but maybe it is too much dependency.
Or I can make a new really small module stock_search_supplier_code in stock-logistics-warehouse
Or if the module is too small, I can keep this part in akretions's odoo-usability repo.
Your opinion is welcome!

rousseldenis:
> @rousseldenis
> Also, we had a similar module, with just one more feature : the possibility to search a picking by supplier code : https://github.com/akretion/odoo-usability/blob/10.0/product_search_supplier_code/views/picking_view.xml
> 
> Do you think it could be included here ? Or should we have a separated module just for this ?
> IMO, if we want to be able to search by supplier code, it is usually because we use the stock module, but maybe it is too much dependency.
> Or I can make a new really small module stock_search_supplier_code in stock-logistics-warehouse
> Or if the module is too small, I can keep this part in akretions's odoo-usability repo.
> Your opinion is welcome!

Indeed, in a stock module is better as this one pulls a minimum of dependencies.

product_supplierinfo_code/views/product_template.xml
rousseldenis: Mmmh, yes, I think this is a part that should go into our side as project specific.

PR #817, opened 2021-03-03
[13.0] [MIG] product_pricelist_direct_print_website_sale: Migration to v13.0

CarlosRoca13:
cc @Tecnativa TT28443

Please @sergio-teruel @carlosdauden review this

PR #839, opened 2021-03-24
[13.0][FIX] product_packaging_unit_price_calculator: Tests broken when website_sale installed

CarlosRoca13:
cc @Tecnativa TT28443

The test of this module were failing when website_sale is installed. It is caused for a problem with Forms... As this module is not using `onchange` we can avoid using Forms.

Please @chienandalu @joao-p-marques, review this changes.


CarlosRoca13:
We can see the problem on this PR https://github.com/OCA/product-attribute/pull/817

chienandalu (approved):

joao-p-marques (approved):

chienandalu:
As far as we found out, it looks like a side effect of some kind with Forms and transient models, but we didn't dive deep into it

PR #842, opened 2021-03-29
[12.0][IMP] product_pricelist_revision: Allow create pricelist item

carlosdauden:
This change is made to enable the option to import records.

@Tecnativa TT27524

I will port this change to v13 after merge

### DIFF
--- a/product_pricelist_revision/views/pricelist_view.xml
+++ b/product_pricelist_revision/views/pricelist_view.xml
@@ -10,10 +10,9 @@
         <field name="arch" type="xml">
             <tree position="attributes">
                 <attribute name="editable">bottom</attribute>
-                <attribute name="create">false</attribute>
             </tree>
             <field name="name" position="before">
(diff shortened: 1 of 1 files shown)


pedrobaeza (approved):

victoralmau (approved):

pedrobaeza:
Cherry-picked for v13 at 95bdfb7687c5ec2c66e54decccb161fdafc342ab

francesco-ooops:
@carlosdauden @pedrobaeza I think this feature was not complete AFAICS as it allows user to create a pricelist item without specifying to which pricelist it should belong to - would be good to fix or add to roadmap (currently affecting 13 and 14)

---

I've found this module as I was looking for a menu to import pricelist items, and in general to keep under control pricelists with many lines. Would be good IMHO to split this module in two - a base with only the pricelist item menu/import/record creations, and another module with the revision features

Do you think that's doable on 14? Or should this be kept for future portings?

PR #843, opened 2021-04-01
[13.0][IMP] product_secondary_unit: Add a tree view for secondary UoM

ernestotejeda:
Cc @Tecnativa TT28726

 Add a tree view for secondary UoM so that it can be used by other modules


pedrobaeza (approved):

pedrobaeza:
@sergio-teruel We are extracting the secondary view tree to an isolated view to be reusable for other things.

sergio-teruel (approved):

PR #725, opened 2020-11-17
[14.0][MIG]product_state

watthanun:

PR #846, opened 2021-04-13
[14.0][MIG] product_state_history

rousseldenis:
Depends on:

- [x] #725


hailangvn (approved):
Code review: LGTM. Thank you.

product_state_history/models/product_template.py
```diff
@@ -0,0 +1,42 @@
...
+        }
+        return vals
+
+    def _set_product_state_history(self, vals):
+        """Create a list of history values"""
+        history_obj = self.env["product.state.history"]
+        history_vals = []
+        state_id = vals["product_state_id"]
+        for template in self:
+            history_vals.append(
```
xavier-bouquiaux, line 26: @rousseldenis 
Line are duplicated for now (the change in history are shown 2 times)
in debug we saw that the write is called 2 times when we change a product_state
could you add this if please ?

```
     if state_id != template.product_state_id.id:
                history_vals.append(
                    template._prepare_product_state_history_values(state_id)
                )
```

rousseldenis: @xavier-bouquiaux @sbejaoui I agree with this protection.

In fact, it should have been a good point to check why two write() are called.

This is due to product_state implementation as both state and state_id are in the view.

I will add a readonly=True to state field there https://github.com/OCA/product-attribute/blob/14.0/product_state/models/product_template.py#L15

product_state_history/models/product_template.py
```diff
@@ -0,0 +1,43 @@
...
+        }
+        return vals
+
+    def _set_product_state_history(self, vals):
+        """Create a list of history values"""
+        history_obj = self.env["product.state.history"]
+        history_vals = []
+        state_id = vals["product_state_id"]
+        # We check if new state is different from the current one
+        for template in self.filtered(lambda t: t.product_state_id.id != state_id):
```
sbejaoui, line 25: it's better to add the if inside the for loop, otherwise it will loop twice

```suggestion
        for template in self:
            if template.product_state_id.id != state_id:
```

ivantodorovich (approved):

PR #847, opened 2021-04-13
[14.0][MIG] product brand

xavier-bouquiaux:


youring:
FYI, version 14.0 already migrated by @joao-p-marques https://github.com/OCA/brand/pull/58
product_brand has been move to [OCA/brand](https://github.com/OCA/brand ) repo since version 13.0, although IMHO it has nothing related to the [brand](https://github.com/OCA/brand/tree/13.0/brand) module.

joao-p-marques:
Thanks for the reference @youring 
Indeed, this is already migrated there, so I guess we can close this...

> although IMHO it has nothing related to the brand module.

Yes, it does not depend on that module. However, the purpose of the https://github.com/OCA/brand repo is to aggregate modules that “Manage brands for products and companies”, as seen in its README, so it also makes sense to put it there...
You can always check the migration issue and respective comments to follow the status of a migration: https://github.com/OCA/product-attribute/issues/508#issuecomment-542244676

xavier-bouquiaux:
Thanks to both of you @youring @joao-p-marques 

I'll delete this

KR

xbo

PR #849, opened 2021-04-20
[14.0][ADD] rating_moderation

david-popa:
This module adds a menu and allows the moderation of ratings by a Rating Moderator


rousseldenis:
Hi @david-popa, thanks for this.

IMHO this has to be moved to server-ux repository. Could you check that?

PR #851, opened 2021-04-28
[13.0][FIX]product_code_unique:check unique code in all companies

PabloPaezSheridan:
The default_code have to be unique across all products
no matters the company their belong

### DIFF
--- a/product_code_unique/models/product.py
+++ b/product_code_unique/models/product.py
@@ -1,7 +1,7 @@
 # Copyright (C) 2018 - TODAY, Open Source Integrators
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
 
-from odoo import models
+from odoo import models, api
 
 
 class ProductProduct(models.Model):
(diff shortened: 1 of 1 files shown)


legalsylvain (changes requested):
I see it redondant with the current sql constraints. Or did i missed something ?

PabloPaezSheridan:
@legalsylvain no is not redundant, sql constraint wasn't working properly when products have different companies, in that case same default_code was allowed. This FIX have into account those scenaries

legalsylvain:
Sorry to insist, but I don't get how a sql constrains ``"unique(default_code)`` allows the same ``default_code`` for various companies. 
I just tested on runbot 13.0 and the module works properly without your patch, preventing the creation of product with same default_code on different companies.

I guess that your instance doesn't have the sql constrains defined due to duplicates before the installation.

PabloPaezSheridan:
@legalsylvain sorry me i have had a raugh day, i will edit the pr description, and the  commit name, the problem here is that default_code is only verified unique when it has a company. We are implementing in our clients products without companies in order to make visilble all product across all companies, but in that situations this module was not working

Sorry again, i have misunderstood the problem of this fix when i made the PR. Tomorrow i will double check the functionality again and update this PR description and name

pedrobaeza:
@PabloPaezSheridan maybe when you installed the module, there was already duplicated codes, so the SQL constraint was not applied, and thus now duplicates are allowed meanwhile you don't remove all the existing duplicates and update the module again. There's code for handling that on the hook when installing, but maybe it's not working correctly.

PabloPaezSheridan:
@legalsylvain @pedrobaeza sorry about your time, this was all a misunderstanding, this lines were supposed to be in our fork solving a specific scenarie, believed it was for here. I close the PR. Thanks and sorry again

PR #852, opened 2021-04-29
[12.0] new module product_uom_po_domain

legalsylvain:
By default in Odoo, on the ``product.template`` and ``product.product`` form views, the list of the available Purchase UoM is not filtered. (``uom_po_id`` field)

So the UI is proposing all the uoms, even those that are not consistent with the selected main UoM. (defined in the ``uom_id`` field.)

without_module

This module extends the functionallity of product module, adding a domain on this field.

with_module

CC : @quentinDupont


product_uom_po_domain/__manifest__.py
```diff
@@ -0,0 +1,22 @@
+# Copyright (C) 2021 - Today: GRAP (http://www.grap.coop)
+# @author: Sylvain LE GAL (https://twitter.com/legalsylvain)
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+
+{
+    "name": "Product - Domain on Purchase UoM",
+    "version": "12.0.1.0.1",
+    "category": "Product",
+    "author": "GRAP",
```
rvalyi: you should add 'Odoo Community Association (OCA)', this is what is making Travis fail for now.

rvalyi (changes requested):

product_uom_po_domain/__manifest__.py
legalsylvain: thanks. done.

product_uom_po_domain/models/product_template.py
```diff
@@ -0,0 +1,20 @@
...
+
+
+class ProductTemplate(models.Model):
+    _inherit = "product.template"
+
+    uom_category_id = fields.Many2one(
+        comodel_name="uom.category", related="uom_id.category_id")
+
+    uom_measure_type = fields.Selection(
+        related="uom_id.category_id.measure_type")
```
hparfr, line 15: useful ?

legalsylvain: not for this module, but as a technical field that can be used for other purpose in the same spirit as ``uom_category_id``. I see it a little heavy to have two modules for two fields. But if you think it's not acceptable, I'll create another module for this related field.
Let me know.

Note : it is described in the DEVELOP.rst section.

hparfr: It's ok for me. Just wanted the why.

PS: this field has been dropped in v14.

hparfr (approved):
LGTM

product_uom_po_domain/models/product_template.py
legalsylvain: > PS: this field has been dropped in v14.

Yes. It's a pity, because it is usefull.

quentinDupont (approved):
reviewed and tested in clean Odoo :heavy_check_mark:

rvalyi (approved):

marcelsavegnago (approved):

PR #853, opened 2021-05-05
[13.0][IMP] product_cost_security: Add security to Purchase page on product sheet

CarlosRoca13:
cc @Tecnativa TT29577

Please @sergio-teruel @victoralmau review this

### DIFF
--- a/product_cost_security/README.rst
+++ b/product_cost_security/README.rst
@@ -25,8 +25,8 @@ Product Cost Security
 
 |badge1| |badge2| |badge3| |badge4| |badge5| 
 
-Adds a security group for product cost price field.
-Only users with this group can view this field.
+Adds a security group for product cost price field and purchase tab on products
+sheet. Only users with this group can view this two.
 
 **Table of contents**
 
@@ -47,7 +47,7 @@ Usage
 To use this module you need to:
 
 #. Go to product form view logged with this user and you will see the
-   standard_price field.
+   standard_price field and Purchase tab.
 
 Bug Tracker
 ===========
@@ -70,7 +70,10 @@ Authors
 Contributors
 ~~~~~~~~~~~~
 
-* Sergio Teruel <sergio.teruel@tecnativa.com>
+* `Tecnativa <https://www.tecnativa.com>`_:
+
+  * Sergio Teruel
+  * Carlos Roca
 
 Maintainers
 ~~~~~~~~~~~
--- a/product_cost_security/__manifest__.py
+++ b/product_cost_security/__manifest__.py
@@ -12,6 +12,6 @@
     "license": "AGPL-3",
     "application": False,
     "installable": True,
-    "depends": ["product", "purchase"],
+    "depends": ["product"],
     "data": ["security/product_cost_security.xml", "views/product_views.xml"],
 }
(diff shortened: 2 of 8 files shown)


product_cost_security/views/product_views.xml
```diff
@@ -10,6 +10,11 @@
                     name="groups"
                 >product_cost_security.group_product_cost</attribute>
             </xpath>
+            <xpath expr="//page[@name='purchase']" position="attributes">
+                <attribute
+                    name="groups"
+                >product_cost_security.group_product_cost</attribute>
+            </xpath>
```
victoralmau, line 17:
```suggestion
            <xpath expr="//page[@name='purchase']" position="attributes">
                <attribute
                    name="groups"
                >purchase.group_purchase_manager</attribute>
            </xpath>
```

Although its correct, maybe it's a good idea create new addon to set groups to purchase page. What do you think?

CarlosRoca13: On the Purchase page we have the cost associated to each vendor, so I think that security group is correct... But I prefer the opinion of 1 more... What do you think @sergio-teruel ??
image

chienandalu: Or maybe just set the group to the field in the tree element

sergio-teruel: Normally, purchase users must have permission to view the cost of the product, so I think that this is not problem.
Maybe You can try add to group_purchase_user the security cost group as implied_ids, by this way, a purchase user always will have the security_cost group.

product_cost_security/README.rst
```diff
@@ -47,7 +47,7 @@ Usage
 To use this module you need to:
 
 #. Go to product form view logged with this user and you will see the
-   standard_price field.
+   standard_price field and Purchase window.
```
sergio-teruel: and Purchase tab.

product_cost_security/readme/DESCRIPTION.rst
```diff
@@ -1,2 +1,2 @@
-Adds a security group for product cost price field.
-Only users with this group can view this field.
+Adds a security group for product cost price field and purchase window on products
```
sergio-teruel: purchase tab

product_cost_security/readme/USAGE.rst
```diff
@@ -1,4 +1,4 @@
 To use this module you need to:
 
 #. Go to product form view logged with this user and you will see the
-   standard_price field.
+   standard_price field and Purchase window.
```
sergio-teruel: the same

sergio-teruel (changes requested):

CarlosRoca13:
Okey so, changes done

product_cost_security/__manifest__.py
```diff
@@ -12,6 +12,6 @@
     "license": "AGPL-3",
     "application": False,
     "installable": True,
-    "depends": ["product"],
+    "depends": ["product", "purchase"],
```
pedrobaeza: I don't see correct that this module depends on purchase.

CarlosRoca13:
What do you suggest @pedrobaeza ?

product_cost_security/__manifest__.py
pedrobaeza: A new glue module should be done.

PR #856, opened 2021-05-20
[14.0] [NEW] product_tier_validation: New module to validate the product.

Chandresh-OSI:
CC @dreispt


oca_dependencies.txt
```diff
@@ -1 +1,2 @@
 # See https://github.com/OCA/odoo-community.org/blob/master/website/Contribution/CONTRIBUTING.rst#oca_dependencies-txt
+server-ux
```
rousseldenis, line 2: don't need this (see other comment in https://github.com/OCA/partner-contact/pull/1108)

product_tier_validation/__manifest__.py
```diff
@@ -0,0 +1,20 @@
...
+{
+    "name": "Product Tier Validation",
+    "summary": "Support a tier validation process for Products",
+    "version": "14.0.1.0.0",
+    "website": "https://github.com/OCA/product-attribute",
+    "category": "Products",
+    "author": "Open Source Integrators, Odoo Community Association (OCA)",
+    "license": "AGPL-3",
+    "installable": True,
+    "maintainer": "dreispt",
```
rousseldenis:
```suggestion
    "maintainer": ["dreispt"],
```

product_tier_validation/models/product_template.py
```diff
@@ -0,0 +1,32 @@
...
+class ProductTemplate(models.Model):
+    _name = "product.template"
+    _inherit = ["product.template", "tier.validation"]
+    _tier_validation_manual_config = False
+
+    @api.model
+    def create(self, vals):
+        new = super().create(vals)
+        if new.need_validation and new.state != "confirmed":
+            new.active = False
```
rousseldenis, line 20: IMHO, this is a little too intrusive. Maybe should at least be a module option 'Set product inactive before validation'

rousseldenis (changes requested):
LGTM. Some changes

dreispt:
@Chandresh-OSI Travis is red, can you fix the PR please?

Chandresh-SerpentCS:
@dreispt An error is coming from `product_sequence` module

rousseldenis:
@Chandresh-OSI An error in product_secondary_unit module is appearing first:

image

rousseldenis:
@Chandresh-OSI See my comment about setting active == False on each new product creation. This is breaking standard behaviour. That's why other modules are failing. That behaviour should be conditionned to an option.

Chandresh-SerpentCS:
@rousseldenis Yes, thank you for pointing out that. I will correct that.

dreispt:
Yes, I suggest having the default data validation rule to be inactive by default.
And it would be nice for the README to inform that a sample rule is there and can be activated.

product_tier_validation/models/product_template.py
dreispt: IMO the Product should start inactive until is it validated, to be sure it can't be used on transactions.
Still, I now made this easy to configure, and even extend, in case you want specific "product inactivation" conditions.

dreispt:
@rousseldenis I believe your review comments were addressed.

@Chandresh-OSI please add a simple test:
- Activate the default rule
- Create a Product
- It should be inactive
- Run validate
- It should now be active

See https://github.com/OCA/sale-workflow/blob/14.0/sale_tier_validation/tests/test_tier_validation.py#L69

product_tier_validation/__manifest__.py
```diff
@@ -0,0 +1,20 @@
...
+{
+    "name": "Product Tier Validation",
+    "summary": "Support a tier validation process for Products",
+    "version": "14.0.1.0.0",
+    "website": "https://github.com/OCA/product-attribute",
+    "category": "Products",
+    "author": "Open Source Integrators, Odoo Community Association (OCA)",
+    "license": "AGPL-3",
+    "installable": True,
+    "maintainer": ["dreispt"],
```
rousseldenis:
```suggestion
    "maintainers": ["dreispt"],
```

rousseldenis (approved):
Nitpicking. LGTM

Chandresh-OSI:
> @rousseldenis I believe your review comments were addressed.
> 
> @Chandresh-OSI please add a simple test:
> 
> * Activate the default rule
> * Create a Product
> * It should be inactive
> * Run validate
> * It should now be active
> 
> See https://github.com/OCA/sale-workflow/blob/14.0/sale_tier_validation/tests/test_tier_validation.py#L69

@dreispt Tast case file is added in the module. Done !

dreispt (approved):

PR #859, opened 2021-05-27
[14.0] mig product_supplierinfo_for_customer

fredzamoabg:
**product_supplierinfo_for_customer** module migration from 13.0


product_supplierinfo_for_customer/models/product_product.py
```diff
@@ -21,44 +21,48 @@ def _name_search(
...
             .browse(customerinfo_ids)
             .mapped("product_tmpl_id")
             - res_templates
         )
-        product_ids = self._search(
-            [("product_tmpl_id", "in", product_tmpls.ids)],
-            limit=limit,
-            access_rights_uid=name_get_uid,
+        product_ids = list(
+            self._search(
```
tafaRU, line 56: Shouldn't `_search` already return list of ids? (Ref. https://github.com/odoo/odoo/blob/94944e1ed0f3cd7d344ab42788e770ad58d8a49c/odoo/models.py#L4482)

tafaRU (changes requested):

product_supplierinfo_for_customer/models/product_product.py
fredzamoabg: I found that **_search** method returns a **odoo.osv.query.Query** object. 
here I need to extend with ids so I can use list() or **odoo.osv.query.Query._result** (Ref. https://github.com/odoo/odoo/blob/94944e1ed0f3cd7d344ab42788e770ad58d8a49c/odoo/osv/query.py#L147). let me know what do you prefer.

anyway I changed here https://github.com/OCA/product-attribute/pull/859/files#diff-f00b983e1fac4d29d20ae66593f2df2378eb5445c33b11b2ea43421a0a81e463R36 from previous version as **odoo.osv.query.Query** object can be directly pass to **browse** method

tafaRU (approved):

rousseldenis (approved):
Code review

luisg123v (approved):
CC @moylop260

PR #860, opened 2021-05-27
[13.0][FIX] product_pricelist_direct_print: Error when date is not defined

CarlosRoca13:
When we check the price of any product, the date that we pass on the context has to be a Datetime value or None. With this change we avoid to pass a False.

Here you have the reference to the code that check that https://github.com/OCA/OCB/blob/13.0/addons/product/models/product.py#L599-L600

TT29935

### DIFF
--- a/product_pricelist_direct_print/views/report_product_pricelist.xml
+++ b/product_pricelist_direct_print/views/report_product_pricelist.xml
@@ -88,7 +88,7 @@
                                     </td>
                                     <td t-if="pricelist" class="text-right">
                                         <strong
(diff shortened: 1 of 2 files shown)


CarlosRoca13:
Ping @carlosdauden @pedrobaeza

pedrobaeza (approved):

pedrobaeza:
Check always pre-commit.

CarlosRoca13:
Idk why it wasn't launched when I do the commit

dreispt (approved):

PR #864, opened 2021-06-01
[IMP][14.0] base_product_mass_addition: add support for UoM

kevinkhao:
required by https://github.com/OCA/purchase-workflow/pull/1046

### DIFF
--- a/base_product_mass_addition/models/product_product.py
+++ b/base_product_mass_addition/models/product_product.py
@@ -13,16 +13,37 @@ class ProductProduct(models.Model):
 
     qty_to_process = fields.Float(
         compute="_compute_process_qty",
-        inverse="_inverse_set_process_qty",
+        inverse="_inverse_quick_vals",
         help="Set this quantity to create a new line "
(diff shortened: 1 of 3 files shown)


base_product_mass_addition/models/product_product.py
```diff
@@ -13,25 +13,53 @@ class ProductProduct(models.Model):
 
     qty_to_process = fields.Float(
         compute="_compute_process_qty",
-        inverse="_inverse_set_process_qty",
+        inverse="_inverse_quick_vals",
```
rousseldenis: @kevinkhao Is that change wanted as module is already merged ?

kevinkhao: thanks for feedback
reverted https://github.com/OCA/product-attribute/pull/864/commits/f9c49b8d193d248cfdf81e17f938cca793b09142

sebastienbeau (approved):
LGTM code review

bguillot (approved):
Tested

PR #866, opened 2021-06-08
[14.0] product_seasonality: add display_name to config line

simahawk:
Makes easier to reference  existing lines in user messages.

ref: 624

### DIFF
--- a/product_seasonality/models/seasonal_config_line.py
+++ b/product_seasonality/models/seasonal_config_line.py
@@ -30,6 +30,7 @@ class SeasonalConfigLine(models.Model):
     friday = fields.Boolean(default=True)
     saturday = fields.Boolean(default=True)
     sunday = fields.Boolean(default=True)
+    display_name = fields.Char(compute="_compute_display_name")
 
     @api.constrains("date_start", "date_end")
     def _check_date_end(self):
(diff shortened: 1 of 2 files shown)


product_seasonality/models/seasonal_config_line.py
```diff
@@ -59,3 +60,15 @@ def find_for_product(self, prod, config=None):
...
             domain.append(("seasonal_config_id", "=", config.id))
         return self.search(domain)
+
+    def _compute_display_name(self):
+        for rec in self:
+            rec.display_name = rec._name_get()
+
+    def _name_get(self):
+        parts = [
+            f"[{self.seasonal_config_id.display_name}]",
```
jcoux, line 70: As seasonal config as no display name defined, I would have name here, no?
```suggestion
            f"[{self.seasonal_config_id.name}]",
```

product_seasonality/tests/test_seasonal_config.py
```diff
@@ -16,6 +16,13 @@ def _test_conf_line(self, line, date_ok, date_ko):
         for dt in date_ko:
             self.assertFalse(line.is_sale_ok(dt), f"{dt.strftime('%Y-%m-%d')} is wrong")
 
+    def test_display_name(self):
+        line = self.seasonal_conf.config_for_product(self.prod1)
+        self.assertEqual(
+            line.display_name,
+            f"[{self.seasonal_conf.display_name}] {self.prod1.display_name} ({line.id})",
```
jcoux, line 23:
```suggestion
            f"[{self.seasonal_conf.name}] {self.prod1.display_name} ({line.id})",
```

product_seasonality/models/seasonal_config_line.py
simahawk: display_name is the safest in theory. All models have it no matter if you have name or not. It normally falls back to `name`.
That's why I did this fix https://github.com/OCA/search-engine/pull/89 because it's more reliable.

jcoux: Ok, then if we add a custom display name later, it will continue to works here: :+1:

jcoux (approved):

PR #872, opened 2021-06-24
[WIP][14.0][ADD] product_cost_adjustment

patrickrwilson:
Adds product cost adjustment module that allows a user to follow a similar process used for inventory adjustments but for cost adjustments on manual valuation products.

- Needs Tests


product_cost_adjustment/models/cost_adjustment.py
```diff
@@ -0,0 +1,239 @@
...
+                     line:\n\t%s - cost: %s""",
+                    negative.product_id.display_name,
+                    negative.product_cost,
+                )
+            )
+        return True
+
+    def _remove_unchanged_lines(self):
+        for line in self.line_ids:
+            if line.product_cost == line.product_original_cost:
```
dreispt, line 113: Nitpick: comparing floats is tricky and `==` should not be used for this (even if rounded). You can use `odoo.tools.float_compare()`.

product_cost_adjustment/models/cost_adjustment.py
```diff
@@ -0,0 +1,239 @@
...
+                "product_original_cost": product.standard_price,
+                "product_cost": product.standard_price,
+                "qty_on_hand": product.sudo().quantity_svl,
+            }
+            vals.append(line_values)
+        return vals
+
+    def action_open_cost_adjustment_lines(self):
+        self.ensure_one()
+        action = {
```
dreispt, line 203: You could move this to the end, so that it then includes context and domain.
It could even be returned then, without assigning to a variable.

product_cost_adjustment/models/cost_adjustment.py
```diff
@@ -0,0 +1,239 @@
...
+        # Define domains and context
+        domain = [("cost_adjustment_id", "=", self.id)]
+        action["view_id"] = self.env.ref(
+            "product_cost_adjustment.cost_adjustment_line_view_tree"
+        ).id
+        action["context"] = context
+        action["domain"] = domain
+        return action
+
+    def copy_data(self, default=None):
```
dreispt, line 223: You mean `copy()`?
I'm not familiar with `copy_data()`.

product_cost_adjustment/models/cost_adjustment_line.py
```diff
@@ -0,0 +1,189 @@
...
+
+    is_editable = fields.Boolean(
+        string="Editable?", help="Technical field to restrict editing."
+    )
+    cost_adjustment_id = fields.Many2one(
+        "cost.adjustment",
+        string="Cost Adjustment",
+        check_company=True,
+        index=True,
+        ondelete="cascade",
```
dreispt, line 27: restrict?

product_cost_adjustment/models/cost_adjustment_line.py
```diff
@@ -0,0 +1,189 @@
...
+        states={"confirm": [("readonly", False)]},
+        default=0,
+        copy=False,
+    )
+    difference_cost = fields.Float(
+        string="Difference",
+        compute="_compute_difference",
+        help="Indicates the gap between the product's original cost and its new cost.",
+        readonly=True,
+        search="_search_difference_cost",
```
dreispt, line 60: Why not store and save a few lines of code?

product_cost_adjustment/models/cost_adjustment_line.py
```diff
@@ -0,0 +1,189 @@
...
+
+    @api.depends(
+        "product_original_cost", "product_id.standard_price", "product_id.quantity_svl"
+    )
+    def _compute_outdated(self):
+        for line in self:
+            line.price_outdated = (
+                True
+                if line.product_original_cost != line.product_id.standard_price
+                else False
```
dreispt, line 107: Simpler this way:
```suggestion
                line.product_original_cost != line.product_id.standard_price
```

product_cost_adjustment/models/cost_adjustment_line.py
```diff
@@ -0,0 +1,189 @@
...
+            )
+            line.qty_outdated = (
+                True
+                if line.product_id.sudo().quantity_svl != line.qty_on_hand
+                else False
+            )
```
dreispt, line 113:
```suggestion
            line.qty_outdated = line.product_id.sudo().quantity_svl != line.qty_on_hand
```

product_cost_adjustment/models/cost_adjustment_line.py
```diff
@@ -0,0 +1,189 @@
...
+            if existings:
+                raise UserError(
+                    _(
+                        """There is already one cost adjustment line for this product,
+                         you should rather modify this one instead of creating a
+                         new one."""
+                    )
+                )
+
+    def _search_difference_cost(self, operator, value):
```
dreispt, line 167: Can storing the field avoid this code?

product_cost_adjustment/models/product.py
```diff
@@ -0,0 +1,153 @@
...
+from odoo import _, api, models
+from odoo.exceptions import UserError
+from odoo.tools import float_is_zero
+
+
+class ProductProduct(models.Model):
+    _inherit = "product.product"
+
+    @api.model
+    def _change_standard_price(self, new_price, account=False):
```
dreispt, line 13: AFAICT this feature is out of the box.
Just set a value on the `standard_price` field
See https://github.com/odoo/odoo/blob/f11c807217cccd96fa18fe24c0cbb436bb1830d1/addons/stock_account/models/product.py#L105

PS: we will want to be able to use specific accounts, depending on the Adjustment Type.
For that we need to extend  `get_product_accouns` (https://github.com/odoo/odoo/blob/f11c807217cccd96fa18fe24c0cbb436bb1830d1/addons/stock_account/models/product.py#L85)

No useful arguments to do this, so instead  we will need to set a context key to let the extension code to look for accounts on the right Adjustment Type record. This could be in another PR, I can help with it.

dreispt (commented):
Lots of attention to detail here, good job.
Most comments are minor suggestions.
Only significant concern is regarding `_change_standard_price()`.

patrickrwilson:
This will be moving to stock-logistics-warehouse repo under new name stock_inventory_revaluation

PR #874, opened 2021-06-24
[FIX] product_template_tags: fix rule because multicompany scheme change

jcadhoc:


jjscarafia (approved):
LGTM

product_template_tags/__manifest__.py
```diff
@@ -4,7 +4,7 @@
 {
     "name": "Product Template Tags",
     "summary": "This addon allow to add tags on products",
-    "version": "13.0.2.0.0",
+    "version": "13.0.2.1.0",
```
simahawk, line 7: you should not change version, the bot will do it. Also, this is a patch not a feature ;)

simahawk (approved):
could you drop the version change?

rousseldenis:
@jcadhoc

PR #869, opened 2021-06-17
[14.0] product_assortment: misc fixes

Kev-Roche:

PR #875, opened 2021-07-01
[14.0] product_assortment: fix default model_id

simahawk:
Do not fail if model_id is present but falsy.


rousseldenis (approved):

hparfr:
please have a look to #869

rousseldenis:
> please have a look to #869

Indeed, I was thinking that I've already seen that fix @simahawk

jcoux (approved):

PR #878, opened 2021-07-07
[13.0][IMP] product_assortment: Allow to use assortment based on partner domain

sergio-teruel:
cc @Tecnativa TT30584
ping @carlosdauden @joao-p-marques

### DIFF
--- a/product_assortment/models/ir_filters.py
+++ b/product_assortment/models/ir_filters.py
@@ -1,7 +1,11 @@
 # Copyright 2018-2019 ACSONE SA/NV
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+import datetime
+
 from odoo import _, api, fields, models
 from odoo.osv import expression
+from odoo.tools.safe_eval import safe_eval
 
 
 class IrFilters(models.Model):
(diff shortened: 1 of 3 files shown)


product_assortment/models/ir_filters.py
```diff
@@ -39,6 +43,25 @@ def _get_default_is_assortment(self):
     record_count = fields.Integer(compute="_compute_record_count")
 
     is_assortment = fields.Boolean(default=lambda x: x._get_default_is_assortment())
+    partner_domain = fields.Text(default="[]", required=True)
+    all_partner_ids = fields.Many2many(
```
joao-p-marques, line 47: Currently this is never seen by the user. Is it intended? Or could another button be added like this?
image

joao-p-marques (commented):
Code and functional review. Just a small doubt, otherwise looks good :+1:

product_assortment/models/ir_filters.py
sergio-teruel: Yes @joao-p-marques This is only a technical field to be use in other modules like as sale_order_product_assortment
Maybe we could add the smartbutton to open related partners

carlosdauden: The domain widget shows a button to access the records, so to make it homogeneous, perhaps the current smartbutton can be removed.

pedrobaeza:
@joao-p-marques @carlosdauden please finish the review

carlosdauden (approved):

joao-p-marques (approved):

PR #880, opened 2021-07-08
[14.0][IMP] product_template_tags: usability improvements

ivantodorovich:
A batch of small improvements on this module

* Added sequence field
* Fixed company security rules
* Improved tags form view

before:
form_before

after:
form_after

* Improved tags tree view
* Improved product.template kanban view

before:
before

after:
after

* Added tags product.product kanban view
* Removed menu item from general configuration
* Added glue modules to show Product Tags menues in specific apps

| _before: (in general settings)_ | _after: (in glue modules)_ |
| --- | --- |
| removed_menu | image
 |


product_template_tags/models/product_template_tag.py
```diff
@@ -5,11 +5,11 @@
 
 
 class ProductTemplateTag(models.Model):
-
     _name = "product.template.tag"
     _description = "Product Tag"
 
     name = fields.Char(string="Name", required=True, translate=True)
+    sequence = fields.Integer(default=10)
```
legalsylvain, line 13: Sequencecomes generaly with the _order option. Dont' you think

purchase_product_template_tags/__manifest__.py
```diff
@@ -0,0 +1,16 @@
+# Copyright 2021 Camptocamp SA
+# @author Iván Todorovich <ivan.todorovich@gmail.com>
```
legalsylvain, line 2: If the name is product_tempkate_tags_purchase, le module will be near in the folder.

legalsylvain (commented):
Minor remarks / partial review.

product_template_tags/models/product_template_tag.py
ivantodorovich: Oops! Forgot that one! Thanks!

purchase_product_template_tags/__manifest__.py
ivantodorovich: Yes but that's kind of breaking the guidelines

https://github.com/OCA/odoo-community.org/blob/master/website/Contribution/CONTRIBUTING.rst#id6
> When combining an Odoo module with another from OCA, Odoo's name goes before. I.e., if you want to combine Odoo's crm with OCA's partner_firstname, the name should be crm_partner_firstname.

legalsylvain: Ow i didn't know that rule. Dont like it ! But ok !

simahawk: I don't like it neither :rofl: 

but let's stick to it :wink:

simahawk (changes requested):
nice work, thanks!

Please split changes in different commits, like:

1. 1 commit per new module eg: `Add purchase_product_template_tags`
2. 1 commit for the fix of the company rule
3. 1 commit for all the other UI improvements on the base module

purchase_product_template_tags/__manifest__.py
ivantodorovich: Yeah, I'm with you too :sweat_smile: 
Maybe we can propose a change in the guidelines. Like, all of the above and:

> [...] unless it's a very small glue module, in which case it's better to keep them closer together when sorted alphabetically

:smile:

simahawk (approved):
@legalsylvain good for you?

TDu (approved):

PR #881, opened 2021-07-09
[12.0][FIX] product_pricelist_supplierinfo: Avoid min qty in Supplierinfo

CarmenMiranda:
Before this commit, using the field `no_supplierinfo_min_quantity` in the `product.pricelist.item` only allow us to use `product.supplierinfo` with `min_qty` equals 0 or 1, even when there is a supplierinfo with the same sequence but a higher `min_qty` and the order of that model is 'sequence, min_qty desc, price' which tell us that the seller with a higher `min_qty` should be first.

For example in the test `test_pricelist_min_quantity` when the field `no_supplierinfo_min_quantity` is set to True the order of the
supplierinfo arrive first the one with the vendor "cls.supplier1" and the price 50 because both sellers have the same sequence but this has the higher `min_qty, but with the method `_select_seller` sending the qty with 1 only allows us to use the first supplierinfo with the qty of 1 or 0 so that is why the price that was obtained was 10 instead of 50.

Test Supplierinfo Error

Adding a param to the method `_select_seller` in the model `product.product` 'avoid_min_qty' that can be sent in the parameter
`params` where we will avoid the validation of the `min_qty` of the supplierinfo to use on the method `_get_supplierinfo_pricelist_price` of the model `product.template` when the field `no_supplierinfo_min_quantity` is set as True.

### DIFF
--- a/product_pricelist_supplierinfo/models/product_product.py
+++ b/product_pricelist_supplierinfo/models/product_product.py
@@ -3,7 +3,7 @@
 # Copyright 2019 Tecnativa - Carlos Dauden
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
 
-from odoo import models
+from odoo import models, fields
 
 
 class ProductProduct(models.Model):
@@ -24,3 +24,34 @@ class ProductProduct(models.Model):
             return dict.fromkeys(self.ids, 1.0)
         return super().price_compute(
             price_type, uom=uom, currency=currency, company=company)
+
+    def _select_seller(
+            self, partner_id=False, quantity=0.0, date=None, uom_id=False,
+            params=False):
+        if not params or not params.get('avoid_min_qty'):
+            return super()._select_seller(partner_id, quantity, date, uom_id, params)
+
+        # Section based on the original Odoo's method '_select_seller' but in here the
+        # validation of the min_qty on the seller is being avoided with the use of the
+        # field `no_supplierinfo_min_quantity` on the model `product.pricelist.item`.
+        self.ensure_one()
+        if date is None:
+            date = fields.Date.context_today(self)
+
+        res = self.env['product.supplierinfo']
+        sellers = self._prepare_sellers(params)
+        company_id = self.env.context.get('force_company')
+        if company_id:
+            sellers = sellers.filtered(
+                lambda s: not s.company_id or s.company_id.id == company_id)
+        for seller in sellers:
+            if (
+                (not seller.date_start or seller.date_start <= date)
+                and (not seller.date_end or seller.date_end >= date)
+                and (not partner_id
+                     or seller.name in [partner_id, partner_id.parent_id])
+                and (not seller.product_id or seller.product_id == self)
+            ):
+                res |= seller
+                break
+        return res
--- a/product_pricelist_supplierinfo/models/product_template.py
+++ b/product_pricelist_supplierinfo/models/product_template.py
@@ -18,17 +18,22 @@ class ProductTemplate(models.Model):
         product = self.product_variant_id
         if product_id:
             product = product.browse(product_id)
-        if rule.no_supplierinfo_min_quantity:
-            quantity = 1.0
         # The product_variant_id returns empty recordset if template is not
         # active, so we must ensure variant exists or _select_seller fails.
         if product:
             if type(date) == datetime:
                 date = date.date()
+            # If the field `no_supplierinfo_min_quantity` is set as True, we need
+            # to use the select seller that doesn't take into account the `min_qty`
+            # for the validation to know if the seller can be used (Using the param
+            # `avoid_min_qty`).
             seller = product._select_seller(
                 partner_id=rule.filter_supplier_id,
                 quantity=quantity,
-                date=date)
+                date=date,
+                params={
+                    'avoid_min_qty': rule.no_supplierinfo_min_quantity,
+                })
             if seller:
                 price = seller._get_supplierinfo_pricelist_price()
         if price:
--- a/product_pricelist_supplierinfo/tests/test_product_supplierinfo.py
+++ b/product_pricelist_supplierinfo/tests/test_product_supplierinfo.py
@@ -2,7 +2,7 @@
 # Copyright 2018 Tecnativa - Pedro M. Baeza
 # License AGPL-3 - See http://www.gnu.org/licenses/agpl-3.0.html
 
-from datetime import date
+from datetime import date, timedelta
 
 from odoo.tests import common
 
@@ -106,7 +106,60 @@ class TestProductSupplierinfo(common.SavepointCase):
         self.assertAlmostEqual(
             self.pricelist.get_product_price(self.product, 10, False), 50,
         )
+        # As the sequence of both supplierinfos is the same, the next field to order
+        # is the min_qty (desc), so the price must be the one with the higher min_qty
+        # no matter if the qty is lower than the min_qty.
         self.pricelist.item_ids[0].no_supplierinfo_min_quantity = True
+        self.assertAlmostEqual(
+            self.pricelist.get_product_price(self.product, 1, False), 50,
+        )
+        self.assertAlmostEqual(
+            self.pricelist.get_product_price(self.product, 5, False), 50,
+        )
+
+    def test_pricelist_no_min_quantity(self):
+        """Testing the validations of the `_select_seller_without_min_qty` method.
+
+        As the sequence of both supplierinfos in the product is the same, the next
+        field to order is the min_qty (desc), so the price must be the one with the
+        higher min_qty (the one with the price 50), but we are going to test the
+        following cases in the supplierinfo with the higher min_qty:
+        - The start date is tomorrow (not current).
+        - The end date is yesterday (not current).
+        - The supplierinfo is current, but there is a 'Vendor' set in the field
+        `filter_supplier_id` of the item in the pricelist.
+        """
+        product = self.product
+        # Case 1: The start date is not a current one so it will be taken the next one.
+        tomorrow = date.today() + timedelta(days=1)
+        product.seller_ids[0].date_start = tomorrow
+        self.pricelist.item_ids[0].no_supplierinfo_min_quantity = True
+        self.assertAlmostEqual(
+            self.pricelist.get_product_price(self.product, 1, False), 10,
+        )
+        self.assertAlmostEqual(
+            self.pricelist.get_product_price(self.product, 5, False), 10,
+        )
+
+        # Case 2: The end date is an expired one so it will be taken the next one.
+        product.seller_ids[0].date_start = False
+        yesterday = date.today() - timedelta(days=1)
+        product.seller_ids[0].date_end = yesterday
+        self.assertAlmostEqual(
+            self.pricelist.get_product_price(self.product, 1, False), 10,
+        )
+        self.assertAlmostEqual(
+            self.pricelist.get_product_price(self.product, 5, False), 10,
+        )
+
+        # Case 3: Current seller but a `filter_supplier_id` is set in the item, so it
+        # will will be taken the next one.
+        product.seller_ids[0].date_start = False
+        product.seller_ids[0].date_end = False
+        self.pricelist.item_ids[0].filter_supplier_id = self.supplier2
+        self.assertAlmostEqual(
+            self.pricelist.get_product_price(self.product, 1, False), 10,
+        )
         self.assertAlmostEqual(
             self.pricelist.get_product_price(self.product, 5, False), 10,
         )


CarmenMiranda:
Hi @JulioSerna This is the PR of the issue that you report to me early today, for the avoidance of the min_qty. Could you review it? 😄

CarmenMiranda:
Hi, @hugho-ad Could you review it, please?? I'm working to add more tests to cover the missing lines of coverage 🙈

CarmenMiranda:
Hi, @moylop260 Could you review it, please?? 😄

JulioSerna:
LGTM 👍 

@moylop260  Could you check this mr?

product_pricelist_supplierinfo/models/product_product.py
```diff
@@ -24,3 +24,35 @@ def price_compute(self, price_type, uom=False, currency=False,
...
         return super().price_compute(
             price_type, uom=uom, currency=currency, company=company)
+
+    @api.multi
+    def _select_seller_without_min_qty(self, partner_id=False, date=None, params=False):
+        """Method base on the original Odoo's method '_select_seller' but in here the
+        validation of the min_qty on the seller is being avoided with the use of the
+        field `no_supplierinfo_min_quantity` on the model `product.pricelist.item`.
+        Odoo's _select_seller() method:
+        https://github.com/odoo/odoo/blob/d17c20e28879d4abd951d56f367e4c1fd21df/addons/product/models/product.py#L566
```
luisg123v: Shorter URL

```suggestion
        https://github.com/odoo/odoo/blob/d17c20e28879/addons/product/models/product.py#L566
```

product_pricelist_supplierinfo/models/product_product.py
```diff
@@ -24,3 +24,35 @@ def price_compute(self, price_type, uom=False, currency=False,
...
+            sellers = sellers.filtered(
+                lambda s: not s.company_id or s.company_id.id == company_id)
+        for seller in sellers:
+            if seller.date_start and seller.date_start > date:
+                continue
+            if seller.date_end and seller.date_end < date:
+                continue
+            if partner_id and seller.name not in [partner_id, partner_id.parent_id]:
+                continue
+            if seller.product_id and seller.product_id != self:
```
luisg123v: Too much if's increase cyclomatic complexity, you may joim all of them:

```
if (
    (seller.date_start and seller.date_start > date)
    or (seller.date_end and seller.date_end < date)
    or (partner_id and seller.name not in [partner_id, partner_id.parent_id])
    or (seller.product_id and seller.product_id != self)
):
    continue
```

product_pricelist_supplierinfo/models/product_product.py
```diff
@@ -24,3 +24,35 @@ def price_compute(self, price_type, uom=False, currency=False,
...
+        company_id = self.env.context.get('force_company')
+        if company_id:
+            sellers = sellers.filtered(
+                lambda s: not s.company_id or s.company_id.id == company_id)
+        for seller in sellers:
+            if seller.date_start and seller.date_start > date:
+                continue
+            if seller.date_end and seller.date_end < date:
+                continue
+            if partner_id and seller.name not in [partner_id, partner_id.parent_id]:
```
luisg123v: Consider using `commercial_partner_id` instead of `parent_id`

product_pricelist_supplierinfo/models/product_product.py
```diff
@@ -24,3 +24,35 @@ def price_compute(self, price_type, uom=False, currency=False,
...
+                continue
+            if seller.product_id and seller.product_id != self:
+                continue
+
+            res |= seller
+            break
```
luisg123v: If ifs are joined, you may do this inside the if

product_pricelist_supplierinfo/models/product_template.py
```diff
@@ -18,17 +18,23 @@ def _get_supplierinfo_pricelist_price(
...
                 date = date.date()
-            seller = product._select_seller(
-                partner_id=rule.filter_supplier_id,
-                quantity=quantity,
-                date=date)
+            # If the field `no_supplierinfo_min_quantity` is set as True, we need
+            # to use the select seller that doesn't take into account the `min_qty`
+            # for the validation to know if the seller can be used.
+            if rule.no_supplierinfo_min_quantity:
+                seller = product._select_seller_without_min_qty(
```
luisg123v: Why don't you inherit the original method and add that extra behavior?

product_pricelist_supplierinfo/tests/test_product_supplierinfo.py
```diff
@@ -107,6 +107,58 @@ def test_pricelist_min_quantity(self):
             self.pricelist.get_product_price(self.product, 10, False), 50,
         )
         self.pricelist.item_ids[0].no_supplierinfo_min_quantity = True
+        # As the sequence of both of the supplieinfo is the same the next field to
```
luisg123v:
```suggestion
        # As the sequence of both supplierinfos is the same, the next field to
```

product_pricelist_supplierinfo/tests/test_product_supplierinfo.py
```diff
@@ -107,6 +107,58 @@ def test_pricelist_min_quantity(self):
...
+        # order is the min_qty (desc), so the price must be the one with the higher
+        # min_qty no mather if the qty is lower that the min_qty.
```
luisg123v:
```suggestion
              # min_qty no matter if the qty is lower than the min_qty.
```

product_pricelist_supplierinfo/tests/test_product_supplierinfo.py
```diff
@@ -107,6 +107,58 @@ def test_pricelist_min_quantity(self):
...
+        self.assertAlmostEqual(
+            self.pricelist.get_product_price(self.product, 1, False), 50,
+        )
+        self.assertAlmostEqual(
+            self.pricelist.get_product_price(self.product, 5, False), 50,
+        )
+
+    def test_pricelist_no_min_quantity(self):
+        """Testing the validations of the `_select_seller_without_min_qty` method.
+        As the sequence of both of the supplieinfo in the product is the same the
```
luisg123v:
```suggestion
        As the sequence of both supplierinfos in the product is the same, the
```

Also leave a blank line between title and description (it's the [docstring convention](https://www.python.org/dev/peps/pep-0257/#multi-line-docstrings)).

product_pricelist_supplierinfo/tests/test_product_supplierinfo.py
```diff
@@ -107,6 +107,58 @@ def test_pricelist_min_quantity(self):
...
+        - The start date is tomorrow (not current).
+        - The end date is yesterday (not current).
+        - The supplierinfo is current, but there is a 'Vendor' set in the field
+        `filter_supplier_id` of the item in the pricelist.
+        """
+        product = self.product
+        tomorrow = date.today() + timedelta(days=1)
+        product.seller_ids[0].date_start = tomorrow
+        self.pricelist.item_ids[0].no_supplierinfo_min_quantity = True
+        # Case 1: The start date is not a current one so it will takes the next one.
```
luisg123v:
```suggestion
        # Case 1: The start date is not a current one so it will take the next one.
```

or even better, "will be taken"

product_pricelist_supplierinfo/tests/test_product_supplierinfo.py
```diff
@@ -107,6 +107,58 @@ def test_pricelist_min_quantity(self):
...
+            self.pricelist.get_product_price(self.product, 1, False), 10,
+        )
+        self.assertAlmostEqual(
+            self.pricelist.get_product_price(self.product, 5, False), 10,
+        )
+
+        product.seller_ids[0].date_start = False
+        yesterday = date.today() - timedelta(days=1)
+        product.seller_ids[0].date_end = yesterday
+        # Case 2: The end date is an expired one so it will takes the next one.
```
luisg123v: - Same here
- Shouldn't this comment go just above `product.seller_ids[0].date_start = False`?

luisg123v (changes requested):

product_pricelist_supplierinfo/models/product_template.py
CarmenMiranda: Because the original has the for cycle inside and it will return me the supplier info that is available, I can't add an extra validation, but I'm open to suggestions

product_pricelist_supplierinfo/models/product_product.py
CarmenMiranda: As I'm using the original code from the `_select_seller` should I change it?

product_pricelist_supplierinfo/models/product_product.py
CarmenMiranda: So, should I reverse the if to do this inside?

product_pricelist_supplierinfo/models/product_template.py
luisg123v: I mean, you may pass an extra parameter. If it's not set, fallback to the default behavior. If set, perform your custom stuf

product_pricelist_supplierinfo/models/product_product.py
luisg123v: I just checked the original code. Even though you would be changing it, I think it's kind of clear you would be doing the same thing.

product_pricelist_supplierinfo/models/product_product.py
luisg123v: Yes. I wanted to write that, but I didn't find the correct word (invert/reverse/inverse)

product_pricelist_supplierinfo/models/product_product.py
luisg123v: Sorry, I replied to the wrong comment. I think the original one should use the commercial partner. But, since it does not let's stick with that way.

product_pricelist_supplierinfo/models/product_product.py
CarmenMiranda: Done

product_pricelist_supplierinfo/models/product_product.py
CarmenMiranda: Done

product_pricelist_supplierinfo/models/product_product.py
CarmenMiranda: Ok, done

product_pricelist_supplierinfo/models/product_product.py
CarmenMiranda: Haha, oki, it's done

product_pricelist_supplierinfo/models/product_template.py
CarmenMiranda: Ohh, oki, now is done

product_pricelist_supplierinfo/tests/test_product_supplierinfo.py
CarmenMiranda: Done

product_pricelist_supplierinfo/tests/test_product_supplierinfo.py
CarmenMiranda: Done

product_pricelist_supplierinfo/tests/test_product_supplierinfo.py
CarmenMiranda: Done

product_pricelist_supplierinfo/tests/test_product_supplierinfo.py
CarmenMiranda: Done

product_pricelist_supplierinfo/tests/test_product_supplierinfo.py
CarmenMiranda: Done

product_pricelist_supplierinfo/models/product_product.py
```diff
@@ -24,3 +24,33 @@ def price_compute(self, price_type, uom=False, currency=False,
             return dict.fromkeys(self.ids, 1.0)
         return super().price_compute(
             price_type, uom=uom, currency=currency, company=company)
+
+    def _select_seller(self, partner_id=False, quantity=0.0, date=None, uom_id=False,
```
luisg123v: Params? Better using the actual param name (`avoid_min_qty`), this is too generic

product_pricelist_supplierinfo/tests/test_product_supplierinfo.py
```diff
@@ -106,7 +106,60 @@ def test_pricelist_min_quantity(self):
         self.assertAlmostEqual(
             self.pricelist.get_product_price(self.product, 10, False), 50,
         )
+        # As the sequence of both supplieinfos is the same, the next field to order
```
luisg123v: This also happens below

```suggestion
        # As the sequence of both supplierinfos is the same, the next field to order
```

product_pricelist_supplierinfo/tests/test_product_supplierinfo.py
```diff
@@ -106,7 +106,60 @@ def test_pricelist_min_quantity(self):
...
+        yesterday = date.today() - timedelta(days=1)
+        product.seller_ids[0].date_end = yesterday
+        self.assertAlmostEqual(
+            self.pricelist.get_product_price(self.product, 1, False), 10,
+        )
+        self.assertAlmostEqual(
+            self.pricelist.get_product_price(self.product, 5, False), 10,
+        )
+
+        # Case 3: Current seller but a `filter_supplier_id` is set in the iem, so it
```
luisg123v: "iem"?

product_pricelist_supplierinfo/models/product_product.py
luisg123v: break after parenthesis

luisg123v (changes requested):

product_pricelist_supplierinfo/models/product_product.py
```diff
@@ -24,3 +24,33 @@ def price_compute(self, price_type, uom=False, currency=False,
...
+                       params=False):
```
pedrobaeza: You can't change the arguments of a method of upstream code. This will provoke countless compatibility problems. Use context for propagating values between methods.

product_pricelist_supplierinfo/models/product_product.py
CarmenMiranda: I'm using the [original parameters](https://github.com/odoo/odoo/blob/f23e07dcb0606dbc4bf835b8266323d5a515662d/addons/product/models/product.py#L566), and `params` is one of them, I'm only using it as is used on the [`purchase` module](https://github.com/odoo/odoo/blob/f23e07dcb0606dbc4bf835b8266323d5a515662d/addons/purchase/models/purchase.py#L632)

product_pricelist_supplierinfo/models/product_product.py
CarmenMiranda: Hi, I'm using the [original parameters](https://github.com/odoo/odoo/blob/f23e07dcb0606dbc4bf835b8266323d5a515662d/addons/product/models/product.py#L566), and `params` is one of them, I'm only using it as is used on the [`purchase` module](https://github.com/odoo/odoo/blob/f23e07dcb0606dbc4bf835b8266323d5a515662d/addons/purchase/models/purchase.py#L632) 
I can change it to a context if is better for you

product_pricelist_supplierinfo/models/product_product.py
luisg123v: Oh, I see. Kind of weird that original method, but it's OK for me then.

product_pricelist_supplierinfo/tests/test_product_supplierinfo.py
CarmenMiranda: Done

product_pricelist_supplierinfo/tests/test_product_supplierinfo.py
CarmenMiranda: It was "item" 🙈

product_pricelist_supplierinfo/models/product_product.py
CarmenMiranda: Done

pedrobaeza:
I think I'm losing something, as the behavior of `no_supplierinfo_min_quantity` is just for ignoring the min quantity field, and so to return the first supplierinfo record found, even if the quantity doesn't match. The sequence will decide which applies. Maybe you can describe a full example to check the problem.

CarmenMiranda:
> I think I'm losing something, as the behavior of `no_supplierinfo_min_quantity` is just for ignoring the min quantity field, and so to return the first supplierinfo record found, even if the quantity doesn't match. The sequence will decide which applies. Maybe you can describe a full example to check the problem.

Hi @pedrobaeza Sorry for the delay, but here is a video with an example:
https://youtu.be/ZOkVYiLivb8

The case is when we have a product only with a supplierinfo with a min_qty > 1 and in the pricelist item we have selected the field `no_supplierinfo_min_quantity` on True, the price of the supplierinfo will not appear in the sale, it only appears when the `min_qty` is 0 or 1, and is avoiding the other quantities

Also, another case is the one in the test `test_pricelist_min_quantity`, in there the product has 2 supplierinfos, one with the vendor "cls.supplier1", the `min_qty` with 5 and the `price` is 50, and another one with the vendor "cls.supplier2", the `min_qty` with 1 and the `price` of 10.

As the order in the model `product.supplierinfo` is 'sequence, min_qty desc, price', and both supplierinfos have the same `sequence`,  the price that should be taken is 50 and not 10, because as you said the method should return the first one in line.
But right now, without this change, it was only sending 1 to the method _select_seller` and that only allows us to use the first supplierinfo with the qty of 1 or 0 (because [this validation](https://github.com/odoo/odoo/blob/5372bedacf42edd423834609b447c8c750c672ca/addons/product/models/product.py#L588) is not being avoided ) so that is why the price that was obtained was 10 instead of 50.

The runbot where I did the example is [this one](http://3492685-12-0-a7b66e.runbot2-2.odoo-community.org/?).

Cc. @luisg123v @JulioSerna

PR #883, opened 2021-07-14
[13.0][ADD] product_list_pricelist_price

JordiMForgeFlow:
This module adds two menu items to easily display the sales price of product
templates and product variants for the different pricelists in the system.

Each menu corresponds to one of the two models, and the prices for each
pricelists are added as optional columns to be displayed in the list view.

See the Usage section to know how to configure the Pricelists you want to show
in the views.


keshrath:
What happens if a pricelist is removed?

JordiMForgeFlow:
> What happens if a pricelist is removed?

On every pricelist creation or write, the registry is invalidated so the register hook is executed again, computing dynamically again the fields for each active pricelist: https://github.com/OCA/product-attribute/pull/883/files#diff-c99de111aa989d8d9251b74463362185c6523456adaf10e056cd40e5e5ddaabfR12

product_list_pricelist_price/models/product_template.py
```diff
@@ -0,0 +1,87 @@
...
+    def _compute_product_template_pricelist_price(self):
+        pricelists = self.env["product.pricelist"].search([("active", "=", True)])
+        for product in self:
+            result = pricelists._compute_price_rule_multi([(product, 1.0, False)])[
+                product.id
+            ]
+            for pricelist in pricelists:
+                field_name = "product_tmpl_price_pricelist_%s" % (pricelist.id)
+                product[field_name] = pricelist.currency_id.round(
+                    result[pricelist.id][0]
```
JordiBForgeFlow: @JordiMForgeFlow perhaps you should consider the case where pricelist.id is not in result, or that result[pricelist.id] does not contain an element? Just to avoid exceptions raised

JordiMForgeFlow: Based on the standard code it should always contain the result, as by default it will at least set a price of 0.0 for each product (https://github.com/odoo/odoo/blob/13.0/addons/product/models/product_pricelist.py#L170) and for each pricelist passed to the method _compute_price_rule_multi (https://github.com/odoo/odoo/blob/13.0/addons/product/models/product_pricelist.py#L91).

Nevertheless it doesn't hurt to add extra security, so I can add it 😄

legalsylvain:
H Jordi. FYI, I proposed a module that provides complentary features to see prices and related margins for a product : here 
https://github.com/OCA/margin-analysis/pull/96

Maybe you could be interested...

product_list_pricelist_price/models/product_template.py
JordiMForgeFlow: Done!

JordiBForgeFlow (approved):
:+1: functional and technical review. Works like a charm!

JordiMForgeFlow:
Two main changes have been added in the new commits:

1. The prices are displayed in two separate menus (one for templates, one for variants), because loading times can get larger when the number of products is high. Thus, we prevent interfering in the standard menus.
2. A field has been added to mark the pricelists that want to be displayed in the views.

AdriaGForgeFlow (approved):
Functional and code review :+1:

PR #885, opened 2021-07-15
[14.0][FIX] product_manufacturer: access form view from tree view

andreampiovesana:
Product manufacturer not visible if open from tree view of product variant


marcelofrare (approved):
LGFM

OmniaGit (approved):

PR #886, opened 2021-07-15
[14.0] [MIG] product_lot_sequence

remi-filament:


LoisRForgeFlow (approved):
Functional review :+1:

LoisRForgeFlow:
@remi-filament The migration commit seems to be blended with "back, isort, prettier" one. It would be good if you can fix that in the commit history.

remi-filament:
Thanks for your review @LoisRForgeFlow 
Actually, except for changes introduced by pre-commit ans version in manifest, I did not have to change anything from v13 to get it working.
Then I made an improvement to update name by défaut with product lot sequence instead of generic lot sequence.
I could rename any of these 2 commits as migration one, which one do you prefer ?

LoisRForgeFlow:
@remi-filament That's fine with me. Perhaps the only thing I would ask you is to reword your second commit to `"[IMP] product_lot_sequence: Change name on create"`. This way it will include a valid tag and the module name which is always desirable an more aligned with the [OCA guidelines](https://github.com/OCA/odoo-community.org/blob/master/website/Contribution/CONTRIBUTING.rst#71commit-message).

remi-filament:
@LoisRForgeFlow I just rewrote history, let me know if OK for you ?
Best Regards

LoisRForgeFlow:
@remi-filament Thanks! LGTM

rousseldenis (approved):
LGTM

LoisRForgeFlow:
@rousseldenis I think you can merge this if there is nothing else to change.

PR #889, opened 2021-07-22
[13.0][ADD] product_dimension_volumetric_weight: Add Product Uom Volumetric Weight

victoralmau:
This module extends the functionality of product_dimension to add Volumetric weight in products.

Please @pedrobaeza and @chienandalu can you review it?

@Tecnativa TT30722


pedrobaeza:
If this finally depends on `product_dimension`, shouldn't it be `product_dimension_volumetric_weight`?

chienandalu (commented):
¿No support for variants?

hparfr:
Hi,

May be a dumb question but, why there is no weight in the formula of volumetric_weight ? 

About the UI, the unit of "volumetric_weight" is not explicit. Shoulnd't it be "dimensional_uom / Kg" ?

victoralmau:
Changes done.
IMO volumetric weight is need to be added in extra addon because `product_dimension` is only related to dimension and field name is correct.

chienandalu (approved):

product_dimension_volumetric_weight/models/product.py
```diff
@@ -0,0 +1,40 @@
...
+
+    weight_volumetric = fields.Float(
+        string="Volumetric weight",
+        compute="_compute_weight_volumetric",
+        store=True,
+        readonly=True,
+    )
+
+    @api.depends("dimensional_uom_id", "volume")
+    def _compute_weight_volumetric(self):
```
pedrobaeza: No need to repeat this method if it's the same as upstream `product.template` one.

product_dimension_volumetric_weight/readme/DESCRIPTION
```diff
@@ -0,0 +1 @@
+This module extends the functionality of product_dimension to add Volumetric weight in products.
```
pedrobaeza:
```suggestion
This module extends the functionality of product_dimension to add Volumetric weight in products.

It's used for example by some carrier agencies to compute the cost price of the delivery service.
```

product_dimension_volumetric_weight/tests/test_product_uom.py
```diff
@@ -0,0 +1,32 @@
...
+        self.uom_m.write({"weight_volumetric_ratio": 400})
+        self.uom_cm = self.env.ref("uom.product_uom_cm")
+        self.uom_cm.write({"weight_volumetric_ratio": 400})
+
+    def test_volumetric_weight_in_m(self):
+        product_form = Form(self.env["product.template"])
+        product_form.name = "Test product"
+        product_form.product_length = 1
+        product_form.product_height = 1
+        product_form.product_width = 1
```
pedrobaeza: This is testing a functionality of other module, so no need to include it here.

product_dimension_volumetric_weight/tests/test_product_uom.py
```diff
@@ -0,0 +1,32 @@
...
+        product_form.dimensional_uom_id = self.uom_m
+        self.assertAlmostEqual(400, product_form.weight_volumetric)
+
+    def test_volumetric_weight_in_cm(self):
+        product_form = Form(self.env["product.template"])
+        product_form.name = "Test product"
+        product_form.product_length = 40
+        product_form.product_height = 50
+        product_form.product_width = 60
```
pedrobaeza: Again testing something that is not of this module. Why don't you set `volume` and just that?

product_dimension_volumetric_weight/models/product.py
```diff
@@ -0,0 +1,41 @@
...
+        for item in self:
+            item.weight_volumetric = (
+                item.volume * item.dimensional_uom_id.weight_volumetric_ratio
+            )
+
+
+class ProductProduct(models.Model):
```
pedrobaeza: Split both models in their corresponding files.

product_dimension_volumetric_weight/models/product.py
```diff
@@ -0,0 +1,41 @@
...
+
+class ProductTemplate(models.Model):
+    _inherit = "product.template"
+
+    weight_volumetric = fields.Float(
+        string="Volumetric weight",
+        compute="_compute_weight_volumetric",
+        store=True,
+        readonly=True,
+        digits="Volume",
```
pedrobaeza: This is weight, not volume.

product_dimension_volumetric_weight/models/product.py
```diff
@@ -0,0 +1,41 @@
...
+
+
+class ProductTemplate(models.Model):
+    _inherit = "product.template"
+
+    weight_volumetric = fields.Float(
+        string="Volumetric weight",
+        compute="_compute_weight_volumetric",
+        store=True,
+        readonly=True,
```
pedrobaeza: A compute is always readonly

pedrobaeza (changes requested):
Add also configure file (and generate README), stating how to set the ratio at UoM.

pedrobaeza:
Another thing: why fields are called `weight_volumetric`, but the module is `volumetric_weight`. The only reasoning I see is for having together `weight` and `weight_volumetric`, but not sure if that reason is important enough for having this duality in the name. What others think about this?

victoralmau:
> Another thing: why fields are called `weight_volumetric`, but the module is `volumetric_weight`. The only reasoning I see is for having together `weight` and `weight_volumetric`, but not sure if that reason is important enough for having this duality in the name. What others think about this?

Yes, probably `volumetric_weight` is the better option to field names.

rousseldenis (approved):
Code review

pedrobaeza (approved):

PR #890, opened 2021-07-22
[13.0][IMP] product_supplierinfo_for_customer: introduce _select_customerinfo

LoisRForgeFlow:
This method mimics [standard "brother" `_select_seller`](https://github.com/odoo/odoo/blob/14.0/addons/product/models/product.py#L602) to make the
module more consistent and to introduce a standard way to get the
customerinfo in modules depending on this one.

@ForgeFlow

### DIFF
--- a/product_supplierinfo_for_customer/models/product_product.py
+++ b/product_supplierinfo_for_customer/models/product_product.py
@@ -1,10 +1,12 @@
 # Copyright 2015 OdooMRP team
 # Copyright 2015 AvanzOSC
 # Copyright 2015 Tecnativa
+# Copyright 2021 ForgeFlow S.L. (https://www.forgeflow.com)
 # License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl.html).
 import datetime
 
-from odoo import api, models
+from odoo import api, fields, models
+from odoo.tools import float_compare
 
 
 class ProductProduct(models.Model):
@@ -64,18 +66,7 @@ class ProductProduct(models.Model):
         self.ensure_one()
         if not partner_id:
             return 0.0
-        customerinfo = self.env["product.customerinfo"].search(
-            [
-                ("name", "=", partner_id),
-                "|",
-                ("product_id", "=", self.id),
-                "&",
-                ("product_tmpl_id", "=", self.product_tmpl_id.id),
-                ("product_id", "=", False),
-            ],
-            limit=1,
-            order="product_id, sequence",
-        )
+        customerinfo = self._select_customerinfo(partner_id=partner_id)
         if customerinfo:
             return customerinfo.price
         return 0.0
@@ -85,8 +76,8 @@ class ProductProduct(models.Model):
             partner_id = self.env.context.get(
                 "partner_id", False
             ) or self.env.context.get("partner", False)
-            if partner_id and isinstance(partner_id, models.BaseModel):
-                partner_id = partner_id.id
+            if partner_id and isinstance(partner_id, int):
+                partner_id = self.env["res.partner"].browse(partner_id)
             prices = super(ProductProduct, self).price_compute(
                 "list_price", uom, currency, company
             )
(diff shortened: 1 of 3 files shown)


LoisRForgeFlow:
cc @pedrobaeza

product_supplierinfo_for_customer/models/product_product.py
```diff
@@ -85,8 +76,8 @@ def price_compute(self, price_type, uom=False, currency=False, company=False):
             partner_id = self.env.context.get(
                 "partner_id", False
             ) or self.env.context.get("partner", False)
-            if partner_id and isinstance(partner_id, models.BaseModel):
-                partner_id = partner_id.id
+            if partner_id and isinstance(partner_id, int):
+                partner_id = self.env["res.partner"].browse(partner_id)
```
pedrobaeza, line 80: Why changing that?

product_supplierinfo_for_customer/models/product_product.py
```diff
@@ -114,3 +105,66 @@ def price_compute(self, price_type, uom=False, currency=False, company=False):
...
+            if customer.product_id and customer.product_id != self:
+                continue
+            if not res or res.name == customer.name:
+                res |= customer
+        # NOTE: This differs from the standard sellers logic. This is done to
+        # keep the old behavior of this module.
+        variant_specific = res.filtered(lambda r: r.product_id)
+        if variant_specific:
+            res = variant_specific
+        return res.sorted("price")[:1]
```
pedrobaeza, line 170: You are not preparing a hook, but changing more things here, and that should be properly documented and tested. I think we should start with the basic one first.

product_supplierinfo_for_customer/models/product_product.py
LoisRForgeFlow: `_select_customerinfo` like standard `_select_seller` requires a res.partner as argument (later in the funcion it does partner.parent_id), therefore we need to pass the record, not the id

product_supplierinfo_for_customer/models/product_product.py
LoisRForgeFlow: I did this to keep the previous behavior in with a variant specific customerinfo has preference over a non-specific one. This is used to get the price. What do you suggest then? that I isolate this "patch" into a hook?

product_supplierinfo_for_customer/models/product_product.py
pedrobaeza: Then don't call the variable `partner_id`, but `partner`.

product_supplierinfo_for_customer/models/product_product.py
pedrobaeza: No, I see the diff, and you are replacing a simple `search` by a lot of code, and that's why I don't get.

LoisRForgeFlow: @pedrobaeza Something like this?
```python
...
        return self._select_customerinfo_hook(res)

    @api.model
    def _select_customerinfo_hook(self, customers):
        # NOTE: This differs from the standard sellers logic. This is done to
        # keep the old behavior of this module.
        variant_specific = customers.filtered(lambda r: r.product_id)
        if variant_specific:
            customers = variant_specific
        return customers.sorted("price")[:1]

```

pedrobaeza: Selección_010

LoisRForgeFlow: ah, I see, maybe I should have put a link to the standard "brother": https://github.com/odoo/odoo/blob/14.0/addons/product/models/product.py#L602.

I'm doing the same Odoo does with supplieinfos but for customers, which makes more sense IMO.

pedrobaeza: And that's what I say that should go into another PR to be discussed, as this can have side effects, changing the selector for the proper customer info on existing installations.

LoisRForgeFlow: why not discuss it in this one? I pinged you because that was also my concern, if you have installations with this module it can affect you. But I tried to keep the behavior the same, do you see any change that can affect your installations?

pedrobaeza: Is there any reason for you to change the current behavior?

Anyways, if changed in this one, some tests for each edge case should be introduced, and compare before and after to be able to see.

About our current customers using this module, @sergio-teruel and @carlosdauden can say better. @chienandalu do you TCR cases?

LoisRForgeFlow: I think adding a method to get customerinfo is almost mandatory, I was reviewing child modules of this one and each of them searches the customerinfo in a different way... 

Once I was doing it, I thought it was good to it good, and for me the best way is to mimic Odoo; I think is an UX improvement if Odoo searches customerinfo and supplierinfo the same way. But this is optional, if you see issues with this I will prepare an alternative.

LoisRForgeFlow: Alternative: https://github.com/OCA/product-attribute/pull/891.

LoisRForgeFlow:
This is not going to be done in v13 (hopefully in v14), in the meanwhile it is substituted by https://github.com/OCA/product-attribute/pull/891.

PR #891, opened 2021-07-22
[13.0][IMP] product_supplierinfo_for_customer: add `_select_customerinfo

LoisRForgeFlow:
The idea of this method is to centralize the searches of customerinfo
to have more consistent usage in child modules.

@ForgeFlow

### DIFF
--- a/product_supplierinfo_for_customer/models/product_product.py
+++ b/product_supplierinfo_for_customer/models/product_product.py
@@ -64,18 +64,8 @@ class ProductProduct(models.Model):
         self.ensure_one()
         if not partner_id:
             return 0.0
-        customerinfo = self.env["product.customerinfo"].search(
-            [
-                ("name", "=", partner_id),
-                "|",
-                ("product_id", "=", self.id),
-                "&",
-                ("product_tmpl_id", "=", self.product_tmpl_id.id),
-                ("product_id", "=", False),
-            ],
-            limit=1,
-            order="product_id, sequence",
-        )
+        partner = self.env["res.partner"].browse(partner_id)
+        customerinfo = self._select_customerinfo(partner=partner)
         if customerinfo:
             return customerinfo.price
         return 0.0
@@ -114,3 +104,31 @@ class ProductProduct(models.Model):
         return super(ProductProduct, self).price_compute(
             price_type, uom, currency, company
         )
+
+    def _prepare_domain_customerinfo(self, params):
+        self.ensure_one()
+        partner_id = params.get("partner_id")
+        return [
+            ("name", "=", partner_id),
+            "|",
+            ("product_id", "=", self.id),
+            "&",
+            ("product_tmpl_id", "=", self.product_tmpl_id.id),
+            ("product_id", "=", False),
+        ]
+
+    def _select_customerinfo(
+        self, partner=False, _quantity=0.0, _date=None, _uom_id=False, params=False
+    ):
+        """Customer version of the standard `_select_seller`. """
+        # TODO: For now it is just the function name with same arguments, but
+        #  can be changed in future migrations to be more in line Odoo
+        #  standard way to select supplierinfo's.
+        if not params:
+            params = dict()
+        params.update({"partner_id": partner.id})
+        domain = self._prepare_domain_customerinfo(params)
+        res = self.env["product.customerinfo"].search(
+            domain, order="product_id, sequence", limit=1
+        )
+        return res


LoisRForgeFlow:
cc @pedrobaeza

product_supplierinfo_for_customer/models/product_product.py
```diff
@@ -114,3 +103,30 @@ def price_compute(self, price_type, uom=False, currency=False, company=False):
...
+    ):
+        """Customer version of the standard `_select_seller`. """
+        # TODO: For now it is just the function name with same arguments, but
+        #  can be changed in future migrations to be more in line Odoo
+        #  standard way to select supplierinfo's.
+        if not params:
+            params = dict()
+        params.update({"partner_id": partner.id})
+        res = self._prepare_customerinfo(params)
+        return res[:1]
```
chienandalu: Maybe a limit=1 in the domain would suffice?

chienandalu (commented):
Looks less risky indeed :sweat_smile:

product_supplierinfo_for_customer/models/product_product.py
```diff
@@ -60,38 +60,27 @@ def _name_search(
         res.extend(self.browse(product_ids).name_get())
         return res
 
-    def _get_price_from_customerinfo(self, partner_id):
+    def _get_price_from_customerinfo(self, partner):
```
pedrobaeza: It's still risky, as you are changing the signature of one method, and thus other extensions will fail in this. Please keep the same signature as previously. You can change the signature on a version migration.

product_supplierinfo_for_customer/models/product_product.py
```diff
@@ -114,3 +103,30 @@ def price_compute(self, price_type, uom=False, currency=False, company=False):
         return super(ProductProduct, self).price_compute(
             price_type, uom, currency, company
         )
+
+    def _prepare_customerinfo(self, params):
```
pedrobaeza: Why this double method? If you want to inject things on the domain for performing only one search, then put a `_prepare_domain` hook, but this is not adding anything useful to the inheritance (nothing that you can't done on the main method).

product_supplierinfo_for_customer/models/product_product.py
LoisRForgeFlow: I want to keep the structure like [_sellect_sellers](https://github.com/odoo/odoo/blob/14.0/addons/product/models/product.py#L602) as much as possible

product_supplierinfo_for_customer/models/product_product.py
LoisRForgeFlow: Done

product_supplierinfo_for_customer/models/product_product.py
LoisRForgeFlow: Done.

LoisRForgeFlow:
@pedrobaeza does it look good now?

product_supplierinfo_for_customer/models/product_product.py
pedrobaeza: This is not adding anything useful, as the return result is going to be the same 1 record, but with less performance and readability. Not always the upstream code is the one to imitate...

product_supplierinfo_for_customer/models/product_product.py
```diff
@@ -64,34 +64,23 @@ def _get_price_from_customerinfo(self, partner_id):
...
-                "|",
-                ("product_id", "=", self.id),
-                "&",
-                ("product_tmpl_id", "=", self.product_tmpl_id.id),
-                ("product_id", "=", False),
-            ],
-            limit=1,
-            order="product_id, sequence",
-        )
+        if partner_id and isinstance(partner_id, int):
```
pedrobaeza: `partner_id` shouldn't be polymorphic and only receives an integer IMO.

LoisRForgeFlow: can you make this concession at least? this won't affect in any way any installation...

pedrobaeza: Yes, it does. Any subsequent inherited method will need to check the same each time.

product_supplierinfo_for_customer/models/product_product.py
LoisRForgeFlow: Ok, done.

product_supplierinfo_for_customer/models/product_product.py
pedrobaeza: That's why type hinting was added in Python 3.x, for avoiding these ugly hacks.

LoisRForgeFlow: Ok, what should I put then?

LoisRForgeFlow:
```
...
partner = self.env["res.partner"].browse(partner_id)
customerinfo = self._select_customerinfo(partner=partner)
```

will this be ok?

LoisRForgeFlow: pushed, hope it is good now

LoisRForgeFlow:
@chienandalu could you check again?

product_supplierinfo_for_customer/models/product_product.py
pedrobaeza: As the argument is the integer, always pass the ID, and do the browse if you need the recordset. On a migration to next non migrated version, the argument may be changed to `partner`.

LoisRForgeFlow: it is done, could you check?

product_supplierinfo_for_customer/models/product_product.py
```diff
@@ -64,34 +64,22 @@ def _get_price_from_customerinfo(self, partner_id):
...
         if customerinfo:
             return customerinfo.price
         return 0.0
 
     def price_compute(self, price_type, uom=False, currency=False, company=False):
         if price_type == "partner":
-            partner_id = self.env.context.get(
-                "partner_id", False
-            ) or self.env.context.get("partner", False)
-            if partner_id and isinstance(partner_id, models.BaseModel):
```
pedrobaeza: This polymorphism seems to be still needed (what I was talking about is to not create new ones).

product_supplierinfo_for_customer/models/product_product.py
```diff
@@ -64,34 +64,22 @@ def _get_price_from_customerinfo(self, partner_id):
...
-                partner_id = partner_id.id
+            partner = self.env.context.get("partner_id", False) or self.env.context.get(
+                "partner", False
+            )
```
MiquelRForgeFlow, line 77: you shouldn't touch this. `_get_price_from_customerinfo` receives int, not a record

MiquelRForgeFlow: Oh, pedro was faster haha

LoisRForgeFlow:
@pedrobaeza Can you check again?

product_supplierinfo_for_customer/models/product_product.py
```diff
@@ -114,3 +104,30 @@ def price_compute(self, price_type, uom=False, currency=False, company=False):
         return super(ProductProduct, self).price_compute(
             price_type, uom, currency, company
         )
+
+    def _prepare_domain_customerinfo(self, params):
+        partner_id = params.get("partner_id")
```
pedrobaeza, line 110:
```suggestion
        self.ensure_one()
        partner_id = params.get("partner_id")
```

pedrobaeza (approved):

product_supplierinfo_for_customer/models/product_product.py
LoisRForgeFlow: added.

MiquelRForgeFlow (approved):

LoisRForgeFlow:
@pedrobaeza Good to merge?

chienandalu (approved):

pedrobaeza:
/ocabot merge patch

I think we are in time for changing params in v14, as it has been just merged.

PR #892, opened 2021-07-22
14.0 imp product packaging type display code and qty in name

estekadigital:
Added two options in product packaging type to hide or display the code and/or the quantity (when possible) in the product packaging name

- Display code in name: e.g. BOX (BX)
- Display quantity in name: e.g. BOX (2.0KG) (BX)

### DIFF
--- a/product_packaging_type/models/product_packaging_type.py
+++ b/product_packaging_type/models/product_packaging_type.py
@@ -13,6 +13,7 @@ class ProductPackagingType(models.Model):
 
     name = fields.Char(required=True, translate=True)
     code = fields.Char(required=True)
+    display_code_in_name = fields.Boolean(default=True)
     sequence = fields.Integer(required=True)
     has_gtin = fields.Boolean()
     active = fields.Boolean(default=True)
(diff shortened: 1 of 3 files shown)


product_packaging_type/models/product_packaging_type.py
```diff
@@ -138,7 +143,15 @@ def name_get(self):
...
-                result.append((record.id, record.packaging_type_id.display_name))
+                if record.packaging_type_id.display_qty_in_name:
+                    result.append((record.id, "{} ({})".format(
+                        record.packaging_type_id.display_name,
+                        str(record.product_uom_id._compute_quantity(
+                            record.qty,
+                            record.product_uom_id)) + record.product_uom_id.name
+                    )))
+                else:
+                    result.append((record.id, record.packaging_type_id.display_name))
```
ajaniszewska-dev, line 154: Maybe more elegant way would be to modify the value you want to use inside append()  within if-else block (in both cases you use record.packaging_type_id.display_name alone or with qty data) and do append only once with final value at the end.

ajaniszewska-dev (approved):

PR #894, opened 2021-08-03
 [ADD][12.0] new module product_net_weight

legalsylvain:
~This module extends the functionality of stock module to support gross weight.~
~Updated : See conversation below~

This module extends the functionality of stock module to support net weight. (container excluded)

It also compute the Weight Price (Sale Price / Net Weight) that is a mandatory information
in certain businesses.

* Go to 'Inventory > Master Data > Product' and edit items.

* Go to 'Inventory' tab, and fill the "Net Weight" value.

product_form

* If the net weight defined, the weight price is computed and displayed on the main tab

product_form_price

**Note**: The original weight field is renamed to "Gross Weight", to be more explicit.


dreispt:
You probably mean "Gross Weight".
I could be wrong, but AFAIU the out-of-the-box "Weight" is the gross weight.
Because when computing shipping weight, the "Weight" field is used.

So what we might need to add is the "Net Weight" instead.

legalsylvain:
Thanks for your insight, @dreispt.

> You probably mean "Gross Weight".

Yes ! sorry. I'll fix that.

>  I could be wrong, but AFAIU the out-of-the-box "Weight" is the gross weight.

No the ``weight`` field is the net weight, AFAIRTD. (As Far As I read the doc).

https://github.com/odoo/odoo/blob/12.0/addons/product/models/product.py#L148 See : ``packaging not included``

>  Because when computing shipping weight, the "Weight" field is used.

in odoo or in oca modules ? (i don't use shipping weight)


In V8, there was two fields ``weight`` and ``weight_net``, but : 
- the second field disappeared
- the helper of the first field changed (``The gross weight in Kg.`` --> ``Weight of the product, packaging not included.``)
https://github.com/odoo/odoo/blob/8.0/addons/product/product.py#L545

In Openupgrade, we consider that the new ``weight`` field is the old ``weight_net`` field : 

https://github.com/OCA/OpenUpgrade/blob/9.0/addons/product/migrations/9.0.1.2/pre-migration.py#L129

help welcome on that point !

legalsylvain:
Hi @dreispt : I replaced "Brut Weight" per "Gross Weight".

legalsylvain:
Hi @amoyaux. Could you give your point of view, regarding the ``weight`` field. 
- Is it a net or a gross weight ? 
- Why odoo removed the ``weight_net`` field between 8.0 and 9.0 version ? 

(I ping you because you're set as responsible of Logistics topic here (https://github.com/odoo/odoo/wiki/BugManagement) feel free to ping another guy if I'm wrong)

Thanks.

amoyaux:
Gross weight (at least from delivery point of view)
It's probably because it's not really used by the majority of use and a lot of module (like this one) exist for people with the need.

legalsylvain:
> Gross weight (at least from delivery point of view)
>  It's probably because it's not really used by the majority of use and a lot of module (like this one) exist for people with the need.

Well, if you say that the odoo ``weight`` field is the gross weight : 
-> I'll rename this module to ``product_net_weight``. @dreispt : are you ok with that proposal ?
-> I think odoo SA should add an ``help`` on this field, mentionning it is a gross weight. (https://github.com/odoo/odoo/blob/master/addons/product/models/product.py#L123). On master description is too generic. Don't you think ?

Thanks @amoyaux for this quick response !

dreispt:
@legalsylvain Yes, that would match my experience.
Let me elaborate with an example:
- A bottle of wine weights 1250g (gross) = 750g wine (net weight) + 500g bottle
- When I ship the bottle, for my transport company the product weight I need to use is 1250.
- When I ship 6 bottles in a wood crate, the crate package weights 500g. That's what the "excluding package" means. For my transport operation, total weight will then be 6*1250 + 500 (including package).

For Odoo shipping to work correctly, I need the "weight" field to be set to 1250g, not 750g.

Does this make sense?

legalsylvain:
@dreispt : perfect ! thanks.

rousseldenis:
@legalsylvain Just one question, if you want to manage packaging weight, why not putting it there as for one product, you can have different weights - but maybe is too keep it simple ?

legalsylvain:
Hi @rousseldenis. 
Well, I made an error regarding "packaging" world. Let's take @dreispt exemple, with bottle : 

- The content (wine) weighs ``750g``. This is the net weight. until now, I was using the ``weight`` field, that was wrong, according to what @amoyaux says . So, i'll rename the current module to ``product_net_weight``, and i will switch to use this new field. I need this field to compute price per unit value. That is mandatory in food business to display on label in your shop the price per liter or price per kilo of food product. (see below)

- The container (bottle) weighs ``400g``, so the gross weight (``weight`` in odoo, if i understand correctly) is ``750 + 400 = 1150g``. 

- The packaging (box) weighs ``300g``, so a box of 6 bottles weighs ``(6 * 1150) + 500 = 7400g``, that means ``1233 gr per Unit`` (``weight_packaging_included``) I need this field for logistic purpose because we have developed an in-house delivery service at grap, and we need to know if the truck is overloaded.
 But it's to hard to know the weigh of each package. (We have a database of 10.000 products, and each product has it packaging). (I understand your solution of creating a packaging product.) 
To have this value, we decide to make an approximation that is ``weight + 10%``. (In my example it is 7%).

Hope I answered your question.


Label edited by Odoo.
image

pedrobaeza:
Sometime ago, there was net weight field in Odoo core: https://github.com/odoo/odoo/blob/9e8f70e4849b0eeaca8b5cf51372ecfa23dc561b/addons/product/product.py#L546 (this is 8.0)

legalsylvain:
Hi all. 
- I updated the module to introduce the "Net Weight" field.
- Also added the computed field "Weight Price".
- added tests & updated documentation and PR description.

It's ready for review. 

Thanks.

product_net_weight/models/product_product.py
```diff
@@ -0,0 +1,30 @@
...
+    )
+
+    weight_price = fields.Float(
+        string="Weight Price",
+        digits=dp.get_precision("Product Price"),
+        compute="_compute_weight_price"
+    )
+
+    # Explicit field, renaming it
+    weight = fields.Float(string="Gross Weight")
```
rousseldenis, line 19: @legalsylvain I would have done something else (as I suppose it's for display purpose):

```suggestion
    gross_weight = fields.Float(related="weight", readonly=False)
```

And make the normal one invisible.

legalsylvain: Hi. I'm not sure of this proposal. I see the current one lighter. With your design : 

- all the views displaying ``weight`` field will continue to display "Weight" instead of "Gross Weight". 
- to display the gross weight, we'll have to overload each view.
- the models has two identical fields, that is less understandable IMO.

what is the point of your solution ?

rousseldenis: Yes, that was the bottleneck. Ok for me

legalsylvain: Talking about bottleneck : it makes sense, on a wine bottle topic...

hparfr:
Thanks for this module.

I prefer (not blocking) if this module only adds a net_weight field and do not any calculation about price. 
Be dependent of product_logistics_uom is also a great feature.

- product_logistics_uom (product_attribute, v12, v14) takes care of expressing volume and weights in different units, it allows to express bottle of 75cl instead of 0,00075m³ (rounded at 0 by odoo) or 30gr of pepper without changing decimal precision. product_net_weight is not dependent of product_logistics_uom
- price calculation is more complex than that and, IMO, it should not be included in this module. Because, it should be done against a different unit of measure (like volume for the liquids, or per unit for like eggs).
- price display, is another story, on e-commerce and labels, you may want to display €/gr or €/100gr instead of €/kg.

Because I know, you don't want to change now, I hope, you agree to let these changes happen in v14.

legalsylvain:
Hi @hparfr. 

I removed Weight Price, as indeed in next version, it should be maybe in other module. (refactor with ``product_logistics_uom``, or other).

Could you review and approve this PR, and we can plan to work in the next oca days on that topic ? 

kind regards.

rvalyi:
Hello @sylvainc I was about to ask you the same as @hparfr . We have a customer selling air-conditionners who need this and price per Kg won't interrest them. I'm glad you removed the calculation. Now you have a trivial flake8 issue to fix (remove uselss api import) and then I promise to approve it.

mbcosta (approved):

mbcosta:
Hi @legalsylvain 
Good work, I also had doubt about what weight field refered after v8 Gross or Net and by openupgrade thought it was Net, thanks to make it clear. One suggestion is includ the wine example at README.

Should this https://github.com/OCA/OpenUpgrade/blob/9.0/addons/product/migrations/9.0.1.2/pre-migration.py#L133 be fixed?

I already approved because is simple module and the travis error is just an import not used after fixed the PR should be green.

legalsylvain:
hi @mbcosta. thanks for your review.
- I fixed travis, should be green now.
- I made the PR against openupgrade, could you review it ? https://github.com/OCA/OpenUpgrade/pull/2892

thanks.

rvalyi (approved):
thank you @legalsylvain for having clarified this weight mess!!

rvalyi (approved):

PR #899, opened 2021-08-05
Init copier template

joao-p-marques:
Also, add the `swig` package in Travis, needed when installing `reporting-engine`.

ping @pedrobaeza @Yajo @CarlosRoca13


pedrobaeza:
As said, please don't do that for all repos, but change OCA/reporting-engine

yajo (approved):

pedrobaeza:
The `swig` thing should be fixed on OCA/reporting-engine, not here

joao-p-marques:
Changed and ready to merge @pedrobaeza

PR #902, opened 2021-08-09
[14.0][FIX] product_secondary_unit : ZeroDivisionError when qty is null

Kev-Roche:
@sergio-teruel, @rousseldenis, @hparfr 
As mentioned by  @Bilbonet (https://github.com/OCA/sale-workflow/pull/1671#issuecomment-894473115), a quantity set as 0 brings the ZeroDivisionError: float division by zero


pedrobaeza:
Shouldn't you tackle the underlying problem (as the operation you are protecting is a multiplication, not division)?

Kev-Roche:
> Shouldn't you tackle the underlying problem (as the operation you are protecting is a multiplication, not division)?

Double mistake on my side, no change are requested here, the problem was elsewhere. And you were right about my modif. 
Sorry for the trouble.

PR #903, opened 2021-08-11
[13.0] product_packaging_type_required: improve + add test

simahawk:
The limit was useless because never used.
That's also the reason why the cron was working.
Limit is dropped now and the code improved a little bit.
A basic test and logging has been added too.

### DIFF
--- a/product_packaging_type_required/__manifest__.py
+++ b/product_packaging_type_required/__manifest__.py
@@ -3,7 +3,7 @@
 
 {
     "name": "Product Packaging Type Required",
-    "version": "13.0.1.0.0",
+    "version": "13.0.1.1.0",
     "development_status": "Beta",
     "category": "Product",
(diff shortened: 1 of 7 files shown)


product_packaging_type_required/__manifest__.py
```diff
@@ -3,7 +3,7 @@
 
 {
     "name": "Product Packaging Type Required",
-    "version": "13.0.1.0.0",
+    "version": "13.0.1.1.0",
```
mmequignon, line 6: Shouldn't it be done by the bot?

mmequignon (approved):

product_packaging_type_required/__manifest__.py
simahawk: no because there's a migration script

sebalix (approved):

PR #905, opened 2021-08-24
[FIX] product_pricelist_supplierinfo: public users

chienandalu:
In the e-commerce, if we have a pricelist with a rule based on
supplierinfo for a determined supplier partner, such partner could be
not accesible for public users and the affected products could not be
reached.

cc @Tecnativa TT31476

ping @pedrobaeza @CarlosRoca13

### DIFF
--- a/product_pricelist_supplierinfo/models/product_template.py
+++ b/product_pricelist_supplierinfo/models/product_template.py
@@ -26,9 +26,12 @@ class ProductTemplate(models.Model):
             if type(date) == datetime:
                 date = date.date()
             seller = product._select_seller(
-                partner_id=rule.filter_supplier_id,
+                # For a public user this record could be not accessible, but we
+                # need to get the price anyway
+                partner_id=rule.sudo().filter_supplier_id,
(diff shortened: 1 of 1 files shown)


pedrobaeza (approved):

pedrobaeza:
@eLBati I remember you faced something similar. Can it be? Or it was @SimoRubi ?

CarlosRoca13 (approved):

SimoRubi:
> @eLBati I remember you faced something similar. Can it be? Or it was @SimoRubi ?

I'm sorry I don't recall anything like this

chienandalu:
After proper testing of the patch, I changed the approach.

eLBati:
@pedrobaeza was it this https://github.com/OCA/product-attribute/pull/667 ?

pedrobaeza:
> @pedrobaeza was it this #667 ?

Not exactly. I think my memory has failed this time.

I'm merging for avoiding the error 500:

/ocabot merge patch

Is this needed in v13?

PR #907, opened 2021-09-07
[14.0][FIX] secondary_uom: independent behavior

Kev-Roche:
@hparfr, @sergio-teruel 
Fix minor behavior of the independent mode : 
- put 1 as primary unit quantity by default
- Change of the secondary unit quantity doesn't change the primary unit quantity when both UoM are independent.


sergio-teruel (approved):

hparfr (approved):
LGTM

PR #911, opened 2021-09-17
[14.0][ADD] product_pricelist_item_button

rousseldenis:
Depends on:
- [ ] #910 

- [ ] Remove test-requirements.txt entry

Sélection_660


product_pricelist_item_button/models/product_pricelist_item.py
```diff
@@ -0,0 +1,21 @@
+# Copyright 2021 ACSONE SA/NV (<http://acsone.eu>)
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+from odoo import _, models
+
+
+class ProductPricelist(models.Model):
+    _inherit = "product.pricelist"
+
+    def show_rules(self):
```
legalsylvain: Maybe show_items could be a better name

rousseldenis: I've followed Odoo display names :-)

rousseldenis: But I can change

legalsylvain: matter of taste. i usually use technical name for technical name function. but really not a blocking point !

product_pricelist_item_button/__manifest__.py
```diff
@@ -0,0 +1,21 @@
...
+    "name": "Product Pricelist Item Button",
+    "summary": """
+        Allows to access to pricelist items tree view to get search features""",
+    "version": "14.0.1.0.0",
+    "license": "AGPL-3",
+    "author": "ACSONE SA/NV,Odoo Community Association (OCA)",
+    "maintainers": ["rousseldenis"],
+    "website": "https://github.com/OCA/product-attribute",
+    "depends": [
+        "product",
```
dalonsod, line 14: Unnecessary dependency (already included in `product_pricelist_button_box`)
```suggestion
```

dalonsod (changes requested):
Locally tested OK. See comment, and test should be provided.

Non-blocking suggestions:
* Maybe a useful `active_item_ids_count` should be added (and used in the link button), indicating the current items active by date, with the related filter enabled by default. Anyway, I don't really know if such thing is easy to achieve...
* Hiding the current tab with rules should be considered, because these data view is replaced by the new button. It should take the advantage of a faster load of `product.pricelist` form when item list becomes too long.

@rousseldenis I'm interested in porting this addon to v13, so I can help with this PR if you need help to finish it

PR #913, opened 2021-09-22
[IMP] product_packaging_dimension: have product_uom_mm for dependants…

hailangvn:
… to use

Module `shopfloor_batch_automatic_creation` is depending on this module and need `product_uom_mm` uom for testing. The uom is in `stock_measuring_device` module but the former module does not need the later module. So this commit is to move the uom to this module for better dependency.

### DIFF
--- a/product_packaging_dimension/__manifest__.py
+++ b/product_packaging_dimension/__manifest__.py
@@ -13,6 +13,7 @@
     ],
     "website": "https://github.com/OCA/product-attribute",
     "data": [
+        "data/uom.xml",
         "views/product_packaging.xml",
     ],
     "installable": True,
--- a/product_packaging_dimension/data/uom.xml
+++ b/product_packaging_dimension/data/uom.xml
@@ -1,0 +1,13 @@
+<?xml version="1.0" encoding="UTF-8" ?>
+<!--
+    This is from https://github.com/OCA/stock-logistics-warehouse/blob/14.0/stock_measuring_device/data/uom.xml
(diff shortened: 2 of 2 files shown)


sebalix (commented):
Probably this UoM deserves its own module (`uom_mm`), as it could be used outside of packaging management?

hailangvn:
Thanks @sebalix for the suggestion. IMHO it is too small to have its own module. How about adding it to https://github.com/OCA/product-attribute/tree/14.0/product_logistics_uom?

sebalix:
Let's ask to a PSC of this repository: what do you think about this @rousseldenis @JordiBForgeFlow ?

sebalix:
@hailangvn we prefer to get a dedicated module `uom_mm` to host this new UoM (even if it is a small one, it's better to manage in the future)

hailangvn:
Thanks @sebalix for the advice. I will make the module.

hailangvn:
How about we use the name `oca_uom` and add some common uom like `ml`, `Week` (7 days), `Work Week` (5 days)?

sebalix:
With a bit a chance the `mm` could be integrated in the Odoo standard later, meanwhile I think it's easier to get a small module including only this UoM (and hopefully it won't be migrated to further versions one day).
A module depending on such an addon won't include undesirable UoM in the project (like the ones you suggest).
Having a small module for that is OK :)

hailangvn:
Close this as https://github.com/OCA/product-attribute/pull/914 is created.

PR #914, opened 2021-09-24
[ADD] uom_mm: have mm uom

hailangvn:
Module `shopfloor_batch_automatic_creation` need `mm` uom for testing. The uom is in `stock_measuring_device` module but the former module does not need the later module. The uom is also commonly used so this module is to add it for better dependency.


legalsylvain (changes requested):
Hi. 

I'm not sure such module should exist under OCA umbrella. AFAIU, it just add a data (``uom.uom`` milimitter). If we go in that direction, a lot of module will exist.
why don't you just create a new uom in the database ?

sorry if my question is not relevant.

kind regards.

hailangvn:
> I'm not sure such module should exist under OCA umbrella. AFAIU, it just add a data (`uom.uom` milimitter). If we go in that direction, a lot of module will exist.
> why don't you just create a new uom in the database ?

Thanks @legalsylvain for your comment. It is correct that we can simply add it but then later we will have it in different modules, as in `stock_measuring_device` one. This module can be dropped if Odoo add the `mm` uom.

legalsylvain:
> Thanks @legalsylvain for your comment. It is correct that we can simply add it but then later we will have it in different modules, as in stock_measuring_device one. This module can be dropped if Odoo add the mm uom.

Thanks for your answer.

that's not my point. The question is : why adding such record by data ? I mean, we could also removing it from ``stock_measuring_device``. why it is required ?


thanks !

pedrobaeza:
At last, if such measures are needed, we can add a `uom_extra_data` or similar that gathers all of them.

hailangvn:
> that's not my point. The question is : why adding such record by data ? I mean, we could also removing it from stock_measuring_device. why it is required ?

The `mm` uom is actually used  in `stock_measuring_device` (1). So after this is merged, we can update the module to use this instead of having its own.
(1) https://github.com/OCA/stock-logistics-warehouse/blob/d1e2219b3ebeb669ff1d610df00d20b380027628/stock_measuring_device/wizard/measuring_wizard.py#L40

> At last, if such measures are needed, we can add a uom_extra_data or similar that gathers all of them.

Thanks @pedrobaeza. Maybe some common uom like `ml`, `Week` (7 days), `Work Week` (5 days)?

pedrobaeza:
> Thanks @pedrobaeza. Maybe some common uom like ml, Week (7 days), Work Week (5 days)?

Yes, for both having the need fulfilled depending on this module, and being a hub for any of this non common UoMs that are not included in Odoo standard.

sebalix:
It's me who ask @hailangvn  to create `uom_mm` instead of a bigger addon, but OK for `uom_extra_data`.

legalsylvain:
Well. Not a blocking point, but I don't understand why you don't use the default existing meter ``uom.uom``. if you use it, you don't have to have such new uom.

I don't know deeply the module, but why ``width`` are in mm and not meter here ?
https://github.com/OCA/stock-logistics-warehouse/blob/d1e2219b3ebeb669ff1d610df00d20b380027628/stock_measuring_device/wizard/measuring_wizard_line.py#L18

it doesn't seem very consistent with the volume which is expressed in meters. (L20)

and so it requires extra conversion : https://github.com/OCA/stock-logistics-warehouse/blob/d1e2219b3ebeb669ff1d610df00d20b380027628/stock_measuring_device/wizard/measuring_wizard_line.py#L41

Well, I'm not very confortable having fields that you never know in which unit they fit. There was a similar problem in odoo before with the percentages. 5% was sometimes written 0.05 and sometimes 5.0 in the database. But this is now corrected in recent versions. I have the impression that having length fields expressed sometimes in meters, and sometimes in millimeters reproduces the problem.

But well, just an opinion, and as said before, I don't know the module. so :+1: if other PSC are OK.

pedrobaeza:
Good reasoning, Sylvain. Don't know either about the use case for mm.

legalsylvain:
Reading again the code of the other module, i think that adding dimensional_uom_id on the wizard should avoid to hardcode millimeter unit and could be more modular.
What do you think ?

hailangvn:
Thanks @legalsylvain and @pedrobaeza for the great discussion. I updated `shopfloor_batch_automatic_creation` to use `cm` instead to reduce dependency. For `stock_measuring_device`, I think there is a reason for `mm`. It may be used to increase accuracy in calibration. And yes, it is good idea to have `dimensional_uom_id` in the wizard.

I close this PR for now until the need is firmed.

PR #916, opened 2021-10-07
[ADD] product_sold_by_delivery_week: New module

chienandalu:
This module allows to compute a graphical hint stored in the product itself to display
whether or not that product was selled and delivered in the past weeks.

image

Pending:

- [ ] Tests

cc @Tecnativa TT31742


pedrobaeza:
I think sale-reporting repo seems a better target. About the name, at least it should be `sold`, not `selled`, but may I suggest `product_sale_report_by_delivery_week` or similar, but I don't know exactly the goal.

product_selled_delivered_weekly/models/product_template.py
```diff
@@ -0,0 +1,74 @@
+# Copyright 2021 Tecnativa - David Vidal
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+from odoo import api, fields, models
+
+
+class ProductTemplate(models.Model):
+    _inherit = "product.template"
+
+    weekly_selled_delivered = fields.Char(
+        string="Internal Reference",
```
legalsylvain: I think that the string is not correct. (X2)

product_selled_delivered_weekly/views/product_views.xml
```diff
@@ -0,0 +1,33 @@
+<?xml version="1.0" encoding="utf-8" ?>
+<!-- Copyright 2021 Tecnativa - David Vidal
+     License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl). -->
+<odoo>
+    <record id="product_template_tree_view" model="ir.ui.view">
+        <field name="model">product.template</field>
+        <field name="inherit_id" ref="product.product_template_tree_view" />
+        <field name="groups_id" eval="[(4, ref('base.group_user'))]" />
```
legalsylvain: Why this group ?

legalsylvain (commented):
- Some minor remarks inline.
- adding some screenshots in the readme part could be great.
- looks great ! Nice design, and I guess quite user friendly.

product_selled_delivered_weekly/views/product_views.xml
chienandalu: I'm going to change the groups to the basic sales permission, but I didn't remember the xml_id at the moment so I just placeholded it so once I could replace them all in one go

pedrobaeza: `sales_team.group_salesman`

product_selled_delivered_weekly/models/product_template.py
chienandalu: Fixed

product_selled_delivered_weekly/views/product_views.xml
chienandalu: Also fixed

carlosdauden (approved):

rousseldenis:
Nice module.

@chienandalu  Shouldn't some code from stock_demand_estimate one be reused ?

https://github.com/OCA/stock-logistics-warehouse/tree/14.0/stock_demand_estimate

chienandalu:
> @chienandalu Shouldn't some code from stock_demand_estimate one be reused ?
> 
> https://github.com/OCA/stock-logistics-warehouse/tree/14.0/stock_demand_estimate

I don't really see how. This module doesn't estimate nothing but reflect the fact of the already sold :)

chienandalu:
Moving to https://github.com/OCA/sale-reporting/pull/121 as it's a more appropiate repository

PR #922, opened 2021-10-19
[14.0][ADD] New module 'product_main_supplierinfo'

sebalix:
This module computes the main vendor of a product, as we likely duplicate this feature in different modules.

- [x] <s>I first named the `product_main_supplier` and then renamed it to `product_main_vendor` as Odoo uses this term on product forms, not sure which term fits the best.</s>


pedrobaeza:
As this is the technical name of the module, we stick to this for putting name, so I would go with `product_main_supplierinfo`.

product_main_supplierinfo/views/product_product.xml
```diff
@@ -0,0 +1,30 @@
...
+    </field>
+  </record>
+
+  <record id="product_product_tree_view" model="ir.ui.view">
+    <field name="name">product.product.tree.inherit</field>
+    <field name="model">product.product</field>
+    <field name="inherit_id" ref="product.product_product_tree_view" />
+    <field name="arch" type="xml">
+      <field name="uom_id" position="after">
+        <field name="main_vendor_id" optional="show" />
```
simahawk:
```suggestion
        <field name="main_vendor_id" optional="hide" />
```
to avoid loading seller_ids in the tree view by default

simahawk (approved):

product_main_supplierinfo/views/product_product.xml
pedrobaeza: AFAIK, they are loaded anyways as soon as you add it in the tree, no matter if hidden or not.

sebalix: done

product_main_supplierinfo/models/product_product.py
```diff
@@ -0,0 +1,54 @@
...
+        ondelete="restrict",
+        string="Main Vendor",
+        compute="_compute_main_vendor_id",
+    )
+
+    @api.depends(
+        "seller_ids.name.active",
+        "seller_ids.sequence",
+        "seller_ids.min_qty",
+        "seller_ids.price",
```
legalsylvain, line 20: Why do you depends on all that fields ?

legalsylvain (changes requested):
Thanks for this module. One question inline. 

Regards

product_main_supplierinfo/models/product_product.py
sebalix: They are all the fields used by the computation:
- the standard `_prepare_sellers` method is using `name.active`, `sequence`, `min_qty` and `price`
- and the rest by `_get_sellers`.

sebalix: `ddmrp` is also computing this main vendor here: https://github.com/OCA/ddmrp/blob/e7487ea43b5ff9d7be0c21cdff7f1d97b5a9c1c6/ddmrp/models/stock_buffer.py#L825-L842
I'll open a PR afterwards to make it use this module instead.

TDu (approved):

product_main_supplierinfo/models/product_product.py
legalsylvain: Oh I see. I didn't know that price was used by _prepare_sellers. (I thought that it was only order by sequence.) Thanks !

product_main_supplierinfo/models/product_product.py
```diff
@@ -0,0 +1,54 @@
...
+
+class ProductProduct(models.Model):
+    _inherit = "product.product"
+
+    main_supplierinfo_id = fields.Many2one(
+        comodel_name="product.supplierinfo",
+        ondelete="restrict",
+        string="Main Vendor",
+        compute="_compute_main_supplierinfo_id",
+    )
```
legalsylvain, line 14: For consistency reason with Odoo product module, I propose to rename the field : 

    main_seller_id = fields.Many2one(
        comodel_name="product.supplierinfo",
        ondelete="restrict",
        string="Main Vendor",
        compute="_compute_main_seller_id",
    )

See : https://github.com/odoo/odoo/blob/14.0/addons/product/models/product_template.py#L113

Not a blocking point, but I see it more consistent, don't you think ?

legalsylvain (approved):
Looks great. One non blocking point inline.

Other question : Do you think it is usefull to set a ``ondelete="restrict"`` on a non stored field ? I thought it had an impact only on the database side. (so, not required in that case).

Otherwise :+1: thanks.

sebalix:
@legalsylvain the ondelete"restrict" is a mistake indeed, it comes from a bad copy-paste where this field was stored.

BTW storing the value to be able to search on it is another need we have, but it is another story... I will probably propose something later in a new PR.
About the name I could rename it yes (I already did twice :D), some OCA addons talk about "supplier", Odoo uses "seller" as technical term and "vendor" on UI, while the underlying data model uses "supplier", it's quite confusing, that's why I renamed it like the data model it was pointing and used "supplier" + uses "Vendor" term as label.

Any other opinion on this? @pedrobaeza maybe? :)

pedrobaeza:
In fact, not being stored, the ondelete has no effect AFAIK.

sebalix:
@pedrobaeza yes of course, that's why it has been removed :) I ping you more about the name suggested by @legalsylvain , but if everyone is fine with the current name I won't touch it.
Ready for a merge then.

pedrobaeza:
Ah, OK, thinking twice, indeed the correspondence `seller_ids` > `main_seller_id` seems clearer, but the module name is still correct. It's a bit annoyance that Odoo is not consistent even on its technical names, but also understandable if we split model name with field name.

product_main_supplierinfo/models/product_product.py
sebalix: Updated!

pedrobaeza:
Thanks for doing all the requested changes:

/ocabot merge nobump

pedrobaeza:
Relaunching the merge:

/ocabot merge nobump

PR #925, opened 2021-10-22
[14.0] Add product_status

sonhd91:
This module compute a status on product based on some date field displayed on the Sales tab.
There are some change apply with this module:
- `product_state` updated: depend on [PR#931](https://github.com/OCA/product-attribute/pull/931)
- Use `product_state` features for this module instead
- Add `product_state` new data , these data can not be modified, deactive default data from `product_state`
- Tests updated, add prevent modified test case

Depends on:

- [x] #961
- [x] #962

PR #931, opened 2021-10-28
[14.0] product_state: move default states to demo data

sonhd91:
@simahawk suggestion in [issue comment](https://github.com/OCA/product-attribute/pull/925#issuecomment-950631523)
- Add fields, `active`, `default` in `product.state`
- Default `product_state_id` only get records which `default=True`

### DIFF
--- a/product_state/data/product_state_data.xml
+++ b/product_state/data/product_state_data.xml
@@ -9,6 +9,7 @@
         <field name="code">sellable</field>
         <field name="name">Normal</field>
         <field name="sequence">20</field>
+        <field name="default">True</field>
     </record>
     <record id="product_state_end" model="product.state">
         <field name="code">end</field>
--- a/product_state/models/product_state.py
+++ b/product_state/models/product_state.py
@@ -1,7 +1,8 @@
 # Copyright 2017 ACSONE SA/NV (<http://acsone.eu>)
 # License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl.html).
 
-from odoo import api, fields, models
+from odoo import _, api, fields, models
+from odoo.exceptions import ValidationError
 
 
 class ProductState(models.Model):
@@ -14,6 +15,7 @@ class ProductState(models.Model):
     sequence = fields.Integer(
         string="Sequence", help="Used to order the States", default=25
     )
+    active = fields.Boolean(default=True)
     description = fields.Text(translate=True)
     product_ids = fields.One2many(
         comodel_name="product.template",
@@ -24,7 +26,7 @@ class ProductState(models.Model):
         string="Number of products",
         compute="_compute_products_count",
     )
-
+    default = fields.Boolean("Default state")
     _sql_constraints = [
         ("code_unique", "UNIQUE(code)", "Product State Code must be unique.")
     ]
@@ -42,3 +44,8 @@ class ProductState(models.Model):
         }
         for state in self:
             state.products_count = mapped_data.get(state.id, 0)
+
+    @api.constrains("default")
+    def _check_default(self):
+        if self.search_count([("default", "=", True)]) > 1:
+            raise ValidationError(_("There should be only one default state"))
--- a/product_state/models/product_template.py
+++ b/product_state/models/product_template.py
@@ -28,9 +28,7 @@ class ProductTemplate(models.Model):
 
     @api.model
     def _get_default_product_state_id(self):
-        return self.env.ref(
-            "product_state.product_state_sellable", raise_if_not_found=False
-        )
+        return self.env["product.state"].search([("default", "=", True)], limit=1).id
 
     @api.depends("product_state_id")
     def _compute_product_state(self):
--- a/product_tier_validation/data/product_state_data.xml
+++ b/product_tier_validation/data/product_state_data.xml
@@ -1,7 +1,4 @@
 <odoo noupdate="1">
-    <record id="product_state.product_state_draft" model="product.state">
-        <field name="code">draft</field>
-    </record>
     <record id="product_state.product_state_sellable" model="product.state">
         <field name="code">confirmed</field>
     </record>
--- a/product_state/README.rst
+++ b/product_state/README.rst
@@ -50,6 +50,11 @@ To add a product to a state:
 #. Go to the product itself and edit.
 #. You can select the desired status in the list of buttons above the form.
 
+Known issues / Roadmap
+======================
+
+File data/product_state_data.xml will be moved to demo/product_state_demo.xml since version 15.0
+
 Bug Tracker
 ===========
 
--- a/product_state/readme/ROADMAP.rst
+++ b/product_state/readme/ROADMAP.rst
@@ -1,0 +1,1 @@
+File data/product_state_data.xml will be moved to demo/product_state_demo.xml since version 15.0
--- a/product_state/static/description/index.html
+++ b/product_state/static/description/index.html
@@ -3,7 +3,7 @@
 <html xmlns="http://www.w3.org/1999/xhtml" xml:lang="en" lang="en">
 <head>
 <meta http-equiv="Content-Type" content="text/html; charset=utf-8" />
-<meta name="generator" content="Docutils 0.15.1: http://docutils.sourceforge.net/" />
+<meta name="generator" content="Docutils: http://docutils.sourceforge.net/" />
 <title>Product State</title>
 <style type="text/css">
 
@@ -379,11 +379,12 @@ ul.auto-toc {
 <div class="contents local topic" id="contents">
 <ul class="simple">
 <li><a class="reference internal" href="#usage" id="id1">Usage</a></li>
-<li><a class="reference internal" href="#bug-tracker" id="id2">Bug Tracker</a></li>
-<li><a class="reference internal" href="#credits" id="id3">Credits</a><ul>
-<li><a class="reference internal" href="#authors" id="id4">Authors</a></li>
-<li><a class="reference internal" href="#contributors" id="id5">Contributors</a></li>
-<li><a class="reference internal" href="#maintainers" id="id6">Maintainers</a></li>
+<li><a class="reference internal" href="#known-issues-roadmap" id="id2">Known issues / Roadmap</a></li>
+<li><a class="reference internal" href="#bug-tracker" id="id3">Bug Tracker</a></li>
+<li><a class="reference internal" href="#credits" id="id4">Credits</a><ul>
+<li><a class="reference internal" href="#authors" id="id5">Authors</a></li>
+<li><a class="reference internal" href="#contributors" id="id6">Contributors</a></li>
+<li><a class="reference internal" href="#maintainers" id="id7">Maintainers</a></li>
 </ul>
 </li>
 </ul>
@@ -401,8 +402,12 @@ ul.auto-toc {
 <li>You can select the desired status in the list of buttons above the form.</li>
 </ol>
 </div>
+<div class="section" id="known-issues-roadmap">
+<h1><a class="toc-backref" href="#id2">Known issues / Roadmap</a></h1>
+<p>File data/product_state_data.xml will be moved to demo/product_state_demo.xml since version 15.0</p>
+</div>
 <div class="section" id="bug-tracker">
-<h1><a class="toc-backref" href="#id2">Bug Tracker</a></h1>
+<h1><a class="toc-backref" href="#id3">Bug Tracker</a></h1>
 <p>Bugs are tracked on <a class="reference external" href="https://github.com/OCA/product-attribute/issues">GitHub Issues</a>.
 In case of trouble, please check there if your issue has already been reported.
 If you spotted it first, help us smashing it by providing a detailed and welcomed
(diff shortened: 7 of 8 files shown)


test-requirements.txt
```diff
@@ -1 +1,2 @@
 odoo_test_helper
+odoo14-addon-product_tier_validation @git+https://github.com/OCA/product-attribute@refs/pull/932/head#subdirectory=setup/product_tier_validation
```
simahawk: what is this? :thinking:

simahawk:
pls move #932 here and drop the dependency

sonhd91:
> pls move #932 here and drop the dependency

Thanks,  branch updated

test-requirements.txt
sonhd91: Dependency removed, branch updated

simahawk (approved):

simahawk:
can you rebase to fix pre-commit issue?

simahawk:
@rousseldenis would you mind having a look?

rousseldenis:
I'm not sure at all to understand the needs here.

Why not having the default state as 'sellable' as this should correspond to default behaviour ?

If you don't put default if you are in real, you need to go through all your products and change it. Moreover, this change would require the default method surcharge in every case. This is less convenient.

And last but not least, I would say that putting reference to demo data in real code is at the opposite of all logic...

I'm not in favour of this change.

If you have specific requirements in your project, just override the default method.

simahawk:
> I'm not sure at all to understand the needs here.
> 
> Why not having the default state as 'sellable' as this should correspond to default behaviour ?
> 
> If you don't put default if you are in real, you need to go through all your products and change it. Moreover, this change would require the default method surcharge in every case. This is less convenient.
> 
> And last but not least, I would say that putting reference to demo data in real code is at the opposite of all logic...

@rousseldenis who said you should use demo data? It's demo data, you don't use it for real.
The key point here is that this module should not define default states because they have no functional reason to be there and they can be suitable for 50% of the cases but not for the rest.
Hence, you will always have to hide or delete them w/ the risk of getting them back on an upgrade.

See https://github.com/OCA/product-attribute/pull/925#issuecomment-950631523


> I'm not in favour of this change.
> 
> If you have specific requirements in your project, just override the default method.

If you have project specific requirements just define your own states.

The only workaround I see is to make states archive-able and disable all the ones that you don't need but TBH I don't see any reason to keep useless records.
Plus, we should probably have a flag to set the default one.

product_state/models/product_template.py
```diff
@@ -29,7 +29,7 @@ class ProductTemplate(models.Model):
     @api.model
     def _get_default_product_state_id(self):
         return self.env.ref(
-            "product_state.product_state_sellable", raise_if_not_found=False
+            "product_state.demo_product_state_sellable", raise_if_not_found=False
```
simahawk: this is wrong, indeed.

simahawk (changes requested):

product_state/models/product_template.py
sonhd91: I had fixed default function, thanks

simahawk:
@rousseldenis would you agree with the following?

1. have an `active` flag
2. have a `default` flag and change default get to look for it
3. in `product_status` we set all existing records as inactive for v14
4. in `product_state` make them demo data for v15 (to no have bad surprises for existing instances on 14.0)
5. document this in both modules' roadmap

rousseldenis:
> @rousseldenis would you agree with the following?
> 
> 1. have an `active` flag
> 2. have a `default` flag and change default get to look for it
> 3. in `product_status` we set all existing records as inactive for v14
> 4. in `product_state` make them demo data for v15 (to no have bad surprises for existing instances on 14.0)
> 5. document this in both modules' roadmap

Yes, that sounds acceptable. Thanks. I'll try to test if it fits existing flows.

simahawk:
@rousseldenis tnx!

>     1. have an `active` flag
>     2. have a `default` flag and change default get to look for it
>     3. in `product_status` we set all existing records as inactive for v14
>     4. in `product_state` make them demo data for v15 (to no have bad surprises for existing instances on 14.0)
>     5. document this in both modules' roadmap

@sonhd91 could you please follow these steps?

product_state/__manifest__.py
```diff
@@ -12,10 +12,12 @@
     "depends": ["product", "sale"],
     "data": [
         "security/ir.model.access.csv",
-        "data/product_state_data.xml",
         "security/ir.model.access.csv",
         "views/product_template.xml",
     ],
+    "demo": [
```
simahawk: as stated in the roadmap, this cannot happen in v14 :wink:

product_state/demo/product_state_demo.xml
```diff
@@ -1,21 +1,22 @@
 <?xml version="1.0" encoding="utf-8" ?>
 <odoo noupdate="1">
-    <record id="product_state_draft" model="product.state">
+    <record id="demo_product_state_draft" model="product.state">
```
simahawk: pls rollback these changes

product_state/models/product_state.py
```diff
@@ -24,7 +25,7 @@ class ProductState(models.Model):
         string="Number of products",
         compute="_compute_products_count",
     )
-
+    default = fields.Boolean("Default state")
```
simahawk, line 29: we should have an `@api.constrains` check to make sure we have only 1 default at time

product_state/models/product_template.py
```diff
@@ -26,11 +26,13 @@ class ProductTemplate(models.Model):
...
 
+    # Choose from state has default=True, needed override when new states implemented
     @api.model
     def _get_default_product_state_id(self):
-        return self.env.ref(
-            "product_state.product_state_sellable", raise_if_not_found=False
+        res = self.env["product.state"].search(
+            [("default", "=", True), ("active", "=", True)]
         )
+        return res[0] if res else False
```
simahawk: the "active" leaf is useless as it's done by default

```suggestion
        return self.env["product.state"].search([("default", "=", True)], limit=1).id
```

product_tier_validation/models/product_template.py
```diff
@@ -28,9 +28,3 @@ def write(self, vals):
         if "state" in vals:
             vals["active"] = self._is_tier_validated_active(vals["state"])
         return super().write(vals)
-
-    @api.model
-    def _get_default_product_state_id(self):
```
simahawk: this probably should be kept as is on v14

simahawk (changes requested):

product_state/models/product_state.py
```diff
@@ -42,3 +44,14 @@ def _compute_products_count(self):
...
+    @api.constrains("default")
+    def constrains_default_state(self):
+        rec = self.search([("default", "=", True)])
+        if len(rec) != 1:
+            raise ValidationError(
+                _(
+                    "There should be only one default state, current default state: '%s'"
+                    % (rec - self).name
+                )
+            )
```
simahawk:
```suggestion
    def _check_default(self):
        if self.search_count([("default", "=", True)]) > 1:
            raise ValidationError(
                _(
                    "There should be only one default state, current default state: '%s'"  
                )
            )
```

product_tier_validation/data/product_state_data.xml
```diff
@@ -1,7 +1,4 @@
 <odoo noupdate="1">
-    <record id="product_state.product_state_draft" model="product.state">
```
simahawk, line 2: this state should stay no?

product_tier_validation/models/product_template.py
simahawk: As per my comment, we should keep the direct ref because we cannot create another default state.
Otherwise, you have to add migration and installation script to swap the default state on "draft" and trash this override here.

simahawk (changes requested):

product_tier_validation/data/product_state_data.xml
sonhd91: This data already have one in `product_state` module

product_state/models/product_state.py
sonhd91: I think we should let user know which state is default when error raised

product_state/models/product_template.py
```diff
@@ -26,11 +26,10 @@ class ProductTemplate(models.Model):
         tracking=10,
     )
 
+    # Choose from state has default=True, needed override when new states implemented
```
simahawk: this comment is not correct IMO. The point is not that you need an override.
The point is that on the tier module, to avoid messing up w/ defaults and avoiding writing hooks + migration steps just for v14, it's easier to override it. Can you drop it pls?

The alternative is that you provide in the tier module an installation hook AND a migration step.
If you don't, pls add a comment to drop that override on v15.

sonhd91: Thanks for review, branch updated

rousseldenis (approved):

rousseldenis:
I think we should sync between us as it seems I have now the same requirements on product_tier_validation.

I don't like the way it's done currently for the active field. That behaviour should be activated through an option in settings ('[x] Set product as inactive as soon as not validated')

@simahawk

product_tier_validation/data/product_state_data.xml
rousseldenis: I think we should go further here as I don't understand why we should change the product states codes...

rousseldenis:
> I think we should sync between us as it seems I have now the same requirements on product_tier_validation.
> 
> I don't like the way it's done currently for the active field. That behaviour should be activated through an option in settings ('[x] Set product as inactive as soon as not validated')
> 
> @simahawk

After thinking about, that option should be here.

So, to summarize:

- Add an option to product.state 'Set product to inactive when reaching this state'.
- Remove in product.tier.validation the 'draft' state stuff.
What do you think about? I will add a new module that extend this.

simahawk:
> > I think we should sync between us as it seems I have now the same requirements on product_tier_validation.
> > I don't like the way it's done currently for the active field. That behaviour should be activated through an option in settings ('[x] Set product as inactive as soon as not validated')
> > @simahawk
> 
> After thinking about, that option should be here.
> 
> So, to summarize:
> 
>     * Add an option to product.state 'Set product to inactive when reaching this state'.
> 
>     * Remove in product.tier.validation the 'draft' state stuff.
>       What do you think about? I will add a new module that extend this.

I don't understand. 

There's no logic now for the active field. The active field here is just on the state to be able to archive states that are not needed. Here you talk about product.active being affected by the state and I don't get why.

Regarding the tier validation: I don't know. IMO if it has specific logic for specific states it should have its own states.
Up to you anyway. I don't know it.

rousseldenis:
> > > I think we should sync between us as it seems I have now the same requirements on product_tier_validation.
> > > I don't like the way it's done currently for the active field. That behaviour should be activated through an option in settings ('[x] Set product as inactive as soon as not validated')
> > > @simahawk
> > 
> > 
> > After thinking about, that option should be here.
> > So, to summarize:
> > ```
> > * Add an option to product.state 'Set product to inactive when reaching this state'.
> > 
> > * Remove in product.tier.validation the 'draft' state stuff.
> >   What do you think about? I will add a new module that extend this.
> > ```
> 
> I don't understand.
> 
> There's no logic now for the active field. The active field here is just on the state to be able to archive states that are not needed. Here you talk about product.active being affected by the state and I don't get why.
> 
> Regarding the tier validation: I don't know. IMO if it has specific logic for specific states it should have its own states. Up to you anyway. I don't know it.

Ok, I put this because I thought you spoke about validation in the thread somewhere. See https://github.com/OCA/product-attribute/pull/948

rousseldenis:
@simahawk This one is for me ok

simahawk:
ahhh I see, ok, thanks! :)

simahawk (approved):

product_state/models/product_state.py
i-vyshnevska: default = fields.Boolean("Default state", default=False)

product_state/data/product_state_data.xml
```diff
@@ -9,6 +9,7 @@
         <field name="code">sellable</field>
         <field name="name">Normal</field>
         <field name="sequence">20</field>
+        <field name="default">True</field>
```
i-vyshnevska, line 12: this file needs to be moved to the demo section in manifest

sonhd91: Thanks for review, I will check it out !

product_state/models/product_state.py
sonhd91: the [boolean](https://github.com/odoo/odoo/blob/14.0/odoo/fields.py#L1204) was inherited default = None from [Field](https://github.com/odoo/odoo/blob/14.0/odoo/fields.py#L234) and it is convert to False when we install the module. May I know why we need to implicitly define it?

product_state/data/product_state_data.xml
sonhd91: We had moved it to demo section but then need to revert back due to comments, [this comment](https://github.com/OCA/product-attribute/pull/931#issuecomment-966268076) and this [comment](https://github.com/OCA/product-attribute/pull/931#issuecomment-967029668)

product_state/models/product_state.py
i-vyshnevska: ah sorry, you don't need to add this, I just thought that was a reason for failed constrain, no change needed indeed

product_state/data/product_state_data.xml
i-vyshnevska: thanks for the ansver

PR #932, opened 2021-10-28
[14.0] product_tier_validation: move default states to demo data

sonhd91:


simahawk:
@sonhd91 pls move this to #931

PR #935, opened 2021-11-05
[14.0][IMP] product_profile: refactor to improve inheritance

sebastienbeau:
Give the possibility to call directly with a domain the method _customize_view
with a domain so you can reuse this method in custom module to add profile filter.

In my case I want to add a filter on mrp.bom so just adding a related
field + a call to _customize_view allow to add only profile related to
manufacturing

Also add a separator to avoid mixing profile filter with existing filter

### DIFF
--- a/product_profile/models/product_profile.py
+++ b/product_profile/models/product_profile.py
@@ -238,7 +238,7 @@ class ProductMixinProfile(models.AbstractModel):
         ]
 
     @api.model
-    def _customize_view(self, res, view_type):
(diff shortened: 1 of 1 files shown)


sebastienbeau:
@kevinkhao @bealdav

bealdav (approved):

simahawk (approved):
LG. @sebastienbeau pls include module name in commit msg and leave long desc for the commit description :wink:

rousseldenis:
@sebastienbeau

PR #942, opened 2021-11-23
[14.0][FW] [FIX] product_packaging_unit_price_calculator: Tests broken when

hailangvn:
website_sale installed

* This is to forward port #839

### DIFF
--- a/product_packaging_unit_price_calculator/tests/test_product_price_packaging_qty.py
+++ b/product_packaging_unit_price_calculator/tests/test_product_price_packaging_qty.py
@@ -1,4 +1,4 @@
-from odoo.tests.common import Form, SavepointCase
+from odoo.tests.common import SavepointCase
 
 
 class TestProductPricePackagingQty(SavepointCase):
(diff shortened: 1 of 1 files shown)


product_packaging_unit_price_calculator/tests/test_product_price_packaging_qty.py
```diff
@@ -36,22 +37,19 @@ def setUpClass(cls):
...
-        self.assertEqual(form.unit_price, 4)
-        form.save()
+        self.assertEqual(self.wizard_1.unit_price, 4)
         self.wizard_1.action_set_price()
         self.assertEqual(self.product.list_price, 4)
 
     def test_set_purchase_pacakge_price(self):
         self.wizard_1.product_supplierinfo_id = self.supplier_info
-        form = Form(self.wizard_1)
+        self.wizard_1.product_tmpl_id = self.product.product_tmpl_id
```
rousseldenis, line 49: @hailangvn Do you have tested with the 'with' context manager form? As the onchanges won't apply with the previous code?

hailangvn: Thanks @rousseldenis for the question. Here are my findings.
* onchanges will be applied when we access the column in `depends` statement so the previous code does work;
* the original test passed with the `with` context manager and without it;
* this change has no positive effect so I am closing this.

hailangvn:
The original test does work with module `website_sale` installed. So this port is not needed for 14.0.

PR #947, opened 2021-11-25
[14.0][IMP] product_state: Add an inverse method on product_state_id

rousseldenis:
Easy non-breaking change.

PR #948, opened 2021-11-25
[14.0][ADD] product_state_active

rousseldenis:
Depends on :

- [ ] #947


rousseldenis:
@simahawk @dreispt This is quite simple and avoid to have such things:

https://github.com/OCA/product-attribute/blob/14.0/product_tier_validation/models/product_template.py#L19

The tier validation IMHO should only do its stuff: doing validation. In combination with this, you can define the state that will trigger the active change.

The point is: that's the state change that should only trigger the active change, not the validation.

What do you think about ?

product_state_active/models/product_state.py
```diff
@@ -0,0 +1,27 @@
...
+from odoo import _, api, fields, models
+from odoo.exceptions import ValidationError
+
+
+class ProductState(models.Model):
+
+    _inherit = "product.state"
+
+    deactivate_product = fields.Boolean(
+        help="Check this if you want to inactive product when reaching this state."
```
simahawk:
```suggestion
        help="Check this if you want to archive products when reaching this state."
```

product_state_active/models/product_state.py
```diff
@@ -0,0 +1,27 @@
...
+    )
+    activate_product = fields.Boolean(
+        help="Check this if you want to active product when reaching this state."
```
simahawk:
```suggestion
        help="Check this if you want to un-archive products when reaching this state."
```

simahawk (approved):

product_state_active/models/product_state.py
rousseldenis: @simahawk Done. Please review too #947

flotho (approved):
LGTM

PR #951, opened 2021-11-25
[14.0][IMP] product_tier_validation: Simplify flow

rousseldenis:
- [x] Don't modify product 'active' field.
- [x] Don't modify initial states as now, you can archive them to use others.
- [x] Don't get default state as now it is configurable on state level.
- [x] Change to 'Beta'

If we want to manage that, it should be done in combination with https://github.com/OCA/product-attribute/pull/948

### DIFF
--- a/product_tier_validation/models/product_template.py
+++ b/product_tier_validation/models/product_template.py
@@ -1,36 +1,10 @@
 # Copyright 2021 Open Source Integrators
 # License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
 
-from odoo import api, models
+from odoo import models
 
 
 class ProductTemplate(models.Model):
     _name = "product.template"
     _inherit = ["product.template", "tier.validation"]
     _tier_validation_manual_config = False
-
-    @api.model
-    def _is_tier_validated_active(self, state_code):
-        return state_code != "draft"
-
-    @api.model
-    def create(self, vals):
-        new = super().create(vals)
-        if new.need_validation and self._is_tier_validated_active(new.state):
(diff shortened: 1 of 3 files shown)


hailangvn (approved):
LGTM, thanks for the improvement.

product_tier_validation/models/product_template.py
```diff
@@ -1,36 +1,10 @@
...
     _tier_validation_manual_config = False
-
-    @api.model
-    def _is_tier_validated_active(self, state_code):
-        return state_code != "draft"
-
-    @api.model
-    def create(self, vals):
-        new = super().create(vals)
-        if new.need_validation and self._is_tier_validated_active(new.state):
```
simahawk, line 19: this will change existing expected behavior, no?
I don't care that much... just raising a possible issue :wink:

simahawk: aaah nevermind #948 does that...

simahawk (approved):
we should go for major release as this is a breaking change

product_tier_validation/models/product_template.py
hailangvn: IIUC The activation/deactivation is now in another module by #948.

rousseldenis: Yes, that code was done by @dreispt in order to have that behaviour we have in product_state now.

rousseldenis:
@simahawk As soon as this is merged, I will do another PR in order to manage the write on 'product_state_id' as the evaluation is incorrect in that case. But I've already the solution

PR #850, opened 2021-04-25
14.0 mig product packaging dimension

Kev-Roche:
Replace migration #736  (inactive) and improve #799, inspired by #775

Instead of hardcoding the packaging volume in cubic meters, it is computed depending of both the dimensional (input) and volumetric (output) UoMs. Also add some tests.

PR #957, opened 2021-12-10
[FIX] product_packaging_dimension: use integer instead of float

i-vyshnevska:
set same field type as defined in core to be compatible with other addons as this field is used all over stock management stack
fix migration step
remove fields converting - idea is that previously value was stored in mm and customer wants to keep this measure in mm, we cannot set length_uom_id to mm as they defined in "stock_measuring_device" addon which is not mandatory installed so update of this field need to be handled by the customer separately

### DIFF
--- a/product_packaging_dimension/migrations/14.0.1.0.0/pre-migrate.py
+++ b/product_packaging_dimension/migrations/14.0.1.0.0/pre-migrate.py
@@ -5,18 +5,9 @@ from odoo.tools.sql import column_exists, rename_column
 
 
 def migrate(cr, version):
-    # Rename lnght into packaging_length
-    if column_exists(cr, "product_packaging", "lnght"):
-        rename_column(cr, "product_packaging", "lnght", "packaging_length")
 
-        # Convert old hard-coded uom values (mm)
-        # into new default uom values (m)
-        cr.execute(
-            """
-        UPDATE product_packaging
-        SET
-        packaging_length = packaging_length/1000,
-        height = height/1000,
-        width = width/1000,
-        """
-        )
+    # Rename lngth into packaging_length
+    if column_exists(cr, "product_packaging", "packaging_length"):
+        cr.execute("UPDATE product_packaging SET packaging_length = lngth")
+    elif column_exists(cr, "product_packaging", "lngth"):
+        rename_column(cr, "product_packaging", "lngth", "packaging_length")
--- a/product_packaging_dimension/models/product_packaging.py
+++ b/product_packaging_dimension/models/product_packaging.py
@@ -10,18 +10,18 @@ class ProductPackaging(models.Model):
     # The redundancy here avoids unnecessary dependencies on sale modules.
 
     _sql_constraints = [
-        ("positive_height", "CHECK(height>=0.0)", "Height must be positive"),
-        ("positive_width", "CHECK(width>=0.0)", "Width must be positive"),
-        ("positive_length", "CHECK(packaging_length>=0.0)", "Length must be positive"),
+        ("positive_height", "CHECK(height>=0)", "Height must be positive"),
+        ("positive_width", "CHECK(width>=0)", "Width must be positive"),
+        ("positive_length", "CHECK(packaging_length>=0)", "Length must be positive"),
         (
             "positive_max_weight",
-            "CHECK(max_weight>=0.0)",
+            "CHECK(max_weight>=0)",
             "Max Weight must be positive",
         ),
     ]
-    height = fields.Float("Height")
-    width = fields.Float("Width")
-    packaging_length = fields.Float("Length")
+    height = fields.Integer("Height")
+    width = fields.Integer("Width")
+    packaging_length = fields.Integer("Length")
 
     length_uom_id = fields.Many2one(
         "uom.uom",
--- a/product_packaging_dimension/tests/test_packaging_volume.py
+++ b/product_packaging_dimension/tests/test_packaging_volume.py
@@ -53,28 +53,28 @@ class TestPackagingVolumeCompute(TransactionCase):
         # initial UoM always in meters and Volume in m3, but with different dimensions.
 
         self.packaging.packaging_length = 10
-        self.packaging.height = 8.0
-        self.packaging.width = 10.8
+        self.packaging.height = 8
+        self.packaging.width = 10
         self.packaging.length_uom_id = self.uom_m
         self.packaging.volume_uom_id = self.uom_m3
         self.packaging._compute_volume()
-        self.assertEqual(864, self.packaging.volume)
+        self.assertEqual(800, self.packaging.volume)
 
         self.packaging2.packaging_length = 6.0
         self.packaging2.height = 14.0
-        self.packaging2.width = 1.2
+        self.packaging2.width = 1.0
         self.packaging2.length_uom_id = self.uom_m
         self.packaging2.volume_uom_id = self.uom_m3
         self.packaging2._compute_volume()
-        self.assertAlmostEqual(100.8, self.packaging2.volume)
+        self.assertEqual(84.0, self.packaging2.volume)
 
         self.packaging3.packaging_length = 100.0
-        self.packaging3.height = 50.5
-        self.packaging3.width = 80.0
+        self.packaging3.height = 50
+        self.packaging3.width = 80
         self.packaging3.length_uom_id = self.uom_m
         self.packaging3.volume_uom_id = self.uom_m3
         self.packaging3._compute_volume()
-        self.assertEqual(404000, self.packaging3.volume)
+        self.assertEqual(400000, self.packaging3.volume)
 
     def test_output_uom(self):
         # Tests with both different initial and volume UoMs.


legalsylvain (changes requested):
Hi. Thanks for your contribution. 
Here, i don't understand why you want to change float into integer. Could you elaborate ?

i-vyshnevska:
> Hi. Thanks for your contribution. Here, i don't understand why you want to change float into integer. Could you elaborate ?

fields are copied from the delivery addon, to be able to use this functionality without sale addon, but if the customer wants to use and delivery from core we may have an issue, these fields were integer on the previous version as well

legalsylvain:
Indeed.

It seems there was some error during the migration of this module.
The current implementation conflict with odoo delivery design : https://github.com/odoo/odoo/blob/14.0/addons/delivery/models/product_packaging.py#L17


Seems it has been introduced here : https://github.com/OCA/product-attribute/pull/850
@Kev-Roche : could you take a look on that topic ? 

It's quite annoying, because moving float to int on existing database will loose data precision.

product_packaging_dimension/migrations/14.0.1.0.0/pre-migrate.py
```diff
@@ -5,18 +5,9 @@
...
-        cr.execute(
-            """
-        UPDATE product_packaging
-        SET
-        packaging_length = packaging_length/1000,
-        height = height/1000,
-        width = width/1000,
-        """
-        )
+    # Rename lngth into packaging_length
```
simahawk, line 9: This script won't run on existing installations.
You should add a script for `14.0.2.0.0` (I'd go for major as this can be a breaking change).

simahawk:
> It's quite annoying, because moving float to int on existing database will loose data precision.

This should be tackled by switching the UOM to a lower scale (eg: 1.8m -> 180cm or 1800mm).
Not sure we can do this automatically for all cases in the migration step tho.

product_packaging_dimension/migrations/14.0.1.0.0/pre-migrate.py
```diff
@@ -5,18 +5,9 @@
 
 
 def migrate(cr, version):
-    # Rename lnght into packaging_length
-    if column_exists(cr, "product_packaging", "lnght"):
-        rename_column(cr, "product_packaging", "lnght", "packaging_length")
```
sebalix, line 10: There was an issue there, the column is named `lngth`, not `lnght` that's why we didn't get the values during the migration. We need to move this script to `migrations/14.0.2.0.0/pre-migrate.py` with the right column name (no need to keep the current one).
As by default the length UoM used is meter in v14, the script is probably right to divide the values by 1000 as it was in millimeters in v13. But our custom need is to keep millimeters in v14, should we use a environment variable to change the behavior of this script to not divide the values by 1000? This same environment variable will be used elsewhere where we were storing millimeters values (like in `stock_storage_type`, the migration needs to be checked there as well).
@i-vyshnevska @simahawk What do you think?

i-vyshnevska: @sebalix  I remember the story about "length" "lngth" and in core, it was renamed to "packagin_length" so there is no need to have "lngth" anymore

I changed migration in update 14.0.1.0.0 because it was done without considering of change int -> float and all data is lost so for projects migrated from now it should land for that step (data was lost for width, height as well)

for providing script to consider a change of accuracy with considering of uom - not all of uom are coming from core, for instance, mm defined in "stock_measuring_device" some can be set manually so add them can be tricky

sebalix: I wasn't talking about the `length` and `lngth` story, but the existing typo in the migration script where the last two letters were inverted: `lngth` vs `lnght`, so the script could never work as it couldn't pass the `if column_exists()` condition.

Now I still do not understand why we should delete the conversion of millimeters to meters in this OCA module.

- in v13 the `width` field was hosting millimeters value (no related UoM to change that)
- in v14 the business rule from Odoo is to enforce this kind of fields to meters thanks to the `product.template._get_length_uom_id_from_ir_config_parameter()`), and all OCA v14 addons are now based on this rule (see the different `length_uom_id` field in `stock.location.storage.type`, `product.packaging`, `stock.quant.package`...), so by default why not converting the existing millimeter values to meter? I know we need millimeters in our project but it's up to us to change that, in OCA we should stick to sane & standard defaults.

i-vyshnevska: - data was lost not only for lngth for other dimensions as well, so not only typo is the reason of data lost
- my idea is on v13 it was explicitly called mm for all 3 dimensions so why that should change
- as the field should stay integer we really will have losing of accuracy on width/1000 as if you have 200 mm  on converting to m in integer we will have 0 
- we don't have an mm uom in here and suppose we don't want to bring a dependency here

my point is to keep data untouched and the rest will be handled client by client in dependency on their configuration, which will just require a change of length_uom_id

sebalix: Ok it's clearer, thanks!
That means that currently in core Odoo we could only expressed dimensions of a packaging in meters or foots, with entire values? That's crazy, I miss something...

I also wanted to create a `uom_mm` addon few months ago for this kind of dependency (rejected here https://github.com/OCA/product-attribute/pull/914), together with [product_logistics_uom](https://github.com/OCA/product-attribute/tree/14.0/product_logistics_uom) it could have been possible to define the millimeter as the standard UoM for logistics. Having this UoM in `stock_measuring_device` is not its place, we now have to depend on this addon everywhere we need to reference `mm`... Anyway.

simahawk: > wanted to create a uom_mm addon few months ago for this kind of dependency (rejected here #914), 

We should probably reopen this discussion :wink:

product_packaging_dimension/migrations/14.0.1.0.0/pre-migrate.py
simahawk: what about this?

i-vyshnevska: I don't want any script ran on existing installation only migrated data 13->14
if the current version of the addon is installed, harm is already done, 
about the complexity of handling the current data, I pointed here 
https://github.com/OCA/product-attribute/pull/957#discussion_r767633335

simahawk (approved):

hparfr (changes requested):
Hi, 

@i-vyshnevska  I don't get your needs here. Float + uom allows a lot of usages.

In some industries, it make sense to express packages dimensions in cm with a decimals : 12,5cm or weight like 1,32kg because **users** are more comfortable with it, so less errors are introduced in the system.

>my idea is on v13 it was explicitly called mm for all 3 dimensions so why that should change

It has changed because uom were not managed consistently: sometime hardcoded in the view, xml in the label, sometime hardcoded in the fieldname, or in the helpmessage... It was not good enough for a minimum serious implementation. 
For instance, product_logistics_uom is born because you can't have a volume of 33cl on a product (you will get 0m³) in odoo core.

The actual implementation allow the use of the right uom. It's a good feature. 

In your business case, if you need to force the value of the uom_id you can do as @sebalix suggested:  change the ir_parameter. You can also change the field uom_id to a computed hardcoded value (in a custom module).


> my point is to keep data untouched and the rest will be handled client by client in dependency on their configuration, which will just require a change of length_uom_id

The conversion in the migration script was done in order to be aligned with the default values (so the default uom). It's better to have the whole values (old and new) aligned by default.


>I know we need millimeters in our project but it's up to us to change that, in OCA we should stick to sane & standard defaults.
:+1: 

>We should probably reopen this discussion 
:+1:

i-vyshnevska:
@hparfr  IMO change to float is unsafe as other addons rely on these fields for instance modules in enterprise comparing these values as integers without the usage of float_compare, plus they treating them as an integer so in case of 0.5 they will get a 0

and migration from int to float need to be done with a change of column type (in case if current PR will be blocked)

hparfr:
It's quite hard to guess issues with closed source entreprise.

hparfr (changes requested):
After more thinking about it,  I think it's acceptable to store in db values as integer + different uom.
Computed / Inversed fields can be created in the view to allow the user to enter / view values in float with their uom of choice and convert to a lower (smallest?) uom for db.

rousseldenis (approved):

rousseldenis:
@i-vyshnevska Is this ready ?

i-vyshnevska:
@rousseldenis  ATM I don't have what to add to this PR, as I mentioned before adding data migrating scripts is tricky here 
@hparfr suggestion for enabling float and inverse methods should be done in separate addon

sebalix (approved):

hparfr (approved):
Thanks @i-vyshnevska 

It's ready to be merged.

simahawk:
@i-vyshnevska can you squash?

@legalsylvain mind to update your review?

celm1990:
Hi
some update?. I waiting for merge this changes into migration to V15

celm1990 (approved):

PR #958, opened 2021-12-10
[IMP] add module for adding video on product and category

sebastienbeau:
This module add a new tab in product to be able to add video (same on category)

Note on category it will be great to add a basic module that add a notebook on category (so every module that add feature on category can add it in a cleaner way). I will propose an other PR for this

### DIFF
--- a/product_video_link/__manifest__.py
+++ b/product_video_link/__manifest__.py
@@ -1,0 +1,29 @@
+# Copyright 2021 Akretion (https://www.akretion.com).
+# @author Sébastien BEAU <sebastien.beau@akretion.com>
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+
+{
+    "name": "Product Video",
+    "summary": "Link Video on product and category",
+    "version": "14.0.1.0.0",
+    "category": "Product",
+    "website": "https://github.com/OCA/product-attribute",
+    "author": " Akretion, Odoo Community Association (OCA)",
+    "license": "AGPL-3",
+    "application": False,
+    "installable": True,
+    "external_dependencies": {
+        "python": [],
+        "bin": [],
+    },
+    "depends": [
+        "product",
+        "base_video_link",
+    ],
+    "data": [
+        "views/product_category_view.xml",
+        "views/product_template_view.xml",
+        "security/ir.model.access.csv",
+    ],
+    "demo": ["demo/product_demo.xml"],
+}
--- a/product_video_link/__init__.py
+++ b/product_video_link/__init__.py
@@ -1,0 +1,1 @@
+from . import models
--- a/product_video_link/demo/product_demo.xml
+++ b/product_video_link/demo/product_demo.xml
@@ -1,0 +1,11 @@
+<?xml version="1.0" encoding="UTF-8" ?>
+<odoo>
+
+    <record id="product.product_order_01_product_template" model="product.template">
+        <field
+            name="video_ids"
+            eval="[(6, 0, [ref('base_video_link.video_video_1')])]"
+        />
+    </record>
+
+</odoo>
--- a/product_video_link/models/__init__.py
+++ b/product_video_link/models/__init__.py
@@ -1,0 +1,2 @@
+from . import product_template
+from . import product_category
--- a/product_video_link/models/product_category.py
+++ b/product_video_link/models/product_category.py
@@ -1,0 +1,10 @@
+# Copyright 2021 Akretion (https://www.akretion.com).
+# @author Sébastien BEAU <sebastien.beau@akretion.com>
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+
+from odoo import models
+
+
+class ProductCategory(models.Model):
+    _inherit = ["product.category", "video.link.mixin"]
+    _name = "product.category"
--- a/product_video_link/models/product_template.py
+++ b/product_video_link/models/product_template.py
@@ -1,0 +1,10 @@
+# Copyright 2021 Akretion (https://www.akretion.com).
+# @author Sébastien BEAU <sebastien.beau@akretion.com>
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+
+from odoo import models
+
+
+class ProductTemplate(models.Model):
+    _inherit = ["product.template", "video.link.mixin"]
+    _name = "product.template"
--- a/product_video_link/readme/CONTRIBUTORS.rst
+++ b/product_video_link/readme/CONTRIBUTORS.rst
@@ -1,0 +1,1 @@
+* Sébastien Beau <sebastien.beau@akretion.com>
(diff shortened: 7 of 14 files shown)


clementmbr:
@sebastienbeau It looks like it could be interesting to separate the media_video code from the product_video code, don't you think so?
I got a client which could be interested in having Videos thumbnail on crm.lead for instance...

Thanks for the module anyway! :pray:

pedrobaeza:
Shouldn't this be `product_video_link` according to the feature?

product_video/models/media_video_provider.py
```diff
@@ -0,0 +1,15 @@
...
+from odoo import fields, models
+
+
+class MediaVideoProvider(models.Model):
+    _name = "media.video.provider"
+    _description = "Media Video Provider"
+
+    name = fields.Char()
+    code = fields.Char()
+    pattern_video_url = fields.Char(default="https://example.org/video/%s")
```
legalsylvain: all that 3 fields should be required, don't you think ?

product_video/models/media_video_provider.py
```diff
@@ -0,0 +1,15 @@
...
+
+from odoo import fields, models
+
+
+class MediaVideoProvider(models.Model):
+    _name = "media.video.provider"
+    _description = "Media Video Provider"
+
+    name = fields.Char()
+    code = fields.Char()
```
legalsylvain: what is the interest of the "code" field?

product_video/models/product_category.py
```diff
@@ -0,0 +1,11 @@
...
+# @author Sébastien BEAU <sebastien.beau@akretion.com>
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+
+from odoo import fields, models
+
+
+class ProductCategory(models.Model):
+    _inherit = "product.category"
+
+    video_ids = fields.One2many("media.video", "product_tmpl_id", "Video")
```
legalsylvain: replace ``product_tmpl_id`` by ``category_id``. don't you think ?

legalsylvain (changes requested):
That's look great ! 

2 General design question : 

A) what about to have : 
- ``media_video_base``  : a generic module that introduce ``media.video``, and ``media.video.provided``. and a mixin : ``media.video.mixin``, that has ``video_ids`` field. 
- then a module ``media_video_product`` that makes product.category and product.template inherits from ``media.video.mixin`` and overload views. (auto install false)
- we can then imagine other module  ``media_video_partner``, etc... depending on the business requirements. A good candidate is to have a ``media_video_event`` with the link of the video of conference.

B) did you thought to have a M2M fields between ``media.video`` and `` media.video.mixin`` (currently ``product.product`` for exemple) ? 
I mean, a video maybe can talk about many products, or many categories. Don't you think ?

regards.

product_video/views/media_video_provider_view.xml
```diff
@@ -0,0 +1,48 @@
...
+            <field name="name" />
+        </tree>
+    </field>
+</record>
+
+<record id="media_video_provider_view_form" model="ir.ui.view">
+    <field name="model">media.video.provider</field>
+    <field name="arch" type="xml">
+        <form string="Video Provider">
+            <field name="name" />
```
legalsylvain: other fields are missing.

product_video/views/media_video_provider_view.xml
```diff
@@ -0,0 +1,48 @@
...
+</record>
+
+<record model="ir.actions.act_window" id="media_video_provider_action">
+    <field name="name">Video Provider</field>
+    <field name="type">ir.actions.act_window</field>
+    <field name="res_model">media.video.provider</field>
+    <field name="view_mode">tree,form</field>
+    <field name="search_view_id" ref="media_video_provider_view_search" />
+    <field name="domain">[]</field>
+    <field name="context">{}</field>
```
legalsylvain: line 36-38 are useless. (default values).

product_video/views/media_video_provider_view.xml
```diff
@@ -0,0 +1,48 @@
...
+</record>
+
+<menuitem
+        id="media_video_provider_menu"
+        parent="sale.prod_config_main"
+        sequence="20"
+        action="media_video_provider_action"
+    />
```
legalsylvain: if a more generic approach is used, we could imagine an extra new menu item, named video, with all the videos, and the providers.
so, no need to depends on sale module.

product_video/demo/media_video_demo.xml
```diff
@@ -0,0 +1,11 @@
+<?xml version="1.0" encoding="UTF-8" ?>
+<odoo>
+
+<record id="product_video" model="media.video">
+    <field name="provider_id" ref="media_video_provider_youtube" />
+    <field name="name">Talk</field>
+    <field name="video_idendifier">6JByIDOnuv8</field>
```
legalsylvain: did you meant AKlQuetlyn4 ? ;-)

product_video/models/media_video_provider.py
sebastienbeau: It's a technical name, for example if you can use it in your front end (shopinavder, odoo, or whatever) to use the right wigdet

product_video/demo/media_video_demo.xml
sebastienbeau: héhé :)

sebastienbeau: maybe : sTbTzINcj1w

product_video/models/product_category.py
sebastienbeau: fixed as we use a m2m

product_video/models/media_video_provider.py
sebastienbeau: done

product_video/models/media_video_provider.py
legalsylvain: Why not a sélection better ? YouTube / vimeo / peertube / etc ?
It could also allow in a second time to add an onchange on url video to have the possibility to get the name and the thumbnail of the video. Depending on the technology.
(Not a blocking point)

legalsylvain (approved):
Thanks.

product_video/models/media_video_provider.py
sebastienbeau: the idea is be able to create new provider without customizing the code, so all provider can be added with data

simahawk:
pls check https://github.com/OCA/server-tools/pull/2230

sebastienbeau:
@simahawk done

product_video_link/__manifest__.py
```diff
@@ -0,0 +1,29 @@
...
+    "category": "Product",
+    "website": "https://github.com/OCA/product-attribute",
+    "author": " Akretion, Odoo Community Association (OCA)",
+    "license": "AGPL-3",
+    "application": False,
+    "installable": True,
+    "external_dependencies": {
+        "python": [],
+        "bin": [],
+    },
```
simahawk, line 18:
```suggestion
```

simahawk (approved):

PR #961, opened 2021-12-14
[14.0] [FIX] product_state

sonhd91:
* Fix bug when inverting non-exist new state
* Show default and active fields in product state form
* Refactor inverse state to use with variant state

### DIFF
--- a/product_state/models/product_template.py
+++ b/product_state/models/product_template.py
@@ -23,7 +23,7 @@ class ProductTemplate(models.Model):
         help="Select a state for this product",
         group_expand="_read_group_state_id",
         inverse="_inverse_product_state_id",
-        default=lambda self: self._get_default_product_state(),
+        default=lambda self: self._get_default_product_state().id,
         index=True,
         tracking=10,
     )
--- a/product_state/models/product_template.py
+++ b/product_state/models/product_template.py
@@ -1,7 +1,8 @@
 # Copyright 2017-2021 ACSONE SA/NV (<http://acsone.eu>)
 # License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl.html).
 
-from odoo import api, fields, models
+from odoo import _, api, fields, models
+from odoo.exceptions import UserError
 
 
 class ProductTemplate(models.Model):
@@ -22,7 +23,7 @@ class ProductTemplate(models.Model):
         help="Select a state for this product",
         group_expand="_read_group_state_id",
         inverse="_inverse_product_state_id",
-        default=lambda self: self._get_default_product_state_id(),
+        default=lambda self: self._get_default_product_state(),
         index=True,
         tracking=10,
     )
@@ -34,8 +35,8 @@ class ProductTemplate(models.Model):
         """
 
     @api.model
-    def _get_default_product_state_id(self):
-        return self.env["product.state"].search([("default", "=", True)], limit=1).id
+    def _get_default_product_state(self):
+        return self.env["product.state"].search([("default", "=", True)], limit=1)
 
     @api.depends("product_state_id")
     def _compute_product_state(self):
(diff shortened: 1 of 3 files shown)


product_state/models/product_template.py
```diff
@@ -35,22 +35,26 @@ def _inverse_product_state_id(self):
 
     @api.model
     def _get_default_product_state_id(self):
-        return self.env["product.state"].search([("default", "=", True)], limit=1).id
+        return self.env["product.state"].search([("default", "=", True)], limit=1)
```
simahawk, line 39: bad naming of the method then :)

product_state/models/product_template.py
```diff
@@ -35,22 +35,26 @@ def _inverse_product_state_id(self):
...
             product_tmpl.state = product_tmpl.product_state_id.code
 
     def _inverse_product_state(self):
-        ProductState = self.env["product.state"]
         for product_tmpl in self:
-            product_state = ProductState.search(
-                [("code", "=", product_tmpl.state)], limit=1
+            self._set_product_state_id(product_tmpl)
+
+    def _set_product_state_id(self, record):
```
simahawk:
```suggestion
            product_tmpl._set_product_state_id()

    def _set_product_state_id(self):
```

product_state/models/product_template.py
```diff
@@ -35,22 +35,26 @@ def _inverse_product_state_id(self):
...
+        """ The record param is for similar state field at product.product model. """
+        ProductState = record.env["product.state"]
+        product_state = ProductState.search([("code", "=", record.state)], limit=1)
+        if record.state and not product_state:
+            product_state = ProductState.create(
```
simahawk: I wouldn't do this :thinking:

simahawk (changes requested):
I don't get what's the point of creating a state on the fly

product_state/views/product_template.xml
```diff
@@ -57,6 +57,8 @@
                     </group>
                     <group string="Description">
                         <field name="description" nolabel="1" />
+                        <field name="default" widget="boolean_toggle" />
+                        <field name="active" widget="boolean_toggle" />
                     </group>
```
i-vyshnevska, line 62: pls add a fix on line 26
` <field name="domain">[('id', '=', product_state_id)]</field>`

product_state/models/product_template.py
sonhd91: As we are removing the method from `product_tier_validation` (https://github.com/OCA/product-attribute/pull/962), we can rename it.

product_state/models/product_template.py
sonhd91: The method is being used by both `product.template` and `product.product` so `record` is needed. It is now promoted to `api.model` level.

product_state/models/product_template.py
sonhd91: Thanks for the advice, it has been fixed.

product_state/views/product_template.xml
sonhd91: Thanks for the advice, it has been fixed.

hailangvn:
Thanks @simahawk  and @i-vyshnevska for your review. The PR has been updated.

product_state/models/product_template.py
```diff
@@ -22,7 +23,7 @@ class ProductTemplate(models.Model):
         help="Select a state for this product",
         group_expand="_read_group_state_id",
         inverse="_inverse_product_state_id",
-        default=lambda self: self._get_default_product_state_id(),
+        default=lambda self: self._get_default_product_state(),
```
i-vyshnevska:
```suggestion
        default=lambda self: self._get_default_product_state().id,
```

i-vyshnevska (approved):

simahawk (approved):

PR #962, opened 2021-12-14
[14.0][IMP] product_tier_validation: get default from product_state

hailangvn:
- With updated product_state, default state can be get instead of hard code.
- The code `confirmed` does not match with state name `Sellable`. It
  does little value in code. So it can be removed to avoid failure in
  tests of other modules.

### DIFF
--- a/product_tier_validation/data/product_state_data.xml
+++ b/product_tier_validation/data/product_state_data.xml
@@ -1,8 +1,21 @@
 <odoo noupdate="1">
-    <record id="product_state.product_state_sellable" model="product.state">
-        <field name="code">confirmed</field>
-    </record>
-    <record id="product_state.product_state_end" model="product.state">
+    <record id="product_state_cancel" model="product.state">
         <field name="code">cancel</field>
+        <field name="name">Cancelled</field>
+        <field name="sequence">50</field>
     </record>
+
+    <function name="write" model="product.state">
+        <function name="search" model="product.state">
+            <value eval="[('default', '=', True), ('active', 'in', [False, True])]" />
+        </function>
+        <value eval="{'default': False}" />
+    </function>
+
+    <function name="write" model="product.state">
+        <function name="search" model="product.state">
+            <value eval="[('code', '=', 'draft'), ('active', 'in', [False, True])]" />
+        </function>
+        <value eval="{'active': True, 'default': True}" />
+    </function>
 </odoo>
--- a/product_tier_validation/models/product_template.py
+++ b/product_tier_validation/models/product_template.py
@@ -28,9 +28,3 @@ class ProductTemplate(models.Model):
         if "state" in vals:
             vals["active"] = self._is_tier_validated_active(vals["state"])
         return super().write(vals)
-
-    @api.model
-    def _get_default_product_state_id(self):
-        return self.env.ref(
-            "product_state.product_state_draft", raise_if_not_found=False
-        )
--- a/product_tier_validation/tests/test_product_tier_validation.py
+++ b/product_tier_validation/tests/test_product_tier_validation.py
@@ -34,6 +34,7 @@ class TestProductTierValidation(common.SavepointCase):
                 "list_price": 120.00,
             }
         )
+        self.assertEqual(product.state, "draft")
         product.request_validation()
         product.with_user(self.test_user_1).validate_tier()
         product.write({"product_state_id": self.normal_state.id})


product_tier_validation/models/product_template.py
```diff
@@ -28,9 +28,3 @@ def write(self, vals):
         if "state" in vals:
             vals["active"] = self._is_tier_validated_active(vals["state"])
         return super().write(vals)
-
-    @api.model
-    def _get_default_product_state_id(self):
-        return self.env.ref(
-            "product_state.product_state_draft", raise_if_not_found=False
```
simahawk, line 35: where is the draft state then? This module should add it as default, no?

product_tier_validation/data/product_state_data.xml
```diff
@@ -1,8 +1,7 @@
 <odoo noupdate="1">
-    <record id="product_state.product_state_sellable" model="product.state">
```
simahawk, line 2: is `confirmed` a state this module expects?

simahawk (commented):
@dreispt can we have your feedback on `product_tier_validation` change?

product_state/__manifest__.py
```diff
@@ -17,6 +17,7 @@
         "views/product_template.xml",
     ],
     "application": False,
+    "installable": False,
```
i-vyshnevska: as 961 is merged can we set it back to True

test-requirements.txt
```diff
@@ -1 +1,2 @@
 odoo_test_helper
+odoo14-addon-product_state @git+https://github.com/OCA/product-attribute@refs/pull/961/head#subdirectory=setup/product_state
```
i-vyshnevska: no need anymore

product_tier_validation/data/product_state_data.xml
hailangvn: I could not found any place that uses product code `confirmed` (https://github.com/OCA/product-attribute/search?q=confirmed&type=code). AFAIK, `code` is used by py code or xml. With existing dbs those have the module installed, join between `product.template` and `product.state` is based on id, not on code. In theory, we should not change code.

product_tier_validation/models/product_template.py
hailangvn: Thanks for spotting that. I am updating the `data/product_state_data.xml`.

hailangvn:
Thanks @simahawk and @i-vyshnevska for your review. This PR has been updated.

product_tier_validation/data/product_state_data.xml
```diff
@@ -1,8 +1,30 @@
...
+        <field name="name">Cancelled</field>
+        <field name="sequence">50</field>
+    </record>
+
+    <function name="write" model="ir.model.data">
+        <function name="search" model="ir.model.data">
+            <value eval="[('module', '=', 'product_state')]" />
+        </function>
+        <value eval="{'noupdate': False}" />
+    </function>
```
simahawk: instead of 3 writes can't you just write default=False on the matching record?
Not a blocker ;)

simahawk (approved):

product_tier_validation/data/product_state_data.xml
```diff
@@ -1,8 +1,30 @@
 <odoo noupdate="1">
+    <record id="product_state_cancel" model="product.state">
```
rousseldenis, line 2: I'm always wondering why adding new states here ?

Seems to be project-specific ...

product_tier_validation/data/product_state_data.xml
```diff
@@ -1,8 +1,30 @@
...
+    </function>
+
     <record id="product_state.product_state_sellable" model="product.state">
-        <field name="code">confirmed</field>
+        <field name="default" eval="False" />
     </record>
-    <record id="product_state.product_state_end" model="product.state">
-        <field name="code">cancel</field>
+
+    <record id="product_state.product_state_draft" model="product.state">
```
rousseldenis: same as above. People has to configure on their instances.

rousseldenis:
@hailangvn @simahawk Could you please review this instead ?

https://github.com/OCA/product-attribute/pull/951/files

product_tier_validation/data/product_state_data.xml
rousseldenis: Ok, to fit 'cancel' behaviour from base_tier_validation

product_tier_validation/data/product_state_data.xml
hailangvn: Yes, it is their responsibility. On other hand we can utilize existing records to help them understand more about logic.

product_tier_validation/data/product_state_data.xml
hailangvn: Thanks for the great advice! The PR has been updated.

i-vyshnevska (approved):

hailangvn:
* Cancelled as https://github.com/OCA/product-attribute/pull/951 has been merged.

PR #968, opened 2021-12-17
[14.0][IMP] product_tier_validation: Take into account state Many2one

rousseldenis:
Depends on:

- [ ] https://github.com/OCA/server-ux/pull/421

Remove:

- [ ] test-requirements.txt entry

### DIFF
--- a/product_tier_validation/models/product_template.py
+++ b/product_tier_validation/models/product_template.py
@@ -8,3 +8,34 @@ class ProductTemplate(models.Model):
     _name = "product.template"
     _inherit = ["product.template", "tier.validation"]
     _tier_validation_manual_config = False
+
+    def _check_state_conditions(self, vals):
+        """
+        As product state write is done on product_state_id field if
+        changed through interface. We need to get the correspondong state
+        """
+        self.ensure_one()
+        if "product_state_id" in vals:
+            new_state = self.env["product.state"].browse(vals["product_state_id"]).code
+            return (
+                getattr(self, self._state_field) in self._state_from
+                and new_state in self._state_to
+            )
+        else:
+            return super()._check_state_conditions(vals)
+
+    def _check_tier_state_transition(self, vals):
+        """
+        As product state write is done on product_state_id field if
+        changed through interface. We need to get the correspondong state
+        """
+        self.ensure_one()
+        if "product_state_id" in vals:
+            new_state = self.env["product.state"].browse(vals["product_state_id"]).code
+            return getattr(
+                self, self._state_field
+            ) in self._state_from and new_state not in (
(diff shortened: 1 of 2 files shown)


rousseldenis:
@simahawk

hailangvn:
I am wondering what if we change https://github.com/OCA/product-attribute/blob/14.0/product_state/views/product_template.xml#L134 to L138 as below.
```
-                    name="product_state_id"
+                    name="state"
                     widget="statusbar"
                     options="{'clickable': '1'}"
                 />
-                <field name="state" invisible="1" />
```

hailangvn:
Updated: as state is `fields.Char`, widget will not work with [this change](https://github.com/OCA/product-attribute/pull/968#issuecomment-997087916).

product_tier_validation/models/product_template.py
```diff
@@ -8,3 +8,34 @@ class ProductTemplate(models.Model):
...
+        self.ensure_one()
+        if "product_state_id" in vals:
+            new_state = self.env["product.state"].browse(vals["product_state_id"]).code
+            return getattr(
+                self, self._state_field
+            ) in self._state_from and new_state not in (
+                self._state_to + [self._cancel_state]
+            )
+        else:
+            return super()._check_tier_state_transition(vals)
```
hailangvn, line 41: With the improvement at https://github.com/hailangvn/server-ux/pull/1, we can rewrite this like below.
```python
    def _get_state_from(self, vals):
        return self.env["product.state"].browse(vals["product_state_id"]).code
```

product_tier_validation/models/product_template.py
```diff
@@ -8,3 +8,34 @@ class ProductTemplate(models.Model):
     _name = "product.template"
     _inherit = ["product.template", "tier.validation"]
     _tier_validation_manual_config = False
+
```
hailangvn, line 11: @simahawk, I found where `confirmed` is used.
@rousseldenis, Should we add this?
```python
    _state_to = ["sellable"]
```

product_tier_validation/models/product_template.py
simahawk: wouldn't be better something like:

```py
[...]
state_code = fields.Char(related="product_state_id.code")`

_state_field = "state_code"
_state_from = ["draft"]
_state_to = ["sellable"]
[...]
```

and we don't override anything?

product_tier_validation/models/product_template.py
rousseldenis: @hailangvn @simahawk Yes indeed. I first did it in my project to confirm flow.

As 'confirmed' state comes from base_tier_validation, we need to override it here

PR #975, opened 2021-12-30
[14.0][IMP] product_code_unique: Allow to inactivate products with same code

rousseldenis:
For reason X or Y (e.g. wrong unit of measure), it would be convenient to archive a product and create a new one with same default_code.


legalsylvain (approved):

pedrobaeza (changes requested):
I don't agree with this change. If such case happens, you should empty the code in the product before doing the other. If not, you may miss the opportunity to reactivate a product with such code because no block is done. You can improve the constraint message to indicate to see also on inactive products.

PR #977, opened 2022-01-03
[FIX] product_pricelist_supplierinfo: override min quantity

chienandalu:
If we don't have a supplierinfo that below 1 unit we won't get any
seller. We want to ensure that the condition is fully ignored to get
whatever price given the other criterias.

cc @Tecnativa TT33659

What do you think @pedrobaeza ?

### DIFF
--- a/product_pricelist_supplierinfo/models/product_product.py
+++ b/product_pricelist_supplierinfo/models/product_product.py
@@ -9,6 +9,17 @@ from odoo import models
 class ProductProduct(models.Model):
     _inherit = 'product.product'
 
+    def _prepare_sellers(self, params):
+        """When we override min qty we want that _select_sellers gives us the
+        first possible seller for every other criteria ignoring the quantity. As
+        supplierinfos are sorted by min_qty descending, we want to revert such
+        order so we get the very first one, which is probably the one to go.
+        """
+        sellers = super()._prepare_sellers(params)
+        if self.env.context.get("override_min_qty"):
+            sellers = sellers.sorted("min_qty")
+        return sellers
+
     def _get_supplierinfo_pricelist_price(
             self, rule, date=None, quantity=None):
         return self.product_tmpl_id._get_supplierinfo_pricelist_price(
--- a/product_pricelist_supplierinfo/models/product_template.py
+++ b/product_pricelist_supplierinfo/models/product_template.py
@@ -19,13 +19,17 @@ class ProductTemplate(models.Model):
         if product_id:
             product = product.browse(product_id)
         if rule.no_supplierinfo_min_quantity:
-            quantity = 1.0
+            # No matter which minimum qty, we'll get every seller. We set a
+            # number absurdidly high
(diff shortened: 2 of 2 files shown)


sergio-teruel (approved):

carlosdauden (approved):

product_pricelist_supplierinfo/models/product_product.py
```diff
@@ -9,6 +9,17 @@
...
+    def _prepare_sellers(self, params):
+        """When we override min qty we want that _select_sellers gives us the
+        first possible seller for every other criteria ignoring the quantity. As
+        supplierinfos are sorted by min_qty descending, we want to revert such
+        order so we get the very first one, which is probably the one to go.
+        """
+        sellers = super()._prepare_sellers(params)
+        if self.env.context.get("override_min_qty"):
+            sellers = sellers.sorted("min_qty")
+        return sellers
```
carlosdauden, line 21: Maybe this code is not necessary as it exits the loop on the first match:
https://github.com/odoo/odoo/blob/5ab778227dfbc93b2ff415ebb0e40ce35953f13e/addons/product/models/product.py#L594
Luckily the default order is as expected:
https://github.com/odoo/odoo/blob/f5751e24be857d5be6969d85225677d7da5f0baa/addons/product/models/product.py#L740

product_pricelist_supplierinfo/models/product_template.py
```diff
@@ -19,13 +19,17 @@ def _get_supplierinfo_pricelist_price(
...
+            quantity = 1e9
         # The product_variant_id returns empty recordset if template is not
         # active, so we must ensure variant exists or _select_seller fails.
         if product:
             if type(date) == datetime:
                 date = date.date()
-            seller = product._select_seller(
+            seller = product.with_context(
+                override_min_qty=rule.no_supplierinfo_min_quantity
+            )._select_seller(
```
carlosdauden, line 32: Remove this if remove the one above

carlosdauden (changes requested):
With this changes the returned price will be the cheapest:
Sorted don't work because exit the loop on the first match:
https://github.com/odoo/odoo/blob/5ab778227dfbc93b2ff415ebb0e40ce35953f13e/addons/product/models/product.py#L594
The default order is min_qty **DESC**:
https://github.com/odoo/odoo/blob/f5751e24be857d5be6969d85225677d7da5f0baa/addons/product/models/product.py#L740

chienandalu:
Yes, that's the purpose. <strike>To get the cheapest one as the main price.</strike> To get the smallest min qty as the supplierinfo reference

pedrobaeza:
Is it then this completed to be merged?

carlosdauden (approved):

carlosdauden:
Ready to merge

pedrobaeza:
Can you please fw-port it to v13 and prepare a test case for avoiding regressions?

PR #983, opened 2022-01-13
[ADD] product_category_supplier module

Kev-Roche:
Add supplier Type on product Categories and supplier_categ_id on product template

### DIFF
--- a/product_category_supplier/__init__.py
+++ b/product_category_supplier/__init__.py
@@ -1,0 +1,1 @@
+from . import models
--- a/product_category_supplier/__manifest__.py
+++ b/product_category_supplier/__manifest__.py
@@ -1,0 +1,20 @@
+# Copyright 2022 Akretion (https://www.akretion.com).
+# @author Kévin Roche <kevin.roche@akretion.com>
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+
+{
+    "name": "Product Category Supplier",
+    "summary": "Add supplier Type and supplier_categ_id on product Categories",
+    "version": "14.0.1.0.0",
+    "website": "https://github.com/OCA/product-attribute",
+    "author": "Akretion, Odoo Community Association (OCA)",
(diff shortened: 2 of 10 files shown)


legalsylvain (commented):
Hi kev. Could you elaborate the use case? Not sure to understand what is the supplier category.

Thanks!

Kev-Roche:
@legalsylvain 
The main idea is to separate the sale products (by the current company) and the supplier's products (purchased by the current company) categories. IMO, the most regular cases will be 1) if the company build new products with purchase components/products  and 2) in case of multi-company and multi-supplier infos, it will propose specific product categories for external uses. Hope to be clear, @sebastienbeau have more info than me

legalsylvain:
thanks for your explanation.
As far as I understand, it is usefull in a MRP context so ? 
if we just buy and sell products does this modules makes sense ?
@sebastienbeau : any extra comments welcome.

sebastienbeau:
Hi the customer case is the following.
The customer import a huge catalog of product from his suppliers (Some supplier have API so odoo automatically import all the product)
We want to store the categories of the supplier (to be able know in which supplier categ is the product), and them we will need to add a logic of mapping (if a new product of the supplier is create in categ "Y" we want to push this product in the internal categ "A"
So the first step is to know the supplier categ and to be able to store this data in odoo

legalsylvain:
Thanks both for your answer. ! 

A general question : 
- Why did you put the category in the ``product.template`` level ? and not in the ``product.supplierinfo`` level ? 
I usually have products sold by many suppliers and categories are decided by each supplier.
- In the same spirit maybe we should put a ``partner_id`` for the supplier in the ``product.category`` to filter categories of the current supplier.

What do you think ? 
kind regards.

Rad0van:
Wouldn't using https://github.com/OCA/product-attribute/tree/14.0/product_multi_category solve the problem in a bit different way?

PR #984, opened 2022-01-14
[13.0][IMP]product_attribute_archive: Added Attribute Values view

GuillemCForgeFlow:
@ForgeFlow
Added new menu item that will be displaying all the Attribute Values in it, also it will have a default group by **attribute**.


AaronHForgeFlow:
In my opinion this deserves a new module,  because the menus are a base for other modules to build on top.

GuillemCForgeFlow:
> In my opinion this deserves a new module, because the menus are a base for other modules to build on top.

I think it's a great idea! Thanks :)

PR #987, opened 2022-01-20
[FIX] [14.0] [BUG] product_supplierinfo_for_customer - variant select…

ilyasProgrammer:
…ion not possible #986

PR #990, opened 2022-01-21
[IMP] Make module compatible with product configurator and variant ma…

ilyasProgrammer:
…trix

Need to merge #987 first


GSLabIt (commented):
I think it is better to move to https://github.com/OCA/sale-workflow/tree/14.0/product_supplierinfo_for_customer_sale

ilyasProgrammer:
Moved to https://github.com/OCA/sale-workflow/pull/1887

PR #991, opened 2022-01-21
[13.0][IMP] product_pricelist_direct_print: Improve inheritance at excel + add new fields

CarlosRoca13:
cc @Tecnativa TT34038

ping @carlosdauden @sergio-teruel

### DIFF
--- a/product_pricelist_direct_print/i18n/es.po
+++ b/product_pricelist_direct_print/i18n/es.po
@@ -8,8 +8,8 @@ msgid ""
 msgstr ""
 "Project-Id-Version: Odoo Server 9.0c\n"
 "Report-Msgid-Bugs-To: \n"
-"POT-Creation-Date: 2021-07-28 11:51+0000\n"
-"PO-Revision-Date: 2021-07-28 13:54+0200\n"
+"POT-Creation-Date: 2022-01-21 15:22+0000\n"
+"PO-Revision-Date: 2022-01-21 16:26+0100\n"
 "Last-Translator: Carlos <carlos.roca@tecnativa.com>\n"
 "Language-Team: Spanish (https://www.transifex.com/oca/teams/23907/es/)\n"
 "Language: es\n"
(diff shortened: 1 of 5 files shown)


sergio-teruel (approved):

product_pricelist_direct_print/i18n/es.po
```diff
@@ -95,7 +95,7 @@ msgstr "<strong>Descripción</strong>"
 #. module: product_pricelist_direct_print
 #: model_terms:ir.ui.view,arch_db:product_pricelist_direct_print.report_product_pricelist_document
 msgid "<strong>List Price</strong>"
-msgstr "<strong>Precio tarifa</strong>"
+msgstr "<strong>Precio de venta</strong>"
```
carlosdauden: What's wrong with the current term?

product_pricelist_direct_print/i18n/es.po
```diff
@@ -262,7 +278,7 @@ msgstr "Última actualización en"
 #: code:addons/product_pricelist_direct_print/report/product_pricelist_xlsx.py:0
 #, python-format
 msgid "List Price"
-msgstr "Precio de coste"
+msgstr "Precio de venta"
```
carlosdauden: This could also be translated as "Precio tarifa"

product_pricelist_direct_print/i18n/es.po
pedrobaeza: In any moment, it was "Precio de coste", and I suggest this as better term.

carlosdauden (approved):
Non-blocking comments

product_pricelist_direct_print/i18n/es.po
pedrobaeza: Thinking twice, it's true that this field is showing the specific price of this pricelist, so it can be more accurate to say "tarifa", but the proper Spanish form is `Precio de tarifa` (with `de`). The other thing is not correct, although tolerated for abbreviating things. Also calling it simply `Precio` can be another option. Which one do you prefer?

carlosdauden: We can show 3 distinct prices with this module:
https://github.com/OCA/product-attribute/blob/8a43eda8f5c47fc8d7c68714bd146f21c55dc11a/product_pricelist_direct_print/views/report_product_pricelist.xml#L57-L65
https://github.com/OCA/product-attribute/blob/8a43eda8f5c47fc8d7c68714bd146f21c55dc11a/product_pricelist_direct_print/views/report_product_pricelist.xml#L84-L94
So I prefer "Precio tarifa" without "de" to get a little abbreviation.

PR #994, opened 2022-01-24
[FIX] product_dimension: Always make dimensions visible in product.product

carmenbianca:
For reasons that are a little beyond me, dimensions are not visible on product variants. Explicitly setting invisible as false makes them visible.

There probably is a better way to do this.

Although I have only tested this against 12.0, a quick glance at 14.0 shows that the issue might exist on that branch as well.


carmenbianca:
Some testing went wrong somewhere; ignore this PR :)

PR #1005, opened 2022-01-31
[14.0] fix base_product_mass_addition

damdam-s:


damdam-s:
cc @kevinkhao @PierrickBrun

PierrickBrun:
Hi @damdam-s , sorry but I don't understand what's fixed with this commit. Could you explain the before/after ? Thanks

damdam-s:
**context :** 
sale_quick (https://github.com/OCA/sale-workflow/pull/1863)
adding 2 products from an emply SO from the dedicated product view

**before:**

- adding product1 is OK and correctly plays onchange to have discount for example
- adding product2 is NOK because it updates the qty/discount/unit_price/... of the first created line

**after:**

- adding product1 is OK and correctly plays onchange to have discount for example for the first line
- adding product2 is OK and correctly plays onchange to have discount for example for the second line

damdam-s:
cc @gurneyalex

rousseldenis (changes requested):
@damdam-s Could you add a test that illustrates this change  ? Thanks

gurneyalex (approved):

legalsylvain:
Hi @damdam-s, could you add a test, as proposed by @rousseldenis ? 

thanks !

PR #1006, opened 2022-02-01
[13.0][FIX] product_abc_classfication: fix _compute_abc_classification method using the accumulated value

albariera:
The _compute_abc_classification was using the value_percentage without being accumulated. The classification was not done correctly because it was very difficult to obtain a product classified on level A. It only happened when we had a product with a consumption_value much higher than the other products.
In this new algorithm, the accumulated value_percentage is used instead, so the classification is done at all levels ( when a lot of products are used).

I have also added a condition to better classify the products that are near the thresholds depending on if they are nearer the upper or lower threshold. But if you consider that it is not necessary (because it affects in fact few products), I can remove it.  In case it was removed, it would be assigned always the higher level ( the one that needs more inventory control).

### DIFF
--- a/product_abc_classification/models/abc_classification_profile.py
+++ b/product_abc_classification/models/abc_classification_profile.py
@@ -138,19 +138,21 @@ class ABCClasificationProfile(models.Model):
             levels = profile.level_ids.sorted(
                 key=_get_sort_key_percentage, reverse=True
             )
-            percentages = levels.mapped("percentage")
-            level_percentage = list(zip(levels, percentages))
+            level_percentage = [[level, level.percentage] for level in levels]
+            current_value = 0
+            accumulated_percentage = level_percentage[0][1]
             for product_data in product_list:
-                product_data["value_percentage"] = (
-                    (100.0 * product_data["value"] / totals["value"])
-                    if totals["value"]
-                    else 0.0
-                )
+                # Accumulated current value
+                current_value += product_data["value"] or 0.0
+                # This comparison would be the same as:
+                # current_value * 100 / totals["value"] > accumulated_percentage,
+                # but it is written in the next way to avoid division and decimal lost.
                 while (
-                    product_data["value_percentage"] < level_percentage[0][1]
+                    current_value * 100 > accumulated_percentage * totals["value"]
                     and len(level_percentage) > 1
                 ):
                     level_percentage.pop(0)
+                    accumulated_percentage += level_percentage[0][1]
                 product_data["product"].abc_classification_level_id = level_percentage[
                     0
                 ][0]


etobella:
Thanks @albariera!

@MiquelRForgeFlow as original creator of the code, Can we kindly ask for your review? :pray:

product_abc_classification/models/abc_classification_profile.py
```diff
@@ -140,17 +140,39 @@ def _get_sort_key_percentage(rec):
...
+                    # The first part of the condition is just the mean of the previous and
+                    # current value, but written in this way:
+                    # (Pv + Cv)/2  = Cv - (Cv - Pv)/2.
+                    # Where Cv- Pv is the value of the product.
+                    if (
+                        current_value - (product_data["value"] or 0.0) / 2
+                    ) * 100 > accumulated_percentage * totals["value"]:
+                        product_data[
+                            "product"
+                        ].abc_classification_level_id = level_percentage[1][0]
```
MiquelRForgeFlow:
```suggestion
```
this is not needed if you apply the other two changes.

product_abc_classification/models/abc_classification_profile.py
```diff
@@ -140,17 +140,39 @@ def _get_sort_key_percentage(rec):
...
                     level_percentage.pop(0)
-                product_data["product"].abc_classification_level_id = level_percentage[
-                    0
-                ][0]
+                    accumulated_percentage += level_percentage[0][1]
```
MiquelRForgeFlow, line 155:
```suggestion
                    accumulated_percentage += level_percentage[0][1]
                product_data["product"].abc_classification_level_id = level_percentage[
                    0
                ][0]
```

product_abc_classification/models/abc_classification_profile.py
```diff
@@ -140,17 +140,39 @@ def _get_sort_key_percentage(rec):
...
+            accumulated_percentage = level_percentage[0][1]
             for product_data in product_list:
-                product_data["value_percentage"] = (
-                    (100.0 * product_data["value"] / totals["value"])
-                    if totals["value"]
-                    else 0.0
-                )
+                product_data["product"].abc_classification_level_id = level_percentage[
+                    0
+                ][0]
```
MiquelRForgeFlow:
```suggestion
```

MiquelRForgeFlow (commented):
> The _compute_abc_classification was using the value_percentage without being accumulated. The classification was not done correctly

Right, my bad. Thanks for the fix. I don't know why it was not detected before when my old PR was opened a lot of time and had some reviewers. Sorry for my part, but when I did my PR, I remember testing a bit, but not enough.

I propose some changes to reduce code. I have fully tested them, and also replicated the exact example of the eazystock doc in an empty database.

MiquelRForgeFlow:
Once merged, I will do another PR to apply some other improvements that everybody will like.

product_abc_classification/models/abc_classification_profile.py
```diff
@@ -140,17 +140,39 @@ def _get_sort_key_percentage(rec):
             )
             percentages = levels.mapped("percentage")
             level_percentage = list(zip(levels, percentages))
```
MiquelRForgeFlow:
```suggestion
            level_percentage = [[level, level.percentage] for level in levels]
```
more cleanup

product_abc_classification/models/abc_classification_profile.py
albariera: In the case of the easystock your code and mine do the exact classification, you are right. But this extra code that I added was thought for the case where a product had an accumulated percentage slightly over the threshold. So maybe, it would be better classified at the previous level. For example, imagine this situation: 
Captura de pantalla de 2022-02-02 09-56-43
As you can see the results are a little different. However, I agree with you that your solution is better because in a real database with a lot of products, the difference between both codes is insignificant and there is no need to complicate it that much.

product_abc_classification/models/abc_classification_profile.py
albariera: Done!

product_abc_classification/models/abc_classification_profile.py
albariera: :+1:

product_abc_classification/models/abc_classification_profile.py
albariera: Done!

albariera: :+1:

albariera:
> Once merged, I will do another PR to apply some other improvements that everybody will like.

Great! Looking forward to it! Thank you :blush:

MiquelRForgeFlow (approved):

product_abc_classification/models/abc_classification_profile.py
MiquelRForgeFlow: Oh, thanks for your explanation and comprehension :)

AaronHForgeFlow (approved):
LGTM

PR #1010, opened 2022-02-02
[13.0][IMP] product_abc_classification: several improvements

MiquelRForgeFlow:
Several improvements here:

- No need to have unique percentages
- ~~Move profiles to templates~~
- And more improvements.

cc: @emagdalenaC2i, @legalsylvain, @albariera

### DIFF
--- a/product_abc_classification/models/abc_classification_profile.py
+++ b/product_abc_classification/models/abc_classification_profile.py
@@ -33,6 +33,12 @@ class ABCClassificationProfile(models.Model):
     past_period = fields.Integer(
         default=365, string="Past demand period (Days)", required=True
     )
+    product_variant_ids = fields.One2many(
+        "product.product", inverse_name="abc_classification_profile_id"
+    )
+    product_count = fields.Integer(
+        string="Product Count", compute="_compute_product_count", readonly=True
+    )
 
     @api.depends("level_ids")
     def _compute_representation(self):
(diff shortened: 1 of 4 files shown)


legalsylvain:
Sorry i still don't use this module, so don't have the knowlegde to review. But thanks for the improvment !

product_abc_classification/models/abc_classification_profile.py
```diff
@@ -54,13 +64,31 @@ def _check_levels(self):
...
-    def write(self, vals):
-        return super().write(vals)
+    @api.depends("product_template_ids")
+    def _compute_template_count(self):
+        for profile in self:
+            profile.template_count = len(profile.product_template_ids)
+
+    def action_view_templates(self):
+        templates = self.mapped("product_template_ids")
+        action = self.env.ref("product.product_template_action").read()[0]
```
rousseldenis:
```suggestion
        action = self.env["ir.actions.act_window"]._for_xml_id("product.product_template_action")
```

rousseldenis (changes requested):
Code review.

This should be updated to major.

rousseldenis (approved):

product_abc_classification/models/abc_classification_profile.py
```diff
@@ -55,8 +61,32 @@ def _check_levels(self):
...
-    def write(self, vals):
-        return super().write(vals)
+    @api.depends("product_template_ids")
+    def _compute_template_count(self):
+        for profile in self:
+            profile.template_count = len(profile.product_template_ids)
+
+    def action_view_templates(self):
+        templates = self.mapped("product_template_ids")
+        action = self.env["ir.actions.act_window"].for_xml_id(
```
albariera, line 75: When clicking the button it gives me this error:
TypeError: for_xml_id() missing 1 required positional argument: 'xml_id'
 
The method has to be _for_xml_id("product.product_template_action"), with the underscore or for_xml_id("product", "product_template_action") without the underscore

albariera (changes requested):
Thanks for the PR :blush: 
But I don't really understand why we are moving the profiles to the template. I suppose that the functionality was intended to be the same but to simplify the code. However, wouldn't it make sense that each product variant could have a different classification? Maybe one product variant is more sold than another and needs more inventory control. Just asking, you'll have more knowledge about it. 
But anyway if you want to move it to the template, at the moment it is not working fine:
- The profile is not being assigned to the product.template, the method _fill_data_from_stock_moves should be modified to product.template
- If you try to group from the product.product view by the level or profile you get this error: Fields in 'groupby' must be database-persisted fields (no computed fields)
- Why do we make invisible at product.template the level and not the profile? We should make invisible both or any, shouldn't we?

MiquelRForgeFlow:
> I don't really understand why we are moving the profiles to the template.

That was a suggestion of @legalsylvain done in https://github.com/OCA/product-attribute/pull/623#discussion_r608669761.

> However, wouldn't it make sense that each product variant could have a different classification? Maybe one product variant is more sold than another and needs more inventory control. Just asking, you'll have more knowledge about it.

Well, that's something about I am not sure, something that @legalsylvain asked about. I think I will delete that commit for now, to let the other commits get merged. If in the future any of you want to put profile in the template, then another PR can be done.

product_abc_classification/models/abc_classification_profile.py
MiquelRForgeFlow: yeah, the confusion comes because `for_xml_id` of v13 and `_for_xml_id` of v14 are different. Thanks for testing.

albariera:
Ok! We can merge these improvements and talk about the variant/template in other PR.
Also, the smart button is not working as expected
test_smart_button
You should remove the action context

albariera (approved):
LGTM :+1:

PR #1012, opened 2022-02-09
[13.0][FIX]product_supplierinfo_archive

GuillemCForgeFlow:
Added field on view and changed filters. Now only the **archived** filter will appear.


JordiMForgeFlow:
The active field is missing in the form view so you can't archive supplier from there. It would be nice to add :)

AaronHForgeFlow (approved):
Code review LG!

GuillemCForgeFlow:
> The active field is missing in the form view so you can't archive supplier from there. It would be nice to add :)

Done! Thanks for the review :+1:

MiquelRForgeFlow (approved):

PR #1013, opened 2022-02-10
[13.0][IMP] product_abc_classification: Add posibility to create a profile with fixed prices

CarlosRoca13:
cc @Tecnativa TT31774

PR #1015, opened 2022-02-15
[13.0][ADD] product_abc_classification_sale: New module to classify products from sale orders

CarlosRoca13:
cc @Tecnativa TT31774

This module depends of: 

- [x] https://github.com/OCA/product-attribute/pull/1013


product_abc_classification_sale/views/product_views.xml
```diff
@@ -0,0 +1,39 @@
+<?xml version='1.0' encoding='utf-8' ?>
+<odoo>
+    <record id="product_template_only_form_view" model="ir.ui.view">
+        <field name="model">product.template</field>
```
MiquelRForgeFlow, line 4: why not add `<field name="name">`?

MiquelRForgeFlow (commented):
@CarlosRoca13 Have you rebased properly? checking travis error seems like last update in product_abc_classification is not taken.

product_abc_classification_sale/views/product_views.xml
CarlosRoca13: It is not necessary, it will be computed by the system

chienandalu (approved):
Functional and code review :+1: Thanks!

MiquelRForgeFlow (approved):

PR #670, opened 2020-08-27
[10.0][ADD] stock_production_lot_expiry: New addon to manage lot expiry

lmignon:
This module extends the functionality of 'Products Expiration Date'
(product_expiry) to easily detect expired lot.

It adds two new fields on the Lot/Serial model.

* Expiry date (expiry_date)
* Is expired (is_expired)

The 'Expiry date' is a field filled at te creation of a lot/serial with the date
on which the lot/serial is expired. This field is filled from one of the
date fields present on the lot/serial ( 'end of life' or 'best before date' or
'removal date' or 'alert date' or ...). The field to use can be configured at
the product level or the category level.

The 'Is expired' field is a computed field that can be used into the business
code to easily know if a product is expired.

It also allows you to easily search for expired lots.

PR #1017, opened 2022-02-18
[13.0][ADD] product_expiry_configurable

albariera:
This model allows setting expiry times on category and to use the 'end_of_life' date for the computation of lot dates.
Read usage for more information.


albariera:
@LoisRForgeFlow @etobella

albariera:
@lmignon I have used in this module the same inheritance method that you have used in #670. It was very useful thank you! I hope you like this new addon.

lmignon:
> @lmignon I have used in this module the same inheritance method that you have used in #670. It was very useful thank you! I hope you like this new addon.

@albariera Happy it help you.

Thank you for your proposal, nevertheless tests are missing. Can you add tests for this new addon PLZ?

albariera:
@lmignon Done!! :blush:

etobella (approved):

product_expiry_configurable/models/stock_production_lot.py
```diff
@@ -0,0 +1,69 @@
...
+import datetime
+
+from odoo import api, fields, models
+
+
+class StockProductionLot(models.Model):
+
+    _inherit = "stock.production.lot"
+
+    product_qty = fields.Float(store=True)
```
etobella: Why do you need this stored?

albariera: I was searching among lots using this field, but it is true that it does not have much sense. I remove it. Thank you!

rousseldenis:
@albariera Thanks for this. I would say the module name should be a little bit improved as 'configurable' is a little bit too generic (but I don't have a specific idea - as your module does a lot of things...)

albariera:
@rousseldenis I agree with you that the name is too generic, but I could not come up with a better name. Maybe product_expiry_dates_extended could be better? Do you have any proposals?

product_expiry_configurable/models/stock_production_lot.py
```diff
@@ -0,0 +1,112 @@
...
+        }
+        product = self.env["product.product"].browse(product_id) or self.product_id
+        res = dict.fromkeys(mapped_fields, False)
+        life_date = self.life_date or self.env.context.get("life_date")
+        if product and product.tracking != "none":
+            if product.compute_dates_from == "life_date" and life_date:
+                mapped_fields.pop("life_date", None)
+                res = self._get_dates_from_life_date(product, mapped_fields, life_date)
+            elif product.compute_dates_from == "current_date":
+                res = self._get_dates_from_current_date(product, mapped_fields)
```
LoisRForgeFlow: You should use super here

LoisRForgeFlow (approved):
LGTM

product_expiry_configurable/models/stock_production_lot.py
albariera: Done!! Thank you for your review :)

etobella:
@OCA/product-maintainers Can you proceed with merge?

olgamarcocb:
LGTM :+1:

MiquelRForgeFlow (approved):

PR #1023, opened 2022-02-25
[ADD] product_pricelist_direct_print_company_group: New module

CarlosRoca13:
This module is an integration of product_company_group with product_pricelist_direct_print

This module depends on:

- [x] https://github.com/OCA/product-attribute/pull/1022

cc @Tecnativa TT34816

please @sergio-teruel @carlosdauden review this


product_pricelist_direct_print_company_group/wizards/product_pricelist_print.py
```diff
@@ -0,0 +1,20 @@
...
+
+class ProductPricelistPrintCompanyGroup(models.TransientModel):
+    _inherit = "product.pricelist.print"
+
+    @api.model
+    def _get_sale_order_domain(self, partner):
+        domain = super()._get_sale_order_domain(partner)
+        # If any parter has company_group_id defined with the selected partner, it means
+        # that the partner selected is a company group.
+        partners_grouped = self.env["res.partner"].search(
```
carlosdauden: Now you can use the field added in:
https://github.com/OCA/partner-contact/pull/1239

I use it at here:
https://github.com/OCA/contract/pull/797/files#diff-4cbf45e260f90f16486df50dd7c169ad2918d55b0abc9603f86c9298643f544aR21

CarlosRoca13:
Changed @carlosdauden

product_pricelist_direct_print_company_group/wizards/product_pricelist_print.py
```diff
@@ -0,0 +1,15 @@
...
+
+class ProductPricelistPrintCompanyGroup(models.TransientModel):
+    _inherit = "product.pricelist.print"
+
+    @api.model
+    def _get_sale_order_domain(self, partner):
+        domain = super()._get_sale_order_domain(partner)
+        if partner.company_group_member_ids:
+            domain.remove(("partner_id", "child_of", partner.id))
+            domain.append(("company_group_id", "=", partner.id))
```
carlosdauden:
```suggestion
            domain.append(("partner_id", "in", partner.company_group_member_ids.ids))
```
With this change we exploit the partner field index:
https://github.com/odoo/odoo/blob/739da6b8b2ba60965728ffdc1ba133203fd65a2d/addons/sale/models/sale.py#L190
company_group_id field are not indexed:
https://github.com/OCA/partner-contact/blob/8f615c685205bb44ea785f5e3e64efd54ce5efae/partner_company_group/models/sale_order.py#L11

product_pricelist_direct_print_company_group/wizards/product_pricelist_print.py
```diff
@@ -0,0 +1,15 @@
...
+
+
+class ProductPricelistPrintCompanyGroup(models.TransientModel):
+    _inherit = "product.pricelist.print"
+
+    @api.model
+    def _get_sale_order_domain(self, partner):
+        domain = super()._get_sale_order_domain(partner)
+        if partner.company_group_member_ids:
+            domain.remove(("partner_id", "child_of", partner.id))
```
carlosdauden: Just as an informative note, if remove doesn't find the value it throws an exception.

I'm considering if it is really useful to add this complexity...
Perhaps the "normal" thing would be to simply return the domain of the super or the one defined here depending on the condition.
If the original domain is modified, we will analyze what we want to do, because at this moment we cannot know if the result after that modification will be the one that interests us.

I also plead guilty :sweat_smile: 
https://github.com/OCA/contract/pull/797/files#diff-4cbf45e260f90f16486df50dd7c169ad2918d55b0abc9603f86c9298643f544aR13-R22

carlosdauden (changes requested):

product_pricelist_direct_print_company_group/wizards/product_pricelist_print.py
CarlosRoca13: Hmmm... IDK too

I corrected what I have made to avoid possible errors, but is true that If any module modifies the domain changing the partners domain the result can take a different direction...

carlosdauden (approved):

sergio-teruel (approved):

PR #1032, opened 2022-03-09
[13.0][OU-FIX] product_abc_classification_sale: migration script

chienandalu:
cc @Tecnativa TT34408


pedrobaeza (approved):
Remember to put `[OU-FIX]` in these cases and the version on the PR title

PR #1033, opened 2022-03-09
[15.0] [MIG] product_category_active

tgrellety:


legalsylvain (approved):
Diff review / no test : lgtm.

thanks.

product_category_active/tests/test_product_category_active.py
```diff
@@ -0,0 +1,52 @@
+# Copyright 2020 Tecnativa - Ernesto Tejeda
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl.html).
+
+from odoo.exceptions import ValidationError
+from odoo.tests import SavepointCase
+
+
+class TestProductCategoryActive(SavepointCase):
```
ajaniszewska-dev: as per https://github.com/OCA/maintainer-tools/wiki/Migration-to-version-15.0:

Replace SavepointCase by TransactionCase in tests, as they are now the same. The old one still exists as an alias, but a warning will arise, and next version will remove such alias.

ajaniszewska-dev:
@tgrellety  small comment regarding test case, can you please fix pre commit as well? use  `pre-commit run --all `command for that and commit modified files.

i-vyshnevska (approved):

Rad0van:
@tgrellety pre-commit is failing probably because of that SavepointCase not being replaced by TransactionCase. Would you please fix that so the PR could be merged? If you're not interested in continuing this someone else can take over.

product_category_active/tests/test_product_category_active.py
Rad0van: This needs to be fixed.

Rad0van (changes requested):
Please fix the SavepointCase -> TransactionCase.

product_category_active/tests/test_product_category_active.py
tgrellety: Thank you for your feedback.
Can you take care of this ?

Rad0van: Nope, I cannot change content of your PR. Sorry.

Rad0van:
@tgrellety better, but still not OK. You need to do 2 things:
1) run `pre-commit run -a` to have the sources formatted correctly (that's why pre-commit here fails). Then push. Will create another commit.
2) Squash the commits those 2 commits into the previous one so only the one with message _[[MIG] product_category_active: Migration to 15.0]_ remains. Like `git rebase -i HEAD~3`

Rad0van:
@tgrellety can you please do those last things to get this merged? Thank you.

product_category_active/models/product_category.py
```diff
@@ -0,0 +1,31 @@
...
+
+from odoo import _, api, fields, models
+from odoo.exceptions import ValidationError
+
+
+class ProductCategory(models.Model):
+    _inherit = "product.category"
+
+    active = fields.Boolean(
+        string="Active",
```
Rad0van: Unfortunately there is another one. You always need to check that the pre-coomit ends up without error. In this case it's complaining about these 2 lines. The second one is redundant as the same string would be rendered just from the field's name.

So please remove the line 12, rerun pre-commit locally to be sure everything else is fine. Then commit, squash with the previous one and should be fine to go.

Rad0van:
Now it seems fine!

Rad0van (approved):
Looks fine, tested.

tgrellety:
> Looks fine, tested.

Yes, thanks for your help @Rad0van !

victoralmau (approved):
Code and functionality OK.

PR #1027, opened 2022-03-01
[FIX] #1026

ilyasProgrammer:

PR #1034, opened 2022-03-10
[FIX] fix missing dependency

sebastienbeau:
The following PR add a new security rule but the dependency is missing https://github.com/OCA/product-attribute/pull/1027/files


sebastienbeau:
@pedrobaeza @ilyasProgrammer @francesco-ooops @iwkse

leemannd (approved):

pedrobaeza (approved):
Sorry for not seeing it

PR #1042, opened 2022-03-25
[12.0][FIX] product_lot_sequence: Avoid generate extra sequences on duplicates

HviorForgeFlow:
A weird behavior of this module is that anytime you update a product template field the write method generate a new ir.sequence because it does not check if an lot_sequence_id is already set on the template.

This fix aims to avoid this duplicates, it is more like a patch because in my opinion lot_sequence generation has to be moved outside write and create method

CC @ForgeFlow

### DIFF
--- a/product_lot_sequence/models/product.py
+++ b/product_lot_sequence/models/product.py
@@ -86,11 +86,14 @@ class ProductTemplate(models.Model):
         for template in self:
             tracking = vals.get("tracking", False) or template.tracking
             if tracking in ["lot", "serial"]:
-                if not vals.get("lot_sequence_id", False):
+                if (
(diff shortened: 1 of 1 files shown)


flotho (approved):
Code review appropriate with comment in the PR.
Thanks

AaronHForgeFlow (approved):
makes sense

MiquelRForgeFlow (approved):

erpstolenstores (approved):

newtratip (approved):
LGTM :+1:

HviorForgeFlow:
@OCA/product-maintainers could we get this merged? Thanks!

HviorForgeFlow:
@simahawk ready as forward port has been merged recently

PR #1044, opened 2022-03-28
[MIG] product_order_noname: Migration to 15.0

nicomacr:


florenciafrigieri2 (approved):
LGTM

product_order_noname/models/product.py
```diff
@@ -0,0 +1,6 @@
+from odoo import models
+
+
+class ProductProduct(models.Model):
+    _inherit = "product.product"
+    _order = "default_code, id"
```
jjscarafia: please use new order criteria
    _order = 'priority desc, default_code, id'

https://github.com/odoo/odoo/blob/13.0/addons/product/models/product.py#L70

product_order_noname/models/product_template.py
```diff
@@ -0,0 +1,8 @@
+from odoo import fields, models
+
+
+class ProductTemplate(models.Model):
+    _inherit = "product.template"
+    _order = "default_code, id"
```
jjscarafia: please use [new order](https://github.com/odoo/odoo/blob/15.0/addons/product/models/product_template.py#L19) criteria
    _order = "priority desc"

may be, I would add "default_code" here because if not the order will be only regarding the priority field

So, I would leave     _order = "priority desc, default_code"

jjscarafia (changes requested):

jjscarafia (approved):
LGTM

nicomacr:
@legalsylvain could you check this PR?
 Thanks

florenciafrigieri2:
@legalsylvain @rousseldenis Hi! Can you please check this PR? Thanks!

legalsylvain (approved):
Please add a section in the roadmap that mention to check if the module is still interesting in V16. (Due to ir.translation refactoring.)
Otherwise lgtm. Thanks.

nicomacr:
> Please add a section in the roadmap that mention to check if the module is still interesting in V16. (Due to ir.translation refactoring.)
> Otherwise lgtm. Thanks.

I've added this and updated the commit!

PR #1055, opened 2022-04-13
[13.0][IMP] product_restricted_type: apply onchanges on product variant

JordiMForgeFlow:
Apply the onchanges on product variant so that they are applied when user creates a variant directly.


AaronHForgeFlow (approved):
Code review LGTM!

rousseldenis (approved):

PR #1061, opened 2022-04-21
[15.0][mig] product_seasonality

damdam-s:


rousseldenis (approved):

damdam-s:
@rousseldenis a refactoring of this is coming to separate "product_allowed_list"(without dates) from seasonal config (with dates)

simahawk:
@damdam-s I would include the refactoring commit here, it's pointless to have v15 w/out it no?

simahawk (changes requested):
blocking just to see how it evolves on #1063

damdam-s:
> @damdam-s I would include the refactoring commit here, it's pointless to have v15 w/out it no?

yes of course but I follow the migration process :) migrate 1:1 and then propose the refactoring

simahawk:
For refactoring I always find better this approach:

1. open a new PR w/ product_seasonality AS IS + installable = False -> we fast track it
2. rebase this PR and add the refactoring here (or in the other PR and close this)

jcoux (approved):

PR #1063, opened 2022-04-22
[15.0] product_seasonality refactoring

damdam-s:
depends on/includes #1061 

product_seasonality refactoring:

- refactor module to add a base module that handles allowed product lists.
- Product_seasonality comes to add dates rules on top of this base module


damdam-s:
@rousseldenis the refactoring mentionned in #1061 ;)
If you can have a look that could be great

product_seasonality/models/product_allowed_list_line.py
```diff
@@ -0,0 +1,41 @@
...
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+
+from odoo import _, api, fields, models
+from odoo.exceptions import ValidationError
+
+
+class ProductAllowedListLine(models.Model):
+    _inherit = "product.allowed.list.line"
+
+    date_start = fields.Datetime(required=True)
```
simahawk: not sure why we should keep this module as `product_seasonality`.
I'd rename this as `product_allowed_list_[date|weekly|calendar]`.

product_seasonality/models/res_company.py
```diff
@@ -0,0 +1,42 @@
...
+
+from odoo import _, api, fields, models
+from odoo.exceptions import ValidationError
+
+
+class ResCompany(models.Model):
+
+    _inherit = "res.company"
+
+    default_seasonal_config_id = fields.Many2one(
```
simahawk: As mentioned, I would not keep this as "seasonal"

simahawk (changes requested):

product_seasonality/models/product_allowed_list_line.py
damdam-s: hum I think the name `seasonality` reflects what it does but I can change if you want

damdam-s: done.
renamed to `product_allowed_list_date`

product_allowed_list/tests/test_product_allowed_list.py
```diff
@@ -0,0 +1,24 @@
+# Copyright 2021 Camptocamp SA
+# @author: Simone Orsi <simone.orsi@camptocamp.com>
+# @author: Damien Crier <damien.crier@camptocamp.com>
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+
+from .common import CommonCaseWithLines
+
+
+class TestSeasonalityCase(CommonCaseWithLines):
```
jcoux, line 9:
```suggestion
class TestProductAllowedListCase(CommonCaseWithLines):
```

jcoux (approved):
1 little suggestion: pre-approved

product_allowed_list/models/product_list_line.py
```diff
@@ -0,0 +1,52 @@
...
+    product_template_id = fields.Many2one(
+        comodel_name="product.template",
+        domain=[("sale_ok", "=", True)],
+        required=True,
+    )
+    product_id = fields.Many2one(
+        comodel_name="product.product",
+        domain="[('product_tmpl_id', '=', product_template_id)]",
+    )
+    display_name = fields.Char(compute="_compute_display_name")
```
ivantodorovich, line 27:
```suggestion
```

No need to add it explicitly, it's considered a "magic field" on every model

https://github.com/odoo/odoo/blob/75dc83184b27f671abe5684bd7556fbb88011d55/odoo/models.py#L203-L204

product_allowed_list/models/product_list_line.py
```diff
@@ -0,0 +1,52 @@
...
+
+    def find_for_product(self, prod, config=None):
+        domain = [
+            "|",
+            ("product_id", "=", prod.id),
+            "&",
+            ("product_id", "=", False),
+            ("product_template_id", "=", prod.product_tmpl_id.id),
```
ivantodorovich, line 35: This domain evaluates to something like:

```(product_id == prod.id) or (product_id == False and product_template_id == prod.product_tmpl_id.id)```

As we can (or at least should [^1]) trust the template will always correspond to the product_id, if present. IMO it should rather be:

```suggestion
            "|",
            ("product_id", "=", prod.id),
            ("product_id", "=", False),
            ("product_template_id", "=", prod.product_tmpl_id.id),
```
meaning:
```(product_id == prod.id or product_id == False) and product_template_id == prod.product_tmpl_id.id```


But for a shorter more readable version I'd go for this:

```suggestion
            ("product_id", "in", [False, prod.id]),
            ("product_template_id", "=", prod.product_tmpl_id.id),
```


[^1]: It may be a good idea to add an `_sql_constraint` for this

product_allowed_list/models/product_list_line.py
```diff
@@ -0,0 +1,52 @@
...
+class ProductListLine(models.Model):
+    _name = "product.allowed.list.line"
+    _description = "Product allowed list configuration Line"
+
+    product_allowed_list_id = fields.Many2one(
+        string="Product list configuration",
+        comodel_name="product.allowed.list",
+        required=True,
+    )
+    product_template_id = fields.Many2one(
```
ivantodorovich, line 18:
```suggestion
    product_tmpl_id = fields.Many2one(
```

Since it's a refactoring, I suggest to do this change too.
This way it's consistent with the `product.product` field name

product_allowed_list/models/product_list_line.py
```diff
@@ -0,0 +1,52 @@
...
+        domain=[("sale_ok", "=", True)],
+        required=True,
+    )
+    product_id = fields.Many2one(
+        comodel_name="product.product",
+        domain="[('product_tmpl_id', '=', product_template_id)]",
+    )
+    display_name = fields.Char(compute="_compute_display_name")
+
+    def find_for_product(self, prod, config=None):
```
ivantodorovich, line 29:
```suggestion
    def _find_for_product(self, prod, config=None):
```
If not exposed through xmlrpc, better to have it as a private method

product_allowed_list_date/models/product_allowed_list_line.py
```diff
@@ -0,0 +1,44 @@
...
+
+    @api.constrains("date_start", "date_end")
+    def _check_date_end(self):
+        for line in self:
+            if line.date_start and line.date_end and line.date_end < line.date_start:
+                raise ValidationError(
+                    _("The end date cannot be earlier than the start date.")
+                )
+
+    def is_sale_ok(self, date):
```
ivantodorovich, line 32:
```suggestion
    def _is_sale_ok(self, date):
```

ivantodorovich (approved):
If it's not too late, I suggest `product_whitelist`
IMO it's shorter and concise

`s/product_allowed_list_id/product_whitelist_id`

legalsylvain:
Cc @quentinDupont maybe this one could interest you.

rousseldenis:
@damdam-s

PR #1065, opened 2022-04-26
[14.0] [IMP] `base_product_mass_addition`: avoid writing on product's metadata + add tests

ivantodorovich:
Supress `LOG_ACCESS_COLUMNS` writes if we're only writing on quick magic fields, as they could lead to concurrency issues.

Moreover, from a functional perspective, these magic fields aren't really modifying the product's data so it doesn't make sense to update its metadata.

We achieve it by reverting the changes made by ``write`` [^1], before [^2] reaching any explicit flush [^3] or inverse computation [^4].

[^1]: https://github.com/odoo/odoo/blob/f74434c6f/odoo/models.py#L3652-L3663
[^2]: https://github.com/odoo/odoo/blob/f74434c6f/odoo/models.py#L3686
[^3]: https://github.com/odoo/odoo/blob/f74434c6f/odoo/models.py#L3689
[^4]: https://github.com/odoo/odoo/blob/f74434c6f/odoo/models.py#L3703

Basically, if all we're modifying are quick magic fields, and we don't have any other column to flush besides the `LOG_ACCESS_COLUMNS`, clear it.

### DIFF
--- a/base_product_mass_addition/models/product_product.py
+++ b/base_product_mass_addition/models/product_product.py
@@ -6,6 +6,7 @@
 
 
 from odoo import fields, models
+from odoo.models import LOG_ACCESS_COLUMNS
 
 
 class ProductProduct(models.Model):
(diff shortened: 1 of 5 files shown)


base_product_mass_addition/models/product_product.py
```diff
@@ -37,6 +38,34 @@ def _inverse_set_process_qty(self):
...
+        # reaching any explicit flush [^2] or inverse computation [^3].
+        #
+        # [^1]: https://github.com/odoo/odoo/blob/f74434c6f/odoo/models.py#L3652-L3663
+        # [^2]: https://github.com/odoo/odoo/blob/f74434c6f/odoo/models.py#L3686
+        # [^3]: https://github.com/odoo/odoo/blob/f74434c6f/odoo/models.py#L3689
+        # [^4]: https://github.com/odoo/odoo/blob/f74434c6f/odoo/models.py#L3703
+        #
+        # Basically, if all we're modifying are quick magic fields, and we don't have
+        # any other column to flush besides the LOG_ACCESS_COLUMNS, clear it.
+        quick_fnames = {"qty_to_process", "quick_uom_id"}
```
simahawk: pls use a `tuple` in these cases

base_product_mass_addition/tests/models/__init__.py
```diff
@@ -0,0 +1 @@
+from . import order
```
simahawk: This must be imported ONLY in the test class setup

simahawk (changes requested):

base_product_mass_addition/models/product_product.py
ivantodorovich: done 👍🏻

base_product_mass_addition/tests/models/__init__.py
ivantodorovich: done 👍🏻

gurneyalex (commented):
The code looks ok to me. I'm just wondering if this is the right place to put this log silencing code? Should this not live in either a customer specific module or in an entirely separate specific module?

ivantodorovich:
> The code looks ok to me. I'm just wondering if this is the right place to put this log silencing code? Should this not live in either a customer specific module or in an entirely separate specific module?

IMO this is the right place because these "magic fields" aren't really meant to write product data. It's a rather hacky way to achieve what we really want: being able to create/update `sale|purchase.order.line` from the product tree view.

damdam-s (approved):
thanks

legalsylvain:
@simahawk, could you update your review ? your remarks looks answered.

thanks !

simahawk (approved):

PR #1066, opened 2022-04-27
[14.0][FIX] product_stock_state : product_template_view

Kev-Roche:
Fix missing labels in product_template_view for "stock_state_threshold" and "manual_stock_state_threshold".
cc @sebastienbeau , @legalsylvain , @simahawk 

Before fix : 

After fix :

### DIFF
--- a/product_stock_state/views/product_template_view.xml
+++ b/product_stock_state/views/product_template_view.xml
@@ -4,10 +4,12 @@
         <field name="model">product.template</field>
         <field name="inherit_id" ref="stock.view_template_property_form" />
         <field name="arch" type="xml">
-            <field name="route_ids" position="before">
-                <field name="stock_state_threshold" />
-                <field name="manual_stock_state_threshold" class="oe_edit_only" />
-            </field>
+            <xpath expr="//group[@name='group_lots_and_weight']" position="after">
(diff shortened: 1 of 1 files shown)


product_stock_state/views/product_template_view.xml
```diff
@@ -4,10 +4,10 @@
         <field name="model">product.template</field>
         <field name="inherit_id" ref="stock.view_template_property_form" />
         <field name="arch" type="xml">
-            <field name="route_ids" position="before">
```
simahawk, line 7: I think it would be better to inherit from the base product view as stock does and hook to the same place:

` <xpath expr="//group[@name='group_lots_and_weight']" position="[before|after]">`

See https://github.com/odoo/odoo/blob/a7f7233e0eae8ee101d745a9813cba930fd03dcb/addons/stock/views/product_views.xml#L124

Kev-Roche: good idea, thanks

simahawk:
This one fell off my radar, sorry. @Kev-Roche can you please:
1. rebase on 14.0
2. run `pre-commit run -a`
3. push again

Thanks!

Kev-Roche:
@simahawk, it s done, thank you !

PR #1067, opened 2022-04-27
[IMP] product_tier_validation (Change product state to use stage state field)

patrickrwilson:
This PR improves on the previous changes that updated tier validation to work with stages. There was an issue where the user could move the product to a stage that was later in the list and then move it back to active so this fixes that issue by adding a 'related state' field to the stage so tier validation can properly work with states. This is similar to how partner_stage functions.

### DIFF
--- a/product_tier_validation/models/product_state.py
+++ b/product_tier_validation/models/product_state.py
@@ -1,0 +1,17 @@
+# Copyright 2017 ACSONE SA/NV (<http://acsone.eu>)
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl.html).
+
+from odoo import fields, models
+
+
+class ProductState(models.Model):
+    _inherit = "product.state"
+
+    state = fields.Selection(
+        [("draft", "To Approve"), ("confirmed", "Approved"), ("cancel", "Archived")],
+        string="Related State",
+        default="confirmed",
+        help="""Tier validation uses specific states (To and From) to trigger
+              validation needs. Since the Code field is a unique and user defined
+              field, this state field provides the states used for valdiation.""",
+    )
--- a/product_tier_validation/__manifest__.py
+++ b/product_tier_validation/__manifest__.py
@@ -14,6 +14,7 @@
     "depends": ["product_state", "base_tier_validation"],
     "data": [
         "data/tier_definition.xml",
+        "views/product_state_view.xml",
         "views/product_template_view.xml",
     ],
 }
--- a/product_tier_validation/models/__init__.py
+++ b/product_tier_validation/models/__init__.py
@@ -1,5 +1,6 @@
 # Copyright 2021 Open Source Integrators
 # License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
 
+from . import product_state
 from . import product_template
 from . import tier_definition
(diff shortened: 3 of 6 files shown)


dreispt (approved):

dreispt:
@patrickrwilson Travis is Red

product_tier_validation/models/product_state.py
```diff
@@ -0,0 +1,14 @@
+# Copyright 2017 ACSONE SA/NV (<http://acsone.eu>)
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl.html).
+
+from odoo import fields, models
+
+
+class ProductState(models.Model):
+    _inherit = "product.state"
+
+    state = fields.Selection(
```
rousseldenis, line 10: @patrickrwilson Could you add a help in order to 'help' people to understand what is this for? This is one more layer and can introduce confusion in people mind...

IMHO, I'm not sure about this

product_tier_validation/tests/test_product_tier_validation.py
```diff
@@ -37,4 +37,4 @@ def test_validation_product_template(self):
         product.request_validation()
         product.with_user(self.test_user_1).validate_tier()
         product.write({"product_state_id": self.normal_state.id})
-        self.assertEqual(product.state, self.normal_state.code)
```
rousseldenis: Is this test result changed ?

product_tier_validation/models/product_state.py
patrickrwilson: I can add a help, the reason we can't use 'code' is because of several reasons...
Tier validation needs 'to' and 'from' states in order to compute the validation needs so it needs to be in a selection field. 
The code field is user defined and unique so we can't really use it for the tier validation purpose.

product_tier_validation/tests/test_product_tier_validation.py
patrickrwilson: Not sure I understand your question. I changed this since in this module instead of setting the state from code, we set it from the state field on the stage therefore the test needed to be updated too in order to be correct.

patrickrwilson:
I was able to allow the tests to be green but I'm not sure if my approach is acceptable because basically there is a compatibilty issue with the product_status module which not only requires specific stages but also contrains them to not be edited. Basically the tier validation won't work if that module is used so I added to bypasses to allow the product_status module to work if both are installed.
I'm open to suggestions if this is the wrong approach and there is a better way.

dreispt (approved):
Ideally the state would be introduced in the product_stage module.
But that would need to be discussed with the corresponding maintainers.
I agree with this approach as a short term stop-gap.

product_tier_validation/models/product_state.py
rousseldenis: IMHO, definetely this is a bad implementation. A state... in a state. Quite insane.

rousseldenis: @kittiu What do you think of this ?

dreispt: @rousseldenis Indeed, the Model should have been named ProductStage.
"Stages" are configurable, while "States" are a predefined fixed list.
For business logic, we want "Stages" to be mapped to known "States" that can be safely used in business logic.

rousseldenis:
IMHO this has to be introduced in another module as this adds too much complexity to this

PR #1074, opened 2022-05-10
[IMP] product_state (copy=False)

patrickrwilson:
The product state shouldn't be copied when copying the product so this adds copy=False to the state field.

### DIFF
--- a/product_state/models/product_template.py
+++ b/product_state/models/product_template.py
@@ -26,6 +26,7 @@ class ProductTemplate(models.Model):
         default=lambda self: self._get_default_product_state().id,
         index=True,
         tracking=10,
+        copy=False,
     )
 
     def _inverse_product_state_id(self):
--- a/product_state/tests/test_product_state.py
+++ b/product_state/tests/test_product_state.py
@@ -75,7 +75,6 @@ class TestProductState(SavepointCase):
         Create another default state,
         It should have the existing only one default state at time
         """
-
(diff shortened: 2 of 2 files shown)


patrickrwilson:
@dreispt can you please review.

legalsylvain (commented):
Comment (non blocking) : state is a caracteristic of the product so it is attended by the user to preserve the information when duplicate. Why do you think that empty value is better than the value of the duplicated item ?

Note : i personnaly dont use this module.

patrickrwilson:
@legalsylvain I'm not sure I understand what you mean about 'empty value' as it wouldn't be an empty value but instead the newly copied product would default to the stage that has the is_default set.

rousseldenis (commented):
@patrickrwilson Could you add a little test for this change ?

legalsylvain:
> patrickrwilson

Ah, sorry. I didn't know the interaction between ``default`` and ``copy``.

regards.

rousseldenis (approved):

patrickrwilson:
FYI: I'll be fixing the new test

dreispt:
@legalsylvain "state" is computed based on "Stage" - is groups the dynamic and configurable Stages list into specific state codes.
When duplicating a record the Stage should move the the initial default stage. The state should also follow that.

dreispt (approved):

PR #1075, opened 2022-05-13
[13.0][FIX] product_secondary_unit: Don't alter previous quantity

pedrobaeza:

- Apply the mixin to the purchase orders.
- Create a purchase order line without secondary unit.

Current behavior: the line quantity is 0.

Expected behavior: the line quantity is 1.

Explanation:

ORM framework in v13 has a problem, raising CacheMiss on records that are not assigned in their compute method although being stored, but we have to assign the value in cache, not the one in the `_origin` internal variable, as a newly record doesn't have any, putting 0 in the quantity.

@Tecnativa TT27363


carlosdauden (approved):

sergio-teruel (approved):

pedrobaeza:
Thanks for the quick reaction:

/ocabot merge patch

PR #1076, opened 2022-05-17
[FIX]product_sequence: remove dependence

vib-adhoc:
The dependence is removed that is not required by the module. It does not affect anyone who has this module installed, but in new installations you can choose or not to use uniqueness.


jjscarafia (approved):
LGTM

jjscarafia:
@sergio-teruel @angelmoya
what do you think about this? A simple adjustment.
Thanks

Rad0van:
+1

Rad0van (approved):
LGTM

PR #1077, opened 2022-05-18
product_main_supplierinfo: fix default sellers for variant

TDu:
If no specific record are found by the first filter. The Odoo default
method will return all active supplier info.
Which will include also supplier info record related to other specific
variant.

With this change if some supplier info related to the product
variant exist, they will be returned instead.

### DIFF
--- a/product_main_supplierinfo/models/product_product.py
+++ b/product_main_supplierinfo/models/product_product.py
@@ -35,7 +35,7 @@ class ProductProduct(models.Model):
         They are ordered and filtered like it is done in the standard 'product' addon.
         """
         self.ensure_one()
-        all_sellers = self._prepare_sellers().filtered(
+        all_sellers = self._prepare_sellers(False).filtered(
             lambda s: not s.company_id or s.company_id.id == self.env.company.id
         )
         today = fields.Date.context_today(self)
@@ -49,5 +49,7 @@ class ProductProduct(models.Model):
             )
         )
         if not sellers:
-            sellers = all_sellers
-        return sellers
+            sellers = all_sellers.filtered(lambda s: (s.product_id == self))
+            if not sellers:
+                sellers = sellers = all_sellers.filtered(lambda s: not s.product_id)
+        return sellers.sorted("price")
--- a/product_main_supplierinfo/tests/test_product_main_supplier.py
+++ b/product_main_supplierinfo/tests/test_product_main_supplier.py
@@ -13,16 +13,81 @@ class TestProductMainSupplierInfo(SavepointCase):
         cls.product = cls.env.ref("product.product_product_6")
         cls.seller_1 = cls.env.ref("product.product_supplierinfo_1")
         cls.seller_1.sequence = 1
+        cls.seller_1.price = 700
         cls.seller_2 = cls.env.ref("product.product_supplierinfo_2")
         cls.seller_2.sequence = 2
+        cls.seller_2.price = 720
         cls.seller_2bis = cls.env.ref("product.product_supplierinfo_2bis")
         cls.seller_2bis.sequence = 3
+        cls.seller_2bis.price = 740
         cls.company_2 = cls.env.company.create({"name": "Company2"})
+        # For the test case 4
+        cls.attribute = cls.env["product.attribute"].create({"name": "Size"})
+        cls.attribute_value_1 = cls.env["product.attribute.value"].create(
+            {"name": "L", "attribute_id": cls.attribute.id}
+        )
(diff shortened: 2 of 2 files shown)


product_main_supplierinfo/models/product_product.py
```diff
@@ -49,5 +49,7 @@ def _get_sellers(self):
             )
         )
         if not sellers:
-            sellers = all_sellers
+            sellers = all_sellers.filtered(lambda s: (s.product_id == self))
```
simahawk, line 52: this is already done above.
I would rewrite the whole method like this:
```
        sellers = all_sellers = self._prepare_sellers().filtered(
            lambda s: not s.company_id or s.company_id.id == self.env.company.id
        )
        # Filter by specific variant or none.
        by_variant = sellers.filtered(
            lambda s: s.product_id == self or not s.product_id
        )
        if by_variant:
            sellers = by_variant

        # Filter by dates
        today = fields.Date.context_today(self)
        by_date = sellers.filtered(lambda s: (
                s.date_start <= today if s.date_start else True
            ) and (
                s.date_end >= today if s.date_end else True
            )
        )
        if by_date:
            sellers = by_date
        
        # Fallback to all available sellers
        if not sellers:
            sellers = all_sellers
        return sellers
```

simahawk (changes requested):

product_main_supplierinfo/models/product_product.py
simahawk: also, we must have a test :wink:

sebalix: Seems in std Odoo they do not return a seller if the variant doesn't match the current one... https://github.com/odoo/odoo/blob/5e45d7267ddacee6cf882e1d3be60efc52b12967/addons/product/models/product.py#L621-L625
And only sellers sharing the same partner are selected, then sorted on the `price` to get the cheaper (`_prepare_sellers` is already doing some sort, don't know if we have to do it again...).

TDu: I have added the test.
But I did not change the implementation.
About that line being already done above, not really because above the filtering is including the supplier info related to no variant. And the idea is if no valid (date wise) supplier info are found, then first return all supplier info only related to the variant.
This implementation change does not do that as the `by_variant` also includes the supplier info related to no variant.

TDu: > Seems in std Odoo they do not return a seller if the variant doesn't match the current one... https://github.com/odoo/odoo/blob/5e45d7267ddacee6cf882e1d3be60efc52b12967/addons/product/models/product.py#L621-L625 And only sellers sharing the same partner are selected, then sorted on the `price` to get the cheaper (`_prepare_sellers` is already doing some sort, don't know if we have to do it again...).

It is `prepare_sellers` that is called above and not `select_sellers` which calls `prepare_sellers` in the odoo implementation.
Are you implying that we should change the implementation ?

sebalix, line 52: I would suggest then:
```suggestion
            sellers = all_sellers.filtered(lambda s: (s.product_id == self))
            if not sellers:
                # Fallback on template sellers if there is none for the current variant
                sellers = all_sellers.filtered(lambda s: not s.product_id)
```

product_main_supplierinfo/models/product_product.py
```diff
@@ -49,5 +49,7 @@ def _get_sellers(self):
...
+            if not sellers:
+                sellers = all_sellers
         return sellers
```
sebalix: We could also add this to be in-line with what Odoo does:
```suggestion
        return sellers.sorted("price")
```

TDu: Done, tested on a purchase order and if multiple supplier info record are matching Odoo will select the cheaper alternative even if it has a higher sequence.

product_main_supplierinfo/models/product_product.py
TDu: In the same logic, added.

sebalix:
@TDu your tests are now failing :)

TDu:
@sebalix  Tests are good now...

sebalix (approved):

simahawk:
@TDu can you squash and rewrite the commit msg (eg: = PR title)?

TDu:
Squashed and renamed. But Travis failed and it does not make sense to me https://app.travis-ci.com/github/OCA/product-attribute/jobs/571634140#L2242
Trying a fix with a fixup

TDu:
That fix worked so squashed again...

simahawk (approved):

PR #1078, opened 2022-05-18
[14.0][ADD] product_multi_lot_sequence

HviorForgeFlow:
Inspired in product_lot_sequence this module allow to select multiple serial number sequences for a product.

For the moment I consider creating a new module the easier way to start. It is not possible to use `product_lot_sequence` and `product_multi_lot_sequence` modules at the same time. 

CC @ForgeFlow


product_multi_lot_sequence/__manifest__.py
```diff
@@ -0,0 +1,19 @@
...
+
+{
+    "name": "Product Multi Lot Sequence",
+    "summary": """
+        Adds ability to define a multiple lot sequence from the product""",
+    "version": "14.0.1.0.0",
+    "license": "AGPL-3",
+    "author": "ForgeFlow S.L., Odoo Community Association (OCA)",
+    "website": "https://github.com/OCA/product-attribute",
+    "depends": ["stock", "product"],
```
MiquelRForgeFlow:
```suggestion
    "depends": ["stock"],
```

product_multi_lot_sequence/models/product.py
```diff
@@ -0,0 +1,16 @@
...
+
+
+class ProductTemplate(models.Model):
+    _inherit = "product.template"
+
+    product_lot_sequence_ids = fields.One2many(
+        "product.lot.sequence",
+        "product_tmpl_id",
+        string="Product Lot Sequence",
+        help="This field contains the information related to the " "numbering of lots.",
```
MiquelRForgeFlow:
```suggestion
        help="This field contains the information related to the numbering of lots.",
```

MiquelRForgeFlow (approved):

JoanSForgeFlow (approved):

olgamarcocb (approved):
LGTM :+1:

product_multi_lot_sequence/models/stock_production_lot.py
```diff
@@ -0,0 +1,83 @@
...
+            if rec.product_id:
+                pls_ids = rec.product_id.product_tmpl_id.product_lot_sequence_ids
+                available = pls_ids.filtered(lambda x: x.product_id == self.product_id)
+                if not available:
+                    available = pls_ids.filtered(
+                        lambda x: x.product_id.product_tmpl_id == self.product_tmpl_id
+                    )
+                rec.available_product_lot_sequence_ids = available
+                rec.available_product_lot_sequence_count = len(available)
+            else:
```
rousseldenis, line 37: product_id is required on lot. Is this necessary ?

product_multi_lot_sequence/models/stock_production_lot.py
```diff
@@ -0,0 +1,83 @@
...
+
+    @api.onchange("product_lot_sequence_id")
+    def onchange_product_lot_sequence_id(self):
+        if self.product_lot_sequence_id:
+            self.name = self.product_lot_sequence_id.lot_sequence_id.get_next_char(
+                self.product_lot_sequence_id.lot_sequence_id.number_next_actual
+            )
+
+    def write(self, vals):
+        for _ in self:
```
rousseldenis:
```suggestion
        if vals.get("product_lot_sequence_id", False):
                pls = self.env["product.lot.sequence"].browse(
                            vals.get("product_lot_sequence_id", False)
                        )
                for _ in self:
                    if vals.get("name", False) != pls.lot_sequence_id.get_next_char(
                        pls.lot_sequence_id.number_next_actual
                    ):
                        vals["name"] = pls.lot_sequence_id._next()
                    else:
                        pls.lot_sequence_id._next()
```

rousseldenis: But does your code works in multi ?

As vals["name"] is erased at each loop iteration and only the last one is written.

You should also do a write per iteration

rousseldenis (changes requested):
Your code for write does not work for a multi recordset.

product_multi_lot_sequence/models/stock_production_lot.py
HviorForgeFlow: I was trying to keep product lot sequence field hidden if there is only one option. What do you suggest?

product_multi_lot_sequence/models/stock_production_lot.py
```diff
@@ -0,0 +1,85 @@
...
+            )
+            for rec in self:
+                if vals.get("name", False) != pls.lot_sequence_id.get_next_char(
+                    pls.lot_sequence_id.number_next_actual
+                ):
+                    vals["name"] = pls.lot_sequence_id._next()
+                else:
+                    pls.lot_sequence_id._next()
+                super(ProductionLot, rec).write(vals)
+            return True
```
HviorForgeFlow, line 70: What it should be returned here? @rousseldenis

product_multi_lot_sequence/views/product_views.xml
```diff
@@ -0,0 +1,28 @@
...
+            <page name="inventory" position="after">
+                <page
+                    name="tracking"
+                    string="Tracking"
+                    attrs="{'invisible': [('tracking', '=', 'none')]}"
+                >
+                    <group string="Product Lot Sequence" name="product_lot_sequence">
+                        <field
+                            name="product_lot_sequence_ids"
+                            context="{'default_product_tmpl_id': context.get('product_tmpl_id', active_id)}"
```
HviorForgeFlow, line 19: I'm also having some issues when creating a product lot sequences from the one2many field, product_tmpl_id is not assigned correctly when opening the new record form.

product_multi_lot_sequence/models/stock_production_lot.py
rousseldenis: That's ok

PR #1037, opened 2022-03-18
[MIG] product_sequence: Migration to 15.0

BertVGroenendael:

PR #1086, opened 2022-06-07
[MIG] product_sequence in v15

stvn-vaidehi:


pedrobaeza:
Thanks for the contribution.

Please preserve commit history following technical method explained in https://github.com/OCA/maintainer-tools/wiki/Migration-to-version-15.0.

If the jump is between several versions, you have to modify the source branch in the main command to accommodate it to this circumstance.

Rad0van:
@sts-vaidehi & @pedrobaeza there already is older PR in better shape: #1037

PR #1087, opened 2022-06-08
[ADD] product_template_tags_group

SirTakobi:
Allow to create a group for tags defined by https://github.com/OCA/product-attribute/tree/6e590f70a06dc6ddbd1bae7ce4fe430812c7d7b8/product_template_tags


product_template_tags_group/security/product_template_tag_group_rule.xml
```diff
@@ -0,0 +1,17 @@
...
+    <record id="product_template_tag_group_rule_company" model="ir.rule">
+        <field name="name">Multi-company for Product Tag Group</field>
+        <field name="model_id" ref="model_product_template_tag_group"/>
+        <field name="perm_read" eval="1"/>
+        <field name="perm_create" eval="1"/>
+        <field name="perm_write" eval="1"/>
+        <field name="perm_unlink" eval="1"/>
+        <field name="domain_force">['|', ('company_id', '=', False), ('company_id', 'child_of', [user.company_id.id])]</field>
+    </record>
+</odoo>
```
eLBati, line 17: I think a multi company rule is not need, as the main data are products, which have multi company rules.
I would follow the same approach as `product.category` about this

eLBati (changes requested):

product_template_tags_group/security/product_template_tag_group_rule.xml
eLBati: Nevermind, didn't see https://github.com/OCA/product-attribute/blob/3e94baf1757a5e9d783adf366f4a210660f661dc/product_template_tags/security/product_template_rule.xml

eLBati (approved):
Thanks

product_template_tags_group/security/ir.model.access.csv
```diff
@@ -0,0 +1,2 @@
+id,name,model_id:id,group_id:id,perm_read,perm_write,perm_create,perm_unlink
+manage_product_template_tag_group,manage_product_template_tag_group,model_product_template_tag_group,base.group_user,1,1,1,1
```
carmenbianca, line 2: Comparing this against the permissions in `product`, the permissions should be `1,0,0,0` for `base.group_user` and `1,1,1,1` for `base.group_system`.

carmenbianca (changes requested):
Mostly LGTM, one comment.

product_template_tags_group/security/ir.model.access.csv
SirTakobi: Thanks for the review!
Note that tags groups are not linked to products but only to tags, so these permissions have to be aligned with permissions defined on tags, and right now they are:
> ```
> <odoo>
>     <record id="product_template_tag_access_user" model="ir.model.access">
>         <field name="name">product.template.tag access user (in product_template_tags)</field>
>         <field name="model_id" ref="model_product_template_tag"/>
>         <field name="group_id" ref="base.group_user"/>
>         <field name="perm_read" eval="1"/>
>         <field name="perm_create" eval="1"/>
>         <field name="perm_write" eval="1"/>
>         <field name="perm_unlink" eval="1"/>
>     </record>
> </odoo>
> ```
(ref. https://github.com/OCA/product-attribute/blob/486f3e850158f88e8630a2fba5acda465967fa84/product_template_tags/security/product_template_tag.xml)

Then product tags are used in products (and that might be why you looked at product's permissions), but it looks like who defined product tags chose to use different permissions from products' permissions, here I am sticking with that choice.

carmenbianca: Fair! Resolved.

carmenbianca (approved):

SirTakobi:
@OCA/product-maintainers can this be merged? Thanks!

SirTakobi:
@OCA/product-maintainers can this be merged? Thanks!

eLBati (approved):

SirTakobi:
@OCA/product-maintainers can this be merged? It already has enough reviews, thanks!

SirTakobi:
@OCA/product-maintainers can this be merged? It already has enough reviews, thanks!

PR #1088, opened 2022-06-08
[IMP] product_state (Move Data to Demo)

patrickrwilson:
The stages in the data file is demo data. Having it in data causes issues when the module is updated and the sample stages were deleted as it will repopulate them, even though there is the no update flag.

It was already stated in the roadmap readme for v15, but why not fix it here.

### DIFF
--- a/product_state/__manifest__.py
+++ b/product_state/__manifest__.py
@@ -12,10 +12,10 @@
     "depends": ["product", "sale"],
     "data": [
         "security/ir.model.access.csv",
-        "data/product_state_data.xml",
         "security/ir.model.access.csv",
         "views/product_template.xml",
     ],
+    "demo": ["demo/product_state_demo.xml"],
     "application": False,
     "maintainers": ["emagdalenaC2i"],
     "post_init_hook": "post_init_hook",
--- a/product_state/demo/product_state_demo.xml
+++ b/product_state/demo/product_state_demo.xml
@@ -1,5 +1,5 @@
 <?xml version="1.0" encoding="utf-8" ?>
-<odoo noupdate="1">
+<odoo>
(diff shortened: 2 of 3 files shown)


legalsylvain (commented):
Hi.
question : don't you have to add a pre-migration script to drop the xml_id for people upgrading production database in V14 ?
not sure. Just a question.

rousseldenis:
@Cedric-Pigeon @simahawk What do you think of this ? I'm not sure about it

Cedric-Pigeon (changes requested):
I am not against this proposal but for sure a migration script is required to avoid any breaks on existing installation

rousseldenis:
> I am not against this proposal but for sure a migration script is required to avoid any breaks on existing installation

The problem I see with that is if you install the module on existing installation, existing products will have no state unless you provide too an init_hook... but with which state ?

patrickrwilson:
For existing installations, I don't think we want to drop any existing states because if its currently using any states at all, whether they were installed by this module before the change or not, we wouldn't want to drop the xml's that were previously installed as that can cause issues and impede business workflows already in place.

Also, having no states seems fine, and doesn't cause issues when no states are in the system. Products function just fine without  them.

simahawk:
I missed the ping here...  Arriving late to the party :smile: 

> > I am not against this proposal but for sure a migration script is required to avoid any breaks on existing installation
> 
> The problem I see with that is if you install the module on existing installation, existing products will have no state unless you provide too an init_hook... but with which state ?

For new installations that's not a problem IMO, but it should be documented.

My top concern is to not hinder existing installations.
IMO we should:
1. drop existing XIDs to unlink existing records from the module (should be harmless I think)
2. rename demo records' XIDS w/ `demo_*` prefix

PR #1089, opened 2022-06-14
product_status: fix product_tmpl back to default state

TDu:
Before this change a product would never get back into its default
state. Example: a new proudct with a new until date in the past will keep
a state set as `new`.
Instead it should be set to the default status.


rousseldenis (approved):

ajaniszewska-dev (approved):
LG

PR #1091, opened 2022-06-16
[15.0][MIG] product_custom_info

Rad0van:
Standard migration. Will temporarily fail till https://github.com/OCA/server-tools/pull/2360 gets merged. So anyone interested please review that one as well. Thanx.

A question for those of greater skill than me: as it stands currently, this module provides a way to assign properties to `product.template` and `product.product` as well. But these are different by nature. Now I'm using [base|product]_custom_info to model various devices one of my customers sell - say for example washing machines. So there is a template for them. Some of them have no variants so all the property values are assigned to the product template. However there are some that have variants. In such case most of the values are the same and thus would be assigned to the product template however those specific should be assigned to product variants. But that's currently not possible. I could duplicate the washing machines product template to product variant template but then the properties will be different, etc.
Can anyone please suggest an elegant way to do this? @yajo @rvalyi @pedrobaeza @kittiu Thank you.


Rad0van:
Now that https://github.com/OCA/server-tools/pull/2360 has been merged can someone please reopen this one? Thank you.

bosd:
> Now that https://github.com/OCA/server-tools/pull/2360 has been merged can someone please reopen this one? Thank you.

Quick reply...
There is also this pr:
https://github.com/OCA/product-attribute/pull/1251

Didn't look into differences yet between these pr's.

But it might be quicker to open a new one.

Rad0van:
> > Now that [OCA/server-tools#2360](https://github.com/OCA/server-tools/pull/2360) has been merged can someone please reopen this one? Thank you.
> 
> Quick reply... There is also this pr: #1251
> 
> Didn't look into differences yet between these pr's.
> 
> But it might be quicker to open a new one.

This one is 7 months older than the other which is also closed... Should a new one be created or this one re-opened?  Maybe @yajo could provide an opinion here?

etobella (approved):

PR #1100, opened 2022-07-04
[13.0][IMP] product_sequence: Check for subsequent duplicates

pedrobaeza:

- Install `product_code_unique` OCA module.
- Duplicate a product with a code.
- A new product is created with the code + `-copy`.
- Take the original product and duplicate it again.
- An error is shown about duplicated product code.

Searching previously if the code exists, we prevent this integration problem, and it makes sense to do this distinction even if the other module is not installed.

@Tecnativa

### DIFF
--- a/product_sequence/models/product_product.py
+++ b/product_sequence/models/product_product.py
@@ -59,5 +59,10 @@ class ProductProduct(models.Model):
         if default is None:
             default = {}
         if self.default_code and "default_code" not in default:
-            default.update({"default_code": self.default_code + _("-copy")})
+            code = self.default_code
+            while True:  # check for previous duplicates
+                code += _("-copy")
+                if not self.search([("default_code", "=", code)]):
+                    break
+            default.update({"default_code": code})
         return super().copy(default)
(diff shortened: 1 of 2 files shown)


pedrobaeza:
@cesar-tecnativa please review

product_sequence/models/product_product.py
```diff
@@ -59,5 +59,10 @@ def copy(self, default=None):
         if default is None:
             default = {}
         if self.default_code and "default_code" not in default:
-            default.update({"default_code": self.default_code + _("-copy")})
+            code = self.default_code
+            while True:  # check for previous duplicates
+                code += _("-copy")
+                if not self.search([("default_code", "=", code)]):
+                    break
+            default.update({"default_code": code})
```
cesar-tecnativa, line 67: We can also prevent that the default_code is also inside the db.

`default_code = default.get("default_code", self.default_code)`
`while self.sudo().search([("default_code", "=", default_code)]): default_code += _("-copy")`
`default.update({"default_code": default_code})`

cesar-tecnativa (approved):

pedrobaeza:
Can you review @chienandalu ?

chienandalu (approved):
:+1:

PR #1107, opened 2022-07-08
Bot/14.15

csil300:
Fixed error when changing the state of a product in product_tier_validation.


dreispt:
@csil300 OCA PRs must use their own branches, not aggregation branches.
Cherry picked this change into #1067

PR #862, opened 2021-05-31
[14.0][MIG] product_categ_image

hailangvn:
This is just small migration together with other big one.

PR #1109, opened 2022-07-10
[MIG] product_categ_image: Migration to 13.0

cubells:
Backported from #862 

Added OCA readme files.


product_categ_image/views/product_category_views.xml
```diff
@@ -0,0 +1,13 @@
+<?xml version="1.0" encoding="utf-8" ?>
+<odoo>
+    <record id="product_category_form_view" model="ir.ui.view">
+        <field name="name">product.category.image.form</field>
+        <field name="model">product.category</field>
+        <field name="inherit_id" ref="product.product_category_form_view" />
+        <field name="arch" type="xml">
+            <xpath expr="//field[@name='name']/../.." position="before">
+                <field name="image_1920" widget="image" class="oe_avatar oe_left" />
```
lfreeke: If you do something like this:

`  <group>
                    <div>
                        <field name="image_1920" widget="image" class="oe_avatar oe_left"/>
                    </div>
                 </group>
`

The Category name label won't be next to the image. I think that looks a bit nicer.

lfreeke (approved):
:+1:  Functional test

PR #1110, opened 2022-07-11
[14.0] product_pricelist_revision: fix date_end/date_start are datetime

sebastienbeau:
On version 14 the date_end / date_start on "product.pricelist.item" are datetime, so we should use datetime in the wizard too.

Right now if you set the 2022-06-15 as date_start, you will have (I am in UTC +2)
- date_end: 2022-06-15 02:00:00
- date_stop: 2022-06-14 02:00:00

So you do not have price for the day 2022-06-14

@sergio-teruel @ernestotejeda @pedrobaeza


sergio-teruel (approved):

legalsylvain (approved):
thanks.

pedrobaeza (approved):

PR #1114, opened 2022-07-12
[13.0][MIG] product_pricelist_simulation: Migration to 13.0

ernestotejeda:
Cc @Tecnativa TT37679

Following https://github.com/OCA/margin-analysis/pull/135#issuecomment-1176520345

ping @pedrobaeza @legalsylvain


legalsylvain:
Hi. I'm no sure to understand what you did. The original idea was to use the original module, but it seems you erased all the module and that PR only contain the code of ``product_pricelist_simulation``. Did I missed something ?

product_pricelist_simulation/__manifest__.py
```diff
@@ -1,24 +1,20 @@
...
-    "version": "12.0.1.0.0",
-    "category": "Account",
-    "author": "GRAP,Odoo Community Association (OCA)",
-    "maintainers": ["legalsylvain"],
+    "name": "Product Pricelist Simulation",
+    "summary": "Simulate the product price for all pricelists",
+    "version": "13.0.1.0.0",
+    "category": "Product",
     "website": "https://github.com/OCA/product-attribute",
+    "author": "Tecnativa, Odoo Community Association (OCA)",
```
pedrobaeza: This is missing GRAP

pedrobaeza:
@legalsylvain yes, basically the 13.0 version is based on the "new system", but with hooks for the margin part, that we will do after this merge. Authorship is also lost, which is a mistake of @ernestotejeda

legalsylvain:
> @legalsylvain yes, basically the 13.0 version is based on the "new system", but with hooks for the margin part, that we will do after this merge. Authorship is also lost, which is a mistake.

Thé problem i see is that this PR is loosing a lot of feature of the previous implémentation. Why not take the previous code and just remove the margin part. (And so the sale dependency)

Thanks.

pedrobaeza:
Can you specify which features is losing? We have already coded this version before you told us, so that's why we don't want to start from scratch.

ernestotejeda:
> ... Authorship is also lost, which is a mistake of @ernestotejeda

Fixed

legalsylvain (changes requested):
> Can you specify which features is losing? 

The simulation is incomplete. See readme with screenshots or runbot.

> We have already coded this version before you told us, so that's why we don't want to start from scratch.

I don't ask to start from scratch, but from existing OCA module.
I don't see the interest to ``cherry-pick`` the commits of the OCA module, if then you ``rm`` all the module and replace by your module. (does it make sense ?)
when I'll migrate remaining margin features I'll have to restart from scratch if all the model / views are changed.

In a community way I propose you base your work on the existing one, improving what you think it can be improved. (basically what we talk : module renaming and adding ``show_in_simulation`` in the 
``product.pricelist`` model).

See roadmap : https://github.com/OCA/margin-analysis/pull/135#issuecomment-1176520345

Thanks.

pedrobaeza:
Sorry, @legalsylvain, if you don't specify what is needed, then we'll back again to the initial intention. There can be several OCA modules for similar reasons, but we can't spend more time on this, so decide what you want:

- Tell what are the missing features in the current code and we see the inclusion of them.
- Continue with your module at OCA/margin-analysis and let ours here without convergence.

pedrobaeza:
@legalsylvain our deadline for this module is over, so please tell us what's your final call.

legalsylvain:
> if you don't specify what is needed, then we'll back again to the initial intention.

... On that topic, that's not great. if you had tested on the runboat, the OCA module, or even just read the documentation I mentioned, you would have the answer. The minimum, when an OCA member indicates that a module already exists, is to benchmark it.
But well.

here is the current OCA module : 

image

here is the current implementation : 

image

As you can see, you can not simulate by date, neither by quantity. So the simulation is incomplete.

> Tell what are the missing features in the current code and we see the inclusion of them.

Well, that features are missing. But you'll not have the problem if you had kept the original module
The main problem of my no-go, is that the current PR erase the whole module with the new tecnativa one, changing all model name for no reason. All is lost. Again cherry-pick + rm has no sense.


> @legalsylvain our deadline for this module is over, so please tell us what's your final call.

If you have no time to converge, maybe you should keep your module out of the OCA, and we can plan to converge when you'll have time.
If you don't have time, no problem, when I'll migrate the module in V16, i'll do the required changes splitting basic feature and margin feature in two modules. 

> Continue with your module at OCA/margin-analysis and let ours here without convergence.

Two modules for the same feature, that's not a good idea. It will be a mess to converge after that.

kind regards.

product_pricelist_simulation/wizard/pricelist_simulation.py
```diff
@@ -0,0 +1,78 @@
...
+    pricelist_id = fields.Many2one(
+        comodel_name="product.pricelist", string="Pricelist", readonly=True,
+    )
+    currency_id = fields.Many2one(
+        comodel_name="res.currency",
+        related="pricelist_id.currency_id",
+        string="Currency",
+        readonly=True,
+    )
+    price = fields.Monetary(string="Price Simulated", digits="Price", readonly=True,)
```
pedrobaeza:
```suggestion
    price = fields.Monetary(string="Price", digits="Price", readonly=True,)
```

pedrobaeza:
We can add the features:

- Pricelist date
- Quantity
- Variant when you come from template

The model name used by us is more suitable according the naming scheme. And it's a transient model, so no big deal about migration.

But the price tax included is not valid, as it depends on the customer and their fiscal position, so I don't agree to add such information that may be incorrect.

legalsylvain:
> We can add the features:

* Pricelist date
* Quantity
* Variant when you come from template

Nice ! Thanks !

also could you reuse the current button in the header of the form, instead of the new smart button by convention ? (AFAIK, in Odoo, header button can open popup and smart button make a navigation). 

> The model name used by us is more suitable according the naming scheme. And it's a transient model, so no big deal about migration.

Are you sure that for a transient model, ``pricelist.simulation`` is more suitable as ``wizard.preview.pricelist`` ? That's really a matter of taste... But well, as you want...

> But the price tax included is not valid, as it depends on the customer and their fiscal position, so I don't agree to add such information that it may be incorrect.

You're right, for some fp the display is not correct. (but that's the same as odoo core that is displaying in recent version of odoo price vat excl / vat incl on the product form). 
But anyway as you are removing the account dependency from the current implementation, you'll not have the possibility to compute vat excl and vat incl, so it's not in the scope of what I ask.

in the future module I'll develop ``product_pricelist_simulation_margin`` 
- I'll hide the ``price`` field and display the both prices (vat excl / incl).
- i'll add all the margin computation.
- I'll add a point in the roadmap to say that the vat excl / incl can be wrong in some fp. (that feature can be developped after, if needed).


Deal ?

product_pricelist_simulation/__manifest__.py
```diff
@@ -0,0 +1,20 @@
+# Copyright 2017 Tecnativa - Sergio Teruel
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/lgpl).
+
+{
+    "name": "Product Pricelist Simulation",
+    "summary": "Simulate the product price for all pricelists",
+    "version": "13.0.1.0.0",
+    "category": "Product",
+    "website": "https://github.com/OCA/product-attribute",
+    "author": "GRAP, Tecnativa, Odoo Community Association (OCA)",
```
legalsylvain, line 10: also the maintainer key has been lost.

https://github.com/OCA/margin-analysis/blob/12.0/product_pricelist_margin/__manifest__.py#L10

pedrobaeza:
> also could you reuse the current button in the header of the form, instead of the new smart button by convention ? (AFAIK, in Odoo, header button can open popup and smart button make a navigation).

OK, we can put it as header button.

> Are you sure that for a transient model, pricelist.simulation is more suitable as wizard.preview.pricelist ? That's really a matter of taste... But well, as you want...

Let's rename it to `wizard.pricelist.simulation` !

> You're right, for some fp the display is not correct. (but that's the same as odoo core that is displaying in recent version of odoo price vat excl / vat incl on the product form).

Let's call the column then "Price (taxes excluded)" (cc @ernestotejeda ), and let for the expansion to include the other column, and maybe even adding a selection for the fiscal position to do the tax translation.

legalsylvain:
> OK, we can put it as header button.

:+1: 

> Let's rename it to wizard.pricelist.simulation !

:+1: also.

> Let's call the column then "Price (taxes excluded)" (cc @ernestotejeda ), and let for the expansion to include the other column, and maybe even adding a selection for the fiscal position to do the tax translation.

Are you sure ? 
- without ``account`` module, you can not compute price vat excl. (or vat incl). (Or did I missed something ?)
- if the module is installed in a B2C context, the price vat excl will be useless. (because product.price means price vat incl, so if you display vat excl, the end user will have to recompute manually).

I so think that @ernestotejeda current implementation is correct, and that the introduction of vat excl / vat incl will be done in the margin module. (you can do it in the current module, but IMO, you have to depend on ``account``).
``             "price": variant.with_context(pricelist=pricelist.id).price,``

pedrobaeza:
It's not B2C, but having prices with taxes included, but I agree: let's keep it as simply "Price".

ernestotejeda:
> We can add the features:
> 
> * Pricelist date
> * Quantity
> * Variant when you come from template

in the current implementation, when the button is clicked from a product template, the a price by pricelist is shown for each variants of that template. what do you think about it @legalsylvain ?

image

legalsylvain:
> in the current implementation, when the button is clicked from a product template, the a price by pricelist is shown for each variants of that template. what do you think about it @legalsylvain ?

No idea ! I work only in mono variant context. I développed the feature for variant / template only to be generic to be shared under OCA umbrella.

I like your design because it's possible to see all the variants prices in the same screen. (Not possible in the current OCA module because product-id field is required) In the other hand if we have template with a lot of variant it will be maybe not usable. (Line qty = variant qty * pricelist qty) Why not integrate product-id field but not required ?

ernestotejeda:
Do you mean add in the wizard the product-id field bat editable by the user so that the user can filter by a specific variant?

ernestotejeda:
@legalsylvain @pedrobaeza Could you please check that I haven't missed anything agreed. Thanks.

legalsylvain:
> Do you mean add in the wizard the product-id field but editable by the user so that the user can filter by a specific variant?

Yes, or let empty to have all the variants data. User could have choice.

ernestotejeda:
@legalsylvain , done. review it, please.

product_pricelist_simulation/models/pricelist.py
```diff
@@ -0,0 +1,10 @@
+# Copyright 2017 Tecnativa - Sergio Teruel
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+from openerp import fields, models
+
+
+class ProductPricelist(models.Model):
+    _inherit = "product.pricelist"
+
+    show_in_simulation = fields.Boolean(string="Show in simulation")
```
legalsylvain:
```suggestion
    show_in_simulation = fields.Boolean(string="Show in simulation", default=True)
```

That option is a great addition to hide some useless pricelist.
However, it should be enabled by default. (User installing that module expect that the feature is enabled).

product_pricelist_simulation/wizards/wizard_preview_pricelist_views.xml
```diff
@@ -0,0 +1,67 @@
...
+    <record id="view_pricelist_simulation_view_form" model="ir.ui.view">
+        <field name="name">wizard.preview.pricelist.form</field>
+        <field name="model">wizard.preview.pricelist</field>
+        <field name="arch" type="xml">
+            <form>
+                <field name="variant_count" invisible="1" />
+                <group name="settings_group">
+                    <group name="settings_left_group">
+                        <field name="template_id" />
+                        <field name="product_id" options="{'no_create': True}" />
```
legalsylvain:
```suggestion
                        <field name="product_id" options="{'no_create': True}" attrs="{'invisible': [('variant_count', '<', 2)]}"/>
```

I think this field is useless in mono variant product template and can be hidden.

product_pricelist_simulation/readme/USAGE.rst
```diff
@@ -0,0 +1,3 @@
+To use this module, you need to:
+
+#. Go to product form view and click '*Prices*'' smart button.
```
legalsylvain:
```suggestion
#. Go to product form view and click '*See Price per Pricelist*' button.
```

legalsylvain (approved):
Minor comments inline. 

LGTM, thanks. (code review + test on runboat).

Possibility to display price for all variants or just for one is a great addition.

pedrobaeza (approved):
Pushed the proposed changes

cesar-tecnativa (approved):

pedrobaeza:
@legalsylvain everything OK in your side?

legalsylvain (approved):
Perfect ! 

thanks a lot for this common work ! 

i'll do a refactor work for the current https://github.com/OCA/margin-analysis/tree/12.0/product_pricelist_margin when I'll migrate in V16. 

Is it ok for you if the name is ``product_pricelist_simulation_margin`` ? (margin-analysis repository).

pedrobaeza:
Yeah, seems good.

/ocabot merge nobump

PR #1120, opened 2022-07-20
[15.0][Add] product_manufacturer: Add Polish translation

jacekmichalski:


pedrobaeza:
Please do this through https://translation.odoo-community.org/

PR #1122, opened 2022-07-24
[14.0][IMP] product_sequence: Check for subsequent duplicates

pedrobaeza:

- Install `product_code_unique` OCA module.
- Duplicate a product with a code.
- A new product is created with the code + `-copy`.
- Take the original product and duplicate it again.
- An error is shown about duplicated product code.

Searching previously if the code exists, we prevent this integration problem, and it makes sense to do this distinction even if the other module is not installed.

Forward-port of #1100 

@Tecnativa

### DIFF
--- a/product_sequence/models/product_product.py
+++ b/product_sequence/models/product_product.py
@@ -59,5 +59,10 @@ class ProductProduct(models.Model):
         if default is None:
             default = {}
         if self.default_code and "default_code" not in default:
-            default.update({"default_code": self.default_code + _("-copy")})
+            code = self.default_code
+            while True:  # check for previous duplicates
+                code += _("-copy")
+                if not self.search([("default_code", "=", code)]):
+                    break
+            default.update({"default_code": code})
         return super().copy(default)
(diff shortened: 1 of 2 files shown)


pedrobaeza:
@chienandalu @cesar-tecnativa please review

Rad0van:
@pedrobaeza while the code is there I fail to replicate that behavior. When I duplicate the product the next in sequence code gets assigned to the copy. When I duplicate it again, next (so +2 against the original) is assigned to the second copy.

pedrobaeza:
Well, this problem was detected in 13.0 when in combination with `product_code_unique`, raising the error.

Rad0van:
> Well, this problem was detected in 13.0 when in combination with `product_code_unique`, raising the error.

Well I am testing this on 15.0 with product_code_unique being a depedency (and it's working). However the code seems to be that same so I don't see how that could happen.
Update: I did a quick debugging session and the copy method is not called at all. but create is called instead. I am using "Duplicate" action.

pedrobaeza:
But this is for v14.

Rad0van:
> But this is for v14.

I know it is for v14 but I didn't expect the behavior change in v15. Will test on v14 as well.

chienandalu (commented):
I'd say the issue to reproduce it is that this `copy()` method belongs to the product variant which has UI Duplicate disabled by default. So duplicating the template is not going to enter this method (it creates the variant (`product.product`), as @Rad0van observed).

Should the method be on product template instead? Are the tests wrong?

PR #1123, opened 2022-07-27
[14.0][IMP] product_categ_image: Remove image field leftover from migration

HekkiMelody:
The binary image field is no longer used. As a matter of fact, the image.mixin was added, and the field `image_1920` is used in the view.

the image field was most likely left over from the migration, and can be removed.


HekkiMelody:
Codecov is less than before because I removed a field...

HekkiMelody:
bumping against the stale label

Issue #1121, opened 2022-07-22
[12.0-13.0-14.0][IMP] product_state: improve security groups?


HekkiMelody:
**Is your feature request related to a problem?**
https://github.com/OCA/product-attribute/blob/14.0/product_state/security/ir.model.access.csv
Currently, Create Write and Delete permissions are given to `base.group_partner_manager`. That group's name is "Contact Creation" and doesn't seem at all related with managing product states

**Describe the solution you'd like**
I'd suggest changing the security group to "inventory manager" (`stock.group_stock_manager`) instead. Seems like a better default, unless there's a reason I'm not aware of. I'd be happy to make the PR myself if it seems like a good change.

BT-mgomez:
Hello, I agree on this.
I saw this access right was introduced in v12 and is up to v15. 
Is there a way to track this down and someone who will decide on this issues? 

Was going to create the same issue, but for v15, then just found this one.

PR #1124, opened 2022-07-27
[14.0][IMP] product_state: move access to Sales + Inventory Manager

HekkiMelody:
https://github.com/OCA/product-attribute/issues/1121

### DIFF
--- a/product_state/security/ir.model.access.csv
+++ b/product_state/security/ir.model.access.csv
@@ -1,3 +1,4 @@
 "id","name","model_id:id","group_id:id","perm_read","perm_write","perm_create","perm_unlink"
-"access_product_state_product_manager","product.state","model_product_state","base.group_partner_manager",1,1,1,1
+"access_product_state_sale_manager","product.state","model_product_state","sales_team.group_sale_manager",1,1,1,1
+"access_product_state_stock_manager","product.state","model_product_state","stock.group_stock_manager",1,1,1,1
 "access_product_state_public","product.state.public","model_product_state",,1,0,0,0
--- a/product_status/tests/test_product_status.py
+++ b/product_status/tests/test_product_status.py
@@ -120,7 +120,8 @@ class TestProductStatusCase(TestProductCommon):
 
     def test_modified_default_data(self):
         st_env = self.env["product.state"]
-        demo_user = self.env.ref("base.user_demo").id
+        demo_user = self.env.ref("base.user_demo")
+        demo_user.groups_id = [(4, self.env.ref("sales_team.group_sale_manager").id)]
         default_state = st_env._get_module_data()
         vals = {
             "name": "State change",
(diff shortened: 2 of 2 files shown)


HekkiMelody:
Had to modify the tests of a dependent module, as a result.

lfreeke (approved):
Functional test :+1:

emagdalenaC2i (approved):

Kev-Roche (approved):

HekkiMelody:
@OCA/product-maintainers sorry for the ping, but could you please merge this?

dreispt:
Shouldn't this access be given to the Inventory Manager, rather than the Sales Manager?

Kev-Roche:
> Shouldn't this access be given to the Inventory Manager, rather than the Sales Manager?

Indeed, it's a good question. Maybe the answer is both because some products with service type are not in stock and some projects are more focused on sale than stock (and opposite).

HekkiMelody:
> > Shouldn't this access be given to the Inventory Manager, rather than the Sales Manager?
> 
> Indeed, it's a good question. Maybe the answer is both because some products with service type are not in stock and some projects are more focused on sale than stock (and opposite).

That was my reasoning as well. I've added both.

legalsylvain:
/ocabot merge patch

thanks for your contribution. Feel free to port this patch upstream, if necessary.

regards.

natuan9:
Hi @aleuffre , I noticed you moved access to Inventory Manager but didn’t add a dependency on the `stock` module. Doesn’t this cause any issues? I also see that the migration PRs don’t mention this either.
Currently, I’m migrating this module to version 18.0, and I’m uncertain whether I should add a dependency on the stock module. Could you clarify this for me?

HekkiMelody:
> Hi @aleuffre , I noticed you moved access to Inventory Manager but didn’t add a dependency on the `stock` module. Doesn’t this cause any issues? I also see that the migration PRs don’t mention this either. Currently, I’m migrating this module to version 18.0, and I’m uncertain whether I should add a dependency on the stock module. Could you clarify this for me?

I think you're correct. I thought there was an indirect dependency based on the module's existing dependencies, but it seems I was wrong (or maybe I just missed it at the time, I don't remember after almost a year). You should definitely add this fix when migrating. Thank you!

simahawk:
Guys, this change is breaking the module. The dependency on `stock` has not been declared.
**EDIT:** I've just seen the msg by @natuan9, sorry for the repetition :)

Sounds like it was never tested in isolation (= install only this module).
Also is arguable that we have to add a full dep on `stock` just to use its group.
IMO it should be done like this:
* add a specific group in this module
* add an auto-installable module that depends on stock that add this group as implied on inv manager

BTW even the dep on sale could be ripped out as it's needed only for the menu...

simahawk:
> > Hi @aleuffre , I noticed you moved access to Inventory Manager but didn’t add a dependency on the `stock` module. Doesn’t this cause any issues? I also see that the migration PRs don’t mention this either. Currently, I’m migrating this module to version 18.0, and I’m uncertain whether I should add a dependency on the stock module. Could you clarify this for me?
> 
> I think you're correct. I thought there was an indirect dependency based on the module's existing dependencies, but it seems I was wrong (or maybe I just missed it at the time, I don't remember after almost a year). You should definitely add this fix when migrating. Thank you!

I would proceed the other way around. We remove the need for the dep in v18 and you can backport the fix here :wink:

HekkiMelody:
Apologies again for the mistake on my part.

I don't have a strong opinion on how to fix it either way, in fact I like your solution best @simahawk , but I've definitely seen small modules blocked to avoid multiplying the number of modules in OCA...

PR #1130, opened 2022-08-04
[13.0][FIX] product_pricelist_direct_print: filter products

chienandalu:
- When we set additional filters, we expect those to be applied event if the products are directly set.
- When we check for the last ordered items we better check all the commercial partner ones.

cc @Tecnativa TT38408

please take a look @CarlosRoca13 @carlosdauden

### DIFF
--- a/product_pricelist_direct_print/wizards/product_pricelist_print.py
+++ b/product_pricelist_direct_print/wizards/product_pricelist_print.py
@@ -226,7 +226,8 @@ class ProductPricelistPrint(models.TransientModel):
         if not partner and self.partner_count == 1:
             partner = self.partner_ids[0]
         orders = self.env["sale.order"].search(
-            self._get_sale_order_domain(partner), order="date_order desc"
+            self._get_sale_order_domain(partner.commercial_partner_id),
+            order="date_order desc",
         )
         products = orders.mapped("order_line").mapped("product_id")
         return products[: self.last_ordered_products]
@@ -285,8 +286,10 @@ class ProductPricelistPrint(models.TransientModel):
 
     def get_products_to_print(self):
         self.ensure_one()
+        domain = self.get_products_domain()
         if self.last_ordered_products:
             products = self.get_last_ordered_products_to_print()
+            products = products.filtered_domain(domain)
         else:
             if self.show_variants:
                 products = self.product_ids or self.product_tmpl_ids.mapped(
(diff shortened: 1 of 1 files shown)


product_pricelist_direct_print/wizards/product_pricelist_print.py
```diff
@@ -294,8 +295,16 @@ def get_products_to_print(self):
...
+        domain = self.get_products_domain()
+        # We default to the products model search but if the last ordered products
+        # checkbox is on we don't want to confuse the user with products which don't
+        # belong to their sales. It's better to show the empty report in that case.
+        if not products and not self.last_ordered_products:
             products = products.search(self.get_products_domain())
+        else:
+            # We need to filter the results anyway, as there could be some extra filters
+            # to consider, like categories.
+            products = products.filtered_domain(domain)
```
chienandalu: I'm in doubt here. Should we filter manually selected products?

product_pricelist_direct_print/wizards/product_pricelist_print.py
```diff
@@ -226,7 +226,8 @@ def get_last_ordered_products_to_print(self):
         if not partner and self.partner_count == 1:
             partner = self.partner_ids[0]
         orders = self.env["sale.order"].search(
-            self._get_sale_order_domain(partner), order="date_order desc"
+            self._get_sale_order_domain(partner.commercial_partner_id),
```
chienandalu, line 229: Could it be interesting to make this configurable in the wizard?

CarlosRoca13: I think that the change makes sense and it's not necessary to add an extra configuration.

product_pricelist_direct_print/wizards/product_pricelist_print.py
CarlosRoca13: If the products are selected manually is because they want to print each products, so I think that it shouldn't be filtered

chienandalu: Yes, I feel the same. It's a matter of UI then.

chienandalu:
Changes done

CarlosRoca13 (approved):
:+1:

chienandalu:
Re-ping @carlosdauden

carlosdauden (approved):
Only code review

PR #1133, opened 2022-08-23
15.0 mig product pricelist supplierinfo

Galoperd:
[15.0][MIG] product_pricelist_supplierinfo: migration to 15.0


Galoperd:
Hello, migration ready for review. I would porpoise to change the current margin formula of the module (=fixed margin amount with amount limit) to a **percentage margin over the purchasing price**. From my experience, this is a more common sale price calculation method.

PR #1146, opened 2022-09-15
[15.0][FIX] product_code_mandatory: align code with Odoo way to set default code

augusto-weiss:
The Issue: When install this module, it generate diferents default codes for each product_product and product_template where default_code is not set.

Example: 
for product_product(1) generate the default code: DEFAULT4
for the product template of product_product(1) enerate the default code: DEFAULT9

Odoo:
Odoo has a method called _compute_default_code(), where de default_code of product_template is asseigned depending of its variants

### DIFF
--- a/product_code_mandatory/__init__.py
+++ b/product_code_mandatory/__init__.py
@@ -1,17 +1,27 @@
 from . import models
 
+from odoo import api, SUPERUSER_ID
+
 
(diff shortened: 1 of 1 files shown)


jjscarafia (approved):
LGTM

andreampiovesana (approved):
LGTM

PR #1156, opened 2022-10-03
[13.0][IMP] stock_account_product_cost_security: additional permission

chienandalu:
It could be the case that even if users are allowed to see product costs, they shouldn't be able to edit them.

cc @Tecnativa TT39711

please review @sergio-teruel @LudLaf

### DIFF
--- a/stock_account_product_cost_security/README.rst
+++ b/stock_account_product_cost_security/README.rst
@@ -25,15 +25,28 @@ Stock Account Product Cost Security
 
 |badge1| |badge2| |badge3| |badge4| |badge5| 
 
-This module depends on product_cost_security and stock_account modules to
-help their to hide some elements for product cost security group.
-
-This module is self-installable.
+This module is a complement to `product_cost_security` and `stock_account` to add
+and extra permission level on product cost editing.
 
 **Table of contents**
 
 .. contents::
    :local:
+
+Installation
+============
+
+This module auto installs when `product_cost_security` and `stock_account` are already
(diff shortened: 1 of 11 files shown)


LudLaf (approved):
Module tested. Working as expected.

sergio-teruel (approved):

andres-pcg:
@pedrobaeza @chienandalu This merge is breaking Odoo test since openupgrade is not in the requirements file for this repo.
Screen Shot 2022-10-12 at 12 50 49 PM

pedrobaeza:
This is only needed for updating the module, not for fresh installs, so it shouldn't be in the requirements. Please add it because you will find these requirement through whole Odoo. It's a must.

andres-pcg:
The instance that was broken was a current working database, not a new one. The only thing we did was updating submodules (including this one) and it broke the whole workflow. There's no openupgrade in the requirements for any of the OCA repos (v13) we're currently using.

andres-pcg:
If this repo or any of its dependencies depends on a library, it should be added to the requirements file @pedrobaeza

pedrobaeza:
Please read again. It's just the reverse case.

andres-pcg:
I've found some repos with this dependency, but in different branches. eg: https://github.com/OCA/server-tools/blob/14.0/requirements.txt

andres-pcg:
Anyways, I manually added the dependency to that repo. Thank you for the feedback @pedrobaeza

PR #1151, opened 2022-09-29
[15.0][MIG] product_logistics_uom

flachica:

PR #1160, opened 2022-10-04
[16.0] [MIG] product_logistics_uom

marielejeune:


rousseldenis:
@marielejeune There was already a 15.0 migration in progress.

#1151

product_logistics_uom/models/product_template.py
```diff
@@ -0,0 +1,72 @@
...
+class ProductTemplate(models.Model):
+    _inherit = "product.template"
+
+    volume_uom_id = fields.Many2one(
+        "uom.uom",
+        string="Volume Unit of Measure",
+        domain=lambda self: [
+            ("category_id", "=", self.env.ref("uom.product_uom_categ_vol").id)
+        ],
+        default=lambda x: x._get_volume_uom_id_from_ir_config_parameter(),
```
apineux: Why use the x instead of self? I prefer not to use one character variables.

product_logistics_uom/models/product_template.py
```diff
@@ -0,0 +1,72 @@
...
+    )
+
+    weight_uom_id = fields.Many2one(
+        "uom.uom",
+        string="Weight Unit of Measure",
+        domain=lambda self: [
+            ("category_id", "=", self.env.ref("uom.product_uom_categ_kgm").id)
+        ],
+        compute=False,
+        default=lambda x: x._get_weight_uom_id_from_ir_config_parameter(),
```
apineux: Same comment -> it is better to use self instead of one character variable

apineux (changes requested):

product_logistics_uom/models/product_template.py
marielejeune: Done

product_logistics_uom/models/product_template.py
marielejeune: Done

lmignon (approved):
@marielejeune Thank you for the migration. :+1: (Code review + functional tests)

rousseldenis (approved):

rvalyi (approved):
tested

PR #1163, opened 2022-10-04
product_template_tags: add menu

sebastienbeau:
Add missing menu, so you can see, import, remove tags

### DIFF
--- a/product_template_tags/views/product_template_tag.xml
+++ b/product_template_tags/views/product_template_tag.xml
@@ -93,4 +93,12 @@
             <p class="o_view_nocontent_empty_folder">Add tags</p>
         </field>
     </record>
+
+    <menuitem
+        id="product_template_tag_menu"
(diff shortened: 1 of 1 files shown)


ivantodorovich:
Hello!

`product_template_tags` doesn't depend on `sale`.

This is why the menues are implemented in separate glue modules:
- [sale_product_template_tags](https://github.com/OCA/product-attribute/tree/14.0/sale_product_template_tags)
- [purchase_product_template_tags](https://github.com/OCA/product-attribute/tree/14.0/purchase_product_template_tags)
- [stock_product_template_tags](https://github.com/OCA/product-attribute/tree/14.0/stock_product_template_tags)

I know, it's not great but seems like it's the only way of having the menues without introducing the app dependencies

sebastienbeau:
Ok I didn't see it. Indeed it's really small glue module, but you are right it's the best way to do it

PR #1164, opened 2022-10-06
[13.0][FIX] product_supplierinfo_for_customer: Crash when the price is computed for a template instead of variant

sergio-teruel:
FWPort https://github.com/OCA/product-attribute/pull/777

cc @Tecnativa TT27887

ping @chienandalu


pedrobaeza (approved):

pedrobaeza:
Let's merge this one forward port for not being desynchronized:

/ocabot merge patch

@sergio-teruel can you please check if this should be forward ported to upper versions as well?

PR #1181, opened 2022-10-31
[ADD] product_cost_read_only

diogocsc:
With this PR we add the module product_cost_read_only, which was inspired on product_cost_security, but instead of hiding the field when the user does not have the permission, it just makes it read only.


dreispt (approved):

rafaelbn:
Hello @diogocsc @dreispt  ,

I think we can better improve `product_cost_security` instead of creating a new module

In fact this is already improved and this PR/module is not needed any more: https://github.com/OCA/product-attribute/pull/1186

Isn't it?

Regards

PR #1182, opened 2022-10-31
16.0 mig product secondary unit

otmanelagy:


otmanelagy:
thank for checking the migration @legalsylvain

PR #1145, opened 2022-09-15
[15.0][MIG] product_lot_sequence

i-vyshnevska:
plus include
https://github.com/OCA/product-attribute/pull/1111

PR #1188, opened 2022-11-07
[15.0][IMP] product_lot_sequence: Add sequence policies 

grindtildeath:
Provide multiple configuration options with this module installed:

- "product": This is the default behaviour once you install this module. It's the
  same than in previous Odoo versions with this module installed, i.e. it allows
  to define a dedicated sequence on each product.

- "global": This was the default behaviour from previous Odoo versions when this
  module was not installed, i.e it will always use the same global sequence for every product.

Depends on:
 - [x] #1145

### DIFF
--- a/product_lot_sequence/data/ir_config_parameter.xml
+++ b/product_lot_sequence/data/ir_config_parameter.xml
@@ -1,0 +1,7 @@
+<?xml version="1.0" encoding="utf-8" ?>
+<odoo noupdate="1">
+    <record model="ir.config_parameter" id="lot_sequence_policy">
+        <field name="key">product_lot_sequence.policy</field>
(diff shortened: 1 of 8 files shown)


product_lot_sequence/data/ir_config_parameter.xml
```diff
@@ -0,0 +1,7 @@
+<?xml version="1.0" encoding="utf-8" ?>
+<odoo>
```
i-vyshnevska: noupdate="1"

ajaniszewska-dev (approved):
seems good

yankinmax (approved):

simahawk:
@grindtildeath you can rebase

PR #1198, opened 2022-11-15
[15.0][FIX] product_manufacturer: Change SavepointCase to TransactionCase

victoralmau:
Change SavepointCase to TransactionCase

Please @pedrobaeza can you review it?

@Tecnativa

### DIFF
--- a/product_manufacturer/tests/test_product_manufacturer.py
+++ b/product_manufacturer/tests/test_product_manufacturer.py
@@ -1,10 +1,10 @@
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
 
 
-from odoo.tests.common import SavepointCase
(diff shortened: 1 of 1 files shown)


pedrobaeza (approved):
I think this one is going to fail also, but let's try:

/ocabot merge patch

rousseldenis:
@pedrobaeza We need to update copier for the problem of flake8 pre-commit url to have a functional merge. I can take care of even branches :-)

pedrobaeza:
Yes, fixed in #1200

We are updating those that involve pull requests that we are merging:

/ocabot merge patch

Any help is always appreciated. It's a pity that this didn't come at the same time for not having to do the same twice, but we can't decide that...

victoralmau:
Can you try the merging process again?

PR #1200, opened 2022-11-15
[15.0][FIX] Update copier template

pedrobaeza:
To avoid Flake8 Gitlab problem.

@Tecnativa


legalsylvain:
Thank you for this tedious work...
(and thanks for the solution, I have similar problem on my custom repo today.)

pedrobaeza:
Yeah, 2 errors in 2 weeks that make us to update 2 times the copier template (#1187 and this one), but well, it's something needed...

PR #1205, opened 2022-11-21
[13.0][FIX] Updated from template

victoralmau:
Updated from template

Ping @pedrobaeza 

@Tecnativa


requirements.txt
```diff
@@ -1,2 +1 @@
-odoo_test_helper
```
pedrobaeza, line 1: You should put these both in `test_requirements.txt`

pedrobaeza (approved):

PR #1214, opened 2022-11-28
[MIG][16.0] stock_production_lot_expiry

glitchov:
migration from 10.0


stock_production_lot_expiry/models/product_template.py
```diff
@@ -23,11 +23,12 @@ class ProductTemplate(models.Model):
     category_lot_expiry_field_name = fields.Selection(
         related="categ_id.lot_expiry_field_name",
         readonly=True,
+        string="category_lot_expiry_field_name",
```
rousseldenis: Maybe something more explicit ???

rousseldenis (approved):
LGTM. @glitchov Could you fix string ?

rousseldenis:
@alexis-via Maybe this one could be your interest

rousseldenis (approved):

alexis-via:
Instead of adding a new computed field "is_expiry", you could use the native computed field "product_expiry_alert" (https://github.com/odoo/odoo/blob/16.0/addons/product_expiry/models/production_lot.py#L21) and add a search method on it.

This module adds "yet another" date field on production lot... so it's a total of 5 date fields (4 datetime and 1 date). With some imagination, I think we could reach 10 date fields anytime soon :)
I'm sorry, but I prefer my module "product_expiry_simple", with just 1 date field... so much simpler and easier to understand for users.

stock_production_lot_expiry/models/stock_lot.py
```diff
@@ -0,0 +1,58 @@
...
+
+from odoo import _, api, fields, models
+from odoo.osv.expression import NEGATIVE_TERM_OPERATORS
+
+
+class StockProductionLot(models.Model):
+
+    _inherit = "stock.lot"
+
+    is_expired = fields.Boolean(
```
rousseldenis: @jbaudoux @glitchov  Maybe we could use the `product_expiry_alert`core field as that one is used in some flows, for instance there : https://github.com/odoo/odoo/blob/16.0/addons/product_expiry/models/stock_move_line.py#L16

sbejaoui (approved):
LGTM (code review),

should we rename the module to `stock_lot_expiry`?

lmignon:
@OCA/product-maintainers Someone to merge this one?

jbaudoux (commented):
A lot is expired on its expiration date. And it is supposed to be removed from stock on its removal date.

It is not clear the intent of this module reading the description. Looks weird.

rousseldenis:
@lmignon

lmignon:
@rousseldenis I propose to close this one... (It makes non sense)

Issue #1213, opened 2022-11-23
product_packaging_type: Migrate to 16.0


rousseldenis:
@jgrandguillaume I want your advice here.

As package type is standard now in Odoo, I'm wondering what name given to the product_packaging_type module as some features are missing in core.

Should we keep the original name, choose a name like stock_package_type_oca, or ... ?

rousseldenis:
> @jgrandguillaume I want your advice here.
> 
> As package type is standard now in Odoo, I'm wondering what name given to the product_packaging_type module as some features are missing in core.
> 
> Should we keep the original name, choose a name like stock_package_type_oca, or ... ?

After first discussion, it has more common sense to split per introduced functionallity.

So, at first sight stock_package_type_gtin, stock_package_type_default, stock_pacakge_type_quantity

@sebalix @grindtildeath @jbaudoux @gurneyalex @lmignon

rousseldenis:
And is it too soon to require a separate repository for this ?

jbaudoux:
The right name should be product packaging level.
But as there was already package level in odoo, to prevent confusion, we opt for packaging type. Now that this name exist also in standard odoo, I would go back to the common name "product_packaging_level"

And I would not split the module.

rousseldenis:
> The right name should be product packaging level. But as there was already package level in odoo, to prevent confusion, we opt for packaging type. Now that this name exist also in standard odoo, I would go back to the common name "product_packaging_level"
> 
> And I would not split the module.

@jbaudoux So, to be clear, it's not the same as in core :

![image](https://user-images.githubusercontent.com/19529533/203752392-8b2fc955-2bec-47d8-a2ac-1c306821a8be.png)

rousseldenis:
And it's not comparable as :

![image](https://user-images.githubusercontent.com/19529533/203753334-7eef1e5f-cd3d-44e5-af41-51001babf5b5.png)

jbaudoux:
Indeed. On the product packaging, we will have the core `package_type_id` which defines the put-away rule for storing the goods. And the new `package_level_id` which defines if it is a consumer unit, a transport box or a pallet. That package level will serve in the SO and PO to express if you want to sell / purchase a multiple of consumer unit, box, or pallets

rousseldenis:
[...]

PR #1215, opened 2022-11-29
[MIG][16.0] product_packaging_type + renaming to product_packaging_level

rousseldenis:
Following this discussion:

https://github.com/OCA/product-attribute/issues/1213

Depends on:

- https://github.com/OCA/openupgradelib/pull/312


rousseldenis:
@ChrisOForgeFlow  @LoisRForgeFlow You should be interested by this.

I've put changes in separate commits, so you maybe want to port this to 15

product_packaging_level/readme/USAGE.rst
```diff
@@ -0,0 +1,4 @@
+To create a new packaging level:
+
+#. Go to *Sales > Configuration > Products > Product Packaging Level*.
```
lmignon:
```suggestion
#. Go to *Inventory > Configuration > Products > Product Packaging Level*.
```

lmignon (changes requested):
@rousseldenis The UI isn't working at all.... It's not possible to add new packaging on the product forms.. the packaging type remains always the same etc...

rousseldenis:
> @rousseldenis The UI isn't working at all.... It's not possible to add new packaging on the product forms.. the packaging type remains always the same etc...

Ok, I check

rousseldenis:
@lmignon Seems better now

lmignon (commented):
This code LGTM (also functional tests) but I've some difficulties to understand the difference between packaging type and packaging level. Can you elaborate and add some informations into the readme plz.

rousseldenis:
@jbaudoux Maybe an advice here on how to describe functionally this module and how this is different from packaging type?

product_packaging_level/readme/DESCRIPTION.rst
```diff
@@ -0,0 +1 @@
+This module allows odoo users to declare packaging levels (pallets, ...)
```
jbaudoux:
```suggestion
This module binds a product packaging to a packaging level.
There are usually 3 levels:
* The primary: it is the closest to the product, also called the retail packaging
* The secondary: it allows to ship the products already in the primary packaging, like a box
* The tertiary: it allows to ship large quantities in warehousing, like a pallet

This differentiates from the odoo standard package type that is typing the packaging in a more precise way and that can serve for put-away routing in the warehouse. For example a small box and a large box, or an EPAL pallet and  CHEP pallet.
```

jbaudoux (commented):
It is missing a small feature to ease the configuration. On the package type, we should be able to configure a package level and when we select a package type on the product packaging, it should set the right package level on the product packaging.

product_packaging_level/hooks.py
remytms, line 1: @rousseldenis I have a question. Why putting this update of the table in a `pre-init-hook` instead of putting it in a `pre-migration.py` script in the migration directory of this module ?

Is the `pre-init-hook` launched when performing an update of the module ? I thought that it only runs when installing the module, but I may be wrong.

bealdav: Not sure on my side but I suppose there is no update here.
You install a new module (new name).
Installing it, you move data with this hook.
It seems logic to me but @rousseldenis will give the good response

product_packaging_level/models/product_packaging_level.py
```diff
@@ -0,0 +1,34 @@
...
+from odoo import _, api, fields, models
+from odoo.exceptions import ValidationError
+
+
+class ProductPackagingLevel(models.Model):
+    _name = "product.packaging.level"
+    _description = "Level management for product.packaging"
+    _order = "sequence, code"
+
+    name = fields.Char(required=True, translate=True)
```
bealdav, line 12: Not blocking
I suppose a majority of projects using this module, will use one lang here.
If not, each of the concerns projects may make this fields translatable.
When you copy this record, copy translation comes with nightmares for consistency of naming
I know Odoo use this default settings. They demonstrate its capability. Not sure this is a good choice.

product_packaging_level/readme/DESCRIPTION.rst
bealdav: Really explicit description. Thanks

bealdav (commented):
Thanks for this renaming, make it more explicit.
I would like to test with runboat before make final review.
Could you try to regenerate it ? 
Thanks

product_packaging_level/models/product_packaging_level.py
jbaudoux: You have no reason to copy a package level. You will configure a very limited amount of levels to use on all the product packagings.
As this can be very useful on e-commerce, it makes sense to have it translatable

rousseldenis:
> Thanks for this renaming, make it more explicit. I would like to test with runboat before make final review. Could you try to regenerate it ? Thanks

@bealdav Done

product_packaging_level/models/product_packaging_level.py
bealdav: ok

lmignon (approved):

jbaudoux:
@rousseldenis Can you update the readme and add the missing link between package type & level. Cf https://github.com/OCA/product-attribute/pull/1215#pullrequestreview-1364793239

bealdav (approved):
LGTM, really thanks.

Thanks to update doc.

jbaudoux (changes requested):
https://github.com/OCA/product-attribute/pull/1215#issuecomment-1582772805

product_packaging_level/hooks.py
```diff
@@ -0,0 +1,25 @@
...
+
+from odoo import SUPERUSER_ID, api
+
+
+def pre_init_hook(cr):
+    if openupgrade.table_exists(cr, "product_packaging_type"):
+        env = api.Environment(cr, SUPERUSER_ID, {})
+        # Former version of the module is present
+        models = [("product.packaging.type", "product.packaging.level")]
+        openupgrade.rename_models(env.cr, models)
```
nilshamerlinck, line 13: don't we also need to `rename_tables` to keep data?

```
tables = [("product_packaging_type", "product_packaging_level")]
openupgrade.rename_tables(env.cr, tables)
```

product_packaging_level/hooks.py
```diff
@@ -0,0 +1,25 @@
...
+        fields = [
+            (
+                "product.packaging",
+                "product_packaging",
+                "packaging_level_id",
```
nilshamerlinck, line 18:
```suggestion
                "packaging_type_id",
```

product_packaging_level/models/__init__.py
```diff
@@ -0,0 +1 @@
+from . import product_packaging, product_packaging_level
```
bealdav: Your code is fine with a fresh empty database, no problem.

But in the case of my project in production with many modules installed I've this error
level

```suggestion
from . import product_packaging_level, product_packaging
```

With this suggestion I've no more error because product_packaging_level table is already a created 
when trying to define packaging_level_id m2o.

This is a really rare case that I have encountered once regarding the import order of odoo models for about ten years

cc @kevinkhao @rousseldenis

rousseldenis: @bealdav Done

product_packaging_level/hooks.py
rousseldenis: @remytms @bealdav Because as the module is renamed, it is not yet installed. So, pre-init is the good entry point.

product_packaging_level/models/__init__.py
bealdav: Thanks a lot, good job

lmignon (approved):

bealdav (approved):

flotho:
hi @rousseldenis , any chance to mark solved the comments and add test to be fully green ?
Or maybe as we have already have 2 positive review we could merge this one ?

rousseldenis:
@jbaudoux Done

jbaudoux (approved):
Let's merge this. The small missing feature can be done in a later PR

product_packaging_level/__manifest__.py
```diff
@@ -0,0 +1,24 @@
+# Copyright 2019-2020 Camptocamp (<https://www.camptocamp.com>).
+# License LGPL-3.0 or later (https://www.gnu.org/licenses/lgpl).
+
+{
+    "name": "Product Packaging Type",
```
jbaudoux:
```suggestion
    "name": "Product Packaging Level",
```

product_packaging_level/hooks.py
MiquelRForgeFlow: this is lacking

product_packaging_level/hooks.py
MiquelRForgeFlow: But the rename should be done in OpenUpgrade (see apriori.py), and with migration scripts in migration folder, you are more in control. For example, you are lacking here the post-migration script to change noupdate=1 data changes.

PR #1161, opened 2022-10-04
[16.0] [MIG] product_packaging_dimension

marielejeune:

PR #1221, opened 2022-11-30
[MIG][16.0] product total weight from packaging

glitchov:
Migration from 14.0


lmignon (approved):
Code review

FrancoMaxime (approved):
code review

xavier-bouquiaux (approved):
LGTM technical side

product_total_weight_from_packaging/models/product.py
```diff
@@ -0,0 +1,37 @@
...
+
+class ProductProduct(models.Model):
+
+    _inherit = "product.product"
+
+    def get_total_weight_from_packaging(self, qty):
+        self.ensure_one()
+        qty_by_packaging_with_weight = self.with_context(
+            **{
+                "_packaging_filter": lambda p: p.max_weight,
```
jbaudoux:
```suggestion
                "_packaging_filter": lambda p: p.weight,
```

product_total_weight_from_packaging/models/product.py
```diff
@@ -0,0 +1,37 @@
...
+        self, packaging_tuple, qty_per_pkg
+    ):
+        res = {
+            "qty": qty_per_pkg,
+        }
+        if packaging_tuple.is_unit:
+            res["weight"] = self.weight
+        else:
+            packaging = self.env["product.packaging"].browse(packaging_tuple.id)
+            res["weight"] = packaging.max_weight
```
jbaudoux:
```suggestion
            res["weight"] = packaging.weight
```

jbaudoux (changes requested):
The max_weight is a concept that was introduced in odoo delivery module to define the transporter weight limit for the shipped packaging. Before v16, the transporter packaging and product packaging was the same model `product.packaging`. Now the transporter packaging is moved to the model `stock.package.type`.
In v13, max_weight defined in delivery module on `product.packaging` was used also for product packaging dimension.
In v14, the weight field was added on the `product.packaging` and the max_weight was still coming from delivery module. This was introducing inconsistency as some modules migrated from v13 to v14 still rely on max_weight and some rely on the new weight field.
In v16, we have to drop this `max_weight` on product.packaging as the delivery module moved it to stock.package.type and we should rely only on the `weight` field.
The migration of the product_packaging_dimension https://github.com/OCA/product-attribute/pull/1161 has to be fixed to not expose the old `max_weight` field and any data on `max_weight` must be moved to `weight` if empty.

This module also deserves a big warning that it is lacking the weight uom conversions as it depends on product_packaging_dimension that depends on product_logistics_uom that allows to set a uom on the weight. The sum in this module is assuming all weights are in kg

victorvermot:
Hello @glitchov, @lmignon I have added a small pr to fix the situation raised by @jbaudoux
- https://github.com/acsone/product-attribute/pull/6

leemannd:
Hello @glitchov , there are conflicts with the branch 16.0, by any chance could you have a look at it and maybe perform a rebase?

lmignon:
@leemannd @victorvermot This branch is now rebased and comments from @jbaudoux are included

jbaudoux (approved):
Thanks @lmignon for the update.

Not related to the migration but this module still deserves a big warning that it is lacking the weight uom conversions as it depends on product_packaging_dimension that depends on product_logistics_uom that allows to set a uom on the weight. The sum in this module is assuming all weights are in kg

lmignon:
@jbaudoux Warning added...
image

leemannd (approved):

PR #1237, opened 2022-12-27
[MIG] product_code_mandatory: Migration to 16.0

matiasperalta1:


rvalyi (dismissed):
LGTM (product default_code works the same in Odoo v15 and Odoo v16)

rvalyi (changes requested):
Sorry I approved too fast before. So actually after testing on Runbot, I see two issues:

1. in the menu, when you edit the product.template (like Sales>Products>Products), the default_code field is not mandatory and you can save without a code. Then some default value is assigned. But I think this defeats the purpose of the module (if you edit a product.product using the Product Variant menu then the field is required). This bug is also present on v15.
2. Even if you use the Product Variant menu (Sales>Products>Product Variants) then some default value is present for the default_code field, which is not the expected and not the same behavior as in v15 (empty required field).

andreampiovesana:
this app is important, can you merge and then fix problems like in version 15?

andreampiovesana (approved):
LGTM

PR #1241, opened 2023-01-09
[15.0][ADD] product_drained_weight

sergio-teruel:
cc @Tecnativa  TT40323


legalsylvain (approved):
hi @sergio-teruel Thanks for your contribution.
code review / no test. LGTM. 

Two remarks : (non blocking)
- looks very similar to product_net_weight ! (same repo / V12 https://github.com/OCA/product-attribute/tree/12.0/product_net_weight#products---net-weight) Maybe you could add reference to the description. Something like "you could be interested by another module ...). (I'll do the same when i'll port product_net_weight in V16)

- did you thought to add a constrains, to check if drained_weight is <= weight to avoid invalid values ? (if weight is defined).

pedrobaeza:
Can't net weight be used as drained weight? Or there can be the 3 weights in the same product?

legalsylvain:
Hi @pedrobaeza 

> Can't net weight be used as drained weight? Or there can be the 3 weights in the same product?

Hi. no. 3 weight. (can be explicited in the readme IMO.)

Exemple :  box of chickpeas : 
- weight (=brut weight) with the box : 800 g
- net weight (without the box) : 600 g
- drained weight (without the water) : 500 g

Note : net weight is a very common information, usefull in many cases. AFAIK, Drained weight is usefull in food industry.

pedrobaeza:
Yeah, but maybe in our use case net weight is not used for its purpose, and we save all the *_drained_weight_* modules for everything.

legalsylvain:
OK. please ping me if you port ``product_net_weight``.

legalsylvain:
Note : not sure, but maybe it could be great to make drained_weight depends on net_weight. Reason : 
- functional : 
1) if drained_weight is used, net_weight will be used for sure also because drained weight is more precise, and weight should be renamed in gross weight.
2) could allow to add constrains with drained_weight < net_weight
- technical : to make the 3 fields ordered in the logic way in the product view.

carlosdauden (approved):

sergio-teruel:
@legalsylvain Thanks for your review.
I have been this module depending on "product_net_weight".
I have added a constraint
I have fixed the computation of drained weight data for template. I think that product_net_weight has an error, net_weight field on product template model can not be related. I will do a PR to fix it.

legalsylvain:
>  I think that product_net_weight has an error, net_weight field on product template model can not be related. I will do a PR to fix it.

totally agree. for the time being on "net" module. 
in V12 this is : 
- product.product : field
- product.template : related field

It should be : 
- product.product : field
- product.template compute field with the same algorithm you implemented in your module, and in the official module : 
https://github.com/odoo/odoo/blob/12.0/addons/product/models/product_template.py#L87

PR #1242, opened 2023-01-10
[13.0][FIX] product_pricelist_supplierinfo: Seller got from other company

carlosdauden:
…because sudo ignore rules

force_company is used here:
https://github.com/odoo/odoo/blob/a72bb463539bacd4a9414a9b7d178ed84fb9d333/addons/product/models/product.py#L612-L613

@Tecnativa TT41130


pedrobaeza (approved):

sergio-teruel (approved):

pedrobaeza:
Please fw-port it to upper versions:

/ocabot merge patch

PR #1246, opened 2023-01-11
[15.0][IMP] product_secondary_unit: Compatibility with product variants

sergio-teruel:
cc @Tecnativa TT38736
ping @carlosdauden @chienandalu

### DIFF
--- a/product_secondary_unit/models/__init__.py
+++ b/product_secondary_unit/models/__init__.py
@@ -1,4 +1,5 @@
 # License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+from . import product_product
 from . import product_second_unit
 from . import product_secondary_unit_mixin
 from . import product_template
--- a/product_secondary_unit/views/product_views.xml
+++ b/product_secondary_unit/views/product_views.xml
@@ -2,10 +2,10 @@
 <!-- Copyright 2018 Tecnativa - Sergio Teruel
      License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl). -->
 <odoo>
-    <record id="product_template_only_form_view" model="ir.ui.view">
+    <record id="product_template_form_view" model="ir.ui.view">
         <field name="name">Product template Secondary Unit</field>
         <field name="model">product.template</field>
-        <field name="inherit_id" ref="product.product_template_only_form_view" />
+        <field name="inherit_id" ref="product.product_template_form_view" />
         <field name="groups_id" eval="[(4, ref('uom.group_uom'))]" />
         <field name="arch" type="xml">
             <xpath expr="//page[@name='general_information']" position="inside">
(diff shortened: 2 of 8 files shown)


product_secondary_unit/views/product_views.xml
```diff
@@ -24,4 +32,34 @@
...
     </record>
+
+    <record id="product_variant_easy_edit_view" model="ir.ui.view">
+        <field name="model">product.product</field>
+        <field name="inherit_id" ref="product.product_variant_easy_edit_view" />
+        <field name="groups_id" eval="[(4, ref('uom.group_uom'))]" />
+        <field name="arch" type="xml">
+            <xpath expr="//group[@name='weight']/.." position="after">
+                <field name="product_tmpl_id" invisible="1" />
+                <group string="Secondary Unit of Measure" name="secondary_uom">
```
chienandalu: Should we remark this?
```suggestion
                <group string="Variant Spefecific Secondary Unit of Measure" name="secondary_uom">
```

sergio-teruel: This view is already a variant so I think that write "Variant Spefecific... " is redundant..

chienandalu: If you have template secondary units along with variant secondary units it can be confusing IMO. Even more when you look into single variants.

sergio-teruel:
@chienandalu Changes done

chienandalu (approved):
:smile: :+1:

carlosdauden (approved):

product_secondary_unit/models/__init__.py
```diff
@@ -1,4 +1,5 @@
 # License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+from . import product
```
pedrobaeza: And sort it alphabetically

```suggestion
from . import product_product
```

product_secondary_unit/models/product.py
```diff
@@ -0,0 +1,15 @@
+# Copyright 2022 Tecnativa - Sergio Teruel
```
pedrobaeza:
```suggestion
# Copyright 2023 Tecnativa - Sergio Teruel
```

pedrobaeza (commented):
Why not doing like Odoo does with `product.supplierinfo`: show the extra `product_id` field in the list to be filled or not, but the same one2many is shown in both the template and the variant, for avoiding misunderstandings if the user accesses from one form or the other (they don't know most of the time the difference).

sergio-teruel:
Changes done!

sergio-teruel:
I do nt remember but I decided do it by this way due to any isuue.. also I want the field in simple variant form too

pedrobaeza:
You can also see both in product.supplierinfo way of working. Please check it. I think it's less confusing that way.

pedrobaeza:
I have tried on runboat, and I insist that this is not user friendly and will lead to confusions. Look, I added a secondary unit on the template (which is the menu "Products", that is the one that people goes by default):

image

And on the variant form, there's no hint of such secondary unit:

TaThe9dHhs

What are the problems of using exactly the same technique of Odoo core with `product.supplierinfo`?

pedrobaeza (changes requested):
Check my comment

pedrobaeza:
@sergio-teruel please review my comment.

sergio-teruel:
@pedrobaeza changes done

pedrobaeza:
Hi, Sergio, thanks for the changes. In a test in runboat, now if I add a secondary unit in the template, it appears also in the variant, and the label says "Variant Specific Secondary Unit of Measure", which should disappear in favor of a neutral label IMO.

If I add a secondary unit for a specific variant, both are seen on the template with the column "Product variant" filled for the specific one, and going to product variants, opening the specific one shows both, but on the rest, only the template one is shown. That can be confusing, and more always showing the column "Product variant". Imagine that you add from the variant 1 a secondary unit for "Variant 2" (it's allowed), you refresh, and you see that such secondary unit disappears.

I think you should follow strictly what Odoo does with product.supplierinfo, showing the same records for both template and variants, and having the product variant column for knowing if it's something specific. Another suggestion is that you show all of them, but put in grey through a decoration, those that applies to other variants where being in a specific variant.

product_secondary_unit/views/product_views.xml
```diff
@@ -24,4 +32,72 @@
...
+                        name="secondary_uom_ids"
+                        nolabel="1"
+                        context="{'default_product_tmpl_id': product_tmpl_id}"
+                    >
+                        <tree editable="bottom" decoration-muted="(not active)">
+                            <field name="active" widget="boolean_toggle" />
+                            <field name="product_tmpl_id" invisible="1" />
+                            <field
+                                name="product_id"
+                                domain="[('product_tmpl_id', '=', parent.id)]"
```
carlosdauden:
```suggestion
                                domain="[('id', '=', parent.id)]"
```

carlosdauden (changes requested):

sergio-teruel:
Changes done!

carlosdauden (approved):

sergio-teruel:
ping @ernestotejeda @chienandalu

ernestotejeda (approved):

pedrobaeza (approved):
/ocabot merge major

PR #1250, opened 2023-01-17
[15.0][MIG]product_packaging_type_pallet: Migration to V15

bizzappdev:


bizzappdev:
Ready for the Review

anajuaristi:
@pedrobaeza 
I see there is 3 closed PR for this module to be migrated to V15, V16 (someone wanted to change the name), V18.
We would like to have it like it is in V14 for V16, so we are going to open a task to do it.

Would OCA consider our PR migration from V14 directly to V16? Do you see any issue we are not aware of?

pedrobaeza:
I don't see any problem in principle. These closed PRs are not approved, so they are not "official". Anyway, check if this PR can be the source instead of 14 if it's correct, to save some time (if it's worth, of course).

PR #1258, opened 2023-01-19
[15.0][FIX+IMP] product_sequence

grindtildeath:
If a sequence prefix is defined on the product category, there's
no way through the UI to remove it. If the prefix field is emptied,
the existing sequence will still be defined on the category without
any possibility for the user to change the setting.
With this change, removing the prefix will remove the link between
the product category and the sequence.
Moreover, if an existing prefix is redefined on a product category
a new sequence was created instead of reusing an existing one.

As the product.category.sequence_id field of product.category is
only displayed in debug mode, we can allow its edition to ease
fixing of wrong configuration.
However, we need to make sure the prefix defined on these product
categories always matches the prefix defined on the sequence and
restrict any change that would introduce a discrepancy.

### DIFF
--- a/product_sequence/models/product_category.py
+++ b/product_sequence/models/product_category.py
@@ -46,7 +46,7 @@ class ProductCategory(models.Model):
                     rec.sudo().sequence_id.prefix = prefix
                 else:
                     seq_vals = self._prepare_ir_sequence(prefix)
-                    rec.sequence_id = self.env["ir.sequence"].create(seq_vals)
+                    vals["sequence_id"] = self.env["ir.sequence"].create(seq_vals)
         return super().write(vals)
 
     @api.model
--- a/product_sequence/models/product_category.py
+++ b/product_sequence/models/product_category.py
@@ -38,22 +38,31 @@ class ProductCategory(models.Model):
         }
         return vals
 
+    def _get_or_create_sequence(self, prefix):
+        seq_vals = self._prepare_ir_sequence(prefix)
+        sequence = self.env["ir.sequence"].search(
+            [("code", "=", seq_vals.get("code")), ("prefix", "=", prefix)]
+        )
+        if not sequence:
+            sequence = self.env["ir.sequence"].create(seq_vals)
+        return sequence
+
     def write(self, vals):
-        prefix = vals.get("code_prefix", False)
-        if prefix:
-            for rec in self:
-                if rec.sequence_id:
-                    rec.sudo().sequence_id.prefix = prefix
-                else:
-                    seq_vals = self._prepare_ir_sequence(prefix)
-                    vals["sequence_id"] = self.env["ir.sequence"].create(seq_vals)
+        if "code_prefix" in vals:
+            prefix = vals.get("code_prefix")
+            if prefix:
+                for rec in self:
+                    if rec.sequence_id:
+                        rec.sudo().sequence_id.prefix = prefix
+                    else:
+                        vals["sequence_id"] = self._get_or_create_sequence(prefix).id
+            else:
+                vals["sequence_id"] = False
         return super().write(vals)
 
     @api.model
     def create(self, vals):
         prefix = vals.get("code_prefix", False)
         if prefix:
-            seq_vals = self._prepare_ir_sequence(prefix)
-            sequence = self.env["ir.sequence"].create(seq_vals)
-            vals["sequence_id"] = sequence.id
+            vals["sequence_id"] = self._get_or_create_sequence(prefix).id
         return super().create(vals)
--- a/product_sequence/models/product_category.py
+++ b/product_sequence/models/product_category.py
@@ -2,7 +2,8 @@
 #   (http://www.forgeflow.com)
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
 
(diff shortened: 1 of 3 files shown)


product_sequence/models/product_category.py
```diff
@@ -38,22 +38,53 @@ def _prepare_ir_sequence(self, prefix):
         }
         return vals
 
+    def _get_or_create_sequence(self, prefix):
+        seq_vals = self._prepare_ir_sequence(prefix)
+        sequence = self.env["ir.sequence"].search(
+            [("code", "=", seq_vals.get("code")), ("prefix", "=", prefix)]
+        )
+        if not sequence:
```
i-vyshnevska, line 46:
```suggestion
        sequence = self.env["ir.sequence"].search(
            [("code", "=", seq_vals.get("code")), ("prefix", "=", prefix)]
        )
        if not sequence:
                    seq_vals = self._prepare_ir_sequence(prefix)
```

product_sequence/models/product_category.py
```diff
@@ -38,22 +38,53 @@ def _prepare_ir_sequence(self, prefix):
...
-        if prefix:
-            for rec in self:
-                if rec.sequence_id:
-                    rec.sudo().sequence_id.prefix = prefix
-                else:
-                    seq_vals = self._prepare_ir_sequence(prefix)
-                    rec.sequence_id = self.env["ir.sequence"].create(seq_vals)
-        return super().write(vals)
+        sequence_ids_to_check = set()
+        if "code_prefix" in vals:
```
i-vyshnevska, line 52: why do not hide "code_prefix" if the sequence is selected? 
and are there so many categories to create sequence instead of the user?

product_sequence/models/product_category.py
```diff
@@ -48,21 +48,43 @@ def _get_or_create_sequence(self, prefix):
...
                     else:
                         vals["sequence_id"] = self._get_or_create_sequence(prefix).id
             else:
                 vals["sequence_id"] = False
-        return super().write(vals)
+        res = super().write(vals)
+        if sequence_ids_to_check:
+            self.env["ir.sequence"].sudo().browse(
+                sequence_ids_to_check
+            )._check_prefix_product_category()
```
TDu, line 70: Why does the function decorated with the `api.constrains` needs to be called explicitly ?

product_sequence/models/product_category.py
```diff
@@ -38,22 +38,31 @@ def _prepare_ir_sequence(self, prefix):
...
-        if prefix:
-            for rec in self:
-                if rec.sequence_id:
-                    rec.sudo().sequence_id.prefix = prefix
-                else:
-                    seq_vals = self._prepare_ir_sequence(prefix)
-                    vals["sequence_id"] = self.env["ir.sequence"].create(seq_vals)
+        if "code_prefix" in vals:
+            prefix = vals.get("code_prefix")
+            if prefix:
```
TDu, line 54: This change makes the diff larger but does not seems to be needed.

product_sequence/models/product_category.py
grindtildeath: Actually we need `seq_vals.get("code")` to perform the search. So I guess calling the prepare function is enough instead of adding another hook for the "code" preparation or having the string defined twice.

product_sequence/models/product_category.py
grindtildeath: > why do not hide "code_prefix" if the sequence is selected?

That's how the module is implemented, and since access to ir.sequence requires extra rights IMO it's ok to keep it as it is.

> and are there so many categories to create sequence instead of the user?

Not sure to understand your question, can you please rephrase?

product_sequence/models/product_category.py
grindtildeath: As we need to "sync" the prefix value on the `ir.sequence` from a change on `product.category`, the update of the sequence record before the call to super would trigger the constraint which would raise an error, as the category is not updated yet (as it will only be updated after the call to super).
Therefore, we want to "delay" the execution of the constraint check until both records are updated, and that's why we have to call it explicitly afterwards.

product_sequence/models/product_category.py
grindtildeath: We actually need to check the existence of `code_prefix` key in the values dictionary that is provided, to remove the link to the sequence if the prefix is empty. (or at least I don't see any other way for a normal user that doesn't have rights on `ir.sequence` to remove this link)

product_sequence/models/product_category.py
i-vyshnevska: > Not sure to understand your question, can you please rephrase?

I meant that automatization is not needed here, that if the user will need to create a sequence, he can check it manually
but as
> access to ir.sequence requires extra rights

understandable

TDu (approved):

i-vyshnevska (approved):

grindtildeath:
ping @khaeusler @sergio-teruel @LoisRForgeFlow @OCA/product-maintainers

LoisRForgeFlow (approved):

LoisRForgeFlow:
ouch, sorry, I cannot merge. But the improvement looks great to me.

PR #1265, opened 2023-01-31
[IMP] product_manufacturer: Add pre_init_hook for faster install

mt-software-de:
The performance while installing this module is for a huge amount of products and product templates bad. 
The problem is the  _compute_manufacturer_info this gets called for all product template's, 
because odoo creates the missing columns of product.template and tries to get the record values from the compute method. 
Since the method _compute_manufacturer_info returns only empty values on installing,
it is more performant to create the columns with a pre_init hook. 
Because then odoo does not create the columns and doesn't call the compute method.

### DIFF
--- a/product_manufacturer/hooks.py
+++ b/product_manufacturer/hooks.py
@@ -1,0 +1,17 @@
+# Copyright 2022 Michael Tietz (MT Software) <mtietz@mt-software.de>
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl.html).
+
+from odoo.tools import sql
+
+
+def pre_init_hook(cr):
+    table_name = "product_template"
+    for field_name, field_type in [
+        ("manufacturer", "int4"),
+        ("manufacturer_pname", "varchar"),
+        ("manufacturer_pref", "varchar"),
+        ("manufacturer_purl", "varchar"),
+    ]:
+        if sql.column_exists(cr, table_name, field_name):
+            continue
+        sql.create_column(cr, table_name, field_name, field_type)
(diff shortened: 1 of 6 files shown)


product_manufacturer/hooks.py
```diff
@@ -0,0 +1,17 @@
...
+    table_name = "product_template"
+    for field_name, field_type in [
+        ("manufacturer", "int4"),
+        ("manufacturer_pname", "varchar"),
+        ("manufacturer_pref", "varchar"),
+        ("manufacturer_purl", "varchar"),
+    ]:
+        if sql.column_exists(cr, table_name, field_name):
+            continue
+        sql.create_column(cr, table_name, field_name, field_type)
```
rousseldenis, line 17: @mt-software-de Do you have any metrics about that as this involves no data? So, why not relying on ORM for that (this would imply we would do that in every module...)?

mt-software-de: in our case we have a database with 
- 361645 records in product_product
- 355095 in product_template

i didn't measured the time, but in fact on installing this module there will be no data in this columns on product_product. 
Which means the method _compute_manufacturer_info will result in empty values for all of the product.template's fields.

mt-software-de: i updated the comment of the PR maybe it is now more understandable? :-)

rousseldenis: Yes, but the problem is that for existing data, you don't fill in the missing data in product_template table.

So, this is not sufficent and not acceptable as is (as no data will be computed at all).

mt-software-de: oh yes this of course a problem.

PR #1269, opened 2023-02-15
[14.0][IMP] product_tier_validation : force default state on create and improve _state_to

Kev-Roche:
When creating a new product, we can directly set a normal / sellable state on it, without validation. This PR avoid that by forcing the "_state_from" on product creation. 

First commit is the failing test, second test fix it. 

Last commit purpose is to improve _state_to value for products : 
The base module "base_tier_validation" proposes "confirmed" as "_state_to" value, which is not a product state. Furthermore, we can create other product.state. In order to use this module out of the box and to take account of all the states created by users, we propose here to change the _state_to accordingly.

### DIFF
--- a/product_tier_validation/models/product_template.py
+++ b/product_tier_validation/models/product_template.py
@@ -1,8 +1,8 @@
 # Copyright 2021 Open Source Integrators
 # License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
 
-from odoo import models
-
+from odoo import api, models
+from odoo.exceptions import UserError
 
 class ProductTemplate(models.Model):
     _name = "product.template"
@@ -20,3 +20,13 @@ class ProductTemplate(models.Model):
         if "product_state_id" in vals:
             self.restart_validation()
         return res
+
+    @api.model_create_multi
+    def create(self, vals_list):
+        states = self.env["product.state"].search([("code", "in", self._state_from)])
+        if not states:
+            raise UserError("Wrong configuration of '_state_from' in product.state")
+        for vals in vals_list:
+            if vals.get("product_state_id") not in states.ids:
+                vals["product_state_id"] = states[0].id
+        return super().create(vals_list)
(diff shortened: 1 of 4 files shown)


Kev-Roche:
Second commit fix it but need to set product_status as rebel module.

product_tier_validation/models/product_template.py
```diff
@@ -20,3 +20,16 @@ def write(self, vals):
         if "product_state_id" in vals:
             self.restart_validation()
         return res
+
+    @api.model_create_multi
+    def create(self, vals_list):
+        draft_state_id = self.env["product.state"].search(
```
sebastienbeau: it's not an id so name it "draft_state"

sebastienbeau (changes requested):

product_tier_validation/models/product_template.py
```diff
@@ -20,3 +20,16 @@ def write(self, vals):
...
+            [("code", "in", self._state_from)], limit=1
+        )
+        for vals in vals_list:
+            if (
+                "product_state_id" in vals
+                and vals.get("product_state_id") not in self._state_from
+            ):
+                vals["product_state_id"] = draft_state_id.id
+        return super().create(vals)
```
sebastienbeau:
```suggestion
        return super().create(list_vals)
```

sebastienbeau:
```suggestion
    @api.model_create_multi
    def create(self, vals_list):
        states = self.env["product.state"].search([("code", "in", self._state_from)])
        if not states:
            raise UserError("Wrong configuration....")
        for vals in vals_list:
            if vals.get("product_state_id") not in states.ids:
                vals["product_state_id"] = states[0].id
        return super().create(list_vals)
```

sebastienbeau (changes requested):

product_tier_validation/models/product_template.py
Kev-Roche: done

product_tier_validation/models/product_template.py
Kev-Roche: done

Kev-Roche: Thank you, I applide your suggestion.

sebastienbeau:
@gurneyalex @rousseldenis can you reopen this PR? (I do not have the right to do it)

Thanks

sebastienbeau (commented):
LGTM (code review) and tested

PR #1271, opened 2023-02-16
[15.0][FIX] product_supplierinfo_for_customer: Use supplierinfo also for child contacts

dhx:
Often in a sale order a specific sub contact of a customer is selected as the partner. Currently the sub contact does not inherit the supplier info records of the parent. With this patch when there is no supplier info record on the sub contact it will try to fall back to the parent.

### DIFF
--- a/product_supplierinfo_for_customer/models/product_product.py
+++ b/product_supplierinfo_for_customer/models/product_product.py
@@ -132,4 +132,13 @@ class ProductProduct(models.Model):
             .sorted(lambda s: (s.sequence, s.min_qty, s.price, s.id))
         )
         res_1 = res.sorted("product_tmpl_id")[:1]
-        return res_1
+        if res_1 or not partner.parent_id:
+            return res_1
+        else:
+            return self._select_customerinfo(
+                partner=partner.parent_id,
+                params=params,
+                _quantity=_quantity,
+                _date=_date,
+                _uom_id=_uom_id,
+            )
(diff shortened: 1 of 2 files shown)


hildickethan (approved):

hildickethan:
the PR title should include the version for clarity in the PR tree view

product_supplierinfo_for_customer/models/product_product.py
```diff
@@ -132,4 +132,13 @@ def _select_customerinfo(
             .sorted(lambda s: (s.sequence, s.min_qty, s.price, s.id))
         )
         res_1 = res.sorted("product_tmpl_id")[:1]
-        return res_1
+        if res_1 or not partner.parent_id:
+            return res_1
+        else:
+            return self._select_customerinfo(
+                partner=partner.parent_id,
```
sergio-teruel, line 139: Maybe can you use partner.commercial_partner_id ??

sergio-teruel (approved):
little comment

product_supplierinfo_for_customer/models/product_product.py
dhx: @sergio-teruel thanks for your comment, sorry I was not able to respond before this was merged. Considering how the commercial_partner_id field is computed this would not do a fallback to the parent if the sub-contact in question is a company:
```python
            if partner.is_company or not partner.parent_id:
                partner.commercial_partner_id = partner
            else:
                partner.commercial_partner_id = partner.parent_id.commercial_partner_id
```
Maybe if you feel this would be the preferred behaviour you want to make a separate PR for this?

PR #1273, opened 2023-02-21
[14.0][ADD] product_pricelist_item_menu

dessanhemrayev:
New module


pedrobaeza:
Please avoid plural in module name.

legalsylvain (changes requested):
Could you add a description of the module you propose ?

PR #1277, opened 2023-02-24
[ADD] 16.0 product_category_level module

PierrickBrun:
Add a 'Level' field on Product Categories to show the number of parents a category has (recursively).


legalsylvain (changes requested):
Hi. LGTM. 
- Just a point, the screenshot are not the good ones I think. it's a copy paste of the module product_category_type. Could you replace by the correct screenshots ? 

- Naive question : how do you use this new level fields ?

legalsylvain (approved):

PierrickBrun:
@legalsylvain Thanks !

The use is purely informative, if you see that your category does not have the correct level then there is an error somewhere, this is complementary to product_category_type.

Also I use it to extract data grouped by category of level X.

nayatec (approved):
LGTM

PR #1278, opened 2023-02-25
[ADD] 14.0: product_form_pricelist_percent_change

FrancescoBallerini:
I developed an extension of product_form_pricelist with the following features:

- Full access to product pricelist rule form-view by product form-view. Basically with this extension the one2many field 'fixed_pricelist_item_ids' treeview is not editable so the form-view of pricelist rule will open on click. Made some adjustment to setup the right relation product_id/rule_id when rule form is opened by 'fixed_pricelist_item_ids'. 

- In the item form-view added section "Compute Selling Price" that introduce 2 features:
 

1.   A computed field 'Selling Price' shows the selling price for active product template or variant depending on provided parameters

2. Percent change: a button "Set Price Discount %" will be added to rule form as well. It will allow to compute the "percent change"  depending on Selling Price (recomputed for the button function) and a User Input field which will represent the new selling price.

Will use the formula:

     (user input - selling price computed) / selling price computed x 100

  and the result will be stored in Price Discount field or Percent Price, depending on "Compute Price" parameter.

We are not using product variants workflow but I have tried to provide support for it. There are some issue with variant smart btn view because the active_id does not change to product_id (remain tmpl_id) plus I might have missed some things so that part might require extra work to be effective.


FrancescoBallerini:
TODO:

- [ ] Write tests
- [x] support for product_pricelist_supplierinfo: make methods more extendable

PR #1281, opened 2023-02-27
[MIG][16.0] product_net_weight

RabbitJon-S73:
Standard migration to 16.0


RabbitJon-S73:
@legalsylvain

legalsylvain (commented):
Hi. Thanks for porting this module. Diff review OK.

However, in another PR, @sergio-teruel said that the design of the net_weight was bad. (for the time being related field -> should be trivial computed field).
could you take a look on [that message](https://github.com/OCA/product-attribute/pull/1241#issuecomment-1378719018) and take advantage of the migration to correct this point ?
@sergio-teruel : you said you'll do a PR. Did you do it ? 

thanks !

sergio-teruel:
Sorry @legalsylvain Done here: https://github.com/OCA/product-attribute/pull/1288

@RabbitJon-S73 Can you wait to merge the last one and do the migration... ?? Sorry ...

RabbitJon-S73:
@sergio-teruel I cherry-picked your commit and I did some modifications in README version, but runboat isn't working, could you check it?

sergio-teruel:
Hi.
Only one question...
In older version the podition of net_weight field is before than weight... 
Can you keep this?

image

RabbitJon-S73:
I had to depend on stock and the inherit since `Responsible` is in the middle of both weights because of this xpath https://github.com/odoo/odoo/blob/755c517fcab89df51e8b0ee4529894372b4d4ec1/addons/stock/views/product_views.xml#L448
@sergio-teruel

legalsylvain:
> I had to depend on stock and the inherit since Responsible is in the middle of both weights because of this xpath https://github.com/odoo/odoo/blob/755c517fcab89df51e8b0ee4529894372b4d4ec1/addons/stock/views/product_views.xml#L448
@sergio-teruel

that's annoying.
what do you think @sergio-teruel ? 

For me, it's not a problem  that a "weight" module depends on stock. It is a very basic module, and I don't see use case where we want net weight without stock.

OK for all ?

RabbitJon-S73:
> > I had to depend on stock and the inherit since Responsible is in the middle of both weights because of this xpath https://github.com/odoo/odoo/blob/755c517fcab89df51e8b0ee4529894372b4d4ec1/addons/stock/views/product_views.xml#L448
> > @sergio-teruel
> 
> that's annoying. what do you think @sergio-teruel ?
> 
> For me, it's not a problem that a "weight" module depends on stock. It is a very basic module, and I don't see use case where we want net weight without stock.
> 
> OK for all ?

First I tried to modify the view to keep like v14.0 without adding any additional dependency, but its true that this module and stock will be installed always together. Its OK for me

legalsylvain (approved):

rvalyi:
we also use if the Brazil for electronic invoicing and the module where we use it (l10n_br_nfe,_delivery) also depends on stock, so ok for us too.

sergio-teruel:
Ok for me too

RabbitJon-S73:
Hey @sergio-teruel, could you approve?

sergio-teruel (approved):

sergio-teruel:
@RabbitJon-S73 Done!! Thanks

ferran-S73 (approved):

RabbitJon-S73:
@pedrobaeza

ferran-S73:
@pedrobaeza this should be ready to merge now

PR #1282, opened 2023-02-27
[ADD] [15.0] product_sticker

Shide:
# Module _product_sticker_
You can attach images (aka _stickers_) to any **Category**, **Attribute**, **Attribute Value** and **Company**. 
You can put a text below the _sticker_ if you want.
It's a base module to extend with other modules to display these _stickers_ where you want.
Has been designed to be extended by other modules or your custom ones.

---

https://www.loom.com/share/0a858139f381488cbb603d0acd384845

The other PR to review after this PR: https://github.com/OCA/stock-logistics-reporting/pull/263

---

MT-2153 @moduon @rafaelbn @yajo


rafaelbn:
Thank you! Review!

1. - [ ] Optimize the Text. We could insert text on image but then more KB... we need to set as you said a text with lines, only the name is difficult.

imagen


2. - [ ] When printing a delivery slip with 2 products with the same image -> Odoo server Error

imagen

francesco-ooops (changes requested):
Hi @Shide , what is the purpose of a sticker? what's the use case? 

I think description could be improved in this regard.

rafaelbn:
We should rename module to

`product_sticker_stock` -> `stock_picking_report_product_sticker` inside https://github.com/OCA/stock-logistics-reporting

The main objetive is to have stamps, certifications or other images related to a product (but not the image of a product) or an attribute or a value of an attribute to use for reports o other places in Odoo.

In this case we have the use case than depending of the attribute of the product or the value of the product it is mandatory legally print the "sticker" in "Delivery slip" of stock picking.

Use sticker is intended as "image" is really ambiguous and can mislead or confuse.

product_sticker/__manifest__.py
```diff
@@ -0,0 +1,20 @@
...
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+
+{
+    "name": "Product Sticker",
+    "version": "15.0.1.0.0",
+    "author": "Moduon, Odoo Community Association (OCA)",
+    "license": "AGPL-3",
+    "website": "https://github.com/OCA/product-attribute",
+    "category": "Sales Management",
+    "depends": [
```
yajo, line 11: Don't you add yourself as maintainer?

product_sticker/models/product_product.py
```diff
@@ -0,0 +1,24 @@
...
+    def _get_sticker_arguments(self):
+        pavs = self.product_template_variant_value_ids.product_attribute_value_id
+        return {
+            "categories": self.categ_id,
+            "attributes": pavs.attribute_id,
+            "attribute_values": pavs,
+        }
+
+    def get_product_stickers(self):
+        """Product Stickers related to this Product Variant and his Template"""
```
yajo:
```suggestion
        """Product Stickers related to this Product Variant and its Template"""
```

product_sticker/models/product_product.py
```diff
@@ -0,0 +1,24 @@
...
+        self.ensure_one()
+        # Product Template: Common stickers
+        pt_stickers = self.product_tmpl_id.get_product_stickers()
+        # Product Product: Specific stickers
+        pp_stickers = self.env["product.sticker"].get_stickers(
+            **self._get_sticker_arguments()
+        )
+        return pt_stickers | pp_stickers
```
yajo, line 31: We shouldn't return recordsets in public methods. These methods can be called via RPC, and a recordset doesn't make sense in that context.

Please either make the method private or return ids. Or you could add the `@api.returns` decorator on them, to let Odoo know what to do.

product_sticker/models/product_attribute.py
```diff
@@ -0,0 +1,35 @@
...
+    @api.depends("sticker_ids")
+    def _compute_sticker_count(self):
+        for attribute in self:
+            attribute.sticker_count = len(
+                attribute.sudo().sticker_ids.filtered(
+                    lambda i: not i.company_id
+                    or i.company_id.id
+                    in self.env.context.get("allowed_company_ids", self.env.company.ids)
+                )
+            )
```
yajo: IMHO all this code should bo in an `AbstractModel` mixin. Then, apply the mixin to `product.attribute` and `product.attribute.value`.

product_sticker/models/product_attribute.py
```diff
@@ -0,0 +1,35 @@
...
+            ),
+        ],
+    )
+    sticker_count = fields.Integer(
+        compute="_compute_sticker_count",
+        store=False,
+    )
+
+    @api.depends("sticker_ids")
+    def _compute_sticker_count(self):
```
yajo: Don't forget this!

```suggestion
    @api.depends("sticker_ids")
    @api.depends_context("company")
    def _compute_sticker_count(self):
```

product_sticker/models/product_attribute.py
```diff
@@ -0,0 +1,35 @@
...
+        string="Product Stickers",
+        domain=lambda s: [
+            "|",
+            ("company_id", "=", False),
+            (
+                "company_id",
+                "in",
+                s.env.context.get("allowed_company_ids", s.env.company.ids),
+            ),
+        ],
```
yajo:
```suggestion
        domain=lambda rec: rec.sticker_ids._build_sticker_domain_company(),
```

product_sticker/models/product_attribute.py
yajo:
```suggestion
                attribute.sudo().sticker_ids.filtered_domain(attribute.sticker_ids._build_sticker_domain_company())
```

product_sticker/models/product_attribute_value.py
```diff
@@ -0,0 +1,29 @@
...
+    )
+    sticker_count = fields.Integer(
+        compute="_compute_sticker_count",
+        store=False,
+    )
+
+    @api.depends("sticker_ids")
+    def _compute_sticker_count(self):
+        for attribute_value in self:
+            attribute_value.sticker_count = len(attribute_value.sticker_ids)
```
yajo: All this should go away and simply extend the mixin, as commented above.

product_sticker/models/product_sticker.py
```diff
@@ -0,0 +1,176 @@
...
+    )
+    can_image_128_be_zoomed = fields.Boolean(
+        string="Can Image 128 be zoomed",
+        compute="_compute_can_image_128_be_zoomed",
+        store=True,
+    )
+    show_sticker_name = fields.Boolean(
+        string="Display Name with the sticker",
+        help="If checked, the name will be displayed with the sticker",
+    )
```
yajo: Just to make upper case more consistent:

```suggestion
    product_attribute_value_id = fields.Many2one(
        comodel_name="product.attribute.value",
        string="Attribute value",
        ondelete="cascade",
        domain="[('attribute_id', '=', product_attribute_id)]",
    )
    can_image_128_be_zoomed = fields.Boolean(
        string="Allow zooming in 128px image",
        compute="_compute_can_image_128_be_zoomed",
        store=True,
    )
    show_sticker_name = fields.Boolean(
        string="Display name with the sticker",
        help="If checked, the name will be displayed with the sticker",
    )
```

product_sticker/models/product_sticker.py
```diff
@@ -0,0 +1,176 @@
...
+                ]
+            )
+        return attribute_domain
+
+    def _build_sticker_domain(self, **kwargs):
+        company_domain = self._build_sticker_domain_company()
+        category_domain = self._build_sticker_domain_category(kwargs.get("categories"))
+        attribute_domain = self._build_sticker_domain_attributes(
+            kwargs.get("attributes"), kwargs.get("attribute_values")
+        )
```
yajo: Make it easier to reason about the method signature:

```suggestion
    def _build_sticker_domain(self, categories=None, attributes=None, attribute_values=None):
        company_domain = self._build_sticker_domain_company()
        category_domain = self._build_sticker_domain_category(categories)
        attribute_domain = self._build_sticker_domain_attributes(
            attributes, attribute_values
        )
```

product_sticker/models/product_sticker.py
```diff
@@ -0,0 +1,176 @@
...
+        return [company_domain, category_domain, attribute_domain]
```
yajo, line 171: This doesn't seem like a domain. A domain would be:

```suggestion
        return company_domain + category_domain + attribute_domain
```

The method is called `_build_sticker_domain`, so IMHO it should return a domain.

product_sticker/models/product_sticker.py
```diff
@@ -0,0 +1,176 @@
...
+
+    @api.model
+    def get_stickers(self, **kwargs):
+        """Get stickers for given categories, attributes and attribute values"""
+        sticker_domain = expression.AND(self._build_sticker_domain(**kwargs))
+        return self.sudo().search(sticker_domain)
```
yajo, line 183: Why `sudo()`?

Also, remember to _not_ return recordsets in public methods (explained above).

product_sticker/models/product_template.py
```diff
@@ -0,0 +1,19 @@
...
+        )
+        return {
+            "categories": self.categ_id,
+            "attributes": no_variant_attribute_lines.attribute_id,
+            "attribute_values": no_variant_attribute_lines.value_ids,
+        }
+
+    def get_product_stickers(self):
+        """Attribute Stickers related to this Product Template and his variants"""
+        return self.env["product.sticker"].get_stickers(**self._get_sticker_arguments())
```
yajo: Don't return recordsets in public methods.

product_sticker/readme/DESCRIPTION.rst
```diff
@@ -0,0 +1,2 @@
+With this module you can add Stickers to Attributes, Variants or Categories.
```
yajo:
```suggestion
With this module you can add stickers to product attributes, variants or categories.
```

product_sticker/readme/USAGE.rst
```diff
@@ -0,0 +1,7 @@
+You can add Stickers to Attributes resulting in different behaviours:
+- If an Image has not Company, it will be available to all Companies
+- If an Image has not Attribute, it will be available to all Attributes
+- If an Image has not Attribute Value, it will be available to all Attribute Values of the Attribute
+- If an Image has not Category, it will be available to all Categories
```
yajo:
```suggestion
- If an Image has no Company, it will be available to all Companies
- If an Image has no Attribute, it will be available to all Attributes
- If an Image has no Attribute Value, it will be available to all Attribute Values of the Attribute
- If an Image has no Category, it will be available to all Categories
```

product_sticker/tests/common.py
```diff
@@ -0,0 +1,113 @@
...
+                        0,
+                        {
+                            "attribute_id": cls.at_l.id,  # License
+                            "value_ids": [
+                                (
+                                    6,
+                                    0,
+                                    [
+                                        cls.at_l_o.id,  # One-time Payment
+                                        cls.at_l_s.id,  # Subscription
```
yajo: If you need comments to explain your variables... that means that you should use better variable names! 🤣 

```suggestion
                            "attribute_id": cls.att_platform.id,
                            "value_ids": [
                                (
                                    6,
                                    0,
                                    [
                                        cls.attval_linux.id,
                                        cls.attval_windows.id,
                                    ],
                                )
                            ],
                        },
                    ),
                    (
                        0,
                        0,
                        {
                            "attribute_id": cls.att_license.id,
                            "value_ids": [
                                (
                                    6,
                                    0,
                                    [
                                        cls.attval_onetime.id,
                                        cls.attval_subscription.id, 
```

similar across the file.

yajo (changes requested):
Code review of `product_sticker`.

I didn't review the other module because, as @rafaelbn said, it should go to another repo.

product_sticker/__manifest__.py
Shide: Yep, thanks

product_sticker/models/product_attribute.py
Shide: The `inverse_name` on `sticker_ids` fields doesn't allow to put it on a mixin.

product_sticker/models/product_sticker.py
Shide: I will rename the method, but the domains should be returned like this

product_sticker/models/product_sticker.py
Shide: To allow search all images from all companies and not only that the user has activated.

yajo: And what's the use case? 🤔

Shide:
@rafaelbn @yajo changes made.

@francesco-ooops Added a loom video on the PR description

rafaelbn (approved):
👍 Great! Thank you!

francesco-ooops:
@rafaelbn @Shide The video is helping understand a bit more, but the files to improve are DESCRIPTION and USAGE

I would try to define:

- what is a sticker?
- how it can be useful in your vision? (reports, ecommerce?, ...?)
- and even better, a use case

or more in general: just ask a functional to read description and see if they can figure out what this module does and how it can be useful for their work :)

fcvalgar (approved):
LGTM Thank you @Shide

product_sticker/README.rst
```diff
@@ -0,0 +1,95 @@
...
+    :alt: Translate me on Weblate
+.. |badge5| image:: https://img.shields.io/badge/runboat-Try%20me-875A7B.png
+    :target: https://runboat.odoo-community.org/webui/builds.html?repo=OCA/product-attribute&target_branch=15.0
+    :alt: Try me on Runboat
+
+|badge1| |badge2| |badge3| |badge4| |badge5| 
+
+With this module you can add _stickers_ to Product Attributes, Product Attributes Values or Categories.
+This module by itself does not do anything, it is a base module to be extended by other modules.
+
```
rafaelbn, line 38: @Shide please, add this description to the readme. Thank you! 😄 

A sticker could be the image of a stamps, certifications or other images related to a product (but not the image of a product) or an attribute or a value of an attribute to use for reports o other places in Odoo. For example: ISO, Ecology, Security, Resistance, etc.

A general use case is: we have the use case than depending of the attribute of the product or the value of the product it is mandatory legally print the "sticker" in "Delivery slip" of stock picking.

A specific example of use case: when selling ecological products European Union makes mandatory print official ECO Logo (https://agriculture.ec.europa.eu/farming/organic-farming/organic-logo_en) in picking and this depends on the product attribute or values. An apple could be ecology or not and depending the attribute this should be printed or not.

product_sticker/README.rst
```diff
@@ -0,0 +1,95 @@
...
+.. |badge4| image:: https://img.shields.io/badge/weblate-Translate%20me-F47D42.png
+    :target: https://translation.odoo-community.org/projects/product-attribute-15-0/product-attribute-15-0-product_sticker
+    :alt: Translate me on Weblate
+.. |badge5| image:: https://img.shields.io/badge/runboat-Try%20me-875A7B.png
+    :target: https://runboat.odoo-community.org/webui/builds.html?repo=OCA/product-attribute&target_branch=15.0
+    :alt: Try me on Runboat
+
+|badge1| |badge2| |badge3| |badge4| |badge5| 
+
+With this module you can add _stickers_ to Product Attributes, Product Attributes Values or Categories.
```
rafaelbn: Could you please @Shide detail a little bit more like you did in:

> You can attach images (aka stickers) to any Category, Attribute, Attribute Value and Company.
> You can put a text below the sticker if you want.

Like:

With this module you can add _stickers_ (aka Images) to Product Attributes, Product Attributes Values or Categories. Also you can put a text below the sticker if you want.

rafaelbn (commented):
Just improve a little bit more readme (already is great) 😄

product_sticker/__manifest__.py
```diff
@@ -0,0 +1,21 @@
+# Copyright 2023 Eduardo de Miguel <edu@moduon.team>
```
yajo: Please remember to change this on all files:

```suggestion
# Copyright 2023 Moduon Team S.L. <info@moduon.team>
```

cc @rafaelbn

product_sticker/models/product_attribute.py
```diff
@@ -0,0 +1,25 @@
...
+
+class ProductAttribute(models.Model):
+    _inherit = "product.attribute"
+
+    sticker_ids = fields.One2many(
+        comodel_name="product.sticker",
+        inverse_name="product_attribute_id",
+        string="Product Stickers",
+        domain=lambda s: s.env["product.sticker"]._build_sticker_domain_company(),
+    )
```
yajo: You can still do this and benefit from the mixin approach, if you find it useful:

```suggestion
class ProductAttribute(models.Model):
    _name = "product.attribute"
    _inherit = ["product.attribute", "product.sticker.mixin"]
    
    sticker_ids = fields.One2many(
        inverse_name="product_attribute_id",
    )

class ProductStickerMixin(models.AbstractModel):
    _name "product.sticker.mixin"

    sticker_ids = fields.One2many(
        comodel_name="product.sticker",
        string="Product Stickers",
        domain=lambda s: s.env["product.sticker"]._build_sticker_domain_company(),
    )
```

It's just an example showing that it's simple, but of course the mixin should go in a separate file etc.

product_sticker/readme/CONTRIBUTORS.rst
```diff
@@ -0,0 +1,3 @@
+* `Moduon Team <https://www.moduon.team>`_:
+
+  * Eduardo de Miguel
```
yajo: As discussed yesterday:

```suggestion
* Eduardo de Miguel (`Moduon <https://www.moduon.team/>`__)
```

product_sticker/views/product_attribute_views.xml
```diff
@@ -0,0 +1,25 @@
...
+        <field name="model">product.attribute</field>
+        <field name="inherit_id" ref="product.product_attribute_view_form" />
+        <field name="arch" type="xml">
+            <xpath expr="//div[@name='button_box']" position="inside">
+                <button
+                    name="%(action_product_sticker)d"
+                    type="action"
+                    class="oe_stat_button"
+                    icon="fa-picture-o"
+                    attrs="{'invisible': [('id', '=', False)]}"
```
yajo: This domain is strange... I'm curious, is it needed?

product_sticker/views/product_attribute_views.xml
```diff
@@ -0,0 +1,25 @@
...
+                    context="{'search_default_product_attribute_id': id, 'default_product_attribute_id': id, 'search_default_grp_product_attribute_value': 1}"
+                >
+                    <field name="sticker_count" widget="statinfo" string="Stickers" />
+                </button>
```
yajo: I miss a smart button like this in the product view.

product_sticker/views/product_sticker_views.xml
```diff
@@ -0,0 +1,299 @@
+<?xml version="1.0" encoding="UTF-8" ?>
+<odoo>
+    <record id="action_product_sticker" model="ir.actions.act_window">
+        <field name="name">Product Stickers</field>
+        <field name="res_model">product.sticker</field>
+        <field name="view_mode">kanban,tree,form</field>
+    </record>
```
yajo, line 7: I miss a menu that calls this action, to be able to review this module independently.

I know this is just a base module, but since it's defining some views, it's useful to expose them somewhere. The common practice is to put it under _Settings > Technical_. After all, with the permissions exposed in this base module, you need access to settings to be able to manage stickers. And that menu is hidden by default without debug mode.

product_sticker/views/product_sticker_views.xml
```diff
@@ -0,0 +1,299 @@
...
+                                    attrs="{'invisible': [('image_1920_size', '&gt;', 30)]}"
+                                >Aceptable file size</span>
+                                <span
+                                    class="text-warning font-italic font-weight-light"
+                                    attrs="{'invisible': ['|', ('image_1920_size', '&lt;=', 30), ('image_1920_size', '&gt;', 60)]}"
+                                >Huge file size. The image should be optimized/reduced</span>
+                                <span
+                                    class="text-danger font-italic font-weight-light"
+                                    attrs="{'invisible': [('image_1920_size', '&lt;=', 60)]}"
+                                >Optimization required! Reduce the image size</span>
```
yajo: Is this really needed?

When you use `image.mixin` (like you're doing), all you need to care about is the `image_1920` field. [Then you have 4 more sizes automatically produced](https://github.com/odoo/odoo/blob/19ff7a72eee6242e84e242ac0bdbf2d78a1be136/odoo/addons/base/models/image_mixin.py#L15-L19).

So, if you need to use a smaller sticker in the report, all you need to do is use `image_256` field or similar. I don't think we have to bother the user (and dev) about this detail that is solved by Odoo automatically.

product_sticker/models/product_sticker.py
```diff
@@ -0,0 +1,208 @@
...
+    can_image_128_be_zoomed = fields.Boolean(
+        string="Allow zooming in 128px image",
+        compute="_compute_can_image_128_be_zoomed",
+        store=True,
+    )
+    show_sticker_note = fields.Boolean(
+        string="Sticker Note",
+        help="If checked, the note will be displayed with the sticker",
+    )
+    note = fields.Text(
```
yajo, line 40: I don't think it's a good idea to make users write HTML. Let's just use the appropriate field, don't you think?

```suggestion
    note = fields.Html(
```

yajo (changes requested):
Thanks, much better. Code and functional review.

product_sticker/views/product_attribute_views.xml
Shide: It's for the context below, but can be removed with an `oe_read_only` class

product_sticker/models/product_sticker.py
Shide: No, it's better to use a simple text label and display always the same format. It's like a caption on image.

product_sticker/views/product_sticker_views.xml
Shide: It's fine to warn if the image_1920 is too big. The image_128 will be proportionally weighter.

francesco-ooops (approved):
Thank you for the documentation improvement!

LGTM

product_sticker/models/product_sticker.py
```diff
@@ -0,0 +1,208 @@
...
+        translate=True,
+        help="Used to display a note with the sticker.\n"
+        "Will be HTML sanitized and displayed as plain text.",
+    )
+
+    def get_html_sanitized_note(self):
+        self.ensure_one()
+        if self.show_sticker_note and self.note:
+            return "<br/>".join(html2plaintext(self.note).split("\n"))
+        return ""
```
yajo, line 50: Hi! If you don't want to use `fields.Html`, then please remove the HTML feature from this field.

The main point is that we can't expect users to write HTML. So we either use the Html field type, or we just don't use HTML and use only plain text.

Thanks!

product_sticker/views/product_sticker_views.xml
yajo: I double-checked this because I couldn't understand how that could be true... I'm afraid it's not how you say 😕 . I downloaded the same image at scales 4608x3456px (2,9MB) and 474x355px (30,4 kB).

I uploaded one sticker for each, and I ran `read` over their fields, passing the context `bin_size: true` to get the attachments sizes. The results are:

  | big image | small image
-- | -- | --
image_1920 | 709.35 Kb (1) | 29.66 Kb (2)
image_1024 | 247.44 Kb | 29.66 Kb (2)
image_512 | 73.54 Kb | 29.66 Kb (2)
image_256 | 21.48 Kb | 21.79 Kb
image_128 | 6.24 Kb | 6.31 Kb

Notes:

1. Although I uploaded an image of 2,9 MB, even the biggest size `image_1920` got an automatic shrink to 709.35 Kb.
2. When I uploaded the small image, `image_1920`, `image_1024` and `image_512` contained just the same data (29.66 Kb) because the original image was still smaller than those sizes (474x355px).

I hope this helps you see how the size of the image uploaded by the user is not important when using the `image.mixin` class. All we need to do is select the appropriate size field when *using* the image, but use the biggest one when *uploading* it. The rest is automatic. This makes it simpler for the user, and it's also good because it means removing code that is not needed.

Now that this is clear, please remove all the code related to image size warning.

product_sticker/models/product_sticker.py
```diff
@@ -0,0 +1,208 @@
...
+        string="Company",
+        default=lambda s: s.env.company,
+    )
+    sequence = fields.Integer(default=10, index=True)
+    name = fields.Char(required=True, translate=True)
+    image_1920 = fields.Image(required=True)
+    image_1920_size = fields.Integer(
+        compute="_compute_image_1920_size",
+        store=False,
+    )
```
yajo: This is probably not needed as explained in the other comment. Please remove it.

product_sticker/models/product_sticker.py
```diff
@@ -0,0 +1,208 @@
...
+        comodel_name="product.attribute.value",
+        string="Attribute value",
+        ondelete="cascade",
+        domain="[('attribute_id', '=', product_attribute_id)]",
+    )
+    can_image_128_be_zoomed = fields.Boolean(
+        string="Allow zooming in 128px image",
+        compute="_compute_can_image_128_be_zoomed",
+        store=True,
+    )
```
yajo: I don't see this field being used anywhere in this PR or in https://github.com/OCA/stock-logistics-reporting/pull/263.

Please remove it.

product_sticker/models/product_sticker.py
```diff
@@ -0,0 +1,208 @@
...
+                image_size = 0
+            sticker.image_1920_size = image_size
+
+    @api.depends("image_1920", "image_128")
+    def _compute_can_image_128_be_zoomed(self):
+        for image in self:
+            image.can_image_128_be_zoomed = (
+                image.image_1920
+                and tools.is_image_size_above(image.image_1920, image.image_128)
+            )
```
yajo: Seems like all this can be removed too.

product_sticker/readme/USAGE.rst
```diff
@@ -0,0 +1,20 @@
...
+A general use case is: we have the use case than depending of the attribute of
+the product or the value of the product it is mandatory legally print the
+**sticker** in **Delivery slip** of stock picking.
+
+A specific example of use case: when selling ecological products European Union
+makes mandatory print official
+`ECO Logo <https://agriculture.ec.europa.eu/farming/organic-farming/organic-logo_en>`_
+in picking and this depends on the product attribute or values.
+An apple could be ecology or not and depending the attribute this should
+be printed or not.
```
yajo: Minor spell corrections:

```suggestion
A specific example of use case: when selling ecological products, the European Union
makes it mandatory to print official
`ECO Logo <https://agriculture.ec.europa.eu/farming/organic-farming/organic-logo_en>`_
in pickings, and this depends on the product attribute or values.
An apple could be ecologic or not and depending the attribute this should
be printed or not.
```

yajo (changes requested):
Thanks for the enhancements, almost there! 😊 

(please make sure to review https://github.com/OCA/product-attribute/pull/1282#discussion_r1127475506. It's hidden and I can't mark as unresolved, I don't know why)

product_sticker/models/product_sticker.py
Shide: It's just a converter to display the "\n" on reports. 
Also to add some security, I convert HTML syntax to plain text.

yajo (approved):

PR #1286, opened 2023-03-01
[16.0][MIG] product_uom_updatable

glitchov:
Standard migration from 14.0


product_uom_updatable/models/product_template.py
```diff
@@ -0,0 +1,53 @@
...
+        res = super().write(vals)
+        return res
+
+    def _update_uom(self, uom_id, field_name):
+        uom_obj = self.env["uom.uom"]
+        sorted_items = sorted(self, key=lambda r: r[field_name])
+        for key, products_group in groupby(sorted_items, key=lambda r: r[field_name]):
+            products = reduce(
+                lambda x, y: x | y, products_group, self.env["product.template"]
+            )
```
lmignon: @glitchov It's not a good idea to use the operator '|' to merge list of recordset. It's less performant than the previous implementation ` product_ids = [product.id for product in products_group]` and far more difficult to read.

lmignon (changes requested):

product_uom_updatable/models/product_template.py
glitchov: I need the recordset to invalidate after the update and i thought it was easier to get the ids from recordset than having the ids and re-browse for recordset.

lmignon (approved):
Thank you @glitchov   LGTM (Code review)

jbaudoux (approved):

rousseldenis (approved):

PR #1296, opened 2023-03-06
[16.0][MIG] product_lot_sequence: Migration to 16.0

AungKoKoLin1997:
@qrtl


grindtildeath:
@AungKoKoLin1997 Thanks for the mig. Can you please bring in the change from https://github.com/OCA/product-attribute/pull/1297 to fix readme generation?
I'll try to review your PR as soon as I'll have time to have a look on how serials are handled in v16.0.

yostashiro:
Tested locally and the it looks good as far as the migration is concerned, however I notice a few points to improve:

- The default number of digits for sequence generation is fixed to 7, which can be inconvenient if user prefers to use a different number.
- The custom fields on product (lot_sequence_prefix, etc.) do not show in the form view even when tracking by serial/lot is selected for a new produce being created (i.e. record has not been saved yet), because `is _compute_display_lot_sequence_fields()` is not triggered for new records.
- `lot_sequence_prefix` and `lot_sequence_padding` should be made invisible when `lot_sequence_id` is assigned to the product, since these values are no longer relevant (changes on the sequence record is not reflected back on the product) once the sequence is generated.  Therefore showing these fields is unnecessary or misleading.

@AungKoKoLin1997 The above points have been taken care of in https://github.com/qrtl/axls-oca/pull/33/commits/a08d7325692a2d4b3e8a69077cdaf6df06b769a0. Can you please update this PR with this commit content.

yostashiro:
@AungKoKoLin1997 Please squash the last two commits and assign a proper commit message.

yostashiro (approved):

gaikaz (approved):
Code review, lgtm

Mantux11 (approved):
Code review, LGTM.

AungKoKoLin1997:
@product-attribute-maintainers
Can we get this PR merged?

Du-ma (approved):
Code looks good, but I dunno if this was the right approach with adding an IMP to a migration PR.
This could be the reason maintainers r sleeping on this :man_shrugging:
You could also cover the IMP with tests for more cov

PR #1300, opened 2023-03-10
[15.0][FIX] product_abc_classification_sale: Flush data in tests

pedrobaeza:
Using models that rely on an SQL view (like sale.report) requires that data are already dumped to DB before querying it, which is not always the case depending on the ORM stack and installed modules.

We should ensure that dump using flush before operating with such data at SQL level.

@Tecnativa


pilarvargas-tecnativa (approved):
thanks!

carlosdauden (approved):

PR #1302, opened 2023-03-16
[16.0][MIG] product_category_product_link

astirpe:


legalsylvain (commented):
Hi. 
- the name if product_category_product_link, but you don't add a smart button in the view of the category. (to go to the templates). Other OCA modules have that features. Could you add it ?

PR #1303, opened 2023-03-17
product_supplierinfo_group: fix deleting product_supplierinfo_group

sebastienbeau:
The delete cascade was missing so it was not possible to delete a product_supplierinfo_group
@Kev-Roche


Kev-Roche (approved):
LGTM

PR #1311, opened 2023-03-29
14.0 [IMP] product_pricelist_supplierinfo: add currency conversion in seller fetch sorting

FrancescoBallerini:
This PR improve methods _prepare_seller() and _select_seller() in relation to OCA implementations of method _get_supplierinfo_pricelist_price(). 

The returned seller is fetched by sorting by [sequence, min qty, price, id](https://github.com/odoo/odoo/blob/194e57355470565cd3b7a62f74fe93cfb87c06e9/addons/product/models/product.py#L610), but the price key will not sort consistently if product has seller_ids with different currencies. After this PR price key will sort consistently in a multi-currency scenario. 

This implementation uses a param argument that is natively passed to _seller_method() and will propagate params for currency conversion, but might also contains other params (added an hook to provide extensibility). For more informations, read the commit: [9edee3a](https://github.com/OCA/product-attribute/pull/1311/commits/9edee3a3a113ceeb1002aadeffd6c8a24eb3d49a)

#### Note / reminder for me: 
the currency conversion cannot be done directly in [_get_supplierinfo_pricelist_price()](https://github.com/OCA/product-attribute/blob/33330d151400cf1c13b94ad28eecdf59a6056703/product_pricelist_supplierinfo/models/product_template.py#L32-L39) simply because sellers are not sorted there, the conversion has to be done in _prepare_sellers() so we cannot merge the two convert() even if we want to. 

This means that we extend the behaviour of native method with some line of condes which is totally imho fine because the behaviour remains substantially untouched, it's now a little more verbose but it provide some extensibility for later use ; )

Since we touch the native methods though we should consider to make a PR to odoo/odoo to make _prepare_sellers() and _select_sellers() improvement more stable and supported by default, then leave the remaining lines of code in this PR.


FrancescoBallerini:
### Closed


I was going to simplify the PR by making use of context instead of custom hook, to provide parameter for currency conversion. 


However, after self-reviewing my code I realized that I was going to benefit this PR only because I was planning to use it along with other custom implementations, but re-sorting by price with currency conversion does not fit well the way Odoo wants to provide seller price. 

In fact, re-sorting by price (converted) was basically going to override the sort done in `_prepare_seller` ([sequence, min qty, price, id](https://github.com/odoo/odoo/blob/194e57355470565cd3b7a62f74fe93cfb87c06e9/addons/product/models/product.py#L610)), which is obviously not good, plus it's not really needed anyway as the prevalent way Odoo is going to provide you the seller is the `sequence` key: other criterias are only used to provide an ordered list of records in order to be able to filter it properly, in case of multiple purchase from the same seller.

So the `price` key shouldn't need a conversion in this scenario as the same seller is always going to have same currency.

PR #1314, opened 2023-03-31
[16.0][FIX] product_logistics_uom: Store value in system UOM

### LINKED ISSUE #1312: [16.0] product_logistics_uom: Introduces inconsistencies
## Module
product_logistics_uom

## Describe the bug

In odoo 16 at least, default uoms are defined into the config parameters for volume and weight and length.  To ensure consistencies between values for such a field, values into the database should always be stored into the default uom. 
The `product_logistics_uom` addon reintroduce new default config parameter and allows to store the product's volume an weight into an other uom that the system default. 

Even if we want to allow the user to choose the volume and weight uom per product, the value stored into the DB must be expressed into the default uoms.

**Expected behavior**
At least 16.0

lmignon:
Before this change, the volume and weight values on product were no more stored into the expected kg and m3 uoms. To know the effective value you had to convert the value using the specified uom on the record. As side effect the arithmetic operations done on these fields in others addon as for exemple in 'delivery' were not correct.

This change restore the default behavior by always storing the values into the default uoms whatever the uom specified on the product. 2 new fields are added on the product to allows the user to see and managed these values in the specified uom. When values are updated through an update of these new fields, the new values are converted into the default uoms and the result stored into the original fields.

fixes #1312

### DIFF
--- a/product_logistics_uom/models/product_template.py
+++ b/product_logistics_uom/models/product_template.py
@@ -8,6 +8,28 @@ from odoo import api, fields, models
 class ProductTemplate(models.Model):
     _inherit = "product.template"
 
+    product_volume = fields.Float(
+        "Volume in product UOM",
+        compute="_compute_product_volume",
+        inverse="_inverse_product_volume",
+        digits="Volume",
+    )
+    product_weight = fields.Float(
+        "Weight in product UOM",
+        compute="_compute_product_weight",
+        digits="Stock Weight",
+        inverse="_inverse_product_weight",
+        store=True,
+    )
+
+    # remove rounding from volume and weight
+    # this is needed to avoid rounding errors when converting between units
+    # and is safe since we display the volume and weight in the product's
+    # volume and weight UOM. In the same time, we need to keep the volume
+    # we ensure that no information is lost by storing the volume and weight
+    # without rounding.
+    volume = fields.Float(digits=False)
+    weight = fields.Float(digits=False)
     volume_uom_id = fields.Many2one(
         "uom.uom",
         string="Volume Unit of Measure",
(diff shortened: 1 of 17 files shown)


lmignon:
ping @hparfr

product_logistics_uom/models/product_template.py
```diff
@@ -38,11 +60,19 @@ class ProductTemplate(models.Model):
         readonly=True,
     )
 
-    product_uom_readonly = fields.Boolean(compute="_compute_product_uom_readonly")
+    show_volume_uom_warning = fields.Boolean(
+        help="Technical field used to warm the user to change the volume"
```
hparfr:
```suggestion
        help="Technical field used to warn the user to change the volume"
```

product_logistics_uom/models/product_template.py
```diff
@@ -38,11 +60,19 @@ class ProductTemplate(models.Model):
...
+        "uom since the value for product_volume is too small and has been"
+        "rounded.",
+        compute="_compute_show_volume_uom_warning",
+    )
 
-    def _compute_product_uom_readonly(self):
-        # helper for view form
-        self.product_uom_readonly = not self.env.user.has_group("uom.group_uom")
+    show_weight_uom_warning = fields.Boolean(
+        help="Technical field used to warm the user to change the weight"
```
hparfr:
```suggestion
        help="Technical field used to warn the user to change the weight"
```

hparfr (approved):
Thanks. 

It's a bit different from my approach to always specify the unit with the value, and I don't see side effects.

LGTM

FrancoMaxime (approved):
LGTM, code review

product_logistics_uom/readme/DESCRIPTION.rst
```diff
@@ -5,3 +5,16 @@ Without this module, you only have the choice between Kg or Lb(s) and m³ for al
...
+
+Even if you choose another UoM for the weight or volume, the system will still
+store the value for these fields in the Odoo default UoM (Kg or Lb(s) and m³).
+This ensures that the arithmetic operations on these fields are correct and
+consistent with the rest of the addons.
+
+Once this addon is installed values stored into the initial Volume and Weight fields
+on the product and product template models are no more rounded to the decimal
+precision defined for these fields. This could lead to some side effects into
+rapports where these fields are used. You can replace the fields by the new
```
bguillot:
```suggestion
reports where these fields are used. You can replace the fields by the new
```

product_logistics_uom/readme/DESCRIPTION.rst
```diff
@@ -5,3 +5,16 @@ Without this module, you only have the choice between Kg or Lb(s) and m³ for al
...
+ones provided by this addon to avoid this problem (product_volume and product_weight).
+In any cases, since you use different UoM by product, you should most probably
+modify your rapports to display the right UoM.
```
bguillot:
```suggestion
modify your reports to display the right UoM.
```

bguillot (changes requested):

lmignon:
@bguillot Changes done

PR #1318, opened 2023-04-07
[MIG][16.0] product_expiry_configurable

rousseldenis:


rousseldenis:
@albariera @etobella As you worked on this.

Changed the behaviour to use computed/stored/readonly=False fields in order to simplify the code.
Rely now on the use_expiration_date field on products - introduced it on product category level.

rousseldenis:
Missing migration scripts

etobella:
@rousseldenis I am not sure, but I think you removed all original logic... The main idea of the module is to allow to define the removal_time and alert_time from the expiration_date as in some cases it is the data received from purchase orders (for example, GS1 barcodes data sends the expiration_date). It was done here: https://github.com/OCA/product-attribute/pull/1318/commits/e70d1321417f6fb66be1696ce0d80c716690a210#diff-479ba875d5052f19dcbdbb17c53b0169f9dbab2121352407c5865c2e24980eceL31

You removed all, why?

rousseldenis:
> @rousseldenis I am not sure, but I think you removed all original logic... The main idea of the module is to allow to define the removal_time and alert_time from the expiration_date as in some cases it is the data received from purchase orders (for example, GS1 barcodes data sends the expiration_date). It was done here: [e70d132#diff-479ba875d5052f19dcbdbb17c53b0169f9dbab2121352407c5865c2e24980eceL31](https://github.com/OCA/product-attribute/commit/e70d1321417f6fb66be1696ce0d80c716690a210#diff-479ba875d5052f19dcbdbb17c53b0169f9dbab2121352407c5865c2e24980eceL31)
> 
> You removed all, why?

expiration_date is now well set by Odoo in core.

With this implementation, it's possible to have:

- expiration date specific per product (set it manually).
- expiration date coming from the category (with parent change behaviour)

So, IMHO this has to be tested intensively in a functional point of view, but I feel that the current implementation should cover all the previous use cases.

etobella:
I didn't know that!!! Thanks!!!

image

Then, do we really need the module?

rousseldenis:
> I didn't know that!!! Thanks!!!
> 
> image
> 
> Then, do we really need the module?

Yes:

- To allow the product computed/store=True/readonly=False on all date fields.
- To allow default values coming from the category
- To add the 'use_expiration_date' that is in core on product level, on category one.

rousseldenis:
@etobella This one is ready. I've updated the README too

hugo-cordoba (approved):
LGTM (code review)

etobella (approved):

PR #1323, opened 2023-04-13
[15.0[FIX] product_lot_sequence: Fix sequence incrementation opening detailed operations

grindtildeath:
When opening the detailed operations view of a stock move, Odoo is calling stock.production.lot._get_next_serial to set stock.move.next_serial field anytime the product is tracked by serial and the move is assigned.

If we use this module with product or global policy, the respective sequence will therefore be called and incremented anytime this view is open, even if the picking is not creating lots (e.g. delivery orders or internal transfers)

To avoid incrementing the sequence unnecessarily, we only allow to get the next sequence number the first time this view is opened and if the move has to create new serial numbers.
Otherwise, we force the value to be set to stock.move.next_serial field to an empty string if no serial has to be created through this move, or to the next_serial value assigned on the first opening of the view.

### DIFF
--- a/product_lot_sequence/models/stock_move.py
+++ b/product_lot_sequence/models/stock_move.py
@@ -1,0 +1,23 @@
+# Copyright 2023 Camptocamp SA
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl)
+from odoo import models
+
+
+class StockMove(models.Model):
+    _inherit = "stock.move"
+
+    def action_show_details(self):
+        """Avoid calling and incrementing the sequence if not needed or already done"""
+        seq_policy = self.env["stock.production.lot"]._get_sequence_policy()
+        if seq_policy in ("product", "global"):
+            # If move is not supposed to assign serial pass empty string for next serial
+            if not self.display_assign_serial:
+                return super(
+                    StockMove, self.with_context(force_next_serial="")
+                ).action_show_details()
+            # If the sequence was already called once, avoid calling it another time
+            elif self.next_serial:
+                return super(
+                    StockMove, self.with_context(force_next_serial=self.next_serial)
+                ).action_show_details()
+        return super().action_show_details()
--- a/product_lot_sequence/models/__init__.py
+++ b/product_lot_sequence/models/__init__.py
@@ -1,2 +1,3 @@
 from . import product
 from . import stock_production_lot
+from . import stock_move
--- a/product_lot_sequence/models/stock_production_lot.py
+++ b/product_lot_sequence/models/stock_production_lot.py
@@ -56,6 +56,8 @@ class ProductionLot(models.Model):
 
     @api.model
     def _get_next_serial(self, company, product):
+        if "force_next_serial" in self.env.context:
+            return self.env.context.get("force_next_serial")
         seq_policy = self._get_sequence_policy()
         if seq_policy == "product":
             seq = product.product_tmpl_id.lot_sequence_id
--- a/product_lot_sequence/readme/ROADMAP.rst
+++ b/product_lot_sequence/readme/ROADMAP.rst
@@ -1,0 +1,11 @@
+* There is an issue with the use of ir.sequence with the newer version of Odoo.
+
+Mostly, when opening the detailed operations of an assigned picking for a product
+tracked by serial numbers, Odoo systematically calls `_get_next_serial` even
+if there is not any serial number to generate.
+Moreover, the widget allowing to generate the serial numbers will not call
+the sequence but only increment the number according to the next serial,
+potentially leading to a sequence that is not in sync anymore with the created
+serial numbers.
+
+cf https://github.com/OCA/product-attribute/issues/1326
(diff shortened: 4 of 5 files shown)


AungKoKoLin1997 (approved):

HviorForgeFlow (approved):

product_lot_sequence/models/stock_move.py
```diff
@@ -0,0 +1,23 @@
...
+
+    def action_show_details(self):
+        """Avoid calling and incrementing the sequence if not needed or already done"""
+        seq_policy = self.env["stock.production.lot"]._get_sequence_policy()
+        if seq_policy in ("product", "global"):
+            # If move is not supposed to assign serial pass empty string for next serial
+            if not self.display_assign_serial:
+                return super(
+                    StockMove, self.with_context(force_next_serial="")
+                ).action_show_details()
```
florentx, line 17: Variable `self` is looked up in local context by `super()`, so you could simplify:
```suggestion
                self = self.with_context(force_next_serial="")
```

product_lot_sequence/models/stock_move.py
```diff
@@ -0,0 +1,23 @@
...
+            # If the sequence was already called once, avoid calling it another time
+            elif self.next_serial:
+                return super(
+                    StockMove, self.with_context(force_next_serial=self.next_serial)
+                ).action_show_details()
```
florentx, line 22: and here:
```suggestion
                self = self.with_context(force_next_serial=self.next_serial)
```

kanda999:
Thank you for PR.
I was not satisfied with the logic proposed by First SN in the product_lot_sequence module from before.
This PR fixes the problem of the First SN being updated every time the wizard is opened.

However, I still think the first SN proposal is not complete and there are still problems.
It happens in the following operations.

ex)

1, Five (or multiple) serial numbers are assigned to serial-managed products in one delivery. 
2023-04-17_17h34_54


2, This delivery is verified and received.

3, Create a new delivery. But then the first SN proposed is 0000002.
2023-04-17_17h36_15

4, Assign the first SN proposed and verified delivery and then get "The serial number has already been asigned" because 0000002 is already in stock.

To solve this problem, the first SN proposal should be the next largest number the assigned SN.
I may work on this issue. I would then use this PR change.  Do you have any good solutions to this problem?

product_lot_sequence/models/stock_move.py
grindtildeath: I didn't want to propagate the context key outside of the super call scope (what would be the case if we alter `self`)

product_lot_sequence/models/stock_move.py
grindtildeath: cf https://github.com/OCA/product-attribute/pull/1323#discussion_r1168915488

grindtildeath:
Hello @kanda999 

Thanks for your comment. 

I agree there is an issue with the actual implementation of the module and the way this view is working. My gut feeling is that it's probably counter intuitive to use sequence objects to handle generation and creation of lots/SN in the newer version, at least for goods that are purchased and where lots/SN are usually provided by the supplier.

I see two possible solutions:
 - either we remove the ability to set "First SN" on the stock move view, and use the "Assign serial numbers" action to call the sequence as many times as we have "Number of SN", so that the sequence will still be in sync with the serials that are generated. (in that case I'm also wondering if we shouldn't just remove the call I'm modifying here to increment the sequence on the opening of the view)
 - avoid using sequences altogether and rely on the search of last lot/SN as is now the case in standard (https://github.com/odoo/odoo/blob/15.0/addons/stock/models/stock_production_lot.py#L68-L70). Eventually remove the product parameter for the search to use a "global" sequence, or add extra parameters to mimic a "product" specific sequence. 

First one would be the more logical move, but I tend to think the second looks like the better long term solution although that would of course require an extra hook in Odoo (or overriding search with a context key). 

@AdriaGForgeFlow @LoisRForgeFlow what do you think?

yostashiro:
> I see two possible solutions:
> 
> * either we remove the ability to set "First SN" on the stock move view, and use the "Assign serial numbers" action to call the sequence as many times as we have "Number of SN", so that the sequence will still be in sync with the serials that are generated. (in that case I'm also wondering if we shouldn't just remove the call I'm modifying here to increment the sequence on the opening of the view)
> * avoid using sequences altogether and rely on the search of last lot/SN as is now the case in standard (https://github.com/odoo/odoo/blob/15.0/addons/stock/models/stock_production_lot.py#L68-L70). Eventually remove the product parameter for the search to use a "global" sequence, or add extra parameters to mimic a "product" specific sequence.
> 
> First one would be the more logical move, but I tend to think the second looks like the better long term solution although that would of course require an extra hook in Odoo (or overriding search with a context key).

@grindtildeath Thank you very much for this PR. I agree. The second option sounds better for the long term as it's more in line with the standard design.

In the meantime, our workaround for our client for now is to use https://github.com/OCA/stock-logistics-workflow/tree/16.0/stock_picking_auto_create_lot together with this module (containing the update of his PR).  Not sure there is actually a situation where manual intervention is required in SN generation, when numbering rule is pre-defined (either for the product or globally).  Therefore the use of stock_picking_auto_create_lot works well for us to bypass the design incompatibility between this module and the standard SN generation.

simahawk:
@grindtildeath can you create an issue to follow up w/ https://github.com/OCA/product-attribute/pull/1323#issuecomment-1511681322 and link it in the roadmap? Then we can merge I think.

grindtildeath:
@simahawk https://github.com/OCA/product-attribute/issues/1326 added in roadmap

PR #1193, opened 2022-11-10
[16.0][MIG] product_secondary_unit

otmanelagy:

PR #1325, opened 2023-04-18
[16.0][MIG] product_secondary_unit #2

remihb:
Adding changes of #1246 

In replacement of #1193 

### Changes
- Reintroduction of product_product part (as it was in version 15.0)
  - [https://github.com/OCA/product-attribute/tree/15.0/product_secondary_unit](url)
- Use of **`precompute=True`** for ProductSecondaryUnitMixin.secondary_uom_qty instead of `default`
  - Else if used with modules like _sale_order_secondary_unit_ wich depends on this module, it will raise
    `UserWarning: Field sale.order.line.product_uom_qty cannot be precomputed as it depends on non-precomputed field sale.order.line.secondary_uom_qty`
- Use of TransactionCase instead of SavepointCase
  - [product_secondary_unit/tests/test_product_second_unit.py#L3](https://github.com/OCA/product-attribute/blob/3702a72f29bc3ad820d5b651cf33fe59468fadba/product_secondary_unit/tests/test_product_second_unit.py#L3)
- Removal of empty group in view and wrong usage of `group_id` (`groups` instead)
  - [product_secondary_unit/views/product_views.xml#L9](https://github.com/OCA/product-attribute/blob/3702a72f29bc3ad820d5b651cf33fe59468fadba/product_secondary_unit/views/product_views.xml#L9)

### Question

- What about the odoo_test_helper needed in manifest ? (not in base odoo requirements)


BT-aleonard:
Hi @remihb ,
Why is https://github.com/OCA/product-attribute/pull/1193 replaced by this PR?
Thanks.

product_secondary_unit/views/product_views.xml
```diff
@@ -0,0 +1,47 @@
+<?xml version="1.0" encoding="utf-8" ?>
+<odoo>
+    <record id="product_template_form_view" model="ir.ui.view">
+        <field name="name">Product template Secondary Unit</field>
+        <field name="model">product.template</field>
+        <field name="inherit_id" ref="product.product_template_form_view" />
+        <field name="arch" type="xml">
+            <xpath expr="//page[@name='general_information']/group[2]" position="after">
```
BT-aleonard: The only thing I'd change here is to use a relative reference.
And set columns for the group and also for the field as well.
`<xpath
                expr="//page[@name='general_information']/group[last()]"
                position="after"
            >
                <group
                    string="Secondary Unit of Measure"
                    col="4"
                    groups="uom.group_uom"
                >
                    <field name="secondary_uom_ids" colspan="4" nolabel="1">`

BT-aleonard (changes requested):

product_secondary_unit/views/product_views.xml
remihb: Ok done

remihb:
> Hi @remihb , Why is #1193 replaced by this PR? Thanks.

As I mentioned, #1193 doesn't include de product variants part, and more important, without` precompute=True` on ProductSecondaryUnitMixin.secondary_uom_qty, other modules like _sale_order_secondary_unit_ or _stock_secondary_unit_ will fail at install

rousseldenis:
@otmanelagy Do you mind if we swtitch to this ?

rousseldenis (approved):
Code review

BT-aleonard (approved):

product_secondary_unit/tests/test_secondary_unit_mixin.py
```diff
@@ -1,5 +1,3 @@
-# Copyright 2021 Tecnativa - Sergio Teruel
-# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
```
FrancoMaxime: Why did you remove the License?

product_secondary_unit/views/product_views.xml
```diff
@@ -1,16 +1,20 @@
 <?xml version="1.0" encoding="utf-8" ?>
-<!-- Copyright 2018 Tecnativa - Sergio Teruel
-     License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl). -->
```
FrancoMaxime: Why did you remove the License?

requirements.txt
```diff
@@ -1,2 +1,3 @@
 # generated from manifests external_dependencies
+odoo_test_helper
```
FrancoMaxime: see comment in the manifest file

FrancoMaxime (changes requested):

product_secondary_unit/__manifest__.py
```diff
@@ -12,6 +12,7 @@
     "application": False,
     "installable": True,
     "depends": ["product"],
+    "external_dependencies": {"python": ["odoo_test_helper"]},
```
FrancoMaxime: Its a bad idea to add odoo_test_helpder in the external_dependencies key since its only use for the unit test. A better way to do that is to use test-requirements.txt file

FrancoMaxime (changes requested):

BT-aleonard:
If you set `precompute=True` on the field secondary_uom_qty of the mixin, the models that are not using precompute yet (for example purchase.order.line) will fail when this fields depends on them

BT-aleonard:
Hi, 
Can you please check this PR, please? 
@sergio-teruel

sergio-teruel:
@BT-aleonard Have you seen this PR (https://github.com/OCA/product-attribute/pull/1336)? About the secondary_uom_qty default?

product_secondary_unit/views/product_views.xml
```diff
@@ -1,16 +1,20 @@
...
 <odoo>
     <record id="product_template_form_view" model="ir.ui.view">
         <field name="name">Product template Secondary Unit</field>
         <field name="model">product.template</field>
         <field name="inherit_id" ref="product.product_template_form_view" />
-        <field name="groups_id" eval="[(4, ref('uom.group_uom'))]" />
```
sergio-teruel: Why remove the groups? Is the better way to not load the view

product_secondary_unit/views/product_views.xml
```diff
@@ -37,7 +41,6 @@
         <field name="name">Product Secondary Unit</field>
         <field name="model">product.product</field>
         <field name="inherit_id" ref="product.product_normal_form_view" />
-        <field name="groups_id" eval="[(4, ref('uom.group_uom'))]" />
```
sergio-teruel: Idem

sergio-teruel (changes requested):

BT-aleonard:
@sergio-teruel 
I didn't see it before.
But if we implement that PR (default value for secondary_uom_qty), then we can get rid of the `precompute=True.`
I'll suggest the changes.

product_secondary_unit/models/product_secondary_unit_mixin.py
```diff
@@ -0,0 +1,133 @@
...
+    def _get_default_secondary_uom(self):
+        return self.env["product.template"]._get_default_secondary_uom()
+
+    secondary_uom_qty = fields.Float(
+        string="Secondary Qty",
+        digits="Product Unit of Measure",
+        store=True,
+        readonly=False,
+        compute="_compute_secondary_uom_qty",
+        precompute=True,
```
BT-aleonard: Remove this, and stick to the default values.

product_secondary_unit/models/product_secondary_unit_mixin.py
```diff
@@ -0,0 +1,133 @@
...
+            return
+        elif self.secondary_uom_id.dependency_type == "independent":
+            return
+        factor = self._get_factor_line()
+        line_qty = self._get_quantity_from_line()
+        qty = float_round(
+            line_qty / (factor or 1.0),
+            precision_rounding=self.secondary_uom_id.uom_id.rounding,
+        )
+        self.secondary_uom_qty = qty
```
BT-aleonard, line 132: Here, you can add the default_get method.

    @api.model
```
    def default_get(self, fields_list):
        defaults = super().default_get(fields_list)
        if self.secondary_uom_id and not self.env.context.get(
            "skip_default_secondary_uom_qty", False
        ):
            defaults["secondary_uom_qty"] = 1.0
        return defaults

```

BT-aleonard (changes requested):
Hi @remihb 
As suggested by @sergio-teruel , we need to take this PR https://github.com/OCA/product-attribute/pull/1336 into consideration as well.

BT-aleonard (changes requested):
Hi @remihb 
As suggested by @sergio-teruel , we need to take this PR https://github.com/OCA/product-attribute/pull/1336 into consideration as well.

BT-aleonard (changes requested):
Hi @remihb 
As suggested by @sergio-teruel , we need to take this PR https://github.com/OCA/product-attribute/pull/1336 into consideration as well.

requirements.txt
BT-aleonard: The correct way to add this is in a file test-requirements.txt

product_secondary_unit/views/product_views.xml
remihb: Because it will raise this error `Inherited view cannot have 'Groups' define on the record. Use 'groups' attributes inside the view definition`

product_secondary_unit/views/product_views.xml
remihb: Idem :)

product_secondary_unit/tests/test_secondary_unit_mixin.py
remihb: An error, corrected

product_secondary_unit/views/product_views.xml
remihb: idem, corrected

product_secondary_unit/__manifest__.py
remihb: Done

product_secondary_unit/models/product_secondary_unit_mixin.py
remihb: Ok, PR #1336 incorporated

sergio-teruel (approved):

product_secondary_unit/models/product_secondary_unit_mixin.py
remihb: After testing, I still have errors with module **sale_order_secondary_unit** that depends on this module.

`odoo/fields.py:802: UserWarning: Field sale.order.line.product_uom_qty cannot be precomputed as it depends on non-precomputed field sale.order.line.secondary_uom_qty`

remihb: What to do, as setting precompute=True on the mixin model will fail models like purchase.order.line (https://github.com/OCA/product-attribute/pull/1325#issuecomment-1557246440)

BT-aleonard (approved):

product_secondary_unit/models/product_secondary_unit_mixin.py
BT-aleonard: Is that the only warning you get?

FrancoMaxime (approved):
LGTM: code review

john-herholz-dt:
@sergio-teruel , can you merge please? I am working on another migration depending on this module.
Thank you!

sergio-teruel:
ping @pedrobaeza

PR #1327, opened 2023-04-24
[ADD] product_attribute_company_favorite

victor-champonnois:
Possibility to set favorite product attributes per company

Using ideas in https://github.com/OCA/multi-company/pull/428/files#diff-8f691f968003b508d4fe8cc27fbeda3c7edfd64c959e0449ebc195f5ca3f5cdaR8 and https://github.com/OCA/multi-company/pull/428/files#diff-8f691f968003b508d4fe8cc27fbeda3c7edfd64c959e0449ebc195f5ca3f5cdaR8


victor-champonnois:
@bealdav

legalsylvain:
@bealdav ;-)

bealdav:
Hi @victor-champonnois thanks for these next commits

I have just tested with runboat

Here is San Francisco company
image

Here is in Chicago company
image

Here we have missing display in one company.

Do you think it's possible to have in improvement where I only can select favorites product.attribute but see all defined product.attribute associated to product.

victor-champonnois:
@bealdav 
about the attributes not displayed for the product, it's because the attribute chicag_favorite was not set to favorite for san francisco company. When I set it to favorite, it appears in the attribute list.

Note that the option to set new attributes as favorite for all companies is False by default. So since you haven't checked the option, the newly created attribute was only favorite for chicago.

product_attribute_company_favorite/models/product_attribute.py
```diff
@@ -24,26 +18,39 @@ def _name_search(
...
+                    contextual_record = record.with_company(company_id)
+                    contextual_record.is_favorite = True
+        return records
+
+
+class ProductAttribute(models.Model):
+    _inherit = ["product.attribute", "product.attribute.mixin"]
+    _name = "product.attribute"
+
+    CREATION_ACROSS_COMPANY_CONFIG = "product_attribute_company_favorite.new_attribute_favorite_for_all_companies"  # noqa
```
bealdav: You may split string I think to avoid #noqa


```suggestion
    CREATION_ACROSS_COMPANY_CONFIG = "product_attribute_company_favorite."
    "new_attribute_favorite_for_all_companies" 
```

Idem in other strings in the PR.

You may also have shorter word like CREATION_ACROSS_CPNY (not mandatory)

product_attribute_company_favorite/models/product_attribute.py
```diff
@@ -0,0 +1,73 @@
+from odoo import api, fields, models
+
+
+class ProductAttributeIsFavoriteMixin(models.AbstractModel):
+    _name = "product.attribute.mixin"
```
bealdav: With this name, it might have a name collision. An another module could use the same name
```suggestion
    _name = "product.attrib.favorite.mixin"
```

product_attribute_company_favorite/models/res_config_settings.py
```diff
@@ -0,0 +1,20 @@
+from odoo import fields, models
+
+
+class ResConfigSettings(models.TransientModel):
+    _inherit = "res.config.settings"
+
+    new_attribute_favorite_for_all_companies = fields.Boolean(
+        config_parameter="product_attribute_company_favorite.new_attribute_favorite_for_all_companies",  # noqa
```
bealdav: I suppose than that string can be also splitted here.

bealdav (commented):
Just check with runboat.
It works as expected, thanks, nice job.

I have just done remarks in naming strings.
Please could you check

product_attribute_company_favorite/models/product_attribute.py
victor-champonnois: Fixed !

product_attribute_company_favorite/models/product_attribute.py
victor-champonnois: fixed

product_attribute_company_favorite/models/res_config_settings.py
victor-champonnois: fixed

bealdav (approved):

PaulGoubert (approved):

product_attribute_company_favorite/models/product_attribute.py
```diff
@@ -0,0 +1,78 @@
...
+                    contextual_record = record.with_company(company_id)
+                    contextual_record.is_favorite = True
+        return records
+
+
+class ProductAttribute(models.Model):
+    _inherit = ["product.attribute", "product.attribute.favorite.mixin"]
+    _name = "product.attribute"
+
+    CREATION_ACROSS_COMPANY_CONFIG = (
```
legalsylvain: I don't see the interest to create a constant for that item. 

why not just use  the following code, and remove the two definition of CREATION_ACROSS_COMPANY_CONFIG ? 

        new_record_favorite_for_all_companies = (
            self.env["ir.config_parameter"]
            .sudo()
            .get_param("product_attribute_company_favorite.new_attribute_favorite_for_all_companies")
        )

product_attribute_company_favorite/models/product_attribute.py
```diff
@@ -0,0 +1,78 @@
+from odoo import api, fields, models
+
+
+class ProductAttributeIsFavoriteMixin(models.AbstractModel):
```
legalsylvain: please, split the file into 4 files. (one per class).

product_attribute_company_favorite/models/product_attribute.py
```diff
@@ -0,0 +1,78 @@
...
+    is_favorite = fields.Boolean(
+        related="attribute_id.is_favorite",
+    )
+
+
+class ProductAttributeValue(models.Model):
+    _inherit = ["product.attribute.value", "product.attribute.favorite.mixin"]
+    _name = "product.attribute.value"
+
+    is_favorite = fields.Boolean(
```
legalsylvain: I think you can set is_favorite in the mixin.

legalsylvain (changes requested):
Code review. a blocking point. Otherwise, LGTM.

Thanks !

product_attribute_company_favorite/models/product_attribute.py
victor-champonnois: Because the name of the parameter is not the same for attribute and for attribute values (`new_attribute_favorite_for_all_companies` vs `new_attribute_value_favorite_for_all_companies`)
Or maybe I did not understand your solution ?

product_attribute_company_favorite/models/product_attribute.py
victor-champonnois: done

product_attribute_company_favorite/models/product_attribute.py
victor-champonnois: OK, I wasn't sure because the help message was not the same, but I changed it to be more generic. Done

product_attribute_company_favorite/models/product_attribute.py
carmenbianca: Is there a use case for these two things to be configured separately?

If there is, a comment would be well-placed here that gives a vague idea what is going on.

product_attribute_company_favorite/models/product_attribute.py
```diff
@@ -0,0 +1,10 @@
+from odoo import models
+
+
+class ProductAttribute(models.Model):
+    _inherit = ["product.attribute", "product.attribute.favorite.mixin"]
+    _name = "product.attribute"
+
+    CREATION_ACROSS_COMPANY_CONFIG = (
+        "product_attribute_company_favorite." "new_attribute_favorite_for_all_companies"
```
carmenbianca:
```suggestion
        "product_attribute_company_favorite.new_attribute_favorite_for_all_companies"
```

product_attribute_company_favorite/models/product_attribute_is_favorite_mixin.py
```diff
@@ -0,0 +1,43 @@
...
+    _description = """Methods used both in product.attribute and
+    product.attribute.value to implement is_favorite functionalities
+    """
+
+    is_favorite = fields.Boolean(
+        string="Favorite",
+        company_dependent=True,
+        default=True,
+        help="""If checked, this record will appear when searched
+        in product attribute list in the product form""",
```
carmenbianca: This statement is not true for attribute values.

It is also not exclusively applicable to the product form, I think.

product_attribute_company_favorite/models/product_attribute_is_favorite_mixin.py
```diff
@@ -0,0 +1,43 @@
...
+        )
+        records = super().create(vals_list)
+        if new_record_favorite_for_all_companies:
+            company_ids = (
+                self.env["res.company"].with_context(active_test=False).search([]).ids
+            )
+            for record in records:
+                for company_id in company_ids:
+                    contextual_record = record.with_company(company_id)
+                    contextual_record.is_favorite = True
```
carmenbianca: There is a potential bug here, although a little unlikely. If `vals_list` contains `{"is_favorite": False}` for ≥1 record, that instruction will be overwritten by the code below.

Instead of writing `True` to the field, you should probably write `record.is_favorite`.

product_attribute_company_favorite/tests/test_product_attribute.py
```diff
@@ -0,0 +1,165 @@
...
+    def test_new_attribute_is_not_favorite_for_other_company(self):
+
+        attribute_list1 = (
+            self.env["product.attribute"]
+            .with_user(self.user2.id)
+            .search([("is_favorite", "=", True), ("id", "=", self.attribute_1.id)])
+        )
+        self.assertEqual(len(attribute_list1), 0)
+
+    def test_not_favorite_attribute_in_name_searchfor_other_company(self):
```
carmenbianca:
```suggestion
    def test_not_favorite_attribute_in_name_search_for_current_company(self):
```

product_attribute_company_favorite/tests/test_product_attribute.py
```diff
@@ -0,0 +1,165 @@
+from odoo.tests.common import TransactionCase
+
+
+class TestProductAttributeMultiCompany(TransactionCase):
```
carmenbianca, line 4: Should probably be SavepointCase in Odoo 14, TransactionCase in Odoo ≥15

product_attribute_company_favorite/tests/test_product_attribute.py
```diff
@@ -0,0 +1,165 @@
...
+        attribute_value_list = (
+            self.env["product.attribute.value"]
+            .with_user(self.user2.id)
+            .search(
+                [("is_favorite", "=", True), ("id", "=", self.attribute_1_value_1.id)]
+            )
+        )
+        self.assertEqual(len(attribute_value_list), 0)
+
+    def test_not_favorite_attribute_value_in_name_searchfor_other_company(self):
```
carmenbianca:
```suggestion
    def test_not_favorite_attribute_value_not_in_name_search_for_current_company(self):
```

carmenbianca (changes requested):
Small remarks, one bug. Otherwise LGTM; great work.

product_attribute_company_favorite/models/product_attribute.py
legalsylvain: > Because the name of the parameter is not the same for attribute and for attribute values (new_attribute_favorite_for_all_companies vs new_attribute_value_favorite_for_all_companies)

Ow ! I didn't saw. Well : 
- nitpicking : i'm in favor to remove the both definition and just insert directly the strings in the code. it will be lighter. The constant has to be used only once in the create function. no need to make factorization. (let me know if I'm not clear). 
not a blocking point.

carmenbianca: The create method is in the mixin, though, and doesn't know whether it's dealing with an attribute or an attribute value :)

product_attribute_company_favorite/tests/test_product_attribute.py
victor-champonnois: I don't understand, this code is for Odoo 16 so it's OK no ?

product_attribute_company_favorite/tests/test_product_attribute.py
victor-champonnois: fixed (with an abbreviation because line too long)

product_attribute_company_favorite/tests/test_product_attribute.py
victor-champonnois: fixed

product_attribute_company_favorite/models/product_attribute.py
victor-champonnois: Yes, sorry if I don't understand, I don't see a way to make this work without a class constant, if we want to leave the create method in the mixin.

product_attribute_company_favorite/models/product_attribute.py
victor-champonnois: fixed

victor-champonnois: >Is there a use case for these two things to be configured separately?
>If there is, a comment would be well-placed here that gives a vague idea what is going on.

I've already explained in the readme (CONFIGURE.rst), I think it's enough no ?

product_attribute_company_favorite/models/product_attribute_is_favorite_mixin.py
victor-champonnois: nice cath ! fixed

product_attribute_company_favorite/models/product_attribute_is_favorite_mixin.py
victor-champonnois: fixed

product_attribute_company_favorite/tests/test_product_attribute.py
carmenbianca: Oh it is! I was working on Odoo 14 code at the time I reviewed this; never mind this comment.

product_attribute_company_favorite/models/product_attribute.py
carmenbianca: Just a comment saying:

```python
# XMLID of configuration boolean for this model. Used in mixin to distinguish from other models.
```

would greatly help in reading the code.

You could also slightly reword the above comment and place it in the mixin version where the variable is actually used.

```python
# Class constant stores the configuration boolean for its model to distinguish itself from other models.
```

I had to spend a few minutes figuring out what was going on here.

product_attribute_company_favorite/tests/test_product_attribute.py
bealdav: Ok nice, now that's fixed, we may asked to @legalsylvain to refresh its review just before a last squash commits ? Thanks to all ?

product_attribute_company_favorite/models/product_attribute.py
victor-champonnois: fixed ! went for the second option, thanks for the examples :)

victor-champonnois:
@legalsylvain @bealdav rebased and ready to merge

legalsylvain (approved):

product_attribute_company_favorite/models/product_attribute_is_favorite_mixin.py
```diff
@@ -0,0 +1,46 @@
...
+        )
+
+    @api.model_create_multi
+    def create(self, vals_list):
+        param = "product_attribute_company_favorite.%s_enable_for_all_companies" % (
+            self._name.replace(".", "_")
+        )
+        new_record_favorite_for_all_companies = (
+            self.env["ir.config_parameter"].sudo()
+            # Class constant stores the configuration boolean
```
bealdav: Thanks for the improvement, here typo, no more constant

victor-champonnois: fixed

carmenbianca (approved):

victor-champonnois:
@gurneyalex @rousseldenis @JordiBForgeFlow This PR is green with 4 positive reviews, could you merge it ? 
PS : I am asking you because you are PSC of this repo. Is it OK to do that ?

legalsylvain:
please, wait a little bit, I have a last improvment to commit please.

bealdav:
Ha some new features ?

legalsylvain:
> please, wait a little bit, I have a last improvment to commit please.

I'm sorry, I was totally wrong. I have nothing to add. This can be merged!

bguillot:
Thanks !

PR #1328, opened 2023-04-25
[MIG] product_weight : Migration to 16.0

syera94:
@florian-dacosta @bguillot 
A cron was added to the module.
The test have been changed to use some demonstration's products.


product_weight/data/ir_cron_data.xml
```diff
@@ -0,0 +1,19 @@
...
+        <field name="code">model.cron_recompute_bom_weight()</field>
+        <field
+            name="model_id"
+            ref="product.model_product_product"
+        />  <!-- il attend le xml id du model sur lequel on veut pointer -->
+        <field name="interval_number">1</field>
+        <field name="interval_type">months</field>
+        <field name="numbercall">-1</field>
+        <field name="doall" eval="False" />
+        <field name="active" eval="True" />
```
legalsylvain: Nice improvment. Could you set it to False by default ? (as some people does'nt want to have the cron enabled, and the cron could write bad data before user disable it)

legalsylvain (changes requested):
Hi thanks for your contribution.

**Technical point** : I don't think that ``product_weight`` is a good name. too generic in my opinion. Something like ``product_weight_from_bom`` , ``bom_update_product_weight``, etc... or other name that should contain "bom".

**functional question** : You are using the ``weight`` field that is the brut weight. in a BoM context, the sum should be done on ``net_weight`` don't you think ?

the net weight was present in old version of odoo, but it has been removed for unknown reason. The field is now present in a OCA module.

- The module in v14 : https://github.com/OCA/product-attribute/tree/14.0/product_net_weight
- the module is being ported in v16 here : https://github.com/OCA/product-attribute/pull/1281

Nice to have : I guess that a button in the bom form to update the weight could be great. I mean, users are creating (or updating) bom, and when saving, want to update the weight at this step. Don't you think ?

florian-dacosta:
> Technical point : I don't think that product_weight is a good name. too generic in my opinion. Something like product_weight_from_bom , bom_update_product_weight, etc... or other name that should contain "bom".

I agree

> **functional question** : You are using the `weight` field that is the brut weight. in a BoM context, the sum should be done on `net_weight` don't you think ?

Well, that is a good question. In my opinion, this weight compute feature is really more adapter for phantom bom.
I have no case where I use this module with manufacturing, so I am not sure, but I think that for a manufactured BOM, it does not make much sense since you will probably scrap some material and the weight can't be the exact some of the component.

But for phantom (kit) bom, it make total sense, and I guess for this case, using the brut weight is fine.

So I wonder if we just adapt the cron so it recompute the weight only for phantom bom, maybe allowing to pass a domain if someone needs it for manufactured bom and think it will be consistent.

But I don't know if we want to force the use of the product_net_weight module to use this module.

syera94:
@legalsylvain @florian-dacosta I changed the name fro product_weight to product_weight_from_bom. 
If needed, the nice to have can be done later.

legalsylvain:
>  If needed, the nice to have can be done later.

In that case, could you add a sentence in the ROADMAP.rst file ? 

otherwise, LGTM. Thanks !

legalsylvain (approved):

legalsylvain (commented):
note : due to module renaming, you should renamed the .pot file and ``product_weight`` in the .po files I guess. Otherwise, translation will be lost.

syera94:
@legalsylvain @florian-dacosta  done for the roadmap and done for the .po name change.

hildickethan:
image
to fit usual commit naming, i would suggest renaming the migration commit to `[MIG] product_weight: Migration to 16.0` and then squash the rest into the refactor commit where you rename it

PR #1329, opened 2023-04-26
[16.0][MIG] product_multi_image

yibudak:
Note: I had to add "with_context(bin_size=False)" to `models/product_product.py#L34` since we expect base64 decoded data rather than binary size. See: https://github.com/OCA/OCB/blob/94b1e891968dff165b6e0968367dda6c1d9319e3/odoo/fields.py#L2300


yibudak:
I've migrated `base_multi_image` also https://github.com/OCA/server-tools/pull/2618

yibudak:
@tate11 @tarteo @pedrobaeza could you please review the line that i mentioned in the PR?

PR #1334, opened 2023-05-05
product_supplierinfo_group: give the possibility to import supplier group from product.product

sebastienbeau:
Allow to import supplier group from product.product
- autofill product_tmpl_id when product_id is define
- add a field to see only specific variant supplier group


sebastienbeau:
@Kev-Roche

Kev-Roche (approved):

PR #1336, opened 2023-05-16
[IMP] product_secondary_unit: Allow skip set default value for secondary_uom_qty

sergio-teruel:
cc @Tecnativa TT43104

ping @ernestotejeda @carlosdauden

### DIFF
--- a/product_secondary_unit/models/product_secondary_unit_mixin.py
+++ b/product_secondary_unit/models/product_secondary_unit_mixin.py
@@ -42,7 +42,6 @@ class ProductSecondaryUnitMixin(models.AbstractModel):
         store=True,
         readonly=False,
         compute="_compute_secondary_uom_qty",
-        default="1",
     )
     secondary_uom_id = fields.Many2one(
         comodel_name="product.secondary.unit",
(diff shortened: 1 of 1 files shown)


carlosdauden (approved):

product_secondary_unit/models/product_secondary_unit_mixin.py
```diff
@@ -131,3 +130,12 @@ def _onchange_helper_product_uom_for_secondary(self):
             precision_rounding=self.secondary_uom_id.uom_id.rounding,
         )
         self.secondary_uom_qty = qty
+
+    @api.model
+    def default_get(self, fields_list):
+        defaults = super().default_get(fields_list)
+        if self.secondary_uom_id and not self.env.context.get(
```
ernestotejeda, line 137: Is it necessary to check `self.secondary_uom_id`?

sergio-teruel: I ithink yes... I do not set a default value to 1.0 if not secondary uom set...

BT-aleonard (approved):
LGTM,
Thanks.

PR #1342, opened 2023-06-01
[14.0][FIX] pos_product_cost_security: missing flag on `_loadMissingProducts`

zamberjo:
The module functionality is correct. But the **pos_override_cost_security** flag needs to be added in `_loadMissingProducts` (see: https://github.com/odoo/odoo/blob/14.0/addons/point_of_sale/static/src/js/models.js#L831)


belenaures (approved):
LGTM

zamberjo:
@chienandalu can you check this?

chienandalu (approved):
Thanks @zamberjo !

PR #1345, opened 2023-06-02
[14.0][FIX] product_pricelist_supplierinfo: Use given partner to get relevant supplier infos

ypapouin:
Because the current implementation ignore the existing partner from the `products_qty_partner` tuple used to get the pricelist rule when calling `_select_sellers`


ypapouin:
This PR has been edited because we cannot rely on the `partner` variable that is filled with the customer one when used from sales. That's why I introduced the `supplier` context variable.

PR #1352, opened 2023-06-07
product_assortment: add hard link between product and assortment

sebastienbeau:
Hi all

The aim is to store the information of the result of the link between the product and the assortment.

This allow to have "hook" when a product is added/removed automatically

@lmignon @sbidoul 

Feel free to fork it if needed


gdgellatly:
Salut @sebastienbeau I'm trying to understand why here. TBH I am only looking at v14 and some basic enhancements that we want like applying on templates and categories in addition to products. Was checking open PR's and am trying to understand the need/purpose here or what issues current module causes.

PR #1353, opened 2023-06-07
[16.0][MIG] product_main_supplierinfo

aliciagaarzo:
Migration of **product_main_supplierinfo** module to version 16


product_main_supplierinfo/__manifest__.py
```diff
@@ -2,15 +2,15 @@
...
+    "version": "16.0.1.0.0",
     "summary": "Display the main vendor of a product.",
     "website": "https://github.com/OCA/product-attribute",
     "author": "Camptocamp, Odoo Community Association (OCA)",
     "maintainers": ["sebalix"],
     "development_status": "Beta",
     "license": "AGPL-3",
     "category": "Product",
-    "depends": ["product", "purchase"],
+    "depends": ["product", "purchase", "sale"],
```
sebalix, line 13: @aliciagaarzo can you explain why we need `sale` as dependency starting from 16.0?
We should be able to install only `purchase` and manage our suppliers without `sale`, there is probably something to fix if we need it now.

sebalix (changes requested):

bofiltd (approved):

PR #1354, opened 2023-06-07
[MIG] product_pricelist_simulation_margin: Migration to 16.0

hugo-cordoba:
ADD new module product_pricelist_simulation_margin into version 16.0


product_pricelist_simulation_margin/wizards/wizard_preview_pricelist.py
```diff
@@ -0,0 +1,74 @@
...
+    )
+
+    margin_percent = fields.Float(
+        string="Margin (%)",
+        store=True,
+        digits=dp.get_precision("Product Price"),
+        readonly=True,
+    )
+
+    bg_color = fields.Char(compute="_compute_bg_color")
```
legalsylvain: I wonder if we should remove bg_color and computation and so one. That was a try in V12, but in fact, the result is not good, and my user asked me to disable the backgrounds.

Not a blocking point. What do you think ?

legalsylvain (commented):
partial review. 
- could you take a look on the red CI ? 
- ideally, adding a test could be great.
(I think you can recover code base here https://github.com/OCA/margin-analysis/blob/12.0/product_pricelist_margin/tests/test_module.py and adapt with the new name of the transient model).

legalsylvain:
**Side question :** 
do we have to put that module in product-attribute (as it depends on module product_pricelist_simulation) or do we have to put it in margin-analysis, because it is related to margin ? 
No idea on that topic. Maybe @pedrobaeza has a point of view.

hugo-cordoba:
hi @legalsylvain.

I haven't finished migrating the module yet, I stopped for lunch and left the PR uploaded so as not to lose it. That's why I hadn't pinged you yet. I'm going to add the test and fix the pre-commit bugs.

Regarding what you comment about 'bg_color', I agree with you, I don't see sense that this is still there, in my experience, I have never used it, nor do I find that it brings a great added value.

Finally, in my opinion, the module should stay in this repository.

pedrobaeza:
Yes, being such tiny extension to a module being hosted here, it would be weird to not be located in the same place.

legalsylvain:
> I haven't finished migrating the module yet, I stopped for lunch and left the PR uploaded so as not to lose it. That's why I hadn't pinged you yet. I'm going to add the test and fix the pre-commit bugs.

sorry for the nagging! I was impatient...
**Note** : at the start of the PR, you can choose to set "draft PR", so it's explicit. Unfortunately, you can not put a non draft PR to draft state, once the PR is created.

> Yes, being such tiny extension to a module being hosted here, it would be weird to not be located in the same place.

:+1: 

> Regarding what you comment about 'bg_color', I agree with you, I don't see sense that this is still there, in my experience, I have never used it, nor do I find that it brings a great added value.

I think you can so remove it. Less code to maintain.

hugo-cordoba:
@legalsylvain you are correct, I did not put the pr in draft. I have uploaded an update, it should be all correct. I am waiting for your feedback :)

Best regards!

legalsylvain (changes requested):
Just tested on runboat. some things are missing. (or maybe I missed something...)

**Configuration :** 

image

**Result**

image

**Problem**
- The column "Price" should be hidden
- The column "Cost" (standard_price) should be visible
- The margin is not computed
- The margin % is not computed

Thanks !

hugo-cordoba:
Hi @legalsylvain, I haven't forgotten about this, but this week it has been impossible for me to get busy. I'll check it tomorrow!

legalsylvain:
no problem ! take your time. Just ping me when you want a review.

Thanks !

hugo-cordoba:
Hi @legalsylvain, how are you?

I have made the changes you proposed, when you have time, can you check that everything works fine?

Regards!

aliciagaarzo (approved):
Code review LGTM!

product_pricelist_simulation_margin/wizards/wizard_preview_pricelist.py
```diff
@@ -0,0 +1,91 @@
...
+        cost = self.env.user.company_id.currency_id._convert(
+            self.standard_price,
+            to_currency,
+            self.env.user.company_id,
+            self.price_date,
+            round=False,
+        )
+        rslt.update(
+            {
+                "pricelist_id": pricelist.id,
```
legalsylvain: no need to return that value, as the super function return the value. See : 
https://github.com/OCA/product-attribute/blob/16.0/product_pricelist_simulation/wizards/wizard_preview_pricelist.py#L59-L67

don't you think ?

product_pricelist_simulation_margin/wizards/wizard_preview_pricelist.py
```diff
@@ -0,0 +1,91 @@
...
+            }
+        )
+        return rslt
+
+
+class PricelistSimulationLine(models.TransientModel):
+    _inherit = "wizard.preview.pricelist.line"
+    _description = "wizard - Preview Pricelist Line"
+
+    price_vat_excl = fields.Float(string="Unit Sales Price (Excl.)", readonly=True)
```
legalsylvain:
```suggestion
    price_vat_excl = fields.Monetary(
        string="Unit Sales Price (Excl.)",
        readonly=True,
        currency_field="currency_id",
        digits="Price",
    )
```

product_pricelist_simulation_margin/wizards/wizard_preview_pricelist.py
```diff
@@ -0,0 +1,91 @@
...
+    price_vat_incl = fields.Float(string="Unit Sales Price (Incl.)", readonly=True)
```
legalsylvain:
```suggestion
    price_vat_incl = fields.Monetary(
        string="Unit Sales Price (Incl.)",
        readonly=True,
        currency_field="currency_id",
        digits="Price",
    )
```

product_pricelist_simulation_margin/wizards/wizard_preview_pricelist.py
```diff
@@ -0,0 +1,91 @@
...
+
+    margin = fields.Float(
```
legalsylvain:
```suggestion
    margin = fields.Monetary(
```

product_pricelist_simulation_margin/wizards/wizard_preview_pricelist.py
```diff
@@ -0,0 +1,91 @@
...
+        store=True,
+        digits=dp.get_precision("Product Price"),
```
legalsylvain:
```suggestion
        digits="Price",
        currency_field="currency_id",
```

product_pricelist_simulation_margin/wizards/wizard_preview_pricelist.py
```diff
@@ -0,0 +1,91 @@
...
+        readonly=True,
+    )
+
+    margin_percent = fields.Float(
+        string="Margin (%)",
+        store=True,
+        digits=dp.get_precision("Product Price"),
```
legalsylvain:
```suggestion
        digits="Price",
```

product_pricelist_simulation_margin/wizards/wizard_preview_pricelist.py
```diff
@@ -0,0 +1,91 @@
+# Copyright 2023 FactorLibre - Hugo Córdoba
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+
+
+from odoo import fields, models
+
+from odoo.addons import decimal_precision as dp
```
legalsylvain:
```suggestion
```

legalsylvain (changes requested):
functional review. LGTM ! Thanks ! 

Some minor propositions inline.

regards.

hugo-cordoba:
Hi @legalsylvain! Thanks for the review. I have just made the changes.

I have a question, what is the difference between creating a float field and adding the monetary widget or making the field 'monetary' directly?

legalsylvain:
> I have a question, what is the difference between creating a float field and adding the monetary widget or making the field 'monetary' directly?

Float + digits will set allways the same number of digits. Not work in multi currency context. 
Monetary + currency will set the number of digits defined in the current currency. For money fields, monetary field is better.

legalsylvain (approved):
Thanks for your work !

hugo-cordoba:
Understood, thanks for the explanation @legalsylvain.

@pedrobaeza can you check this pr? Thank you very much!!

PR #1362, opened 2023-06-16
[14.0][IMP] product_pricelist_assortment: sales administrator access

ilyasProgrammer:
Let Sales administrator modify assortment items. In order to get it had to include sales_team module into dependency.
Before you had to have system admin to modify it, which is too much.

### DIFF
--- a/product_pricelist_assortment/security/product_pricelist_assortment_item.xml
+++ b/product_pricelist_assortment/security/product_pricelist_assortment_item.xml
@@ -14,7 +14,7 @@
     <record model="ir.model.access" id="product_pricelist_assortment_item_manager">
         <field name="name">product.pricelist.assortment pricelist manager</field>
         <field name="model_id" ref="model_product_pricelist_assortment_item" />
-        <field name="group_id" ref="base.group_system" />
+        <field name="group_id" ref="sales_team.group_sale_manager" />
         <field name="perm_read" eval="1" />
         <field name="perm_create" eval="1" />
         <field name="perm_write" eval="1" />
--- a/product_pricelist_assortment/__manifest__.py
+++ b/product_pricelist_assortment/__manifest__.py
@@ -9,7 +9,7 @@
     "license": "AGPL-3",
     "author": "ACSONE SA/NV, Odoo Community Association (OCA)",
     "website": "https://github.com/OCA/product-attribute",
-    "depends": ["product_assortment"],
+    "depends": ["product_assortment", "sales_team"],
     "data": [
         "security/product_pricelist_assortment_item.xml",
         "data/ir_cron.xml",


francesco-ooops (approved):
Functional ok!

francesco-ooops:
@rousseldenis @acsonefho @Cedric-Pigeon good for you?

product_pricelist_assortment/security/product_pricelist_assortment_item.xml
```diff
@@ -14,7 +14,7 @@
     <record model="ir.model.access" id="product_pricelist_assortment_item_manager">
         <field name="name">product.pricelist.assortment pricelist manager</field>
         <field name="model_id" ref="model_product_pricelist_assortment_item" />
-        <field name="group_id" ref="base.group_system" />
+        <field name="group_id" ref="sales_team.group_sale_manager" />
```
acsonefho, line 17: I think adding `sales_team` just to fill a group is not a good idea.

Maybe create a dedicated group "Assortment manager" and then (into another module) link the Sales manager with Assortment manager.

acsonefho (changes requested):

product_pricelist_assortment/security/product_pricelist_assortment_item.xml
ilyasProgrammer: Maybe. I'm not sure what would be more correct.

acsonefho: I'm not in favor of adding `sales_team` if it's not necessary about functionnal behavior

francesco-ooops: @acsonefho sorry this is not really clear to me - to use this module user needs to be able to edit a pricelist (Sales: administrator), so what would be the correct group to use the module?

francesco-ooops: @acsonefho which group do you think is necessary for functional behavior?

acsonefho: Oh sorry I miss read the module name. I thought that was `product_assortment`. It's ok for me :pray:

acsonefho (approved):

francesco-ooops:
@pedrobaeza good for a quick merge?

pedrobaeza:
This change is a bit controversial, but if it's OK for one of the original contributors:

/ocabot merge major

francesco-ooops:
> This change is a bit controversial, but if it's OK for one of the original contributors:

Can you expand on why? This module adds a feature to pricelist, and users who are allowed to manage pricelists must be in group "sales: administrator" (if I'm not mistaken). 

Also consider there is no restriction (to my knowledge) to assortment management, at least in v14

I don't see the reason for restricting pricelist assortment management to "administration: setting" group, but if there's a better solution that could have been implemented I'm all ears :)

pedrobaeza:
- It's changing the dependencies on a stable module.
- It's changing who can access to the model.

francesco-ooops:
I understand. On the other hand, I think this was an oversight in development stage rather than a security implementation, but I see what you mean, thanks for the clarification!

PR #1364, opened 2023-06-21
[16.0] product_catalog: add module

hparfr:
A Product catalog is a group of products

it's based on assortments to create a list of candidates, then products should be elected to be part of the catalog.

- [ ] add buttons in view to _add_all_candidates and _remove_candidates
- [ ] improve view: change widget, add smart buttons...


legalsylvain:
Hi. could you explain the use case of this module .

lmignon:
@hparfr could you take a look at https://github.com/OCA/product-attribute/pull/1352?

hparfr:
#1352 can be closed. It was a previous attempt.

PR #1375, opened 2023-07-05
[15.0][FIX] product_pricelist_direct_print: Migrate changes of mail template

CarlosRoca13:
cc @Tecnativa TT44172

ping @pedrobaeza @victoralmau


pedrobaeza (approved):
This is partly `[OU-ADD]`, but as it contains also a fix, let's keep it for not rebuilding.

PR #1377, opened 2023-07-05
[FIX] product_pricelist_supplierinfo: inherit the _compute_price method in product_pricelist model

andreagidaltig:
In version 15.0 the _compute_price_rule method was used to obtain the product price because the price was a computed field, however, in version 16.0 we need to inherit _compute_price from product_pricelist_item because this method is used directly when creating a new sale order line.


product_pricelist_supplierinfo/models/product_pricelist.py
```diff
@@ -5,32 +5,6 @@
...
-                        rule,
-                        date=date or context.get("date", fields.Date.today()),
-                        quantity=qty,
-                    ),
-                    rule.id,
-                )
-        return result
-
-
 class ProductPricelistItem(models.Model):
```
luisg123v: Since now the only model included here is `product.pricelist.item`, we should rename the file accordingly

luisg123v (changes requested):

luisg123v (approved):

PR #1381, opened 2023-07-10
[12.0][ADD] product_supplierinfo_uom_rule: initial commit

wtaferner:
Adding an easy way to feed the supplierinfo by a different UoM of the same category.


wtaferner:
@legalsylvain 
You have motivated me to finally suggest directly my first module to the OCA (took me quite a long time to finally suggest, but yeah, hope it was worth the effort). Here  you have the PR for the module we were talking about.

Let the idea evolve and help one or the other 😉

legalsylvain:
Hi @wtaferner. Thanks a lot for sharing this module !


I installed it locally.
the design looks great. However, the new fields are not "used" as I would expect. 

**Settings** 
My supplier sells me a product for €1.33 a dozen. 
image
So far, it's OK regarding the setting.

However, When I try to purchase this product : 

image

**Problem 1** : the Unit of Measure is not the expected one. This is Unit by default, and not Dozen.

If I change the Unit to set Dozen : 

image

**Problem 2** : the Unit price is not the good one. it is 12*0.12 = 1.44. (however the good unit price is 1.33)


**Side problem not blocking** The module is not working with other OCA module. (purchase_discount, purchase_triple_discount). (not a big deal, we can create glue modules, or make lazy dependencies).

image

In the image, the expected unit price should be 1.33 / 12 * 0.5. (0,055416667€)

wtaferner:
@legalsylvain 
**Problem 1**:  Unrelated to what this module does as it depends on the unit given on the product template. I have other modules which deal with several convenience features (like the UoM is taken from the supplierinfo and therefore the proper price is taken from the start, but this was too big for this module).
It is also a rounding issue as you could switch manually...

**Problem 2**: The price is rounded on the 'Product Price' which you need to take care of or we have to override the price precision in this module and of course we also have a inter-dependency issue which for sure need to be handled in a glue module.

legalsylvain:
> I have other modules which deal with several convenience features (like the UoM is taken from the supplierinfo and therefore the proper price is taken from the start, but this was too big for this module).

Is it possible to share that module ? i'm interested to "bypass the bad core design of the  "uom_po_id" on the product.template, in fact. (see : https://github.com/odoo/odoo/pull/70462)

wtaferner:
@legalsylvain 
It is basically a hack into the very awesome inheritable core module and I removed the code from this module which is relying on the hack, so, not useful and ready to share here, but maybe I can do a PR against Odoo core at some point again to get the inheritance relief merged.

legalsylvain (approved):

legalsylvain:
>  It is basically a hack into the very awesome inheritable core module and I removed the code from this module which is relying on the hack, so, not useful and ready to share here, but maybe I can do a PR against Odoo core at some point again to get the inheritance relief merged.

That could be great. My PR was not accepted (https://github.com/odoo/odoo/pull/70462) but maybe your will be simplier.

wtaferner:
@legalsylvain 
Here I proposed two commits which go together somehow for my/our needs which we have forked of...not sure if it will be understood and merged, but feel free to support the effort even in stable.

https://github.com/odoo/odoo/pull/128077

PR #1385, opened 2023-07-13
[IMP] product_price_history : add new graph view

legalsylvain:
- Add the possibility to see a graph with the evolution of the cost of the product over the time.

product_price_history_graph

- add me as comaintainers

- add french translation


product_price_history/models/product_template.py
```diff
@@ -4,6 +4,8 @@
 class ProductTemplate(models.Model):
     _inherit = 'product.template'
 
+    cost = fields.Float(group_operator="avg")
```
ghost: Is it a good idea to have the value of cost propagate to the product variant if there is only one variant ?

rousseldenis (approved):

product_price_history/models/product_template.py
legalsylvain: Indeed. I have to take a look.

legalsylvain: Hi @gbrito-erpopen. I updated the code. I made an error, I wanted to set the group_operator on the ``cost`` field of the ``product.price.history`` model.
Now fixed.

ghost (approved):

product_price_history/models/product_template.py
ghost: @legalsylvain looks good, makes more sense :)

legalsylvain:
@rousseldenis could you call ocabot to merge this one ? 

thanks !

legalsylvain:
thanks !

PR #1386, opened 2023-07-14
[IMP] product_packaging_dimension: Don't use max_weight anymore

rousseldenis:
As max_weight is defined now on stock.package.type, rely on that one.

Migrate data that was filled in to weight field.

Added string label to packaging_length too.

### DIFF
--- a/product_packaging_dimension/models/product_packaging.py
+++ b/product_packaging_dimension/models/product_packaging.py
@@ -21,7 +21,7 @@ class ProductPackaging(models.Model):
     ]
     height = fields.Integer()
     width = fields.Integer()
-    packaging_length = fields.Integer()
+    packaging_length = fields.Integer(string="Length")
 
     length_uom_id = fields.Many2one(
         "uom.uom",
--- a/product_packaging_dimension/models/product_packaging.py
+++ b/product_packaging_dimension/models/product_packaging.py
@@ -14,9 +14,9 @@ class ProductPackaging(models.Model):
         ("positive_width", "CHECK(width>=0)", "Width must be positive"),
         ("positive_length", "CHECK(packaging_length>=0)", "Length must be positive"),
         (
-            "positive_max_weight",
-            "CHECK(max_weight>=0)",
-            "Max Weight must be positive",
+            "positive_weight",
+            "CHECK(weight>=0)",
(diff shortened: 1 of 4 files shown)


rousseldenis:
@marielejeune @jbaudoux

product_packaging_dimension/models/product_packaging.py
```diff
@@ -13,11 +13,6 @@ class ProductPackaging(models.Model):
         ("positive_height", "CHECK(height>=0)", "Height must be positive"),
         ("positive_width", "CHECK(width>=0)", "Width must be positive"),
         ("positive_length", "CHECK(packaging_length>=0)", "Length must be positive"),
-        (
-            "positive_max_weight",
-            "CHECK(max_weight>=0)",
-            "Max Weight must be positive",
```
jbaudoux, line 19: You could keep it but on weight instead of max_weight

rousseldenis: Done

jbaudoux (approved):

product_packaging_dimension/migrations/16.0.1.1.0/pre-migrate.py
```diff
@@ -0,0 +1,21 @@
...
+
+def _move_max_weight(env):
+    """
+    Move max_weight value to weight if not defined
+    """
+    query = """
+        UPDATE product_packaging
+            SET weight = max_weight
+            WHERE weight IS NULL OR weight = 0.0
+            AND max_weight IS NOT NULL
```
jbaudoux: Maybe only for the product packaging linked to a product.
```suggestion
            AND max_weight IS NOT NULL
            AND product_id IS NOT NULL
```

rousseldenis: Done

rousseldenis:
@marielejeune

jbaudoux:
> As max_weight is defined now on stock.package.type, rely on that one.

Don't rely 😉

This PR relates to my comment here
https://github.com/OCA/product-attribute/pull/1161#issuecomment-1490120668

Cc @sebalix

lmignon (approved):

PR #1389, opened 2023-07-24
[MIG][16.0] product_origin

acsonefho:
From https://github.com/OCA/product-attribute/pull/424


legalsylvain (changes requested):
Hi @acsonefho  Thanks for porting this module.

Code review : LGTM. 

could you add some extra constraint : 
- a domain on state_id : to avoid to set a state_id inconsistent with the country_id ? 
(ex : https://github.com/grap/grap-odoo-business/blob/12.0/product_origin/models/product_product.py#L45)
- an onchange that reset state_id if the country_id has changed.
(ex : https://github.com/grap/grap-odoo-business/blob/12.0/product_origin/models/product_product.py#L66)

also, it should be great if the field was available stored at product.product level to allow search / filter / etc...

Thanks !

acsonefho:
Hi @legalsylvain ! Thank you for you review! :)
As it's a migration, I prefer to avoid adding something into the module.
But your changes are legit to me. Maybe doing a new PR with that changes later?

legalsylvain:
>  But your changes are legit to me. Maybe doing a new PR with that changes later?

Can I do a PR against your branch ? 

Other question : what about to set the fields on the product.product model. (as barcode, etc... and other fields ) ?

acsonefho:
> 
> Can I do a PR against your branch ?

Of course!
> 
> Other question : what about to set the fields on the product.product model. (as barcode, etc... and other fields ) ?

If the behavior is exactly like the barcode etc, it's ok for me.

legalsylvain:
> If the behavior is exactly like the barcode etc, it's ok for me.

For fields like barcodes, the field is available on product.template in a mono variant context, and available on product.product form view in a multi variant context.

Is it what you mean ?

acsonefho:
> > If the behavior is exactly like the barcode etc, it's ok for me.
> 
> For fields like barcodes, the field is available on product.template in a mono variant context, and available on product.product form view in a multi variant context.
> 
> Is it what you mean ?

Yes exactly :)

legalsylvain:
Hi @acsonefho. done here : https://github.com/acsone/product-attribute/pull/7
could you take a look ? Thanks ! 

CC : @rousseldenis

legalsylvain (approved):

legalsylvain (approved):

legalsylvain (changes requested):
some improvment done here : https://github.com/acsone/product-attribute/pull/8

legalsylvain (approved):

legalsylvain:
@acsonefho : could you reopen it ? 
Thanks !

PR #1393, opened 2023-07-27
16.0 mig base product mass addition

aiendry-aktivsoftware:
Standard migration to version 16.0


base_product_mass_addition/models/product_product.py
```diff
@@ -38,41 +39,31 @@ def _inverse_set_process_qty(self):
...
-        return super().modified(fnames, create=create, before=before)
+        if self and any(quick_fname in vals.keys() for quick_fname in quick_fnames):
+            cr = self.env.cr
+            Model = self._build_model(self.pool, cr)
+            Model._log_access = set(list([]))
+        elif not self._log_access:
+            cr = self.env.cr
+            Model = self._build_model(self.pool, cr)
+            Model._log_access = set(LOG_ACCESS_COLUMNS)
+        return super(ProductProduct, self).write(vals)
```
legalsylvain: we should revert the changes done in Model after the write, don't you think ?

bealdav: @legalsylvain do you consider changes made by @aiendry-aktivsoftware solve your case ? It seems it's on the way !?

Thanks both for your contributions.

base_product_mass_addition/readme/ROADMAP.rst
```diff
@@ -0,0 +1,2 @@
+Adding new implementations would be great:
+on sale.order or on stock.picking.batch for instance.
```
bealdav, line 2: https://github.com/OCA/sale-workflow/tree/14.0/sale_quick exist for now

and also

https://github.com/acsone/stock-logistics-workflow/tree/14.0/stock_picking_quick

Then ROADMAP.rst may be deleted I suppose

base_product_mass_addition/models/product_product.py
```diff
@@ -0,0 +1,105 @@
...
+        #
+        # Basically, if all we're modifying are quick magic fields, and we don't have
+        # any other column to flush besides the LOG_ACCESS_COLUMNS, clear it.
+        quick_fnames = ("qty_to_process", "quick_uom_id")
+        if self and any(quick_fname in vals.keys() for quick_fname in quick_fnames):
+            cr = self.env.cr
+            Model = self._build_model(self.pool, cr)
+            Model._auto = False
+            Model._log_access = set(list([]))
+        res = super(ProductProduct, self).write(vals)
```
legalsylvain, line 63:
```suggestion
        res = super().write(vals)
```

base_product_mass_addition/models/product_product.py
```diff
@@ -0,0 +1,105 @@
...
+        # https://github.com/odoo/odoo/blob/dab802a939c97603f70c504e98ecc92b0ac552f9/odoo/models.py#L3677C1-L3679C61 # noqa: B950
+        #
+        # Basically, if all we're modifying are quick magic fields, and we don't have
+        # any other column to flush besides the LOG_ACCESS_COLUMNS, clear it.
+        quick_fnames = ("qty_to_process", "quick_uom_id")
+        if self and any(quick_fname in vals.keys() for quick_fname in quick_fnames):
+            cr = self.env.cr
+            Model = self._build_model(self.pool, cr)
+            Model._auto = False
+            Model._log_access = set(list([]))
```
legalsylvain, line 62:
```suggestion
            Model._log_access = set()
```

is the same, no ?

base_product_mass_addition/readme/ROADMAP.rst
legalsylvain: (and moved into the description part).

base_product_mass_addition/models/product_product.py
```diff
@@ -26,6 +26,7 @@ class ProductProduct(models.Model):
         domain="[('category_id', '=', quick_uom_category_id)]",
         compute="_compute_quick_uom_id",
         inverse="_inverse_set_process_qty",
+        store=True,
```
legalsylvain, line 29: **issue**: you try to store a field that should be computed on the fly, and not be stored. I guess something is wrong with that design. could you explain that change ? Thanks !

legalsylvain (commented):
Thanks for porting this module ! 

**nipick**: Trivial proposal inlines.

**question** : you adapted the code for V16, removing "modified" orverload by "write" overload. I fear that rebuilding two times the model (Model = self._build_model(self.pool, cr)) can be slow. Did you faced some performance issue.

bealdav:
Thanks a lot @legalsylvain for this review.
On my side I need to port sale_quick to 16.0 then it could be a great help if we could go ahead in this current process.
Thanks both

base_product_mass_addition/models/product_product.py
legalsylvain: well, on that topic, I'm not an expert of the core framework of odoo. (never played modifying technical field like _log_access and rebuild model). Not sure if there are side effect, or if it is safe.
@pedrobaeza : a point of view, here ?

pedrobaeza: This will have side effects sooner than later for sure, so it should be forbidden.

legalsylvain: thanks a lot for your analysis.

base_product_mass_addition/models/product_product.py
FranzPoize: This is the issue without ```store=True```
```
tests/test_product_mass_addition.py::TestProductMassAddition::test_quick_line_add
  /odoo/src/odoo/fields.py:808: UserWarning: Field product.product.quick_uom_id in dependency of product.product.quick_uom_category_id should be searchable. This is necessary to determine which records to recompute when uom.uom.category_id is modified. You should either make the field searchable, or simplify the field dependency.
```

It seems there is some kind of dependency race on this.

However I'm not sure there is a need to have the ```quick_uom_category_id``` field since it's not use anywhere in the module and it could be accessed via ```quick_uom_id.category_id```

FranzPoize:
The test need ```odoo_test_helper==2.1.1```.

And also it seems it's not possible to directly write into metadata anymore, i.e ```cls.product.write_uid = user_demo``` does not change the ```write_uid``` of the product

FranzPoize:
> The test need `odoo_test_helper==2.1.1`.
> 
> And also it seems it's not possible to directly write into metadata anymore, i.e `cls.product.write_uid = user_demo` does not change the `write_uid` of the product

It's confirmed that we can't change ```write_id``` like this anymore. This leads to this code:
```
        if self._log_access:
            # the superuser can set log_access fields while loading registry
            if not(self.env.uid == SUPERUSER_ID and not self.pool.ready):
                bad_names.update(LOG_ACCESS_COLUMNS)

        # set magic fields
        vals = {key: val for key, val in vals.items() if key not in bad_names}
```
which remove all ```LOG_ACCESS_COLUMNS``` from the written values

FranzPoize:
To fix the rest of the text instead of relying on ```write_uid``` we could store the ```write_date``` and checked if it changed or not

base_product_mass_addition/models/product_product.py
FranzPoize: By the way the ```store=True``` breaks functionnality on downstream module it sometimes lead to a None valued uom_id on the product

HaraldPanten:
Hi @aiendry-aktivsoftware Will you continue this PR?

THX!

legalsylvain:
hi @aiendry-aktivsoftware. Friendly reminder. could you fix the CI so we can move forward ? 

thanks !

legalsylvain:
Done here : https://github.com/OCA/product-attribute/pull/1512

PR #1369, opened 2023-06-26
[FIX] product_pricelist_assortment: fix item form

ilyasProgrammer:
Before:
grafik

After:
grafik

For some reason inherited fields are not translated. Had to inherit them explicitly.

PR #1397, opened 2023-08-14
[14.0][IMP] product_pricelist_assortment: rewrite base field in module, to translate it

solo4games:
Rewrite base field in this module, to be able to translate this field

### DIFF
--- a/product_pricelist_assortment/models/product_pricelist_assortment_item.py
+++ b/product_pricelist_assortment/models/product_pricelist_assortment_item.py
@@ -14,6 +14,17 @@ class ProductPricelistAssortmentItem(models.Model):
     _description = "Product Pricelist Assortment Item"
     _inherit = "product.pricelist.item"
 
+    compute_price = fields.Selection(
+        [
+            ("fixed", "Fixed Price"),
+            ("percentage", "Percentage (discount)"),
+            ("formula", "Formula"),
+        ],
+        index=True,
+        default="fixed",
+        required=True,
+    )
+
     assortment_filter_id = fields.Many2one(
         comodel_name="ir.filters",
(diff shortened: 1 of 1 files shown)


francesco-ooops:
@solo4games check [this](https://github.com/OCA/odoo-community.org/blob/master/website/Contribution/CONTRIBUTING.rst#commit-message)

gdgellatly:
I really don't get it. Why would we do this? Surely this is not right, it just feels like it will break any other modules using compute price or custom formulas and seems so unneeded. What am I missing?

francesco-ooops:
@gdgellatly I don't know if this is the best technical solution, on the functional side the issue is that some fields in pricelist form cannot be translated in other languages, see https://github.com/OCA/product-attribute/pull/1369#issuecomment-1621428838

Do you have any suggestions on how to approach this?

gdgellatly:
> @gdgellatly I don't know if this is the best technical solution, on the functional side the issue is that some fields in pricelist form cannot be translated in other languages, see [#1369 (comment)](https://github.com/OCA/product-attribute/pull/1369#issuecomment-1621428838)
> 
> Do you have any suggestions on how to approach this?

Hi, I have no idea, I only care about English and know nothing of translation. But from a code POV, to me this looks hugely unnecessary and pollutive. I would suggest however, if that in applying a perfectly normal and elegant inheritance solution (which IMHO this module does), that this somehow limits translation functionality, then the solution is not within the module itself, but in Odoo core. But that is just my opinion/suspicion. @pedrobaeza may have some thoughts.

pedrobaeza:
Yes, if this is happening, you should notify/repair it at Odoo core ORM level.

PR #1398, opened 2023-08-14
[16.0][FIX] product_pricelist_revision: use compute_sudo

gdgellatly:
The name field is changed to be stored, however the compute function is multi with an unstored field. Therefore to keep consistent behaviour we add compute_sudo to name field.

### DIFF
--- a/product_pricelist_revision/models/pricelist.py
+++ b/product_pricelist_revision/models/pricelist.py
@@ -8,7 +8,7 @@ class ProductPricelistItem(models.Model):
     _inherit = "product.pricelist.item"
     _rec_names_search = ["name"]  # Initialize here to be able to extend
 
-    name = fields.Char(store=True)
+    name = fields.Char(store=True, compute_sudo=True)
 
     previous_item_id = fields.Many2one(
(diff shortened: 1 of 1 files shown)


product_pricelist_revision/models/pricelist.py
```diff
@@ -8,7 +8,7 @@ class ProductPricelistItem(models.Model):
     _inherit = "product.pricelist.item"
     _rec_names_search = ["name"]  # Initialize here to be able to extend
 
-    name = fields.Char(store=True)
+    name = fields.Char(store=True, compute_sudo=True)
```
rousseldenis: @gdgellatly Do you have access errors ? Because it seems name computation does not involve objects that could have different access rights than pricelist.item itself

gdgellatly: @rousseldenis No this is solely because a multi now combines a stored and a non stored field. Oh, looks like I got the wrong field. I picked the one to changed to stored, I need to declare the other one as compute_sudo.

rvalyi (approved):
LGTM, not tested

PR #1399, opened 2023-08-16
[16.0][MIG][WIP] packaging uom

gaelTorrecillas:


rousseldenis:
@gaelTorrecillas Could you fix tests ?

gaelTorrecillas:
hi,
i'm working on it.

gaelTorrecillas:
@rousseldenis during tests, I reproduce this error.

When we create a product.packaging by python code, with only product and uom, the qty compute isn't called. qty = 1.

In fact, the customer uses the interface, so he doesn't have this error.

We are working on it to come up with code that will work with this operation.

edit : In the odoo core, qty has a default, the compute function isn't call

benwillig:
@gaelTorrecillas can you rebase to refresh tests pipelines please?

PR #1402, opened 2023-08-23
[12.0][MIG] product_attribute_archive

newtratip:


rousseldenis:
@newtratip Could you fix this as it seems.. you haven't migrated it?

PR #1404, opened 2023-08-25
[16.0][MIG] product_total_weight_from_packaging: Migrate to 16.0

QuocDuong1306:


santostelmo (approved):

cyrilmanuel (approved):

rafaelbn:
@moduon testit @fcvalgar OCA contribution

rafaelbn:
@grindtildeath do you want to review this module as contributor? 💚

product_total_weight_from_packaging/models/product.py
```diff
@@ -0,0 +1,37 @@
...
+        self, packaging_tuple, qty_per_pkg
+    ):
+        res = {
+            "qty": qty_per_pkg,
+        }
+        if packaging_tuple.is_unit:
+            res["weight"] = self.weight
+        else:
+            packaging = self.env["product.packaging"].browse(packaging_tuple.id)
+            res["weight"] = packaging.max_weight
```
jbaudoux: The right field is weight in v16. Relates to https://github.com/OCA/product-attribute/pull/1386

```suggestion
            res["weight"] = packaging.weight
```


Also you need to convert the weight in the uom of the product as you can now define a different weight uom on the packaging. So if the product weight uom is `g` as defined on the product and the product packaging weight is `kg` as defined on the product packaging, you need to convert the weight in `kg` to a weight in `g`.

Please fix those 2 things in 2 separate commits

jbaudoux (changes requested):

product_total_weight_from_packaging/models/product.py
QuocDuong1306: thank @jbaudoux , I updated

jbaudoux (approved):
Thanks for the fixes. LGTM

PR #1405, opened 2023-08-27
[14.0][ADD] product_variant_company

Volodiay622:
on this [comment](https://github.com/OCA/stock-logistics-barcode/pull/526#discussion_r1275978467) from @[rousseldenis](https://github.com/rousseldenis)


product_company_id/readme/CONTRIBUTORS.rst
```diff
@@ -0,0 +1,3 @@
+Sylvain LE GAL <https://twitter.com/legalsylvain>
```
legalsylvain: This module comes from a module I développed ?

legalsylvain (commented):
Do we have to add a new rule to target the new company field ?

product_company_id/readme/CONTRIBUTORS.rst
Volodiay622: it's from [this PR](https://github.com/OCA/stock-logistics-barcode/pull/526)

francesco-ooops:
ping @simahawk @rousseldenis in relation to [#526](https://github.com/OCA/stock-logistics-barcode/pull/526)

thanks!

rousseldenis:
> Do we have to add a new rule to target the new company field ?

@legalsylvain What do you mean ?

rousseldenis:
> > Do we have to add a new rule to target the new company field ?
> 
> @legalsylvain What do you mean ?

Oh, ir.rule. Yes, indeed. @Volodiay622 Could you do it ?

legalsylvain:
> @legalsylvain What do you mean ?
> Oh, ir.rule. Yes, indeed. @Volodiay622 Could you do it ?

Well, in fact, I don't know how odoo works ! I _guess_ that an ir.rule is applied, when searching on ``product.procuct`` but I don't see ir.rule related to product.product model. There is one, related to product.template. 
https://github.com/odoo/odoo/blob/14.0/addons/product/security/product_security.xml#L34

I so _guess_ that this rule is applied, but making a join that could cost time. So I _gess_ that we should add a new rule on product.product, based on the new company field. But in fact, I'm not sure it will improve something. Maybe two WHEN will be added in the SQL request...
What do you think ?

product_company_id/models/product_product.py
```diff
@@ -0,0 +1,12 @@
...
+
+from odoo import fields, models
+
+
+class ProductProduct(models.Model):
+    _inherit = "product.product"
+
+    # Store related field
+    company_id = fields.Many2one(
+        related="product_tmpl_id.company_id", store=True, readonly=False
```
legalsylvain:
```suggestion
        related="product_tmpl_id.company_id", store=True, readonly=False, index=True
```

Volodiay622: @legalsylvain I added your suggested change

francesco-ooops:
@legalsylvain @rousseldenis are we good to go?

product_company_id/__manifest__.py
```diff
@@ -0,0 +1,18 @@
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/lgpl).
+{
+    "name": "Product Company ID",
```
simahawk: why not product_company only? I mean, why not removing "id" everywhere?
Not a blocker!

product_company_id/tests/__init__.py
```diff
@@ -0,0 +1,3 @@
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/lgpl).
+
+from . import test_module
```
simahawk:
```suggestion
from . import test_product_company
```

product_company_id/tests/test_module.py
```diff
@@ -0,0 +1,33 @@
...
+from odoo.tests.common import TransactionCase
+
+
+class TestModule(TransactionCase):
+    def setUp(self):
+        super().setUp()
+
+        self.ProductProduct = self.env["product.product"]
+        self.company_1 = self.env["res.company"].create({"name": "Company 1"})
+        self.company_2 = self.env["res.company"].create({"name": "Company 2"})
```
simahawk:
```suggestion
    @classmethod
    def setUpClass(cls):
        super().setUpClass()
        cls.env = cls.env(context=dict(cls.env.context, tracking_disable=True))
        cls.ProductProduct = cls.env["product.product"]
        cls.company_1 = cls.env["res.company"].create({"name": "Company 1"})
        cls.company_2 = cls.env["res.company"].create({"name": "Company 2"})
```

product_company_id/__manifest__.py
Volodiay622: applied

product_company_id/tests/__init__.py
Volodiay622: applied

product_company_id/tests/test_module.py
Volodiay622: applied

francesco-ooops:
@rousseldenis @simahawk could you also provide a feedback on module naming and description?

francesco-ooops:
@simahawk ping

simahawk:
LGTM. Can we squash everything onto "[ADD] product_company"?

simahawk:
> LGTM. Can we squash everything onto "[ADD] product_company"?

Thinking of it twice... Sorry... :grin: Maybe `product_variant_company` makes more sense because by default the product has already a company via its template. No?

simahawk:
@Volodiay622 can you squash? 
@rousseldenis @legalsylvain fine for you?

francesco-ooops:
@simahawk from description I still don't understand what this module does. 

Is it a technical or a functional module? 

For which use case would this be employed?

simahawk:
> @simahawk from description I still don't understand what this module does.
> 
> Is it a technical or a functional module?
> 
> For which use case would this be employed?

It's a technical module. You need a company at variant level? Use this.
The description should be adapted if is not clear.

francesco-ooops:
@simahawk ok, but functionally I really don't understand it. 

Here's a video of a new product template with two variants: when setting a company in a variant, it is changed for all variants and for product template as well

https://recordit.co/ljowC1XRUx

simahawk:
> @simahawk ok, but functionally I really don't understand it.
> 
> Here's a video of a new product template with two variants: when setting a company in a variant, it is changed for all variants and for product template as well
> 
> https://recordit.co/ljowC1XRUx

In fact, as it is it does not add anything if not a field that could be overridden.
IMO it could be already using a computed / writeable field instead of a related one and default to the tmpl one...
@rousseldenis as you suggested this split, what did you have in mind regarding this?

legalsylvain:
> Here's a video of a new product template with two variants: when setting a company in a variant, it is changed for all variants and for product template as well

IMO, it is the expected behaviour. 

As said in this module, it is a just technical module to bring company_id on product_product model. Nothing else. Don't you think ?

francesco-ooops:
> > Here's a video of a new product template with two variants: when setting a company in a variant, it is changed for all variants and for product template as well
> 
> IMO, it is the expected behaviour.
> 
> As said in this module, it is a just technical module to bring company_id on product_product model. Nothing else. Don't you think ?

On a second thought, yes, it's standard behavior indeed; 

it's just that from a ux standpoint, odoo makes it confusing when accessing a variant from the "product variant" menu to understand which fields are at product template level and which fields are at variant level.

francesco-ooops:
@legalsylvain @rousseldenis good to go?

product_variant_company/__manifest__.py
```diff
@@ -0,0 +1,18 @@
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/lgpl).
```
yostashiro: Nitpicking. Same for other .py files.
```suggestion
# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
```

Volodiay622: resolved

francesco-ooops:
@legalsylvain @rousseldenis ping :)

legalsylvain (approved):

product_variant_company/__manifest__.py
francesco-ooops: @yostashiro good for you?

product_variant_company/readme/DESCRIPTION.rst
```diff
@@ -0,0 +1,3 @@
+This module extend Odoo functionalities of the "product.product" model
+by adding a related field that provides information about
+the associated company for each product.
```
simahawk: One last thing: can you please add here the explanation that is a technical module and does not change anything in the std behavior (see related comment)?

simahawk (approved):
Pre-approving

PR #1408, opened 2023-08-30
[16.0][ADD] partner_product_price

norlinhenrik:
- Open a contact.
- Click Actions - Products.
- In list view, show the fields **Contact** and **Contact's Price**.


partner_product_price/data/ir_actions_server.xml
```diff
@@ -0,0 +1,15 @@
+<?xml version="1.0" encoding="utf-8" ?>
+<odoo noupdate="1">
+    <record id="action_product_price_list_report" model="ir.actions.server">
+        <field name="name">Products</field>
+        <field name="groups_id" eval="[(4, ref('resource_booking.group_user'))]" />
```
legalsylvain, line 5: this module ``resource_booking`` is not set in the dependency field of the manifest.

partner_product_price/models/product_product.py
```diff
@@ -0,0 +1,30 @@
...
+                )
+
+    partner_id = fields.Many2one(
+        "res.partner",
+        string="Contact",
+        compute="_compute_partner_and_price",
+    )
+    partner_price = fields.Float(
+        "Contact's Price",
+        compute="_compute_partner_and_price",
```
legalsylvain, line 29: could you add precision ?

partner_product_price/data/ir_actions_server.xml
```diff
@@ -0,0 +1,15 @@
+<?xml version="1.0" encoding="utf-8" ?>
+<odoo noupdate="1">
+    <record id="action_product_price_list_report" model="ir.actions.server">
+        <field name="name">Products</field>
```
legalsylvain: should be a more explicit name, don't you think ? something like 'See product prices"

partner_product_price/data/ir_actions_server.xml
legalsylvain: this module resource_booking is not set in the dependency field of the manifest.

legalsylvain (changes requested):
thanks for your contribution !  
Some blocking point in line.
could you take a look on pre-commit error also ? 

thanks !

norlinhenrik:
@legalsylvain Thank you for the input! See the second commit with fixes plus more.

norlinhenrik:
Do you think that this module should be here in `product-attribute` or in `partner-contact`? I put it here because since other modules related to pricing are here.

partner_product_price/models/product_product.py
norlinhenrik: Done

partner_product_price/data/ir_actions_server.xml
norlinhenrik: Done.

partner_product_price/data/ir_actions_server.xml
norlinhenrik: The reference to resource_booking is removed.

PR #1412, opened 2023-09-06
[16.0][ADD] product_company_default

AungKoKoLin1997:
This module assigns default company during product creation.

@qrtl


rinaldifirdaus (approved):
Functional test: Work as expected!

legalsylvain:
Hi @AungKoKoLin1997 why product_dimension and product_company_default are incompatible ?

AungKoKoLin1997:
> why product_dimension and product_company_default are incompatible ?

@legalsylvain I think this https://github.com/OCA/product-attribute/commit/cbb362c97a103d49420884749a0df24c7df4ae8b is the reason for product_dimension. For product_company_default, the product_company_default sets a default company during product creation, leading to incompatibilities with other module tests. Notably, in product_pricelist_supplierinfo, tests fail because pricelists are created with a default company set to "false," while products have the current company assigned, causing discrepancies and errors.

product_company_default/models/product_template.py
```diff
@@ -0,0 +1,22 @@
...
+from odoo import api, fields, models
+
+
+class ProductTemplate(models.Model):
+    _inherit = "product.template"
+
+    company_id = fields.Many2one(default=lambda self: self._get_default_company())
+
+    @api.model
+    def _get_default_company(self):
```
yostashiro: I guess this is non-blocking but according to the convention: https://github.com/OCA/odoo-community.org/blob/master/website/Contribution/CONTRIBUTING.rst#models
```suggestion
    def _default_company_id(self):
```

product_company_default/readme/CONFIGURE.rst
```diff
@@ -0,0 +1,4 @@
+To configure this module, you need to:
```
yostashiro:
```suggestion
To enable the feature, you need to:
```

product_company_default/readme/DESCRIPTION.rst
```diff
@@ -0,0 +1 @@
+This module assigns default company to product.template.
```
yostashiro:
```suggestion
This module populates the company field with user's company as the default value when creating new products.
```

product_company_default/tests/test_product_company_default.py
```diff
@@ -0,0 +1,17 @@
...
+
+    def test_product_company(self):
+        # Enable the configuration using ir.config_parameter
+        self.env["ir.config_parameter"].sudo().set_param(
+            "product.default_company_enable", "1"
+        )
+        # Create a sample product
+        product = self.env["product.product"].create(
+            {"name": "Test Product", "default_code": "Test"}
+        )
```
yostashiro:
```suggestion
        product = self.env["product.product"].create({"name": "Test Product"})
```

product_company_default/data/config_parameter.xml
```diff
@@ -0,0 +1,8 @@
+<?xml version="1.0" encoding="utf-8" ?>
+<odoo noupdate="1">
+    <!-- System Parameter for Default Company Setting -->
+    <record id="param_product_default_company_enable" model="ir.config_parameter">
+        <field name="key">product.default_company_enable</field>
```
yostashiro: I think the key should start with the module name (as a convention).
```suggestion
        <field name="key">product_company_default.default_company_enable</field>
```

yostashiro (commented):
@AungKoKoLin1997 Looks good. Please squash commits.

legalsylvain (changes requested):
I think that the feature should be enabled by default. That is the expected behaviour, if people install the module. having an ir.config.parameter to enable it adds useless complexity. in data it should be "1"

However, we can disable it by demo data, to be sure there is no problem in demo / CI context. (can be "0")

yostashiro (approved):
Code review.

legalsylvain (approved):

PR #1413, opened 2023-09-12
[15.0][FIX] product_secondary_unit: Don't reassign old value

pedrobaeza:
**Steps to reproduce the problem:**

- Have `sale_product_matrix` installed.
- Add a line with a product with variants in a sales order. It shouldn't have secondary units.
- Save the sales order.
- Use the matrix popup to change the quantity and confirm.

**Current behavior:**

The quantity is reset to the original one.

**Expected behavior:**

The quantity is set to the new value.

**Explanation:**

Using `_origin` for resetting the value is taking the old value previously saved before starting to edit again, and anyway, that was a needed hack for v13 for avoiding cache miss error, but now it's not needed.

Thus, we remove such artifact and just act on lines with secondary unit set.

@Tecnativa TT45028


carlosdauden (approved):
Ok if it is not needed

sergio-teruel (approved):

pedrobaeza:
Thanks for the quick reaction.

Merging then:

/ocabot merge patch

PR #1414, opened 2023-09-13
[15.0][FIX] product_secondary_unit: wrong factor computation for uom's equal between secondary uom and product uom

sergio-teruel:
cc @Tecnativa 

ping @pilarvargas-tecnativa @chienandalu


carlosdauden (approved):

pilarvargas-tecnativa (approved):
LGTM thanks

PR #1415, opened 2023-09-14
[16.0][FIX] product_secondary_unit: wrong factor computation for uom's equal between secondary uom and product uom

pilarvargas-tecnativa:
fordward port of: https://github.com/OCA/product-attribute/pull/1414

cc @Tecnativa TT45087

@sergio-teruel @chienandalu please review :)


sergio-teruel (approved):

pedrobaeza (approved):
/ocabot merge patch

PR #1425, opened 2023-09-28
[16.0][FIX] product_secondary_unit: precompute secondary_uom_qty

BT-cserra:
The field `secondary_uom_qty` is defined in the mixin class `product.secondary.unit.mixin` as being computed, but without precomputation. This makes that when this field is used in the variable `_secondary_unit_fields` for classes using the mixing, it ends up as a dependency of other fields that are computed and do require precomputation. But, not being the field `secondary_uom_qty` precomputed, it disables the precomputation of those other fields.

This has an impact when installing the module `sale` along with the module `product_secondary_unit`, which causes the module `sale_order_secondary_unit` to be auto-installed. As a result, the following errors appear in the log, disabling the precomputation of 11 other fields:

```text
/opt/project/repos/odoo/odoo/fields.py:802: UserWarning: Field sale.order.line.product_uom_qty cannot be precomputed as it depends on non-precomputed field sale.order.line.secondary_uom_qty
/opt/project/repos/odoo/odoo/fields.py:802: UserWarning: Field sale.order.line.price_unit cannot be precomputed as it depends on non-precomputed field sale.order.line.product_uom_qty
/opt/project/repos/odoo/odoo/fields.py:802: UserWarning: Field sale.order.line.discount cannot be precomputed as it depends on non-precomputed field sale.order.line.product_uom_qty
/opt/project/repos/odoo/odoo/fields.py:802: UserWarning: Field sale.order.line.price_reduce cannot be precomputed as it depends on non-precomputed field sale.order.line.price_unit
/opt/project/repos/odoo/odoo/fields.py:802: UserWarning: Field sale.order.line.price_subtotal cannot be precomputed as it depends on non-precomputed field sale.order.line.product_uom_qty
/opt/project/repos/odoo/odoo/fields.py:802: UserWarning: Field sale.order.line.price_tax cannot be precomputed as it depends on non-precomputed field sale.order.line.product_uom_qty
/opt/project/repos/odoo/odoo/fields.py:802: UserWarning: Field sale.order.line.price_total cannot be precomputed as it depends on non-precomputed field sale.order.line.product_uom_qty
/opt/project/repos/odoo/odoo/fields.py:802: UserWarning: Field sale.order.line.price_reduce_taxexcl cannot be precomputed as it depends on non-precomputed field sale.order.line.price_subtotal
/opt/project/repos/odoo/odoo/fields.py:802: UserWarning: Field sale.order.line.price_reduce_taxinc cannot be precomputed as it depends on non-precomputed field sale.order.line.price_total
/opt/project/repos/odoo/odoo/fields.py:802: UserWarning: Field sale.order.line.product_packaging_id cannot be precomputed as it depends on non-precomputed field sale.order.line.product_uom_qty
/opt/project/repos/odoo/odoo/fields.py:802: UserWarning: Field sale.order.line.product_packaging_qty cannot be precomputed as it depends on non-precomputed field sale.order.line.product_packaging_id
```

Making `secondary_uom_qty` precomputed solves these warnings.

### DIFF
--- a/product_secondary_unit/models/product_secondary_unit_mixin.py
+++ b/product_secondary_unit/models/product_secondary_unit_mixin.py
@@ -43,6 +43,7 @@ class ProductSecondaryUnitMixin(models.AbstractModel):
         store=True,
         readonly=False,
         compute="_compute_secondary_uom_qty",
+        precompute=True,
     )
     secondary_uom_id = fields.Many2one(
         comodel_name="product.secondary.unit",


BT-cserra:
Hello @sergio-teruel. I see you were already contacted by our common friend @OCA-git-bot but I wanted to address you directly.

I'm not really sure this is the correct fix. However, this is the only fix I can come up with right now. Why I'm not sure? Well because this is a problem that has its root in the module `product_secondary_unit`, but that only appears when installing (maybe as a side effect, for example after installing `sale`) the module `sale_order_secondary_unit`, so _I think_ the fix should be in the module `sale_order_secondary_unit` ─  somehow. I can envision other modules causing a similar problem, though, because of the same reason.

At first I thought it would be enough to redefine the field `secondary_uom_qty` of the class `SaleOrderLine` in the file `sale_order_secondary_unit/models/sale_order.py`, but doing so yields this other warning:

```text
/opt/project/repos/odoo/odoo/fields.py:460: UserWarning: precompute attribute doesn't make any sense on non computed field sale.order.line.secondary_uom_qty
```

In other words: The issue addressed by this PR is real. _BUT_, maybe I'm not addressing the fix properly, maybe I'm missing something or maybe I missunderstood completely the ─to me, new─  precomputation logic and this PR doesn't make sense at all. 

What do you think? Do you see another way of fixing the issue? Thanks in advance.

sergio-teruel (approved):

sergio-teruel:
@BT-cserra Totally agree!! Thanks...

pedrobaeza (approved):
/ocabot merge patch

PR #1426, opened 2023-10-02
[16.0][FIX] product_packaging_dimension: post-migrate script should only apply query if column max_weight exists

JordiBForgeFlow:
This column is no longer defined in 16.0

cc @rousseldenis


bosd (approved):
LGTM

rousseldenis (approved):

PR #1427, opened 2023-10-03
[16.0][IMP] product_packaging_level: configure product packaging name

anothingguy:
- Add feature to configure if the product packaging name
- On the package type, set a package level same as the package level on the product packaging.

### DIFF
--- a/product_packaging_level/models/product_packaging.py
+++ b/product_packaging_level/models/product_packaging.py
@@ -28,6 +28,7 @@ class ProductPackaging(models.Model):
     qty_per_level = fields.Char(
         compute="_compute_qty_per_level", string="Qty per package level"
     )
+    name_policy = fields.Selection(related="packaging_level_id.name_policy")
 
     def default_packaging_level_id(self):
         return self.env["product.packaging.level"].search(
@@ -36,7 +37,10 @@ class ProductPackaging(models.Model):
 
     @api.constrains("packaging_level_id", "product_id")
     def _check_one_packaging_level_per_product(self):
-        for packaging in self:
+        packages_name_by_level = self.filtered(
+            lambda p: p.name_policy == "by_package_level"
+        )
+        for packaging in packages_name_by_level:
             product = packaging.product_id
             # do not use a mapped/filtered because it would union the duplicates
             packaging_level_ids = [
@@ -116,16 +120,17 @@ class ProductPackaging(models.Model):
             res.append((code, new_qty))
         return res
 
-    @api.onchange("packaging_level_id")
-    def _onchange_name(self):
-        if self.packaging_level_id:
-            self.name = self.packaging_level_id.name
+    @api.onchange("package_type_id")
+    def _onchange_package_type(self):
+        package_type_level_id = self.package_type_id.packaging_level_id
+        if package_type_level_id:
+            self.packaging_level_id = package_type_level_id
 
-    def name_get(self):
-        result = []
-        for record in self:
-            if record.product_id and record.packaging_level_id:
-                result.append((record.id, record.packaging_level_id.display_name))
-            else:
-                result.append((record.id, record.name))
-        return result
+    @api.onchange("packaging_level_id", "package_type_id", "name", "name_policy")
+    def _onchange_name(self):
+        new_name = self.name
+        if self.name_policy == "by_package_level":
+            new_name = self.packaging_level_id.display_name
+        elif self.name_policy == "by_package_type":
+            new_name = self.package_type_id.name
+        self.name = new_name
--- a/product_packaging_level/models/product_packaging_level.py
+++ b/product_packaging_level/models/product_packaging_level.py
@@ -15,6 +15,33 @@ class ProductPackagingLevel(models.Model):
     has_gtin = fields.Boolean()
     active = fields.Boolean(default=True)
     is_default = fields.Boolean()
+    name_policy = fields.Selection(
+        selection=[
+            ("by_package_level", "Package Level Name"),
+            ("by_package_type", "Package Type Name"),
+            ("user_defined", "User Defined"),
+        ],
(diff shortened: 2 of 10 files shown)


product_packaging_level/models/stock_package_type.py
```diff
@@ -0,0 +1,12 @@
+# Copyright 2019-2020 Camptocamp (<https://www.camptocamp.com>).
```
jbaudoux:
```suggestion
# Copyright 2023 Camptocamp (<https://www.camptocamp.com>).
```

jbaudoux (approved):
LG, not tested

product_packaging_level/models/product_packaging.py
```diff
@@ -28,15 +28,46 @@ class ProductPackaging(models.Model):
...
+            ("user_defined", "User Defined"),
+        ],
+        default="by_package_level",
+        help=(
+            "config to set name of product packaging. Three options:"
+            "- The package level name (default)"
+            "- The package type name (if groups='stock.group_tracking_lot')"
+            "- user defined: free text value defined"
+        ),
+    )
```
jbaudoux: This config should be defined on the stock_package_level and not on the product_packaging
(so that we don't have to configure on each packaging and it will depend on which packaging level is chosen).

product_packaging_level/views/product_packaging_view.xml
```diff
@@ -31,6 +31,9 @@
         <field name="model">product.packaging</field>
         <field name="inherit_id" ref="stock.product_packaging_form_view" />
         <field name="arch" type="xml">
+            <xpath expr="//field[@name='product_id']" position="before">
+                <field name="name_type" />
+            </xpath>
```
jbaudoux: Same, to move to packaging level view

jbaudoux (changes requested):

product_packaging_level/models/stock_package_type.py
anothingguy: updated

product_packaging_level/views/product_packaging_view.xml
anothingguy: updated

product_packaging_level/models/product_packaging.py
anothingguy: updated

santostelmo (approved):

product_packaging_level/models/product_packaging_level.py
```diff
@@ -15,6 +15,34 @@ class ProductPackagingLevel(models.Model):
     has_gtin = fields.Boolean()
     active = fields.Boolean(default=True)
     is_default = fields.Boolean()
+    name_type = fields.Selection(
```
simahawk: `name_based_on` or `name_policy` sound better

anothingguy: yes, thanks, sounds good, updated, hehe

cyrilmanuel (approved):
LGTM

product_packaging_level/models/product_packaging.py
```diff
@@ -116,16 +120,19 @@ def _make_qty_per_level(self, qty_per_level_mapping, format_pattern=None):
...
 
-    def name_get(self):
-        result = []
-        for record in self:
-            if record.product_id and record.packaging_level_id:
-                result.append((record.id, record.packaging_level_id.display_name))
-            else:
-                result.append((record.id, record.name))
-        return result
+    @api.onchange("packaging_level_id", "package_type_id", "name")
```
simahawk:
```suggestion
    @api.onchange("packaging_level_id", "package_type_id", "name", "name_policy")
```

simahawk (approved):
Pre-approving.
Please check the onchange and remove the version from the commit msg as is not needed

product_packaging_level/models/product_packaging.py
anothingguy: yes, I updated all, thanks

jbaudoux (approved):
LGTM

simahawk (approved):

product_packaging_level/models/product_packaging.py
```diff
@@ -116,16 +120,17 @@ def _make_qty_per_level(self, qty_per_level_mapping, format_pattern=None):
...
-    def name_get(self):
-        result = []
-        for record in self:
-            if record.product_id and record.packaging_level_id:
-                result.append((record.id, record.packaging_level_id.display_name))
-            else:
-                result.append((record.id, record.name))
-        return result
+    @api.onchange("packaging_level_id", "package_type_id", "name", "name_policy")
+    def _onchange_name(self):
```
lmignon, line 130: @anothingguy Why not a computed fields with read_only="False" This change is annoying since any packaging created from the python code is created with the wrong name. Moreover, by replacing the `name_get` by an `on_change` you  broke existing installation since the name is no more properly displayed for existing records.

lmignon, line 130: @anothingguy @jbaudoux The change from `name_get` to an `onchange` is very annoying. It breaks my use of this addon. Moreover, the name on the package level is `translate=True` but not on the product packaging, that was one of the motivations to override the `name_get` .

lmignon: cc @simahawk

jbaudoux: So we should 
* recover name_get, return value according to name policy and add a comment that this is to respect translations on level name
* replace _onchange_name by making the name computed stored=True

then maybe also
* change _onchange_package_type by making packaging_level_id computed store=True based on package_type_id

correct ?

jbaudoux: @lmignon Fixed here https://github.com/OCA/product-attribute/pull/1527 Can you review?

PR #1432, opened 2023-10-12
[14.0][IMP] p_sale_manufactured_for: imp when customer is archived

TDu:
When a partner being archived is used on the Manufactured For field That information is not visible anymore on the product form, although it is set. And that product can still not be ordered by any other customers.

Removing the assignation when a customer is archived, fixes both problems.

### DIFF
--- a/product_sale_manufactured_for/models/res_partner.py
+++ b/product_sale_manufactured_for/models/res_partner.py
@@ -1,0 +1,59 @@
+# Copyright 2023 Camptocamp SA
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl.html)
+from collections import defaultdict
+
+from odoo import _, models
+
+
+class ResPartner(models.Model):
+    _inherit = "res.partner"
+
+    def write(self, vals):
+        res = super().write(vals)
+        if vals.get("active") is False:
+            self._clean_product_manufactured_for()
+        return res
+
+    def _clean_product_manufactured_for(self):
+        """Update product manufactured for when a partner is archived.
+
+        When a partner being archived is used on the Manufactured For field
+        that information is not visible anymore on the product form, although it
+        is still set.
+        And that product can still not be ordered by any other customers.
+
+        Removing this relation when a partner is being archived solves the issue.
+
+        The change is also being logged in the chatter of the related product.
+
+        """
+        query = (
+            "DELETE FROM product_product_manuf_for_partner_rel "
+            "WHERE partner_id IN %s"
+            "RETURNING "
+            "    product_product_manuf_for_partner_rel.product_id, "
+            "    product_product_manuf_for_partner_rel.partner_id;"
+        )
+        self.env.cr.execute(query, (tuple(self.ids),))
+        result = self.env.cr.fetchall()
+        if result:
+            product_to_update = defaultdict(list)
+            for row in result:
+                product_to_update[row[0]].append(row[1])
+            products = self.env["product.product"].browse(product_to_update.keys())
+            customer_ids = {
+                id for values in product_to_update.values() for id in values
+            }
+            all_customers = self.env["res.partner"].browse(customer_ids)
+            for product in products:
+                customers = all_customers.filtered(
+                    lambda customer: customer.id in product_to_update[product.id]
+                )
+                customers_name = ", ".join(customers.mapped("name"))
+                product.message_post(
+                    body=_(
+                        "The product was manufactured for %(customers_name)s "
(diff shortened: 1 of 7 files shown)


product_sale_manufactured_for/models/res_partner.py
```diff
@@ -0,0 +1,17 @@
...
+    _inherit = "res.partner"
+
+    def write(self, vals):
+        res = super().write(vals)
+        if vals.get("active") is False:
+            query = (
+                "DELETE FROM product_product_manuf_for_partner_rel "
+                "WHERE partner_id IN (%s);"
+            )
+            self.env.cr.execute(query, (tuple(self.ids)))
```
TDu: Maybe, as the information of whom the product was manufactured for is lost when a customer is archived. This could be logged in the chatter ?

product_sale_manufactured_for/tests/test_product_sale_manufactured_for.py
```diff
@@ -0,0 +1,19 @@
+# Copyright 2023 Camptocamp
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+from odoo.tests.common import SavepointCase
+
+
+class TestProductSaleManufacturedFor(SavepointCase):
+    def setUp(self):
```
simahawk: setUpClass ;)

product_sale_manufactured_for/models/res_partner.py
simahawk: Good idea.
Regarding the override of write, wouldn't be less "intrusive" to define and inverse method for the active field only?

product_sale_manufactured_for/tests/test_product_sale_manufactured_for.py
TDu: Fixed

TDu:
I have added the logging in the chatter, as mentioned above. There was also a bug in the sql execute with multiple records.
I did not change the implementation to use an inverse method. I understand the function would be called only when necessary, but is there any performance gain or other reason to do so.
Is it related to the 

> Add the inverse method pattern

 in https://github.com/OCA/odoo-community.org/blob/master/website/Contribution/CONTRIBUTING.rst#id59 ?

product_sale_manufactured_for/models/res_partner.py
```diff
@@ -0,0 +1,43 @@
...
+                for product in products:
+                    customers = all_customers.filtered(
+                        lambda customer: customer.id in product_to_update[product.id]
+                    )
+                    customers_name = ", ".join(customers.mapped("name"))
+                    product.message_post(
+                        body=_(
+                            f"The product was manufactured for {customers_name} "
+                            "but the customer(s) has been archived."
+                        )
```
simahawk: The string won't be translatable because is going to change at run time

```suggestion
                        body=_(
                            "The product was manufactured for %(customers_name)s "
                            "but the customer(s) has been archived.",
                            customers_name=customers_name
                        )
```

simahawk (changes requested):

simahawk:
> I have added the logging in the chatter, as mentioned above. There was also a bug in the sql execute with multiple records. I did not change the implementation to use an inverse method. I understand the function would be called only when necessary, but is there any performance gain or other reason to do so. Is it related to the
> 
> > Add the inverse method pattern
> 
> in https://github.com/OCA/odoo-community.org/blob/master/website/Contribution/CONTRIBUTING.rst#id59 ?

IMO your gain is that is more clear and you don't add yet another override of the write on the partner.
Anyway, fine.... :)

product_sale_manufactured_for/models/res_partner.py
```diff
@@ -0,0 +1,44 @@
...
+from odoo import _, models
+
+
+class ResPartner(models.Model):
+    _inherit = "res.partner"
+
+    def write(self, vals):
+        res = super().write(vals)
+        if vals.get("active") is False:
+            query = (
```
simahawk: A nice comment here would be nice. Even better: an additional method with its own docstring :wink:

simahawk (commented):
can you check the build?

simahawk:
@TDu ping

product_sale_manufactured_for/models/res_partner.py
TDu: Added

TDu:
@simahawk  Updated, will squash if good for you

simahawk (approved):

PR #1435, opened 2023-10-19
[14.0] [ADD] product_attribute_variant_rules

paradoxxxzero:
This module adds a more powerful way to describe your product attributes combinations than the default exclusions.

20231019_16h11m53s_grim

It allows to write rules like:

  * All products with blue or green color in XL size will appear only with a V neck collar
    and a short sleeve.
  * All L size products will never appear with a sailor collar.

The rules are split between a precondition a type and a postcondition.

Different attributes are ANDed and same attributes are ORed.

For instance the rule::
```
All products with blue or green color in XL size will appear only with a V neck collar and a short sleeve.
```

Will be written as
```
Precondition: (color: blue), (color: green), (size: XL)
Type: Only With
Postcondition: (collar: V neck), (sleeve: short)
```


product_attribute_variant_rules/views/product_template_views.xml
```diff
@@ -0,0 +1,43 @@
...
+                    attrs="{'invisible': ['|', ('use_attribute_rules', '=', False), ('id','=', False)]}"
+                >
+
+          <tree string="Product Attribute Rules" editable="bottom">
+            <field name="available_precondition_attribute_ids" invisible="1" />
+            <field name="available_postcondition_attribute_ids" invisible="1" />
+            <field
+                            name="product_attribute_value_precondition_ids"
+                            widget="many2many_tags"
+                            domain="[('attribute_id', 'in', available_precondition_attribute_ids)]"
```
sebastienbeau: Domain is not complete here.
we can not filter only on attribute_id.

For example you can have a attribute color with blue, red, green, yellow.

But on the product you only have added yellow and red.
You should not be able to select blue and green

paradoxxxzero: Indeed

product_attribute_variant_rules/models/product_attribute_rule.py
```diff
@@ -0,0 +1,152 @@
...
+        Check if the combination matches the given conditions.
+        """
+        # If there is no condition, the combination matches
+        # (only possible for the preconditions)
+        if not conditions:
+            return True
+
+        # Check if the combination matches the preconditions ANDed between
+        # different attributes
+        for attribute, attribute_values in self._aggregate_conditions(
```
sebastienbeau: You can use the groupby function of odoo

```
from odoo.tools import groupby


for attribute, values in groupby(combinations, lambda s: s.attribute_id):


```

https://github.com/odoo/odoo/blob/53a3509a10f4ed2f4edf5374bb19ebda57e7530e/odoo/tools/misc.py#L1225

product_attribute_variant_rules/models/product_template.py
```diff
@@ -0,0 +1,72 @@
...
+        Check if the combination is possible with the rules defined on the product template.
+        """
+        # Rules are ANDed together
+        for rule in self.product_attribute_rule_ids:
+            if not rule._is_combination_possible(combination):
+                return False
+        return True
+
+    def write(self, vals):
+        res = super(ProductTemplate, self).write(vals)
```
sebastienbeau: python 3 :) (no more params for super, even in odoo)
```suggestion
        res = super().write(vals)
```

paradoxxxzero: Bad copy/paste :)

sebastienbeau:
I have rename type into rule_type (I had a issue with product_profile) and type is already a python keyword

paradoxxxzero:
> I have rename type into rule_type (I had a issue with product_profile) and type is already a python keyword

That shouldn't be much of a problem except if you want to use original python `type` function in the class declaration scope, a lot of odoo modules have `type` as a field (`product.template` for instance). Care to share what was the issue?

sebastienbeau:
The issue is related to a bug in product_profile (not so easy to fix). The module was adding a invisible condition in the field "type" (it should not do it). So the quick fix is to rename the field, and as "type" is a python keyword, it make sense to rename it.

sebastienbeau (approved):
LGTM (code review + test) used in production

sebastienbeau:
@Kev-Roche can you review ?

Kev-Roche (approved):

PR #1438, opened 2023-10-24
[16.0][ADD] product_pricelist_fixed_currency_rate

LoisRForgeFlow:
This module extends the product pricelists feature to allow to set a fixed
currency rate to convert from a pricelist with a different currency.

Usage
=====

When using advanced price rules:

#. Go to a pricelist and create a pricelist with a given currency.
#. Create a new pricelist rule with computation formula.
#. Make it based on other pricelist that uses a different currency.
#. *Fixed Currency Rate* will show up, set a value.

This will force the currency conversion to always use the specified
*Fixed Currency Rate* rate when using this pricelist.

@ForgeFlow


product_pricelist_fixed_currency_rate/models/product_pricelist_item.py
```diff
@@ -0,0 +1,109 @@
...
+    do_inverse_currency_rate = fields.Boolean(
+        compute="_compute_do_inverse_currency_rate",
+        store=True,
+        readonly=False,
+    )
+    currency_rate_tooltip = fields.Char(
+        compute="_compute_currency_rate_tooltip",
+    )
+
+    @api.depends("base_pricelist_id", "base_pricelist_id.currency_id")
```
JordiMForgeFlow: to be safe maybe we should also take into account that compute_price must be _formula_. Otherwise, if you have a pricelist that you have configured with fixed rate and then change the logic of the pricelist to fixed price, the fields keep showing in the view:

image

LoisRForgeFlow: Indeed, attended!

JordiMForgeFlow (approved):
Code + functional review 👍🏼

JordiBForgeFlow (approved):
:+1: Technical and functional review

PR #1439, opened 2023-10-24
[16.0][FIX] product_pricelist_simulation: switch tests to TransactionCase

LoisRForgeFlow:
SavepointCase is deprecated and it was making CI fail.

### DIFF
--- a/product_pricelist_simulation/tests/test_product_pricelist_simulation.py
+++ b/product_pricelist_simulation/tests/test_product_pricelist_simulation.py
@@ -5,7 +5,7 @@
 from odoo.tests import Form, common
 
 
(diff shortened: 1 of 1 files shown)


pedrobaeza (approved):
I would say this is just a warning, so maybe the failing is in other place, but good to change this.

LoisRForgeFlow:
@pedrobaeza Yes, I was just realizing of that, the warning was so big and ugly that I got fooled :laughing: 

In any case, good to fix.

pedrobaeza:
Let's merge it then:

/ocabot merge path

LoisRForgeFlow:
@pedrobaeza you misspelled the command, could you try again?

PR #1453, opened 2023-11-20
[ADD] product_internal_reference_generator

ilyasProgrammer:
This module allows to generate internal references for Product templates and variants using sequences, setting codes as read-only.

In product template, it's possible to choose among different Internal Reference Templates related to a sequence, and then generate an internal reference with the following structure:

Internal Reference Prefix + progressive number for variant, eg: “Main0001001”, where:

"Main0001" is the prefix generated by sequence and assigned to product template, and

"001" the variant identifier.

Every time a new variant is created, a new internal reference is automatically assigned with progressive variant code.

A specific access rights allows specific users to change internal reference template for a product template once an internal reference has been generated, as well as editing existing ones.


HekkiMelody (approved):
Moving my review over from https://github.com/OCA/product-variant/pull/309

Code review, LGTM!

ilyasProgrammer:
Module is incompatible with product_sequence. Is there a way to specify incompatibility?

HekkiMelody:
There are two steps you need to take:

1. Add the key `excludes` to your `__manifest__.py`. See for example: https://github.com/OCA/account-financial-tools/blob/14.0/account_asset_management/__manifest__.py#L11

2. run `copier update --trust -A -o rej` on the repo (install `copier` if necessary: https://copier.readthedocs.io/en/stable/ ),
    when you get to the question about rebel modules, specify the list of module(s) you want to test separately from the rest of the modules in the repo. This should generate the appropriate files in `.github/workflows/test.yaml` Example:
    image

    Fix conflicts and remove any `.rej` files if any were created, and commit the changes.
    
    If the command doesn't seem to generate the correct files, you can try `copier recopy` instead of `copier update`

HekkiMelody:
Make sure to also push the new `.copier-answers.yml` file!

ilyasProgrammer:
There was too many questions from copier after copier recopy and lots of files changed. So i just copied only tests yml. I dont think my new answers file was correct.
Is it possible to run copier recopy to regenerate only tests yml ?

HekkiMelody:
Aside from the one question that I took a screenshot of, you don't have to change any of the other answers. if you don't commit the `.copier-answers.yml`, your changes may be reverted. In general the commit may not be accepted if you only change `test.yaml`

francesco-ooops:
@ilyasProgrammer why don't you change copier in a separate PR so it gets merged straight away?

ilyasProgrammer:
> @ilyasProgrammer why don't you change copier in a separate PR so it gets merged straight away?

Problem not in copier, but in modules conflicts.

ilyasProgrammer:
But looks like need to update copier as well

francesco-ooops:
@ilyasProgrammer copier files should be removed from here I think

ilyasProgrammer:
rebased.
I guess rebel_module_groups needs to stay in the answers file

francesco-ooops (approved):
Functional review ok!

@pedrobaeza feedback? thanks!

.copier-answers.yml
```diff
@@ -15,7 +15,9 @@ odoo_test_flavor: Both
 odoo_version: 14.0
 org_name: Odoo Community Association (OCA)
 org_slug: OCA
-rebel_module_groups: []
+rebel_module_groups:
+- product_internal_reference_generator
```
pedrobaeza, line 19: You only need to put one of them (yours), not both.

pedrobaeza (changes requested):

.copier-answers.yml
ilyasProgrammer: updated

francesco-ooops:
@pedrobaeza good to go?

pedrobaeza:
No, the change should be done in the copier template and should be reflected in the CI yaml, not manually edited the file `.copier-answers...`.

ilyasProgrammer:
So what i did is:
copier update --trust 
Are there in this repo modules that don't get along with their friends? If so, list them here (YAML format) and they will be tested in separate
Beware, if rebel modules should stay separated in groups, you should join them with commas, which could be misinterpreted by YAML.
Example: ["rebel_module_1,rebel_module_2", even_more_rebel_module]
   [product_internal_reference_generator]

@pedrobaeza is it correct ?

pedrobaeza (approved):
OK, now it is correct

/ocabot merge nobump

PR #1455, opened 2023-11-21
[16.0] product_packaging_level_purchasable: imp readme

simahawk:


simahawk:
Last week we had a major outage of OCA servers and some new modules where merged but not published on pypi.
The merge of this PR should fix the build for other PRs like https://github.com/OCA/purchase-workflow/pull/2051 that rely on the module being available on pypi.

Issue #1157, opened 2022-10-03
Migration to version 16.0


sbidoul:
# Todo

https://github.com/OCA/maintainer-tools/wiki/Migration-to-version-16.0

# Modules to migrate

- [ ] packaging_uom - By @gaelTorrecillas - #1399
- [x] pos_product_cost_security - By @chienandalu - #1661
- [x] product_abc_classification - By @rousseldenis - #1192
- [x] product_abc_classification_base - By @rousseldenis - #1192
- [x] product_abc_classification_sale_stock - By @rousseldenis - #1196
- [x] product_assortment - By @marielejeune - #1158
- [x] product_attachment_zipped_download - By @victoralmau - #1822
- [x] product_attribute_archive - By @Rferri44-S73 - #1301
- [x] product_attribute_value_menu - By @neuby001 - #1347
- [ ] product_categ_image - By @cubells - #1967
- [ ] product_category_active
- [ ] product_category_code
- [x] product_category_code_unique - By @manuelregidor - #1829
- [x] product_category_description - By @Tisho99 - #1842
- [ ] product_category_product_link
- [x] product_code_mandatory - By @ced-adhoc - #1693
- [x] product_code_unique - By @marielejeune - #1159
- [x] product_cost_security - By @bvanjeelharia - #1240
- [x] product_dimension - By @FrancoMaxime - #1168
- [x] product_expiry_configurable - By @rousseldenis - #1318
- [x] product_internal_reference_generator - By @sbiosca-s73 - #1485
- [ ] product_list_pricelist_price - By @MarinaAForgeFlow - #1572
- [x] product_logistics_uom - By @marielejeune - #1160
- [x] product_lot_sequence - By @AungKoKoLin1997 - #1296
- [x] product_main_supplierinfo - By @mmequignon - #1786
- [x] product_manufacturer - By @FrancoMaxime - #1166
- [x] product_multi_category - By @FrancoMaxime - #1167
- [x] product_multi_image - By @IJOL - #1707
- [x] product_net_weight - By @RabbitJon-S73 - #1281
- [x] product_origin - By @acsonefho - #1389
- [x] product_packaging_dimension - By @marielejeune - #1161
- [x] product_packaging_type - By @rousseldenis - #1215
- [ ] product_packaging_type_pallet - By @rousseldenis - #1216
- [ ] product_pricelist_by_contact - By @cubells - #2052
- [x] product_pricelist_direct_print - By @legalsylvain - #1365
[...]

PR #1456, opened 2023-11-21
[16.0] [MIG] product_product_template_navigation

renda-dev:
Migration from 14.0 to 16.0 of **[product_product_template_navigation](https://github.com/OCA/product-attribute/tree/14.0/product_product_template_navigation)**


legalsylvain (changes requested):
Hi. Thanks for porting this module.

could you take this opportunity to rename the module's name, which does not correspond to the standard of other OCA modules? 

the other module that adds a link between a model and another, are suffixed by ``_link``. so ``product_product_template_link`` could be a better name.

See the current 60 OCA modules : 

image

renda-dev:
> Hi. Thanks for porting this module.
> 
> could you take this opportunity to rename the module's name, which does not correspond to the standard of other OCA modules?
> 
> the other module that adds a link between a model and another, are suffixed by `_link`. so `product_product_template_link` could be a better name.
> 
> See the current 60 OCA modules :
> 
> image

Done! Thanks for the quick feedback :)

legalsylvain (approved):
Diff review.  Lgtm

francesco-ooops (approved):
Functional is ok, but why are we using the "archive" icon?

image

I think we could re-use the "variants" odoo standard icon, even if the buttons are not really specular

image

francesco-ooops:
@OCA/product-maintainers could you please add this to https://github.com/OCA/product-attribute/issues/1157 specifying the new module name? 

also: should the PR have this name or the new module's name?

renda-dev:
I totally agree with you @francesco-ooops, thanks for the suggestion! :)

PR #1470, opened 2023-11-28
[IMP] product_packaging_container_deposit update container deposit qauntity with Command

santostelmo:
Using Command delete/update to apply container deposit update operations all at once.


simahawk:
See #1475

PR #1475, opened 2023-12-05
[16.0] product_packaging_container_deposit: fix deletion of lines

simahawk:
Continues the work of #1470  and replaces it.

Lines deletion cannot be done in the same transaction or you'll get a broken cache.
See atomic commits for details.

### DIFF
--- a/product_packaging_container_deposit/models/container_deposit_order_mixin.py
+++ b/product_packaging_container_deposit/models/container_deposit_order_mixin.py
@@ -41,6 +41,7 @@ class OrderMixin(models.AbstractModel):
             deposit_container_qties = (
                 lines_to_comp_deposit._get_order_lines_container_deposit_quantities()
             )
+            values_lst = []
             for line in self[self._get_order_line_field()]:
                 if not line.is_container_deposit:
                     continue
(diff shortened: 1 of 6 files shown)


product_packaging_container_deposit/models/container_deposit_order_mixin.py
```diff
@@ -78,6 +87,22 @@ def update_order_container_deposit_quantity(self):
                     )
                     values_lst.append(Command.create(values))
             order.write({self._get_order_line_field(): values_lst})
+        # Schedule line to delete after commit to avoid caching issue w/ UI
+        if line_ids_to_delete:
+            self.env.cr.postcommit.add(
+                partial(
+                    self._order_container_deposit_delete_lines_after_commit,
+                    sorted(line_ids_to_delete),
```
santostelmo, line 95: Why you need to sort the lines ?

product_packaging_container_deposit/models/container_deposit_order_mixin.py
```diff
@@ -49,17 +54,21 @@ def update_order_container_deposit_quantity(self):
...
-                            Command.update(
-                                line.id,
-                                {
-                                    line._get_product_qty_field(): 0,
-                                },
-                            )
+                        # values_lst.append(Command.delete(line.id))
+                        line_ids_to_delete.append(line.id)
+                        # TODO: check if it is needed for UI only
+                        new_vals["name"] = _("[DEL] %(name)s", name=line.name)
```
santostelmo, line 64: If the container deposit line quantity is increased from 0 to n, you will keep the line name with `[DEL]` ?

simahawk: How can you do that if the line is deleted? Can you test it?

santostelmo: You're right! My bad.  I was thinking about confirmed orders where lines are not deleted => in this case we don't write `DEL` on the line name, just set qty to 0

product_packaging_container_deposit/models/container_deposit_order_mixin.py
```diff
@@ -41,36 +46,63 @@ def update_order_container_deposit_quantity(self):
...
                     )
-            values_lst = []
             for product in deposit_container_qties:
                 if deposit_container_qties[product][0] > 0:
                     values = order.prepare_deposit_container_line(
                         product, deposit_container_qties[product][0]
                     )
                     values_lst.append(Command.create(values))
             order.write({self._get_order_line_field(): values_lst})
+        # Schedule line to delete after commit to avoid caching issue w/ UI
```
jbaudoux, line 90: Don't you get the issue on next call from UI ?
Seems safer to me to keep the lines but set quantity to 0

jbaudoux: Saw a demo, it does the job !
Great solution

santostelmo:
Functionally tested and working

santostelmo (approved):

PR #1480, opened 2023-12-12
[16.0][FIX] product_print_category fix a security issue

nayatec:
Odoo V16 does not allow 'regular' users (ie users with no Admin/Access Right group) to acces to ir.model.fields at all, therefore i added a sudo to allow users with no such right to be able to actually write to products.
Did not find a more elegant solution.


legalsylvain (approved):
Tested on runboat / 16.0 with demo user.
indeed we have an error. (this should be new in V16, I didn't have any problem in V12).

image


LGTM. Thanks for fixing this issue.

/ocabot merge patch

PR #1483, opened 2023-12-14
[16.0][IMP] product_packaging_container_deposit process negative quantities

santostelmo:
Allows negative quantities of product container deposit to be processed. This is particularly important if the user wants to process the return to the customer or product supplier.


simahawk (approved):
LG

PR #1491, opened 2024-01-05
[16.0][ADD] product_attribute_model_link

angelinaanaki:
This module allows to link product attributes to models and populate attribute values from the model records and vice versa.

When a record is created in a model that is linked to an attribute new attribute value will be created automatically.

When a record that is linked to an attribute value is deleted linked attribute value will be deleted if it is not used or archived otherwise.

If a related record field value linked to an attribute value is updated the attribute value name is updated accordingly.


product_attribute_model_link/models/product_attribute_value.py
```diff
@@ -0,0 +1,184 @@
...
+class ProductAttributeValue(models.Model):
+    _inherit = "product.attribute.value"
+
+    linked_record_ref = fields.Reference(
+        selection="_selection_target_model", string="Linked record"
+    )
+
+    @api.model
+    def _selection_target_model(self):
+        models = self.env["ir.model"].search([])
```
geomer198: I think better use search_read by 'model' and 'name' fields. Because search_read will return list of dictionaries. With basic types python works faster then object.

product_attribute_model_link/models/product_attribute_value.py
```diff
@@ -0,0 +1,184 @@
...
+        :param vals_list: List of dictionaries containing attribute values for creating records.
+        :return: Created records.
+        :raise: ValidationError if conversion fails for a field.
+        """
+        for vals in vals_list:
+            attribute_id = vals.get("attribute_id")
+            product_attribute = self.env["product.attribute"].browse(attribute_id)
+            # Create new record based on attribute value if create_from_attribute_values is True
+            if (
+                product_attribute.create_from_attribute_values
```
geomer198: Could you add check exists `product_attribute` variable in this condition like this:
```py
if product_attribute and product_attribute.create_from_attribute_values and not self.env.context.get("operation_from_record"):
    ...
```

product_attribute_model_link/wizards/linked_record_wizard.py
```diff
@@ -0,0 +1,39 @@
...
+    linked_record_ref = fields.Reference(
+        selection="_select_target_model",
+        string="Linked Record",
+    )
+
+    @api.model
+    def _select_target_model(self):
+        if self.env.context.get("default_linked_model"):
+            model_name = self.env.context.get("default_linked_model")
+            model = self.env["ir.model"].search([("model", "=", model_name)])
```
geomer198: Better use `_get()` method from `ir.model` model.
```py
model = self.env["ir.model"]._get(model_name)
```

geomer198 (commented):
Please check my comments

product_attribute_model_link/models/product_attribute.py
```diff
@@ -0,0 +1,337 @@
...
+    def create(self, vals_list, **kw):
+        product_attributes = self.env["product.attribute"].search(
+            [("linked_model_id", "=", self._name)]
+        )
+        if not product_attributes or self.env.context.get(
+            "operation_from_attribute_value"
+        ):
+            return create.origin(self, vals_list, **kw)
+        # Call original method
+        records = create.origin(self, vals_list, **kw)
```
DemchukM:
```python
# Call original method
records = create.origin(self, vals_list, **kw)
if not product_attributes or self.env.context.get("operation_from_attribute_value"):
    return records
        
``` 

To avoid code duplication, you can write like this

product_attribute_model_link/models/product_attribute.py
```diff
@@ -0,0 +1,337 @@
...
+        # Search for the connected product atribute values
+        product_attribute_values = self.env["product.attribute.value"].search(
+            [("linked_record_ref", "=", f"{self._name},{self.id}")]
+        )
+        if not (product_attribute_values and self) or self.env.context.get(
+            "operation_from_attribute_value"
+        ):
+            return write.origin(self, vals, **kw)
+        # Call original method
+        write.origin(self, vals, **kw)
```
DemchukM: Code duplication can also be avoided here

angelinaanaki:
@DemchukM @geomer198 Thank you for reviewing the code. Please take a look.

DemchukM (approved):
LGTM

product_attribute_model_link/models/product_attribute_value.py
```diff
@@ -0,0 +1,185 @@
...
+class ProductAttributeValue(models.Model):
+    _inherit = "product.attribute.value"
+
+    linked_record_ref = fields.Reference(
+        selection="_selection_target_model", string="Linked record"
+    )
+
+    @api.model
+    def _selection_target_model(self):
+        models = self.env["ir.model"].search_read([], ["model", "name"])
```
geomer198: I think need add value in domain `("transient", "=", False)`.

geomer198 (approved):
LGTM

product_attribute_model_link/models/product_attribute_value.py
angelinaanaki: You're correct. Fixed!

product_attribute_model_link/README.rst
```diff
@@ -0,0 +1,142 @@
...
+
+Configuration
+=============
+
+Go to the "Inventory/Configuration/Products/Attributes" menu and select an existing or create a new attribute.
+
+Following configuration fields are available:
+
+- Linked Model. Model which records will be used for the attribute values. Cannot be a transient model. Warning: changing or removing existing value will not affect existing attribute values!
+
```
ivs-cetmix, line 73: This warning appears every time when a new attribute is created. Even if it's not linked to any model. This is really confusing and misguiding. 
It should be shown only when an attribute is linked to a field.
And it should be some message in the form header. But not a pop-up for sure.
image

product_attribute_model_link/views/product_attribute_views.xml
```diff
@@ -0,0 +1,48 @@
...
+        <field name="arch" type="xml">
+            <xpath expr="//group[@name='main_fields']" position="after">
+                <group>
+                    <group name="model_link_fields">
+                        <field name="linked_model_id" />
+                        <field
+                            name="linked_field_id"
+                            attrs="{'required': [('linked_model_id', '!=', False)]}"
+                        />
+                        <field name="domain" />
```
ivs-cetmix: Domain widget must be used here. Not a plain char field.
image

product_attribute_model_link/views/product_attribute_views.xml
```diff
@@ -0,0 +1,48 @@
...
+                </group>
+            </xpath>
+            <xpath expr="//group[@name='model_link_fields']" position="before">
+                <xpath expr="//group[@name='main_fields']" position="move" />
+            </xpath>
+            <xpath
+                expr="//field[@name='value_ids']/tree/field[@name='name']"
+                position="after"
+            >
+                <field name="linked_record_ref" readonly="1" />
```
ivs-cetmix, line 64: It's not possible to link existing record to a new attribute value. Because `name` field is mandatory. However if one enters a value in the `name' field it will create a new record in the model. And we don't want to create a new record. We want to use an existing one.
image

ivs-cetmix (changes requested):
Functional review only no code check

angelinaanaki:
@ivs-cetmix Thanks for the review. I fixed it. Please take a look.

product_attribute_model_link/models/product_attribute.py
```diff
@@ -0,0 +1,353 @@
...
+        # Call original method
+        records = create.origin(self, vals_list, **kw)
+        if not product_attributes or self.env.context.get(
+            "operation_from_attribute_value"
+        ):
+            return records
+
+        for record in records:
+            for product_attribute in product_attributes:
+                matching_domain_record = product_attribute._filter_domain(record)
```
yelizariev: It can be optimized by filtering all records at once. Imagine that user creates hundred of records via import tool...

```suggestion
        for product_attribute in product_attributes:
            for matching_domain_record in product_attribute._filter_domain(records)
```

product_attribute_model_link/models/product_attribute.py
```diff
@@ -0,0 +1,353 @@
...
+        ):
+            return True
+
+        # Modify linked product attribute value
+        for product_attribute_value in product_attribute_values:
+            product_attribute_value.with_context(operation_from_record=True).write(
+                {
+                    "name": self[
+                        product_attribute_value.attribute_id.linked_field_id.name
+                    ]
```
yelizariev: Can `self` be multi record set?

product_attribute_model_link/models/product_attribute.py
```diff
@@ -0,0 +1,353 @@
...
+
+
+def make_unlink():
+    """Instantiate an unlink method that unlinks linked product attribute values
+    or deactivates them if they are associated with a product."""
+
+    def unlink(self, **kwargs):
+        if not self.env.context.get("operation_from_attribute_value"):
+            product_attribute_values = self.env["product.attribute.value"].search(
+                [("linked_record_ref", "=", f"{self._name},{self.id}")]
```
yelizariev: `self` could be multi record set?

product_attribute_model_link/models/product_attribute_value.py
```diff
@@ -0,0 +1,191 @@
...
+                        _("Validation Error: ") + error_message
+                    ) from e
+        # Create the records using the super method
+        res = super().create(vals_list)
+        for product_attribute_value in res:
+            # Create related product variants if apply_to_products_on_create is True
+            if (
+                product_attribute_value.attribute_id.apply_to_products_on_create
+                and not self.env.context.get("operation_from_record")
+            ):
```
yelizariev: You can decrease indents

```suggestion
        res = super().create(vals_list)
        if self.env.context.get("operation_from_record"):
            return res
        for product_attribute_value in res.filtered("attribute_id.apply_to_products_on_create"):
            # Create related product variants if apply_to_products_on_create is True
```

product_attribute_model_link/models/product_attribute_value.py
```diff
@@ -0,0 +1,191 @@
...
+                                product_attribute_value.attribute_id.id,
+                            )
+                        ]
+                    )
+                    .with_context(active_test=False)
+                )
+                if product_template_attribute_lines:
+                    for (
+                        product_template_attribute_line
+                    ) in product_template_attribute_lines:
```
yelizariev: if-check is not needed.

Also, variables name could be simplified,

```suggestion
                for line in product_template_attribute_lines:
                    line.write(...)
```

product_attribute_model_link/models/product_attribute_value.py
```diff
@@ -0,0 +1,191 @@
...
+        :raise: ValidationError if conversion fails for a field.
+        """
+        res = super().write(vals)
+        # Modify linked record if modify_from_attribute_values is True
+        if (
+            (vals.get("name") and not vals.get("linked_record_ref"))
+            and self.linked_record_ref
+            and self.attribute_id.create_from_attribute_values
+            and self.attribute_id.modify_from_attribute_values
+            and not self.env.context.get("operation_from_record")
```
yelizariev: It's better to first make check that don't require reading from DB
 
```suggestion
            not self.env.context.get("operation_from_record")
            and (vals.get("name") and not vals.get("linked_record_ref"))
            and self.linked_record_ref
            and self.attribute_id.create_from_attribute_values
            and self.attribute_id.modify_from_attribute_values
```

product_attribute_model_link/models/product_attribute_value.py
```diff
@@ -0,0 +1,191 @@
...
+        ):
+            linked_field = self.attribute_id.linked_field_id
```
yelizariev: `self` can be multi record set ?

product_attribute_model_link/models/product_attribute_value.py
```diff
@@ -0,0 +1,191 @@
...
+                ).unlink()
+        return super().unlink()
+
+    def action_open_linked_record_wizard(self):
+        """
+        Action to open a wizard for the linked record.
+
+        :return: Action dictionary to open the wizard.
+        """
+        context = {
```
yelizariev, line 175: Maybe add `self.ensure_one()` ?

product_attribute_model_link/static/description/index.html
```diff
@@ -0,0 +1,124 @@
...
+
+<section class="oe_container">
+    <div class="oe_row oe_spaced">
+        <div class="oe_span12">
+            <h2 class="oe_slogan">Configuration</h2>
+        </div>
+        <div class="oe_span6">
+            <p class="oe_mt32">To configure this module, you need to:
+                <ul>
+                    <li>...</li>
```
yelizariev, line 40: ... ?

angelinaanaki: Thanks for the review. Any thoughts on why it's not filled in the pre-commit? Tried a few times, but it's not working. Appreciate your help!

product_attribute_model_link/models/product_attribute.py
```diff
@@ -0,0 +1,359 @@
...
+            "operation_from_attribute_value"
+        ):
+            return True
+
+        # Modify linked product attribute value for each record in the set
+        for record in self:
+            record_values = product_attribute_values.filtered(
+                lambda x: x.linked_record_ref == record
+            )
+            for product_attribute_value in record_values:
```
yelizariev: It can be simplified to a single for-loop

```suggestion
        for product_attribute_value in product_attribute_values:
            record = product_attribute_value.linked_record_ref
```

product_attribute_model_link/models/product_attribute_value.py
```diff
@@ -0,0 +1,195 @@
...
+                    ]
+                )
+                .with_context(active_test=False)
+            )
+            for line in product_template_attribute_lines:
+                line.write(
+                    {
+                        "value_ids": [(4, product_attribute_value.id)],
+                    }
+                )
```
yelizariev: It can be even shorter:

```suggestion
            product_template_attribute_lines.write(
                    {
                        "value_ids": [(4, product_attribute_value.id)],
                    }
                )
```

DemchukM (approved):
LGTM

product_attribute_model_link/models/product_attribute_value.py
```diff
@@ -0,0 +1,194 @@
...
+        :return: Created records.
+        :raise: ValidationError if conversion fails for a field.
+        """
+        for vals in vals_list:
+            attribute_id = (
+                vals.get("attribute_id")
+                if vals.get("attribute_id")
+                else self.env.context.get("default_attribute_id")
+            )
+            product_attribute = self.env["product.attribute"].browse(attribute_id)
```
GabbasovDinar: It is possible to put it outside the loop so that the model is not accessed inside the loop:

```python
product_attribute_obj = self.env["product.attribute"]
```

product_attribute_model_link/models/product_attribute_value.py
```diff
@@ -0,0 +1,194 @@
...
+    @api.model_create_multi
+    def create(self, vals_list):
+        """
+        Override to create records based on attribute values.
+
+        :param vals_list: List of dictionaries containing attribute values for creating records.
+        :return: Created records.
+        :raise: ValidationError if conversion fails for a field.
+        """
+        for vals in vals_list:
```
GabbasovDinar, line 44: It seems that the method is too complicated, maybe it makes sense to do it in a separate method, which will also improve the possibility of extension

GabbasovDinar (approved):

ivs-cetmix (approved):
Functional LGTM

product_attribute_model_link/models/product_attribute.py
```diff
@@ -0,0 +1,352 @@
...
+        patched_models = []
+
+        # Retrieve the models that were patched during _register_hook
+        for product_attribute in self.search(
+            [("linked_model_id", "!=", False), ("linked_field_id", "!=", False)]
+        ):
+            patched_models.append(product_attribute.linked_model_id.model)
+
+        for Model in self.env.registry.values():
+            if Model._name in patched_models:
```
yelizariev: registry contains all the models. Why don't iterate over `patched_models` instead?

```suggestion
        for model in patched_models:
            Model = self.env.registry[model]
```

yelizariev (approved):

ivs-cetmix:
Hi @OCA/product-maintainers would appreciate someone having a look at this promising PR 😄

PR #1492, opened 2024-01-09
[14.0][IMP][product_tier_validation] actions for many products to reject and validate

Kev-Roche:
add action for multiple product template validation / reject

### DIFF
--- a/product_tier_validation/i18n/fr.po
+++ b/product_tier_validation/i18n/fr.po
@@ -1,0 +1,137 @@
+# Translation of Odoo Server.
+# This file contains the translation of the following modules:
+#   * product_tier_validation
+#
+msgid ""
+msgstr ""
+"Project-Id-Version: Odoo Server 14.0\n"
+"Report-Msgid-Bugs-To: \n"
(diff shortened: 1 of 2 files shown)


sebastienbeau (approved):
LGTM (code review)

bosd (approved):

legalsylvain:
@Kev-Roche. FYI, translation should be handled via OCA weblate.

/ocabot merge patch

Kev-Roche:
> @Kev-Roche. FYI, translation should be handled via OCA weblate.
> 
> /ocabot merge patch

@legalsylvain ho ok, also if it's new terms to translate ? If I understand correctly, we have no way to have together a new function / term to translate and its translation at the same time ?

legalsylvain:
/ocabot merge patch


@legalsylvain ho ok, also if it's new terms to translate ? If I understand correctly, we have no way to have together a new function / term to translate and its translation at the same time ?

No that's a current annoying limitation of the OCA translation system.

Note : could you port the changes upstream ?

PR #1493, opened 2024-01-09
[17.0][MIG] product_set: Migration to 17.0

NICO-SOLUTIONS:
standard migration to 17.0


cormaza (approved):
LGTM

product_set/tests/test_product_set.py
```diff
@@ -10,9 +10,6 @@ class TestProductSet(common.TransactionCase):
     @classmethod
     def setUpClass(cls):
         super().setUpClass()
-        cls.env = cls.env(context=dict(cls.env.context, tracking_disable=True))
```
rousseldenis: @NICO-SOLUTIONS Why removing those lines ?

product_set/models/product_set.py
```diff
@@ -32,14 +32,20 @@ class ProductSet(models.Model):
         "it's going to be available for all of them.",
     )
 
+    display_name = fields.Char(
+        compute="_compute_display_name", string="Display Name", store=True
+    )
+
     def name_get(self):
-        return [(rec.id, rec._name_get()) for rec in self]
+        return [(rec.id, rec.display_name) for rec in self]
```
rousseldenis: If you implement a display_name, you don't need to override name_get() (moreover deprecated in 17.0).

See : https://github.com/odoo/odoo/blob/17.0/odoo/models.py#L1676

product_set/views/product_set.xml
```diff
@@ -24,7 +24,7 @@
                         name="web_ribbon"
                         title="Archived"
                         bg_color="bg-danger"
-                        attrs="{'invisible': [('active', '=', True)]}"
+                        invisible="active == 'True'"
```
rousseldenis: You can use 'active' only

rousseldenis (changes requested):

product_set/views/product_set.xml
NICO-SOLUTIONS: thx

product_set/models/product_set.py
NICO-SOLUTIONS: true... thx

product_set/tests/test_product_set.py
NICO-SOLUTIONS: there is no dependency to sale module in manifest. This test will be done in sale_product_set. So i guess this i not necessary here?!

MohamedOsman7:
Code LGTM (however the functionality cannot be tested because of the runboat)

NICO-SOLUTIONS:
@MohamedOsman7 
now you can check th erunboat

MohamedOsman7 (approved):
LGTM

pedrobaeza:
This has been improved recently in v16, and it's worth to include that changes: #1548

NICO-SOLUTIONS:
@pedrobaeza  
I will take care of it soon. Thanks for the information

NICO-SOLUTIONS:
super seeded by PR1554

PR #1504, opened 2024-01-29
[15.0][IMP] `product_template_tag`: add parent tags

SilvioC2C:


product_template_tags/models/product_template_tag.py
```diff
@@ -27,6 +28,8 @@ class ProductTemplateTag(models.Model):
         string="Company",
         default=lambda self: self.env.company,
     )
+    parent_id = fields.Many2one("product.template.tag")
```
ivantodorovich: I suggest:
* `index=True`
* `ondelete='cascade'`

ivantodorovich: Also, for all record hierarchies, it's recommended to use `_parent_store = True` on the model with the field:

```py
parent_path = fields.Char(index=True, unaccent=False)
```

e.g: https://github.com/odoo/odoo/blob/d356abfec0219d0ce4b8ec137cd4f13ef148009f/odoo/addons/base/models/res_partner.py#L113

SilvioC2C: @ivantodorovich thanks, I'll take care of this.

BTW I'm doing the same for purchase tags as well (see https://github.com/OCA/purchase-workflow/pull/2161) so I'll add these suggestions there as well!

SilvioC2C: PS: I didn't add `unaccent=False` since it's not supported until v16

ivantodorovich (approved):
Thanks!

gurneyalex (approved):

PR #1506, opened 2024-01-30
[14.0][FIX]product_supplierinfo_for_customer: properly filter customers

PicchiSeba:
Properly filter customers.

Together with https://github.com/OCA/sale-workflow/pull/2934 fixes broken tests for `sale-workflow/product_supplierinfo_for_customer_sale`

### DIFF
--- a/product_supplierinfo_for_customer/models/product_product.py
+++ b/product_supplierinfo_for_customer/models/product_product.py
@@ -169,7 +169,10 @@ class ProductProduct(models.Model):
     def _select_customerinfo(
         self, partner=False, quantity=0.0, date=None, uom_id=False, params=False
     ):
-        """Customer version of the standard `_select_seller`."""
+        """
+        Customer version of the standard `_select_seller`.
+        If you want not to filter by quantity, explicitly pass quantity=None
+        """
         self.ensure_one()
         if not params:
             params = {}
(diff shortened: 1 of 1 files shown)


HekkiMelody (approved):
Code review

HekkiMelody:
This PR makes it so this module mirrors the functioning of `_select_sellers` better than before by accepting `quantity=None` as an argument in order to bypass filtering by quantity entirely

GSLabIt (approved):

product_supplierinfo_for_customer/models/product_product.py
```diff
@@ -181,16 +184,18 @@ def _select_customerinfo(
...
         )
-        res = self._customers_filter_by_quantity(
-            customers, quantity=quantity, uom_id=uom_id, precision=precision
-        )
-        if res:
-            customer_name = res[0].name
-            res = res.filtered(lambda x, name=customer_name: x.name == name)
+
+        if quantity is not None:
+            customers = self._customers_filter_by_quantity(
```
pedrobaeza, line 189: Wrong variable names lead to later confusions:

```suggestion
            customerinfos = self._customers_filter_by_quantity(
```

and instead of `customer_name` put simply `customer`, as it represents a res.partner.

HekkiMelody: Thanks for the feedback!

And, you're right. We'll make this change in another PR

PR #1509, opened 2024-02-06
[14.0][FIX] product_supplierinfo_for_customer: get customerinfo price quantity

renda-dev:
`_get_price_from_customerinfo` doesn't properly work after the fixes made in https://github.com/OCA/product-attribute/pull/1506

With the recent changes, `_select_customerinfo` uses the `quantity` parameter properly, which is not expected by `_get_price_from_customerinfo`

`price_compute` also rely on this, and doesn't work as expected right now

This also fixes tests in https://github.com/OCA/e-commerce/pull/863


PicchiSeba (approved):
Nice catch! LGTM

HekkiMelody (approved):
Code review

PR #1512, opened 2024-02-09
[16.0] MIG base_product_mass_addition

legalsylvain:
This PR try to finish the migration of base_product_mass_addition.
Previsous work done by @aiendry-aktivsoftware in https://github.com/OCA/product-attribute/pull/1393 that has no activity since July 2023.

- :heavy_check_mark: This commit a1d3eac8770c92a2e9a1fe7cb3d278ebea5f0bc1 fixes the new V16 error _"UserWarning: Field product.product.quick_uom_id in dependency of product.product.quick_uom_category_id should be searchable. This is necessary to determine which records to recompute when uom.uom.category_id is modified. You should either make the field searchable, or simplify the field dependency"_. (alternative to https://github.com/OCA/product-attribute/pull/1393#discussion_r1360789201)

- :heavy_check_mark: The module works, and can be tested on runboat here https://github.com/OCA/purchase-workflow/pull/2171 that migrate purchase_quick in V16.

- :-1: In the meantime, I didn't find a way to avoid the change of LOG_ACCESS_COLUMNS, when writing on the pseudo fields qty_to_process / quick_uom_id. 
    - This feature was introduced by @ivantodorovich in V14 here : https://github.com/OCA/product-attribute/pull/1065
    - Current implementation looks dangerous. See https://github.com/OCA/product-attribute/pull/1393#discussion_r1360928228 by @pedrobaeza 
    - I tried to overload write function / modified function / change boolean _log_access without success.

@ivantodorovich : Do you have a time to take a look on how to reimplement : #1065 ?

Thanks !


base_product_mass_addition/models/product_product.py
```diff
@@ -0,0 +1,114 @@
...
+            and any(quick_fname in fnames for quick_fname in quick_fnames)
+        ):
+            for record in self.filtered("id"):
+                towrite = self.env.all.towrite[self._name]
+                vals = towrite[record.id]
+                if not vals:  # pragma: no cover
+                    continue
+                if all(fname in LOG_ACCESS_COLUMNS for fname in vals.keys()):
+                    towrite.pop(record.id)
+        return super().modified(fnames, create=create, before=before)
```
ivantodorovich: I take this code doesn't work anymore because of the removal of `towrite`, right?
I see this has been refactored in https://github.com/odoo/odoo/commit/384fda2c2adaf1a14942197f0a96b182b5db0345

I believe we can keep the same feature, with roughly the same idea, with the necessary changes to adapt for that refactor (using `get_dirty_fields` and `clear_dirty_field` instead of directly accessing `towrite`).

⚠️ I haven't tested this, but I imagine something like this


```suggestion
        quick_fnames = ("qty_to_process", "quick_uom_id")
        if (
            self
            and fnames
            and any(quick_fname in fnames for quick_fname in quick_fnames)
        ):
            if all(field.name in LOG_ACCESS_COLUMNS for field in self.env.cache.get_dirty_fields()):
                for fname in LOG_ACCESS_COLUMNS:
                    self.env.cache.clear_dirty_field(self._fields[fname])
        return super().modified(fnames, create=create, before=before)
```

The previous implementation did this at a record-level: it conditionally cleared the write cache record by record.

The new suggested implementation would do it at the model level: clearing the cache conditionally if only the log access fields are being updated. It's not as smart as the previous one, but it might work in our specific use case (we won't be updating other product fields anyway)

⚠️ again, not tested

In any case, we could re-implement the exact same behavior if we need to, but we'll have to manipulate the `self.env.cache._dirty` and `self.env.cache._data` directly IMO

legalsylvain: Hi @ivantodorovich. Thanks a lot for your time. (analysis and code proposal)
I tested your implementation (see last fixup commit).

- positive point : it works ! I mean, with purchase_quick installed, if change the field qty_to_process field, it is writing on pos.order.line, but doesn't change write_date field :+1: 
- negative point : the test fails. see https://github.com/OCA/product-attribute/actions/runs/7972609836/job/21764713992?pr=1512#step:8:518 :-1: 

Any idea ? 

```
   File "/__w/product-attribute/product-attribute/base_product_mass_addition/tests/test_product_mass_addition.py", line 55, in test_quick_should_not_write_on_product
    self.assertEqual(self.product.write_uid, user_demo)
AssertionError: res.users(1,) != res.users(6,)
```

ivantodorovich: Hmm, I would put a breakpoint there: 

```
            if all(
                field.name in LOG_ACCESS_COLUMNS
                for field in self.env.cache.get_dirty_fields()
            )
```

And then check why the condition is not matched.

It could be that there's an extra dirty fields in there, during tests. In which case we might need to add some explicit flushes in tests.

---

Alternatively, it may be due to the simplification of the new implementation:

> It's not as smart as the previous one, but it might work in our specific use case (we won't be updating other product fields anyway)

Maybe there are other dirty products in the cache. In which case the solution might be also to flush during tests, or improve the code to restore the previous record-by-record dirty check -- but to do that we'd have to mess with the cache internals as there's no public api for this AFAICS

legalsylvain: > And then check why the condition is not matched.

i put a breakpoint, and the call of clear_dirty_field is done. 
So, self.env.cache.get_dirty_fields() is [write_uid, write_date] at the beginning of the call of ``modified`` and when super is executed, it is []. That looks correct. _However_ via test, it doesn't work.

ivantodorovich: >  However via test, it doesn't work.

What is the output of `self.env.cache.get_dirty_fields()` during tests, on that breakpoint? still `[write_uid, write_date]`? 🤔 

---


Alternatively, try adding a flush after this line: https://github.com/grap/product-attribute/blob/8fd402d307148407b9bee52d754ff4c8daf77197/base_product_mass_addition/tests/test_product_mass_addition.py#L52

Perhaps the issue is there, in the unit tests. As we might be dropping the cache modified by the [previous line](https://github.com/grap/product-attribute/blob/8fd402d307148407b9bee52d754ff4c8daf77197/base_product_mass_addition/tests/test_product_mass_addition.py#L51)

FranzPoize: Hi as said in the original PR (https://github.com/OCA/product-attribute/pull/1393#issuecomment-1766841797) the tests does not work because it is not possible to change the value of all log_access field.

I wrote a PR to your branch @legalsylvain to fix the test https://github.com/grap/product-attribute/pull/2

FranzPoize (approved):

legalsylvain:
@bealdav Could you review this one ? As you need it for sale_quick ? Thanks !

bealdav (approved):
LGTM

bealdav:
Thanks a lot @legalsylvain for your decisive implication on this topic

legalsylvain:
> Thanks a lot @legalsylvain for your decisive implication on this topic

you're welcome ! 

Note : I have updated readme, and I have adopted the module in the last commits.

I think this PR could be merged, CC : @OCA/product-maintainers

PR #1518, opened 2024-02-14
[16.0][FIX] product_packaging_level: Preserve sequence order

p-tombez:
Restore original order to prevent altering behavior when packaging is reordered in the view.
(see explanation here : https://github.com/OCA/sale-workflow/pull/2939 )

### DIFF
--- a/product_packaging_level/models/product_packaging.py
+++ b/product_packaging_level/models/product_packaging.py
@@ -8,8 +8,8 @@ from odoo.exceptions import ValidationError
 
 class ProductPackaging(models.Model):
     _inherit = "product.packaging"
-    _order = "product_id, level_sequence"
 
+    sequence = fields.Integer(compute="_compute_sequence", store=True)
     packaging_level_id = fields.Many2one(
         "product.packaging.level",
         required=True,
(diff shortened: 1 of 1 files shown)


StephaneMangin (approved):

santostelmo (approved):

product_packaging_level/models/product_packaging.py
```diff
@@ -8,8 +8,8 @@
 
 class ProductPackaging(models.Model):
     _inherit = "product.packaging"
-    _order = "product_id, level_sequence"
 
+    sequence = fields.Integer(compute="_compute_sequence", store=True)
```
simahawk, line 12: why not related as the prev field?

simahawk:
Tests fail unrelated on `product_abc_classification`
https://github.com/OCA/product-attribute/actions/runs/7899614130/job/21559449400?pr=1518#step:8:518

product_packaging_level/models/product_packaging.py
santostelmo, line 12:
```suggestion
    sequence = fields.Integer(compute="_compute_sequence", store=True, readonly=False)
```
`readonly=False` is need otherwise is not possible to edit the field.

santostelmo (changes requested):
We need to assign readonly=False in sequence field otherwise field value can't be changed in the interface.

santostelmo:
@p-tombez I created another PR with the fix mentioned above https://github.com/OCA/product-attribute/pull/1525

Issue #1513, opened 2024-02-09
V16 branch looks broken


legalsylvain:
hi all.

Since recent changes, V16.0 looks broken : 
https://github.com/OCA/product-attribute/actions/workflows/test.yml?query=branch%3A16.0

So, all the new PR against V16 branch are failing : 
- https://github.com/OCA/product-attribute/pull/1512
- https://github.com/OCA/product-attribute/pull/1511
- https://github.com/OCA/product-attribute/pull/1505 by @SilvioC2C 
- https://github.com/OCA/product-attribute/pull/1510 by @ChrisOForgeFlow 

I dig a little bit on the logs, but did'nt found the origin of the problem.

@OCA/product-maintainers : could you take a look ? @jjscarafia, @JordiBForgeFlow, @rousseldenis, @gurneyalex

Thanks ! 

Note : BTW, there are also weird warning, related to incorrect call of ``digits``, but I don't understand why...

``2024-01-29 11:40:27,563 425 WARNING odoo odoo.fields: Field wizard.preview.pricelist.line.price: unknown parameter 'digits', if this is an actual parameter you may want to override the method _valid_field_parameter on the relevant model in order to allow it ``

simahawk:
The warning has been fixed in #1521 .
The error seems to come from `abc_classification_product_level`
[OCA/product-attribute/actions/runs/7955666562/job/21715009413?pr=1521#step:8:480](https://github.com/OCA/product-attribute/actions/runs/7955666562/job/21715009413?pr=1521#step:8:480)

simahawk:
Plenty of errors in fact... I've tried to move on in #1521

legalsylvain:
Thanks !

PR #1521, opened 2024-02-19
[16.0] Fix build

simahawk:
Tackling some issues from #1513

### DIFF
--- a/product_abc_classification/models/abc_classification_profile.py
+++ b/product_abc_classification/models/abc_classification_profile.py
@@ -128,10 +128,8 @@ class AbcClassificationProfile(models.Model):
                 },
             )
             level_ids.extend(r[0] for r in self.env.cr.fetchall())
-        self.env["abc.classification.product.level"].invalidate_cache(
-            ["manual_level_id"], level_ids
-        )
         modified_levels = self.env["abc.classification.product.level"].browse(level_ids)
+        modified_levels.invalidate_recordset(fnames=["manual_level_id"])
         # mark field as modified and trigger recompute of dependent fields.
         modified_levels.modified(["manual_level_id"])
         modified_levels._recompute_recordset()
--- a/product_abc_classification/tests/test_abc_classification_product_level.py
+++ b/product_abc_classification/tests/test_abc_classification_product_level.py
@@ -4,6 +4,7 @@
 from psycopg2 import IntegrityError
 
 from odoo.exceptions import ValidationError
+from odoo.tools import mute_logger
 
 from .common import ABCClassificationLevelCase
 
@@ -137,6 +138,7 @@ class TestABCClassificationProductLevel(ABCClassificationLevelCase):
         self.assertEqual(self.product_level.level_id, self.classification_level_a)
         self.assertFalse(self.product_level.flag)
 
+    @mute_logger("odoo.sql_db")
     def test_03(self):
         """
         Data:
--- a/product_pricelist_simulation/wizards/wizard_preview_pricelist.py
+++ b/product_pricelist_simulation/wizards/wizard_preview_pricelist.py
@@ -100,7 +100,5 @@ class PricelistSimulationLine(models.TransientModel):
     )
     price = fields.Monetary(
         string="Unit Price",
-        digits="Price",
         readonly=True,
-        currency_field="currency_id",
     )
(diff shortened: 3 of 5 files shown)


legalsylvain:
Thanks !

simahawk:
There's an unrelated error on `abc_classification_product_level`
https://github.com/OCA/product-attribute/actions/runs/7955666562/job/21715009413?pr=1521#step:8:480

legalsylvain:
> There's an unrelated error on abc_classification_product_level

The error is in the module ``product_abc_classification``

@rousseldenis, @lmignon @MiquelRForgeFlow, @lmarion-source

could you take a look ? 
thanks !

santostelmo (approved):

simahawk:
> > There's an unrelated error on abc_classification_product_level
> 
> The error is in the module `product_abc_classification`
> 
> @rousseldenis, @lmignon @MiquelRForgeFlow, @lmarion-source
> 
> could you take a look ? thanks !

I'm on it

simahawk:
Now the issue is on `product_supplierinfo_for_customer`

simahawk:
To me the domain used to look for the customer info is wrong.
Failing test:

image

This means that the selected customer info should be the one w/ price = 100.0 because is the one assigned to the specific variant. However, the selection gives you back the generic price:
```
(Pdb++) product._select_customerinfo(partner=self.customer).price
30.0
```
When I look at the domain it seems wrong
```
Pdb++) product._prepare_domain_customerinfo({"partner_id": self.customer.id})
[('partner_id', '=', 49), '|', ('product_id', '=', 49), '&', ('product_tmpl_id', '=', 32), ('product_id', '=', False)]
```
By logic, I assume we should have an AND w/ partner and product and fallback w/ the OR. However I don't know this module at all...

@pedrobaeza @sergio-teruel @MiquelRForgeFlow @AaronHForgeFlow anyone can provide some feedback?

pedrobaeza:
In a first sight, I don't see the domain incorrect. It's saying that:

- The partner should be 49
- And then, comply with one on these criteria:
  - The product is 49
  - Or the product is empty and the product template is 32, which should be the template for product 49

The only weird thing is that the partner ID and the product ID is the same, but reading the code, it seems legit.

simahawk:
IMO there's something wrong because even if I have 2 records matching the partner and only one of them matches the variant I get the generic one (matching only the tmpl) as a result. Hence, I assume the domain is wrong.

pedrobaeza:
But for that, there's this sort: https://github.com/OCA/product-attribute/blob/16.0/product_supplierinfo_for_customer/models/product_product.py#L138

victoralmau:
IMO the solution for `product_supplierinfo_for_customer` is to set a `sequnce` for each element of `product.customerinfo` (sequence: 10 and sequence: 20 respectively for example).

AaronHForgeFlow:
I think the matching of the variant is more important than the template. Perhaps the sorting is incorrect. Created PR with this: https://github.com/camptocamp/product-attribute/pull/4

pedrobaeza:
@victoralmau see my previous comments. Anyway, you can exercise the solution in one PR if you see it's the proper one.

pedrobaeza:
Fixed in #1528, although the solution doesn't convince me too much I have to say, as this reveals a flaw in the design.

AaronHForgeFlow:
Tested in my local with this: https://github.com/camptocamp/product-attribute/pull/4 and test were not failing

pedrobaeza:
But that's not very optimal in performance. We should try to avoid several filtered.

AaronHForgeFlow:
@pedrobaeza  after @lmignon suggestion it works the same without the use of filtered: https://github.com/camptocamp/product-attribute/pull/4/files

pedrobaeza:
Can you gather all and rebase for getting the proper solution?

AaronHForgeFlow:
> Can you gather all and rebase for getting the proper solution?

Hi @pedrobaeza is this comment in regards my PR? I am afraid I did not get what you mean.

pedrobaeza:
Yes, I mean to put here in this PR (by @simahawk or someone else) the needed things to get such proper code.

simahawk:
I'll cherry-pick https://github.com/camptocamp/product-attribute/pull/4

simahawk:
Are we all fine w/ these fixes?

pedrobaeza (approved):

AaronHForgeFlow (approved):

PR #1527, opened 2024-02-27
[16.0][IMP]product_packaging_level: recover name_get policy flow

sonhd91:
As suggested in: **https://github.com/OCA/product-attribute/pull/1427#discussion_r1500818061**

No need to update `_onchange_package_type` so I keep it

### DIFF
--- a/product_packaging_level/models/product_packaging.py
+++ b/product_packaging_level/models/product_packaging.py
@@ -10,6 +10,7 @@ class ProductPackaging(models.Model):
     _inherit = "product.packaging"
     _order = "product_id, level_sequence"
 
+    name = fields.Char(compute="_compute_name", store=True, readonly=False)
     packaging_level_id = fields.Many2one(
         "product.packaging.level",
         required=True,
@@ -126,11 +127,30 @@ class ProductPackaging(models.Model):
         if package_type_level_id:
             self.packaging_level_id = package_type_level_id
 
-    @api.onchange("packaging_level_id", "package_type_id", "name", "name_policy")
-    def _onchange_name(self):
+    @api.depends(
+        "package_type_id",
+        "name_policy",
+        "packaging_level_id",
+    )
+    def _compute_name(self):
+        for rec in self:
+            rec.name = rec._get_name_from_policy()
+
+    # Keep this method to respect translations on level name
+    @api.depends("product_id", "packaging_level_id", "name_policy")
+    def _compute_display_name(self):
+        for record in self:
+            if record.product_id and record.packaging_level_id:
+                record.display_name = record._get_name_from_policy(lang=self.env.lang)
+            else:
+                return super()._compute_display_name()
+
+    def _get_name_from_policy(self, lang=None):
         new_name = self.name
         if self.name_policy == "by_package_level":
+            lang = lang or self.packaging_level_id.default_lang_id.code
+            self = self.with_context(lang=lang)
             new_name = self.packaging_level_id.display_name
         elif self.name_policy == "by_package_type":
             new_name = self.package_type_id.name
-        self.name = new_name
+        return new_name
--- a/product_packaging_level/models/product_packaging_level.py
+++ b/product_packaging_level/models/product_packaging_level.py
@@ -9,12 +9,24 @@ class ProductPackagingLevel(models.Model):
     _description = "Level management for product.packaging"
     _order = "sequence, code"
 
+    def _default_language(self):
+        lang_code = self.env["ir.default"].get("res.partner", "lang")
+        def_lang_id = self.env["res.lang"]._lang_get_id(lang_code)
+        return def_lang_id or self._active_languages()[0]
+
     name = fields.Char(required=True, translate=True)
     code = fields.Char(required=True)
     sequence = fields.Integer(required=True)
     has_gtin = fields.Boolean()
     active = fields.Boolean(default=True)
     is_default = fields.Boolean()
+    default_lang_id = fields.Many2one(
+        "res.lang",
+        string="Default Language",
+        default=lambda self: self._default_language(),
+        required=True,
+    )
+
     name_policy = fields.Selection(
         selection=[
             ("by_package_level", "Package Level Name"),
--- a/product_packaging_level/views/product_packaging_level_view.xml
+++ b/product_packaging_level/views/product_packaging_level_view.xml
@@ -30,6 +30,10 @@
                             <field name="code" />
                             <field name="has_gtin" />
                             <field name="name_policy" />
+                            <field
+                                name="default_lang_id"
+                                help="Language option for packaging names used in case policy names is packaging by levels"
+                            />
                         </group>
                         <group name="misc">
                             <field name="sequence" />
--- a/product_packaging_level/tests/test_product_packaging_level.py
+++ b/product_packaging_level/tests/test_product_packaging_level.py
@@ -8,6 +8,8 @@ class TestProductPackagingLevel(common.TransactionCase):
     @classmethod
     def setUpClass(cls):
         super().setUpClass()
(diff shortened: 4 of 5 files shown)


product_packaging_level/models/product_packaging.py
```diff
@@ -126,11 +127,27 @@ def _onchange_package_type(self):
...
+    )
+    def _compute_name(self):
+        for rec in self:
+            new_name = rec.name
+            if rec.name_policy == "by_package_level":
+                new_name = rec.packaging_level_id.display_name
+            elif rec.name_policy == "by_package_type":
+                new_name = rec.package_type_id.name
+            rec.name = new_name
+
```
lmignon: Why do you want to set a name that will not be the same into the database depending on user's lang in use at record create?

sonhd91: I got your points, but since we have set `required=True` for the `name` field, we still need this compute method to store the value into the database. Do you have any suggestions ?

sonhd91: @lmignon branch updated, could you please take a look ?

product_packaging_level/models/product_packaging.py
```diff
@@ -126,11 +127,28 @@ def _onchange_package_type(self):
...
+    # Keep this method to respect translations on level name
+    @api.depends("product_id", "packaging_level_id", "name_policy")
+    def _compute_display_name(self):
+        for record in self:
+            if record.product_id and record.packaging_level_id:
+                record.display_name = record.change_name_policy()
+            else:
+                return super()._compute_display_name()
+
+    def change_name_policy(self):
```
lmignon:
```suggestion
    def _get_name_from_policy(self):
```

lmignon (commented):
@sonhd91 A minor comment... 

Since the name is mandatory and not translatable, to avoid mismatch according to the user language when the package level is created, you should maybe add an additional parameter to select the language to use to compute the name.

sonhd91:
> @sonhd91 A minor comment...
> 
> Since the name is mandatory and not translatable, to avoid mismatch according to the user language when the package level is created, you should maybe add an additional parameter to select the language to use to compute the name.

Hello @lmignon, the `_get_name_from_policy` calculates the name packaging by using the `display_name` which include language context for computation, so it will take the value in the user's language context at the packaging creation step

product_packaging_level/models/product_packaging.py
simahawk: I don't get it if this issue is solved... if it depends on the lang it should not be stored UNLESS you make it translatable too. No?

simahawk:
> > @sonhd91 A minor comment...
> > Since the name is mandatory and not translatable, to avoid mismatch according to the user language when the package level is created, you should maybe add an additional parameter to select the language to use to compute the name.
> 
> Hello @lmignon, the `_get_name_from_policy` calculates the name packaging by using the `display_name` which include language context for computation, so it will take the value in the user's language context at the packaging creation step

yeah but then you should use `@api.depends_context("lang")` no? (even if only to document it)

lmignon:
> > > @sonhd91 A minor comment...
> > > Since the name is mandatory and not translatable, to avoid mismatch according to the user language when the package level is created, you should maybe add an additional parameter to select the language to use to compute the name.
> > 
> > 
> > Hello @lmignon, the `_get_name_from_policy` calculates the name packaging by using the `display_name` which include language context for computation, so it will take the value in the user's language context at the packaging creation step
> 
> yeah but then you should use `@api.depends_context("lang")` no? (even if only to document it)

My main concern is to ensure to store into the database the value always in the same language..

sonhd91:
> > > @sonhd91 A minor comment...
> > > Since the name is mandatory and not translatable, to avoid mismatch according to the user language when the package level is created, you should maybe add an additional parameter to select the language to use to compute the name.
> > 
> > 
> > Hello @lmignon, the `_get_name_from_policy` calculates the name packaging by using the `display_name` which include language context for computation, so it will take the value in the user's language context at the packaging creation step
> 
> yeah but then you should use `@api.depends_context("lang")` no? (even if only to document it)


Hello,  I just updated the branch please checked again

product_packaging_level/models/product_packaging.py
```diff
@@ -126,11 +127,29 @@ def _onchange_package_type(self):
...
             self.packaging_level_id = package_type_level_id
 
-    @api.onchange("packaging_level_id", "package_type_id", "name", "name_policy")
-    def _onchange_name(self):
+    @api.depends(
+        "package_type_id",
+        "name_policy",
+        "packaging_level_id",
+    )
+    @api.depends_context("lang")
```
lmignon: IMO it's a nonsense. You don't want to have this field recomputed every time the user change the language. What I expect is to always have the same value saved into the database whatever the user's language.

lmignon (changes requested):

product_packaging_level/models/product_packaging.py
simahawk: > My main concern is to ensure to store into the database the value always in the same language..

In fact, if we want to make sure about this, we should enforce the language to be always the default one.
Now, I'm not sure why we want to store this information... is it too expensive to compute?

sonhd91: Branch updated, please check again, packaging `name` value that stored in database will based on from `default_language` setted on `product.package.level` so user'slanguage context would not affected them

cyrilmanuel:
@lmignon  little ping, is it good for you ?

lmignon (approved):

product_packaging_level/models/product_packaging_level.py
```diff
@@ -9,12 +9,21 @@ class ProductPackagingLevel(models.Model):
...
+        return def_lang_id or self._active_languages()[0]
+
     name = fields.Char(required=True, translate=True)
     code = fields.Char(required=True)
     sequence = fields.Integer(required=True)
     has_gtin = fields.Boolean()
     active = fields.Boolean(default=True)
     is_default = fields.Boolean()
+    default_lang_id = fields.Many2one(
+        "res.lang", string="Default Language", default=_default_language, required=True
```
simahawk:
```suggestion
        "res.lang", string="Default Language", default=lambda self : self._default_language(), required=True
```

product_packaging_level/views/product_packaging_level_view.xml
```diff
@@ -30,6 +30,10 @@
                             <field name="code" />
                             <field name="has_gtin" />
                             <field name="name_policy" />
+                            <field
+                                name="default_lang_id"
+                                help="Language option for packaging name"
```
simahawk: This help txt is not so clear. You should clarify what is used for.

product_packaging_level/models/product_packaging.py
```diff
@@ -126,11 +127,34 @@ def _onchange_package_type(self):
...
         new_name = self.name
         if self.name_policy == "by_package_level":
-            new_name = self.packaging_level_id.display_name
+            if lang:
+                self = self.with_context(lang=lang)
+                new_name = self.packaging_level_id.display_name
+            else:
+                new_name = self.packaging_level_id.with_context(
+                    lang=self.packaging_level_id.default_lang_id.code
+                ).display_name
```
simahawk:
```suggestion
            lang = lang or self.packaging_level_id.default_lang_id.code
            self = self.with_context(lang=lang)
            new_name = self.packaging_level_id.display_name
```

simahawk (changes requested):

product_packaging_level/models/product_packaging.py
sonhd91: @simahawk , branch updated, please check again

sonhd91: Update test cases

simahawk (approved):

PR #1529, opened 2024-02-28
[ADD][16.0] product_supplierinfo_packaging

benwillig:
This new module allows to use packaging on a supplier info. The mechanism is the same as the purchase order, when we assign a minimum quantity, the packaging quantity is computed and vice versa.


rousseldenis:
@benwillig Thanks for this. Is it a split from this https://github.com/OCA/stock-logistics-warehouse/tree/10.0/purchase_packaging ?

benwillig:
@rousseldenis No it's not a split. I wasn't aware of this module but yes there could be submodules to implement the rest of missing features from purchase_packaging.

jbaudoux (approved):
Thanks. Looks useful module !

benwillig:
@rousseldenis I just remembered I also did this one https://github.com/OCA/purchase-workflow/pull/2188

product_supplierinfo_packaging/models/product_supplierinfo.py
```diff
@@ -0,0 +1,161 @@
...
+class ProductSupplierInfo(models.Model):
+    _inherit = "product.supplierinfo"
+
+    packaging_id = fields.Many2one(
+        comodel_name="product.packaging",
+        ondelete="restrict",
+    )
+    packaging_qty = fields.Float(
+        compute="_compute_packaging_qty",
+    )
```
lmignon, line 19: Why not a related?  `related="packaging_id.qty"`

lmignon (changes requested):
A useful one. Thank you @benwillig A little comment.

product_supplierinfo_packaging/models/product_supplierinfo.py
benwillig: Only for inheritance purpose

PR #1535, opened 2024-03-04
[16.0][FIX] product_supplierinfo_for_customer: Avoid returning None

hildickethan:
Fix for errors in `sale-workflow` repo 
```
  File "/__w/sale-workflow/sale-workflow/product_supplierinfo_for_customer_sale/models/sale_order_line.py", line 23, in _compute_product_customer_code
    code = supplierinfo.product_code
AttributeError: 'NoneType' object has no attribute 'product_code'
```


simahawk:
@hildickethan-S73 thanks for your contrib!
I'm sorry but we are already merging https://github.com/OCA/product-attribute/pull/1537 because we missed this PR.

PR #1538, opened 2024-03-06
[FIX] product_cost_security: ORM-level security

yajo:
Before this patch, inheriting modules had to manually add some logic related to guessing wether the user can access cost information. Also, smartypants users could still obtain the desired information through API calls.

Now:
- You can't read/write cost fields via API anymore.
- Views automatically set those fields to readonly if the user has only read permissions (and the model inherits from the new mixin).
- You don't need to enable debug mode anymore to follow configuration instructions.
- Instructions improved.

@moduon MT-5158


Shide (approved):
Code LGTM

rafaelbn (commented):
@yajo ,

- [ ] As a Inventory "user" with our permisions in sales, purchase, etc., I get "Access Error" in common operations

> You don't have permission to update product costs. Allowed group: Product costs / Modify product costs

imagen

yajo:
Oops... I forgot to drop the security on sudo mode. Fixed now, please review again.

yajo:
Now https://github.com/OCA/margin-analysis/pull/198 depends on this PR to also unify the permissions interface with `sale_margin_security` and `sale_margin_delivered_security`.

amarcosg (approved):
Code review LGTM

PR #1540, opened 2024-03-08
[17.0][MIG] product assortment

WorrapongOngsakul:
Removal of "attrs" and "states" Attributes


pedrobaeza:
Another proposal is #1587, but yours have preference being the first proposed, unless you say the other is better. Can you please squash last 2 commits together?

pedrobaeza:
Please include #1682 as well

rousseldenis:
@WorrapongOngsakul

PR #1458, opened 2023-11-21
[17.0][MIG] product_state: Migration to 17.0

NICO-SOLUTIONS:
Standard Migration to 17.0

PR #1543, opened 2024-03-09
[17.0][mig] product state

mostafabarmshory:


fardinmardani (approved):
All things are Ok

mostafabarmshory:
@fardinmardani 

Here is the same one. Would you check and write review on it, please?

https://github.com/OCA/product-attribute/pull/1458

fardinmardani:
> @fardinmardani
> 
> Here is the same one. Would you check and write review on it, please?
> 
> #1458

Yes, you are right

pedrobaeza:
Migration already existing: #1458, so closing

PR #1545, opened 2024-03-12
[16.0][IMP] product_packaging_container_deposit: Add more context for update order container deposit quantity

QuocDuong1306:


product_packaging_container_deposit/models/container_deposit_order_mixin.py
```diff
@@ -86,7 +86,9 @@ def update_order_container_deposit_quantity(self):
                         product, deposit_container_qties[product][0]
                     )
                     values_lst.append(Command.create(values))
-            order.write({self._get_order_line_field(): values_lst})
+            order.with_context(update_order_container_deposit_quantity=True).write(
```
simahawk: there's already the key `skip_update_container_deposit` at line 36: do you really need this?

nilshamerlinck: not exactly the same purpose:
- skip_update_container_deposit: avoid update on copy()
- update_order_container_deposit_quantity: flag to circumvent the fields protection

simahawk: Fair enough, but if you keep it, can you move it to the same place of the one above to not change ctx on _every_ record?

QuocDuong1306: Thank @nilshamerlinck and @simahawk , I updated

simahawk (approved):

simahawk:
@QuocDuong1306 the msg could be shorter and more explicit: `[IMP] product_packaging_container_deposit: add ctx key for qty update`

QuocDuong1306:
Thank @simahawk , done here

PR #1548, opened 2024-03-12
[16.0][IMP] product_set: Move the base logic of the transient model defined in the module sale_product_set.

pilarvargas-tecnativa:
The reason to move this logic is that there are other modules that extend product_set as for example stock_product_set but at the same time make use of the transient model to define a wizard. For this reason it is better to have the logic available in this base module and avoid duplicating code or inheriting from sale_product_set and what this implies in their respective dependencies with the only need to extend the transient model.
In addition, the transient model is renamed to make it clearer to identify that it is this type of model.

These changes affect the following modules:

- stock_product_set: https://github.com/OCA/stock-logistics-workflow/pull/1538
- sale_product_set: https://github.com/OCA/sale-workflow/pull/3009
- project_stock_product_set: https://github.com/OCA/project/pull/1252

cc @Tecnativa TT48100

@victoralmau @pedrobaeza please review


product_set/wizard/product_set_add.py
```diff
@@ -0,0 +1,47 @@
+# Copyright 2015 Anybox S.A.S
+# Copyright 2016-2020 Camptocamp SA
+# @author Simone Orsi <simahawk@gmail.com>
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+from odoo import api, fields, models
+
+
+class ProductSetAdd(models.TransientModel):
```
pedrobaeza: Shouldn't this be an AbstractModel in the style of a mixin?

pilarvargas-tecnativa: done, thanks!

product_set/wizard/product_set_add.py
```diff
@@ -0,0 +1,47 @@
+# Copyright 2015 Anybox S.A.S
+# Copyright 2016-2020 Camptocamp SA
+# @author Simone Orsi <simahawk@gmail.com>
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+from odoo import api, fields, models
+
+
+class ProductSetAdd(models.AbstractModel):
```
petrus-v:
```suggestion
class ProductSetWizard(models.AbstractModel):
```

product_set/wizard/product_set_add.py
```diff
@@ -0,0 +1,47 @@
+# Copyright 2015 Anybox S.A.S
```
petrus-v: We may rename this file to product_set_wizard?

product_set/wizard/product_set_add.py
```diff
@@ -0,0 +1,47 @@
...
+            yield set_line
+
+    def _check_partner(self):
+        """This method may be extended in other modules that use product_set as a base."""
+        if not self.product_set_id.partner_id:
+            return
+
+    def add_set(self):
+        """This method may be extended in other modules that use product_set as a base."""
+        self._check_partner()
```
petrus-v: If you are motivate to test AbstractModel, you could use [FakeModelLoader](https://pypi.org/project/odoo-test-helper/) that would help you to get full coverage ;)

rousseldenis: Yes, adding tests should be great

pilarvargas-tecnativa: I have renamed and added the tests. Can you please check that everything is correct? thanks

rousseldenis (approved):

petrus-v (approved):
thanks taking care of my reviews !

victoralmau (approved):

pedrobaeza (approved):
/ocabot merge major

pedrobaeza:
@pilarvargas-tecnativa you can remove the dependency commits on each children PR.

PR #1549, opened 2024-03-13
[15.0] [IMP] product_state: move access to Sales + Inventory Manager

renda-dev:
FW-PORT of https://github.com/OCA/product-attribute/pull/1124


dreispt (approved):
FW Port of approved change

renda-dev:
Hello @dreispt, seems like that something went wrong while merging, could you please re-run the command here and on the other FW PRs? Thanks!

PR #1555, opened 2024-03-15
[IMP] product_variant_attribute_name_manager: more precise explanation

zeroheure:
this commit improve module description, helping to understand that it is only for back office

### DIFF
--- a/product_variant_attribute_name_manager/__manifest__.py
+++ b/product_variant_attribute_name_manager/__manifest__.py
@@ -1,6 +1,8 @@
 {
     "name": "Product Variant Attribute Name Manager",
-    "summary": "Manage how to display the attributes on the product variant name in back-office.",
+    "summary": """
+        Manage how to display the attributes on the product variant name 
+        in back-office.""",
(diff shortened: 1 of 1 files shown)


JordiMForgeFlow:
@zeroheure thank you for the proposal. Could you change the commit message following the guidelines here (https://github.com/OCA/odoo-community.org/blob/8afdb5a5ebfd3314e2fc836c9b9f644468ee044e/website/Contribution/CONTRIBUTING.rst#L1114)? You mainly need to add '[IMP] product_variant_attribute_name_manager: ...'.

Also, pre-commit check is not passing, if you could also look into that :)

zeroheure:
@JordiMForgeFlow should be ok but there is still a pre-commit error that I don't understand

JordiMForgeFlow:
@zeroheure I can see that there are 3 commits all with wrong commit message. You should only have one commit with a message like "[IMP] product_variant_attribute_name_manager: more precise module summary".

The pre-commit is complaining about some trailing whitespace.

rousseldenis:
@zeroheure

PR #1556, opened 2024-03-17
[16.0][FIX] product_secondary_unit: use existing value

alexey-pelykh:
The `rec._origin[rec._secondary_unit_fields["qty_field"]]` will still be used via https://github.com/odoo/odoo/blob/f9054640f8640d4449df5f7ea5fb7b4e8de4d89a/odoo/fields.py#L1197-L1201:

```
            elif self.store and record._origin and not (self.compute and self.readonly):
                # new record with origin: fetch from origin
                value = self.convert_to_cache(record._origin[self.name], record, validate=False)
                env.cache.set(record, self, value)
```

### DIFF
--- a/product_secondary_unit/models/product_secondary_unit_mixin.py
+++ b/product_secondary_unit/models/product_secondary_unit_mixin.py
@@ -99,7 +99,7 @@ class ProductSecondaryUnitMixin(models.AbstractModel):
         for rec in self:
             if not rec.secondary_uom_id:
                 rec[rec._secondary_unit_fields["qty_field"]] = (
-                    rec._origin[rec._secondary_unit_fields["qty_field"]]
+                    rec[rec._secondary_unit_fields["qty_field"]]
                     or default_qty_field_value
                 )
                 continue


alexey-pelykh:
Hi @sergio-teruel, please take a look and share your vision on this - it needs specifically you.

alexey-pelykh:
@OCA/product-maintainers anyone?

alexey-pelykh:
@pedrobaeza 🙏 I require help with someone looking into this

sergio-teruel:
Sorry @alexey-pelykh Now I do not remenber the reasson of that... @pedrobaeza Do you remember it?

pedrobaeza:
I think this may be related to the record containing `NewId` references that don't fetch the value correctly in defaults. The commit introducing it was: https://github.com/OCA/product-attribute/commit/5a71b68cba5b380d46013b42c43ea6bb693415ef

alexey-pelykh:
Hm, yet the tests aren't failing so either they aren't covering that or there's nothing to fix now?

pedrobaeza:
Maybe the problem doesn't exist in v16. Anyway, why do you want to change it?

alexey-pelykh:
Because that's what is causing https://github.com/OCA/sale-workflow/pull/3022/commits/589e363d8d4cf4146b95f98132bac40b2cfbeab1 to not work - I did a lot of testing if the https://github.com/OCA/sale-workflow/pull/3022/commits/ad195309d17f424cad14ed4cad6bddfa0c2be0fe as you can see :)

alexey-pelykh:
And what's more important - the current code is already there in v16 just as a branch that's going to get executed in .new() cases anyhow.

pedrobaeza (approved):
OK then, let's fix it and see if there's any side effect:

/ocabot merge patch

PR #1557, opened 2024-03-19
[ADD] product_pricelist_alternative

santostelmo:
It allows you to define alternative price lists to a reference price list.
As a general rule, the price of a given product is obtained from the minimum between its reference price list and the alternative price lists.

However, if the product's reference price has been calculated on the basis of a price rule in which the "Alternative Pricelist Policy = Ignore" field is selected, the alternative price lists will not be taken into account.


product_pricelist_alternative/models/product_pricelist.py
```diff
@@ -0,0 +1,50 @@
...
+from odoo import _, api, fields, models
+
+
+class Pricelist(models.Model):
+    _inherit = "product.pricelist"
+
+    alternative_pricelist_ids = fields.Many2many(
+        comodel_name="product.pricelist",
+        string="Alternative pricelists",
+        relation="alternative_pricelist_rel",
```
simahawk:
```suggestion
        relation="product_pricelist_alternative_rel",
```
or don't provide a value and let the orm choose the std

product_pricelist_alternative/models/product_pricelist.py
```diff
@@ -0,0 +1,50 @@
...
+        column1="origin_id",
+        column2="alternative_id",
+        domain="[('id', '!=', id)]",
+    )
+    is_alternative_to_pricelist_ids = fields.Many2many(
+        comodel_name="product.pricelist",
+        string="Is alternative to pricelists",
+        relation="alternative_pricelist_rel",
```
simahawk: same

product_pricelist_alternative/models/product_pricelist.py
```diff
@@ -0,0 +1,50 @@
...
+        column2="origin_id",
+    )
+    is_alternative_to_pricelist_count = fields.Integer(
+        compute="_compute_is_alternative_to_pricelist_count"
+    )
+
+    @api.depends("is_alternative_to_pricelist_ids")
+    def _compute_is_alternative_to_pricelist_count(self):
+        for pricelist in self:
+            pricelist.is_alternative_to_pricelist_count = len(
```
simahawk: pls use read_group

product_pricelist_alternative/models/product_pricelist.py
```diff
@@ -0,0 +1,50 @@
...
+
+    def action_view_is_alternative_to_pricelist(self):
+        self.ensure_one()
+        action = {
+            "type": "ir.actions.act_window",
+            "name": _("Is Alternative to Pricelist"),
+            "res_model": "product.pricelist",
+            "view_mode": "tree,form",
+            "domain": [("id", "in", self.is_alternative_to_pricelist_ids.ids)],
+            "context": dict(self._context, create=False),
```
simahawk:
```suggestion
            "context": dict(self.env.context, create=False),
```

product_pricelist_alternative/models/product_pricelist_item.py
```diff
@@ -0,0 +1,32 @@
...
+        required=True,
+    )
+
+    def _compute_price(self, product, quantity, uom, date, currency):
+        price = super(PricelistItem, self)._compute_price(
+            product, quantity, uom, date, currency
+        )
+        if self.consider_alternative == "use_lower_price":
+            alternative_pricelists = self.pricelist_id.alternative_pricelist_ids
+            if alternative_pricelists:
```
simahawk: this IF is irrelevant

product_pricelist_alternative/tests/test_pricelist_alternative.py
```diff
@@ -0,0 +1,132 @@
...
+
+from odoo.fields import Command
+from odoo.tests import common, tagged
+
+
+@tagged("post_install", "-at_install")
+class TestPricelistAlternative(common.TransactionCase):
+    @classmethod
+    def setUpClass(cls):
+        super().setUpClass()
```
simahawk, line 14: tracking_disable

product_pricelist_alternative/models/product_pricelist_item.py
```diff
@@ -0,0 +1,32 @@
+# Copyright 2024 Camptocamp (<https://www.camptocamp.com>).
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+
+from odoo import fields, models
+
+
+class PricelistItem(models.Model):
+    _inherit = "product.pricelist.item"
+
+    consider_alternative = fields.Selection(
```
simahawk:
```suggestion
    alternative_pricelist_policy = fields.Selection(
```
Just a suggestion

product_pricelist_alternative/models/product_pricelist_item.py
```diff
@@ -0,0 +1,32 @@
...
+        selection=[
+            ("use_lower_price", "Use lower price"),
+            ("do_not_consider", "Do not consider"),
```
simahawk:
```suggestion
            ("ignore", "Ignore alternatives"),
```

product_pricelist_alternative/models/product_pricelist_item.py
```diff
@@ -0,0 +1,32 @@
...
+        ],
+        default="use_lower_price",
+        required=True,
+    )
+
+    def _compute_price(self, product, quantity, uom, date, currency):
+        price = super(PricelistItem, self)._compute_price(
+            product, quantity, uom, date, currency
+        )
+        if self.consider_alternative == "use_lower_price":
```
simahawk:
```suggestion
        if self.alternative_pricelist_policy == "ignore":
            return price
```
and un-indent the rest

product_pricelist_alternative/views/product_pricelist_view.xml
```diff
@@ -0,0 +1,50 @@
...
+                        <field
+                            string="Alternative to Pricelist"
+                            name="is_alternative_to_pricelist_count"
+                            widget="statinfo"
+                        />
+                    </button>
+                </div>
+            </widget>
+
+            <group name="pricelist_settings" position="after">
```
simahawk, line 25: I would add a specific tab instead and list the pricelists instead of using m2m tags

simahawk (changes requested):

product_pricelist_alternative/models/product_pricelist.py
santostelmo: what is the advantage of read_group in that case ?

product_pricelist_alternative/models/product_pricelist_item.py
```diff
@@ -0,0 +1,33 @@
...
+    )
+
+    def _compute_price(self, product, quantity, uom, date, currency):
+        price = super(PricelistItem, self)._compute_price(
+            product, quantity, uom, date, currency
+        )
+        if self.alternative_pricelist_policy == "ignore":
+            return price
+
+        if self.alternative_pricelist_policy == "use_lower_price":
```
santostelmo: Doing a explicit check on the policy as we could imagine to have other  `alternative_pricelist_policy` values

product_pricelist_alternative/views/product_pricelist_view.xml
santostelmo: Not sure that it is a better option. New tab implies one more click to the user to check alternative pricelists.

santostelmo:
@simahawk I changed the code with some of your requests

product_pricelist_alternative/models/product_pricelist_item.py
santostelmo: I changed the field name like that

ivantodorovich (approved):

product_pricelist_alternative/models/product_pricelist.py
simahawk: Because
1. is faster
2. you never know if you or someone else will have to access this info for more than one record
3. is a well known approach w/ plenty of examples that you can copy paste

If you take the habit to do it you'll never regret it :wink:

product_pricelist_alternative/models/product_pricelist_item.py
simahawk: To do that, you'll have to modify this code anyway.
If you want another approach that won't imply writing an IF for all the cases or overrides to add more options from outside:

```
    def _compute_price(self, product, quantity, uom, date, currency):
        price = super(PricelistItem, self)._compute_price(
            product, quantity, uom, date, currency
        )
        handler = getattr(self, f"_compute_alternative_price_{self.alternative_pricelist_policy}")
        return handler(price)

    def _compute_alternative_price_ignore(self, original_price):
        return original_price

    def _compute_alternative_price_use_lower_price(self, original_price):
        for alternative_pricelist in self.pricelist_id.alternative_pricelist_ids:
            alternative_price = alternative_pricelist._get_product_price(product, quantity, uom, date)
            if alternative_price < original_price:
                return alternative_price
        return original_price
```

product_pricelist_alternative/views/product_pricelist_view.xml
simahawk: I'd find it more clear and is not a feature that you need to use on all PL. Anyway, probably just personal taste. I let you and BA decide :wink:

product_pricelist_alternative/models/product_pricelist_item.py
simahawk: Q: as the order of the pricelists impacts the result -> how do you ensure full control on which alternative pricelist will take precedence?

santostelmo: In your example above you are returning the first alternative_price lower than original_price. But don't want that.
We want the minimum between original_price and alternative prices

In our use cases order of alternative pricelists is not relevant as we take the minimum price between them.

product_pricelist_alternative/models/product_pricelist.py
santostelmo: To me it's a surprise that `read_group` is faster than just simply read M2M field `alternative_pricelist_ids`

product_pricelist_alternative/views/product_pricelist_view.xml
santostelmo: I imagine that price list managers access this forms every day multiple times. And one more click makes the difference. I ask the BA...
If not then it can go to a new tab indeed

product_pricelist_alternative/models/product_pricelist_item.py
simahawk: Correct, you can adapt that. I just wanted to show you a different approach that does not require using an IF every time even when not needed :wink:

simahawk (approved):

cyrilmanuel:
Hi @pedrobaeza , can you merge this PR please ?  :)

product_pricelist_alternative/models/product_pricelist.py
```diff
@@ -0,0 +1,71 @@
...
+    )
+
+    @api.depends("is_alternative_to_pricelist_ids")
+    def _compute_is_alternative_to_pricelist_count(self):
+        for pricelist in self:
+            pricelist.is_alternative_to_pricelist_count = len(
+                pricelist.is_alternative_to_pricelist_ids
+            )
+
+    def action_view_is_alternative_to_pricelist(self):
```
simahawk, line 43: pls add a test for this action too

product_pricelist_alternative/models/product_pricelist.py
simahawk: It should not :wink:

product_pricelist_alternative/models/product_pricelist.py
```diff
@@ -0,0 +1,71 @@
...
+                == "use_lower_price"
+            ):
+                product = self.env["product.product"].browse(product_id)
+                for alternative_pricelist in self.alternative_pricelist_ids:
+                    alternative_price_rule = alternative_pricelist._compute_price_rule(
+                        product, qty, uom=uom, date=date, **kwargs
+                    )
+                    if alternative_price_rule[product_id][0] < res[product_id][0]:
+                        res[product_id] = alternative_price_rule[product_id]
+        return res
```
simahawk: There's no need to browse products nor rules. You have them already and you have a mapping by id.
```suggestion
        for product in products:
            price, pl_rule = res[product.id]
            if (
                pl_rule.alternative_pricelist_policy
                == "use_lower_price"
            ):
                for alternative_pricelist in self.alternative_pricelist_ids:
                    alternative_price_rule = alternative_pricelist._compute_price_rule(
                        product, qty, uom=uom, date=date, **kwargs
                    )
                    # use alternative price if lower
                    if alternative_price_rule[product.id][0] < price:
                        res[product.id] = alternative_price_rule[product.id]
        return res
```
Also, no need to check for the `ignore` case as you do something only for `use_lower_price`.

simahawk (changes requested):

product_pricelist_alternative/models/product_pricelist.py
santostelmo: I will do it and you will show me 😇

product_pricelist_alternative/models/product_pricelist.py
santostelmo: Your example makes sense but `pl_rule` is a `id`. We need to browse it anyway to know  `alternative_pricelist_policy` value
res contains something like: `{product_id: (price, pricelist_item_id)}`

product_pricelist_alternative/models/product_pricelist.py
santostelmo: Test added

santostelmo:
@simahawk code was updated with `read_group` to count and new test

product_pricelist_alternative/models/product_pricelist.py
simahawk: you forgot to remove the 1st implementation :wink:

product_pricelist_alternative/models/product_pricelist.py
simahawk: still, do not browse products, you have them

product_pricelist_alternative/models/product_pricelist.py
santostelmo: I removed 1st implementation

product_pricelist_alternative/models/product_pricelist.py
santostelmo: I refactored the method. I am not browsing products anymore.

product_pricelist_alternative/models/product_pricelist.py
```diff
@@ -0,0 +1,76 @@
...
+        compute="_compute_is_alternative_to_pricelist_count"
+    )
+
+    @api.depends("is_alternative_to_pricelist_ids")
+    def _compute_is_alternative_to_pricelist_count(self):
+        groups = self.read_group(
+            [("alternative_pricelist_ids", "in", self.ids)],
+            ["alternative_pricelist_ids"],
+            "alternative_pricelist_ids",
+            lazy=False,
```
simahawk, line 35: why lazy=false?

You don't need to access anything but ID and count.

simahawk (approved):
pre-approving

PR #1562, opened 2024-03-21
[14.0][IMP] ``product_template_tags``: use tags on both templates and variants

SilvioC2C:
This commit allows using tags on both templates and variants.

Tags can be propagated on each template-variants couple in 2 ways:
- from template to variants: in this case, templates tags are copied onto every variant
- from variants to template: in this case, the template inherits all the variants' tags

The propagation behavior is defined at template's level. Default propagation behavior is "from template to variants" to allow full retrocompatibility with module's previous workflow.

### DIFF
--- a/product_template_tags/README.rst
+++ b/product_template_tags/README.rst
@@ -81,6 +81,7 @@ Contributors
 * `Camptocamp <https://www.camptocamp.com>`_
 
   * Iván Todorovich <ivan.todorovich@gmail.com>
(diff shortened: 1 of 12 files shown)


simahawk (changes requested):
I'd rather have a very simple new module that plays only w/ variants.
I don't think there's the need to customize the behavior by template.
If you manage tags at variant level, it's for all products.

PR #1566, opened 2024-03-26
[16.0][IMP] product_attribute_model_link: Attribute value relation enhancement

angelinaanaki:
Before
------

The field linked_record_ref of type Reference is used to store the information about which record is linked to a particular attribute value. However, Reference fields are stored in db as text fields which renders them impractical for usage in queries.

For example if user wants to get all attribute values that belong to a recordset the following steps should be done:

Get all attributes that belong to the record model using: 
attributes = self.env["product.attribute"].search([("linked_model_id", "=", model_id)])

Use mapped together with filtered to get corresponding attribute values: 
matching_attribute_values = attributes.mapped("value_ids").filtered(lambda v: v.linked_record_ref == record)

This will generate a lot of queries and may affect the performance badly.

Now
------

The linked_record_ref field has been enhanced by becoming a computed field, dynamically created based on the model and res_id fields, similar to the mail.message model. This change allows for faster search queries with just one line of code:

["product.attribute.value"].search([("model", "=", self._name), ("res_id", "in", self.ids)])

This adjustment significantly reduces the complexity and improves search speed.


product_attribute_model_link/models/product_attribute_value.py
```diff
@@ -8,10 +8,25 @@
...
+        string="Linked record",
+        compute="_compute_record_ref",
     )
 
+    @api.depends("model", "res_id")
+    def _compute_record_ref(self):
+        for rec in self:
+            if rec.model and rec.res_id:
+                res = self.env[rec.model].browse(rec.res_id)
+                result = res.check_access_rights("read", raise_exception=False)
```
ivs-cetmix: `result` name looks confusing for such function result. Better use something more explicit. Eg `can_read`.

ivs-cetmix (changes requested):
This is breaking  change so it will require a migration script. Otherwise existing data will be lost after the module update.

product_attribute_model_link/__manifest__.py
```diff
@@ -3,7 +3,7 @@
 {
     "name": "Product Attribute Model Link",
     "summary": "Use any model records as product attribute values",
-    "version": "16.0.1.0.1",
+    "version": "16.0.1.0.2",
```
ivs-cetmix: This is considered a significant change. So need to increment the `x` version component. Here is a reference in the  guidelines: https://github.com/OCA/odoo-community.org/blob/master/website/Contribution/CONTRIBUTING.rst#version-numbers

product_attribute_model_link/migrations/16.0.1.0.2/pre-migrate.py
```diff
@@ -0,0 +1,34 @@
+# Copyright (C) 2024 Cetmix OÜ
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+
+def migrate(cr, version):
+    # Create temporary columns to store data before the field becomes computed
+    cr.execute(
+        """
+        ALTER TABLE product_attribute_value
```
ivs-cetmix: Just a question: can't we create columns with the final names right from the start? Without using those intermediate ones.

ivs-cetmix (changes requested):

product_attribute_model_link/migrations/16.0.1.0.2/pre-migrate.py
angelinaanaki: Thanks for the feedback! I've adjusted the script to use the final column names from the start, simplifying the migration process. Please review it when you have a moment.

product_attribute_model_link/__manifest__.py
```diff
@@ -3,7 +3,7 @@
 {
     "name": "Product Attribute Model Link",
     "summary": "Use any model records as product attribute values",
-    "version": "16.0.1.0.1",
+    "version": "16.0.2.0.1",
```
ivs-cetmix: `16.0.2.0.0`

ivs-cetmix (approved):
Fix the module version.
Otherwise LGTM

product_attribute_model_link/migrations/16.0.2.0.1/pre-migrate.py
```diff
@@ -0,0 +1,33 @@
+# Copyright (C) 2024 Cetmix OÜ
```
alexey-pelykh: Why `16.0.2.0.1`?

angelinaanaki: Thanks! Corrected. Could you check again?

product_attribute_model_link/migrations/16.0.2.0.0/pre-migrate.py
```diff
@@ -0,0 +1,33 @@
...
+    )
+    # Extract and store model name and res ID from the linked_record_ref field
+    cr.execute(
+        """
+        SELECT id, linked_record_ref
+        FROM product_attribute_value
+        WHERE linked_record_ref IS NOT NULL
+    """
+    )
+    for rec_id, linked_ref in cr.fetchall():
```
alexey-pelykh: Consider doing
```sql
UPDATE product_attribute_value SET
    model = split_part(linked_record_ref, ',', 1), 
    res_id = split_part(linked_record_ref, ',', 2)
```

product_attribute_model_link/models/product_attribute_value.py
```diff
@@ -174,9 +191,7 @@ def action_open_linked_record_wizard(self):
         self.ensure_one()
         context = {
             "default_linked_record_ref": (
```
alexey-pelykh, line 193: Why to keep that field in wizard then?

alexey-pelykh: Since everything was refactored, I don't really see the rationale to keep it like that in the wizard - yet I may be missing something

product_attribute_model_link/migrations/16.0.2.0.0/pre-migrate.py
angelinaanaki: Thanks! Corrected. Could you check again?

product_attribute_model_link/models/product_attribute_value.py
angelinaanaki: I think that the single reference field in the wizard significantly simplifies the user experience by streamlining the selection process. It makes it easier for users to navigate and choose the appropriate record without the complexity of multiple fields or the need to input IDs directly. Do you have any concerns or alternative perspectives on this approach? I'm open to hearing your thoughts and discussing any other ideas you might have.

alexey-pelykh (approved):
Code LGTM

ivs-cetmix:
Hi @OCA/product-maintainers any chances to have this merged? 😉

Aldeigja (commented):
Functional LGTM

PR #1576, opened 2024-04-10
[16.0][IMP] product_pricelist_alternative: validate alternative pricelist with based formula pricelist

santostelmo:
Validate the following:
Formulas based on another pricelist are not allowed on alternative pricelists.

### DIFF
--- a/product_pricelist_alternative/models/product_pricelist.py
+++ b/product_pricelist_alternative/models/product_pricelist.py
@@ -2,6 +2,7 @@
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
 
 from odoo import _, api, fields, models
+from odoo.exceptions import ValidationError
 
 
 class Pricelist(models.Model):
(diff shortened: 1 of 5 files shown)


cyrilmanuel (approved):
LGTM

cyrilmanuel (approved):

ajaniszewska-dev (approved):
to rebase

product_pricelist_alternative/models/product_pricelist.py
```diff
@@ -74,3 +75,18 @@ def _compute_price_rule(self, products, qty, uom=None, date=False, **kwargs):
...
         return res
+
+    @api.constrains("alternative_pricelist_ids")
+    def _check_pricelist_alternative_items_based_on_other_pricelist(self):
+        """Alternative pricelists can not contain items based on other pricelist"""
+        for pricelist in self:
+            if pricelist.alternative_pricelist_ids.item_ids.filtered(
+                lambda item: item.compute_price == "formula"
+                and item.base == "pricelist"
+            ):
```
SilvioC2C, line 92: Non-blocking: you can use `any()` instead of `filtered` for better performances:
```suggestion
            if any(
                i.compute_price == "formula" and i.base == "pricelist"
                for i in pricelist.alternative_pricelist_ids.item_ids
            ):
```
The reason is:
- `filtered()` loops over each and every the record in the recordset, and only after going through them all the `if` condition is evaluated on the resulting recordset
- `any()` stops as soon as the iterator passed as argument returns a `True`-ish value

SilvioC2C (approved):
LGTM, only 1 nitpicking and non-blocking comment

PR #1578, opened 2024-04-15
[14.0][FIX] pos_product_cost_security: multi-company issues

chienandalu:
In a multicompany environment, doing a sudo() to override the security in the PoS data loading could bring us more data than needed. Therefore we just use the sudo to retrieve the cost and then join the data altogether retrived with the user's permissions.

cc @Tecnativa TT48609

please check @pedrobaeza 

fyi @zamberjo

### DIFF
--- a/pos_product_cost_security/models/product_product.py
+++ b/pos_product_cost_security/models/product_product.py
@@ -7,12 +7,23 @@ class ProductProduct(models.Model):
     _inherit = "product.product"
 
     def read(self, fields=None, load="_classic_read"):
-        """Make an exception when loading data to PoS"""
-        if self.env.context.get(
-            "pos_override_cost_security"
-        ) and self.env.user.has_group("point_of_sale.group_pos_user"):
-            fields = fields if fields else []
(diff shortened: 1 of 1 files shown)


pos_product_cost_security/models/product_product.py
```diff
@@ -7,12 +7,23 @@ class ProductProduct(models.Model):
...
+            and "standard_price" in fields
+        )
+        if not override_cost_security:
+            return super().read(fields=fields, load=load)
+        other_fields = [f for f in fields if f != "standard_price"]
+        result = super().read(fields=other_fields, load=load)
+        std_price_result = super(ProductProduct, self.sudo()).read(
+            fields=["standard_price"], load=load
+        )
+        # Can we relay in the sorting of both results to make the zip?
```
pedrobaeza: Yes, we can rely as it's the same model.

pedrobaeza (approved):

chienandalu:
@zamberjo you might be interested in reviewing this one

zamberjo (approved):
It makes sense, LGTM

PR #1583, opened 2024-04-17
[16.0][FIX] product_packaging_container_deposit : performance

jbaudoux:
When an order is modified, do not recompute container deposit for each line

### DIFF
--- a/product_packaging_container_deposit/models/container_deposit_order_mixin.py
+++ b/product_packaging_container_deposit/models/container_deposit_order_mixin.py
@@ -1,4 +1,5 @@
 # Copyright 2023 Camptocamp (<https://www.camptocamp.com>).
+# Copyright 2024 Jacques-Etienne Baudoux (BCIM) <je@bcim.be>
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
 import logging
 from functools import partial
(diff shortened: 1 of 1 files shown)


product_packaging_container_deposit/models/container_deposit_order_mixin.py
```diff
@@ -108,3 +109,11 @@ def copy(self, default=None):
         return super(
             OrderMixin, self.with_context(skip_update_container_deposit=True)
         ).copy(default=default)
+
+    def write(self, vals):
+        # When an order is modified, don't recompute container deposit for each line
+        res = super(
+            OrderMixin, self.with_context(skip_update_container_deposit=True)
+        ).write(vals)
+        self.update_order_container_deposit_quantity()
```
simahawk, line 118: what's the point of calling `update_order_container_deposit_quantity` if you always skip it?
Also, wouldn't be better to check if there's any update on values that might impact on deposit container lines?

jbaudoux: I skip it during the call to super and call it once done.
When you modify SO lines in the UI, when you save the form, this calls write on SO which in turns calls write on SO line. 
If you don't skip at this stage, the update will be performed on each SO line. In addition, those write could trigger other calls to write which will again recompute it.

Your second point could be a further improvement in another PR.

cyrilmanuel (approved):

santostelmo (approved):

product_packaging_container_deposit/models/container_deposit_order_mixin.py
simahawk: > Your second point could be a further improvement in another PR.

I would check - at least - that you are writing on `order_line`. That's easy, no?

jbaudoux: iI rechecked and `update_order_container_deposit_quantity` is already managing that properly

jbaudoux:
@simahawk Can we merge this?

jbaudoux:
ping @simahawk

PR #1588, opened 2024-04-19
[16.0][ADD] product_category_product_qty

quentinDupont:
New simple module to add information of how many products variants are in each product categories (and add product template quantity on tree view)

Migration from 12.0 from our custom Grap modules

product_category_tree
product_category_form


bealdav (approved):

product_category_product_qty/static/description/product_category_tree.png
petrus-v, line 1: I can't see the product.template computation, is this screenshot still conform with what the module does?

product_category_product_qty/readme/DESCRIPTION.rst
```diff
@@ -0,0 +1,7 @@
+This module extends the functionality of product module to display
+the quantity of products and variants for the product categories.
```
petrus-v, line 2: As far I understand here only variant are computed, also suggest wording  to avoid to confused user with stock quantities.

```suggestion
the number of products for the product categories.
```

product_category_product_qty/static/description/product_category_tree.png
quentinDupont: It uses native product_count https://github.com/odoo/odoo/blob/16.0/addons/product/models/product_category.py#L23 that's why :) and adds product_qty
To make it less confusing, i'm gonna rename to product_variant_count !

quentinDupont: thanks for the review :pray:

legalsylvain: >  To make it less confusing, i'm gonna rename to product_variant_count !

Are you sure to remove the "category" word from the module name ?

quentinDupont: I'm talking about the field's name :)

legalsylvain: soooory ! 
:+1:

legalsylvain (approved):
thanks !

PR #1590, opened 2024-04-19
[16.0][ADD] product_category_usage_group

quentinDupont:
This module extends the functionality of the product module, to restrict usage
of product categories to users that belong to a given group.

Migration from 12.0 of our custom module that we try to share with the OCA now :)

Use case for us : restrict a category where account code is not set by default

product_category_form
product_template_warning


product_category_usage_group/models/product_template.py
```diff
@@ -0,0 +1,42 @@
...
+from odoo.exceptions import ValidationError
+
+
+class ProductTemplate(models.Model):
+    _inherit = "product.template"
+
+    # Overload Section
+    @api.model_create_multi
+    def create(self, vals_list):
+        templates = super(ProductTemplate, self).create(vals_list)
```
bealdav:
```suggestion
        templates = super().create(vals_list)
```

product_category_usage_group/models/product_template.py
```diff
@@ -0,0 +1,42 @@
...
+
+    def write(self, vals):
+        res = super(ProductTemplate, self).write(vals)
+        if vals.get("categ_id"):
+            for rec in self:
+                rec._check_usage_product_category(vals.get("categ_id"))
+        return res
+
+    # Custom Section
+    def _check_usage_product_category(self, categ_id):
```
rousseldenis: @quentinDupont Why not using a constraint instead ?

quentinDupont: great point :laughing:

quentinDupont: I change the code and it makes it simpler

quentinDupont:
@rousseldenis @bealdav your suggestions have been taken into account, could you review the code again ? :)

product_category_usage_group/models/product_template.py
```diff
@@ -0,0 +1,25 @@
...
+    def _check_usage_product_category(self):
+        for record in self.filtered(lambda x: x.categ_id):
+            category = record.categ_id
+            group = category.usage_group_id
+            if group and group.id not in self.env.user.groups_id.ids:
+                raise ValidationError(
+                    _(
+                        'You can not use the product category "%(categname)s"'
+                        ' because you are not member of the group "%(groupname)s"'
+                        % {"categname": category.name, "groupname": group.name}
```
rousseldenis: You should use translation function arguments instead (look into core code for examples).

product_category_usage_group/tests/test_module.py
```diff
@@ -0,0 +1,45 @@
...
+
+from odoo.exceptions import ValidationError
+from odoo.tests import common
+
+
+class Tests(common.TransactionCase):
+    @classmethod
+    def setUpClass(cls):
+        super().setUpClass()
+        cls.env = cls.env(context=dict(cls.env.context, tracking_disable=True))
```
rousseldenis: FYI you can also use `BaseCommon` class from `odoo.addons.base.tests` to avoid this line

product_category_usage_group/models/product_template.py
legalsylvain: Why ? 
What is the problem with the current implémentation ?

Thanks !

rousseldenis:
```suggestion
                        categname=category.name,
                        groupname=group.name
```

rousseldenis:
@quentinDupont

product_category_usage_group/models/product_template.py
quentinDupont: https://www.odoo.com/documentation/16.0/developer/howtos/translations.html#variables

quentinDupont:
> @quentinDupont

Thanks for the review, I change the translation code and test as you suggest

legalsylvain (approved):
all remarks handled. 
Module used since a year in production, merging.

/ocabot merge nobump

PR #1591, opened 2024-04-23
[IMP] Translate product_category_hr_department to es_MX

SMaciasOSI:
@max3903


max3903:
@SMaciasOSI Translation should be done in https://translation.odoo-community.org

PR #1592, opened 2024-04-24
[17.0][MIG] product_status: Migration to 17.0

nguyenminhchien:


gurneyalex (approved):

product_status/models/product_state.py
```diff
@@ -20,7 +20,7 @@ def unlink(self):
         return super().unlink()
 
     def _check_module_data(self):
-        if self.env.user.id == 1:
+        if self.env.user._is_admin():
```
rousseldenis: IMHO this is not really the same.

Should be `._is_superuser()`

rousseldenis (commented):
Code review. One remark

nguyenminhchien:
> Code review. One remark

Hi, i think `_is_admin()` is better than `_is_superuser()` in this case. It should not check the module data when the user is Admin.

gurneyalex:
> > Code review. One remark
> 
> Hi, i think `_is_admin()` is better than `_is_superuser()` in this case. It should not check the module data when the user is Admin.

@nguyenminhchien @rousseldenis 

I would favor _is_superuser(): this seems more consistent with the rule saying "superuser bypasses security and allows you to shoot your foot off". 

On the other hand:

1. there does not seem to be a rule anywhere specifically requiring these states
2. it seems to me that what we don't want is deleting a state which is used by a product -> this should  be implemented by a ondelete='restrict' on `product.template::product_state_id` and then getting rid of this whole check. 

What do you think?

nguyenminhchien:
> > > Code review. One remark
> > 
> > 
> > Hi, i think `_is_admin()` is better than `_is_superuser()` in this case. It should not check the module data when the user is Admin.
> 
> @nguyenminhchien @rousseldenis
> 
> I would favor _is_superuser(): this seems more consistent with the rule saying "superuser bypasses security and allows you to shoot your foot off".
> 
> On the other hand:
> 
>     1. there does not seem to be a rule anywhere specifically requiring these states
> 
>     2. it seems to me that what we don't want is deleting a state which is used by a product -> this should  be implemented by a ondelete='restrict' on `product.template::product_state_id` and then getting rid of this whole check.
> 
> 
> What do you think?

`ondelete='restrict'` prevents the user from deleting the state, but not prevents from deactivating or switching the "default" value. i will use `._is_superuser()` for this migration. if we change our mind, we can add another commit for it.

PR #1595, opened 2024-04-29
[IMP] product_packaging_level: Rely on 'qty' field too to compute the qty_per_level field

rousseldenis:
In the product packaging Form view, rely also on 'qty' field to have a coherent behaviour for the display of qty_per_level field.

In that case, self.id == NewId, so compare ids field instead which is always a list of integers.


rousseldenis:
@jbaudoux

jbaudoux (approved):
Thanks for the fix

FrancoMaxime (approved):

lmignon:
ping @rousseldenis

PR #1599, opened 2024-05-01
[15.0] [MIG] product_video_link: Migration to 15.0

bizzappdev:
Dependency MR for module `base_video_link`  https://github.com/OCA/server-tools/pull/2930


product_video_link/__manifest__.py
```diff
@@ -20,6 +20,7 @@
     "depends": [
         "product",
         "base_video_link",
+        "sales_team",
```
rousseldenis, line 23: @bizzappdev Could you explain why you need this dependency ?

product_video_link/__manifest__.py
```diff
@@ -0,0 +1,31 @@
+# Copyright 2021 Akretion (https://www.akretion.com).
+# @author Sébastien BEAU <sebastien.beau@akretion.com>
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+
+{
+    "name": "Product Video",
+    "summary": "Link Video on product and category",
+    "version": "15.0.1.0.0",
+    "category": "Product",
+    "development_status": "Alpha",
```
rousseldenis, line 10: IMHO, this can be updated.

product_video_link/__manifest__.py
bizzappdev: In the security CSV file, the group from sales_team has been referenced, 
It seems like the group from sales_team is also referenced in v14, but dependency was not added.
The sales_team module is a base for the team management of the Sales and CRM modules, so either we have to add a dependency or change the group that will have access rights.

product_video_link/__manifest__.py
bizzappdev: should we change it to Beta or Production?

PR #1606, opened 2024-05-06
[17.0] [MIG] product_logistics_uom: Migration to 17.0 

bizzappdev:


product_logistics_uom/models/res_config_settings.py
```diff
@@ -0,0 +1,39 @@
...
+        "uom.uom",
+        "Default Volume Unit of Measure",
+        domain=lambda self: [
+            ("category_id", "=", self.env.ref("uom.product_uom_categ_vol").id)
+        ],
+        config_parameter="product_default_volume_uom_id",
+        help="Default unit of measure to express product volume",
+    )
+
+    product_default_length_uom_id = fields.Many2one(
```
nguyenminhchien, line 31: I don't see this field ( `product_default_length_uom_id` ) on the form view.

nguyenminhchien (changes requested):
I have 1 comment, others look good.

ferran-S73:
LGTM, same as already said if you could please add the product_default_length_uom_id field to the settings view

product_logistics_uom/models/res_config_settings.py
bizzappdev: It was also not added to older versions. But I agree with you it should be visible.
We have added it in the view.

ferran-S73 (approved):

nguyenminhchien (approved):

product_logistics_uom/models/res_config_settings.py
nguyenminhchien: thank you.

PCatinean (approved):

PR #1609, opened 2024-05-13
[ADD] product_sequence: add create new code option if category changed

Youstina-NR:
The inventory manager can generate a new product sequence if the category of a product has changed.


product_sequence/__manifest__.py
```diff
@@ -11,11 +11,15 @@
     "website": "https://github.com/OCA/product-attribute",
     "license": "AGPL-3",
     "category": "Product",
-    "depends": ["product"],
+    "depends": [
+        "product",
+        "stock",
```
rousseldenis, line 16: @Youstina-NR Why do you need this dependency ?

PR #1611, opened 2024-05-17
[17.0][MIG] product_variant_attribute_name_manager: Migration to 17.0

johnny-longneck:


david-s73:
@johnny-longneck As I don't have access to upload my change to your PR I show you the changes I have made and what they are for, I have added a new option in the attributes so that in the product variants also show the attribute values of the variant when it only has one value, you can select that option from the attributes.
[0001-IMP-product_variant_attribute_name_manager-single-at.txt](https://github.com/user-attachments/files/16330424/0001-IMP-product_variant_attribute_name_manager-single-at.txt)

PR #1612, opened 2024-05-17
[IMP] product_print_category: Do not write on 'to_print' when printing on product if 'to_print' is not flagged.

legalsylvain:
it will avoid to change write_uid write_date of product.product model.

CC : @remytms (I think you are using this module in coop it easy).
CC : @quentinDupont


quentinDupont (approved):

legalsylvain:
Thanks for the review ! 

/ocabot merge patch

PR #1625, opened 2024-06-10
[17.0][MIG] product_logistics_uom: Migration to 17.0

ryanc-me:
Minimal changes - the `attrs` change, and a small tweak to `pre_init_hook` (re [this](https://github.com/odoo/odoo/commit/b4a7996e967621aa090dc80346e6c3ef1d032dcf)).

I've kept the migration folder in place, because the script there is actually just running `pre_init_hook`. See here for more info: https://github.com/OCA/product-attribute/commit/0bdcbbd21d84b95ec0c4168b84faa4bdd9b9eeb4


legalsylvain:
could you rename your PR title ? thanks !

ryanc-me:
Whoops, thanks for the heads up!

ryanc-me:
Getting the tests to pass was a real pain here.

Basically, the vanilla Odoo [`_compute_volume`](https://github.com/odoo/odoo/blob/5dafc19f72116f2784c0308b5df40f0ec3f117a9/addons/product/models/product_template.py#L288-L290) function was misbehaving. If the volume on the variant was `0.01`, then the volume computed by this function was `0.00`. The function was working fine for higher values (like `10.0`).

It seems like a rounding issue, but I can't see any explicit rounding happening, and the fields are all `digits=False`.

Also, this issue was *not* present on my local test machine (running a pull of Odoo from ~start of May).
Also 2, the issue wasn't happening with `weight`, even though the functions all look identical.

I think perhaps it's due to how `_compute_volume` sets the value, as an dict-set `record[fname] = val` rather than a "field" set `record.fname = val`?

To fix the issue, I had to re-implement `_compute_volume` - see `models/product_template.py`.

ryanc-me:
I've added myself as a contributor, since this migration was a bit special. May be helpful if anyone has issues in the future. I hope that's okay!

ferran-S73:
This module has already been migrated https://github.com/OCA/product-attribute/pull/1606

ryanc-me:
Closing as this was merged in another PR (see above).

PR #1627, opened 2024-06-10
[16.0][ADD] stock_picking_auto_create_product_lot_sequence

AungKoKoLin1997:
This module acts as a bridge between the `stock_picking_auto_create_lot` and `product_lot_sequence` modules.
Due to the changes introduced in [this commit](https://github.com/OCA/stock-logistics-workflow/commit/1f1d1cd4ba102f35971c074ca869ed6f7113cd2b) for `stock_picking_auto_create_lot`, using these two modules together disrupts the design of product_lot_sequence by not allowing the use of specific sequences for each product. This module restores the intended behavior.
Related: https://github.com/OCA/stock-logistics-workflow/pull/1616

@qrtl QT4624


stock_picking_auto_create_product_lot_sequence/models/stock_move_line.py
```diff
@@ -0,0 +1,19 @@
...
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+
+from odoo import api, models
+
+
+class StockMoveLine(models.Model):
+    _inherit = "stock.move.line"
+
+    @api.model
+    def _get_lot_sequence(self):
```
yostashiro:
```suggestion
    def _get_lot_sequence(self):
        self.ensure_one()
```

yostashiro (commented):
Partial review.

How about `stock_picking_auto_create_lot_sequence` as the module name?

I think it's more appropriate to host this module in https://github.com/OCA/stock-logistics-workflow/.

AungKoKoLin1997:
> I think it's more appropriate to host this module in https://github.com/OCA/stock-logistics-workflow/.

What I thought was this module is needed by `product_lot_sequence` to not break its behavior when used together with `stock_picking_auto_create_lot`. The intention of this module is to support `product_lot_sequence`. Therefore, my idea is to keep this module together with product_lot_sequence in the same repository.

AungKoKoLin1997:
This PR is superseded by https://github.com/OCA/stock-logistics-workflow/pull/1619

PR #830, opened 2021-03-16
[14.0] product variant attributes to template

sebastienbeau:

PR #1636, opened 2024-06-13
14.0 product variant attributes to template

sebastienbeau:
Reopen PR : https://github.com/OCA/product-attribute/pull/830

Allow to import variant with attribute


Kev-Roche (approved):
Functional review

PR #1646, opened 2024-06-13
[14.0] Replace 'Camptocamp SA' with 'Camptocamp'

vvrossem:
supersedes https://github.com/OCA/product-attribute/pull/1639


etobella (commented):
Use the proper commit message please

PR #1647, opened 2024-06-13
[17.0] Replace 'Camptocamp SA' with 'Camptocamp'

vvrossem:


vvrossem:
supersedes https://github.com/OCA/product-attribute/pull/1640

etobella (commented):
Use the proper commit message please

PR #1649, opened 2024-06-13
[12.0] Replace 'Camptocamp SA' with 'Camptocamp'

vvrossem:
supersedes https://github.com/OCA/product-attribute/pull/1641

### DIFF
--- a/product_dimension/__manifest__.py
+++ b/product_dimension/__manifest__.py
@@ -6,7 +6,7 @@
     'name': 'Product Dimension',
     'version': '12.0.3.0.0',
     'category': 'Product',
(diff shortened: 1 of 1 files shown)


legalsylvain:
Please. this quantity of PR is a  little awkward. Maybe we could avoid to make PR against obsolete unmainted branch, and target changes only on recent 16.0 / 17.0 version ? 

thanks !

bosd (approved):

rousseldenis:
Agree with @legalsylvain as I and @pedrobaeza proposed. @vvrossem Do you plan to do it on all old branches ?

PR #1655, opened 2024-06-14
[16.0][FIX] product_pricelist_direct_print: Error when display wizard without administration permision

CarlosRoca13:

1. Log in with a user without Administration permision
2. Open Pricelist Print wizard

An error about permisions is desplayed

cc @Tecnativa TT49637

ping @pedrobaeza @chienandalu

### DIFF
--- a/product_pricelist_direct_print/wizards/product_pricelist_print_view.xml
+++ b/product_pricelist_direct_print/wizards/product_pricelist_print_view.xml
@@ -31,7 +31,11 @@
                     </group>
                     <group string="Order Options">
                         <field name="order_field" />
-                        <field name="group_field_id" nocreate="1" />
+                        <field
+                            name="group_field_id"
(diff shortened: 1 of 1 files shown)


chienandalu (approved):

pedrobaeza (approved):
@legalsylvain FYI

@CarlosRoca13 next time, please put the exact error in the commit message for reference.

/ocabot merge minor

legalsylvain (changes requested):
Hum. At the first sight, that's annoying if a "basic user" without administration right can not select the field to group by. Don't you think ? In a classic use case, people that want to print pricelist are not ERP manager.

Could you imagine to resolve the problem differently ?

pedrobaeza:
@legalsylvain that can't be done. We need to solve the problem you created adding this change. Please explore another way to show the field for basic users if you want.

PR #1656, opened 2024-06-14
[16.0][FIX] product_pricelist_direct_print : Prevent access error if a user that doesn't belong to 'Administration/Access Rights'  try to print pricelists. 

legalsylvain:
Error explained here  https://github.com/OCA/product-attribute/pull/1655 by @CarlosRoca13.

In #1655, this problem has been temporarily overcome by hiding the functionality if the user is not a member of a group.
In the current PR, the problem is fixed and the feature restored for all user perfiles

CC : @chienandalu, @pedrobaeza 

thanks for your review;

### DIFF
--- a/product_pricelist_direct_print/wizards/product_pricelist_print.py
+++ b/product_pricelist_direct_print/wizards/product_pricelist_print.py
@@ -54,14 +54,10 @@ class ProductPricelistPrint(models.TransientModel):
     order_field = fields.Selection(
         [("name", "Name"), ("default_code", "Internal Reference")], string="Order"
     )
-    group_field_id = fields.Many2one(
-        comodel_name="ir.model.fields",
+    group_field = fields.Selection(
+        selection=lambda x: x._selection_group_field(),
+        default="categ_id",
         required=True,
-        domain=[
-            ("model", "=", "product.product"),
-            ("ttype", "=", "many2one"),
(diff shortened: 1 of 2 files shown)


pedrobaeza (commented):
The problem I see with this is when a field disappear (like for example in a migration). You'll get a non recoverable error from the UI, avoiding you to access the record.

legalsylvain:
> The problem I see with this is when a field disappear (like for example in a migration). You'll get a non recoverable error from the UI, avoiding you to access the record.


Sorry, I don't get it. let's imagine : 
- pos_categ_id disappear in V18. 
- we have a db in v17 with records (let's say ids = [7, 8]) with group_field = 'pos_categ_id'
- we run an upgrade to V18 before the cron that vaccum transient model delete all the records 
- we launch the migrated db  in V18 instance
- user launch the wizard in V18, that create new records. (let's say id = 9). in the selection list, pos_categ_id is not available, but user can select other M2O fields.

In which stage is there a problem?  

Thanks for your review !

pedrobaeza:
Ah, OK, it's a transient one. I was thinking in something more persistent (like a template).

pedrobaeza (approved):

legalsylvain:
thanks for your quick review ! 

/ocabot merge patch

PR #1657, opened 2024-06-15
[16.0][ADD] product_usability

legalsylvain:
### Description

This module extends the Odoo CE product module to improve usability.

#### New Menu items

By default, when installing ``product`` module, a lot of menu items are not created and are available only if other module are installed.
For exemple:

* Manage product categories requires to install ``stock`` module, that create the menu entry _"Inventory / Configuration / Product / Product Categories"_.
* Manage product tags requires to install ``sale_management`` module, that create the menu entry _"Sale / Configuration / Products / Tags"_.

This module so create a new main menu item named 'Product' that provides all menu entries to see and manage product, attributes, pricelist, Unit of Measure, etc. So that all the product configuration is available in a unique place.

main_menu

#### New Access Rights

By default, to create product, attributes, pricelist, categories, User should be member of a group that adds a lot of rights, like 'Inventory / Manager' or 'Sale / Manager'.

This module adds new Access rules to fine-tune access rights, creating the following rules:

* _"Product Creation"_: Allow to create Products (``product.template``), Variants (``product.product``), Product Template Attribute Lines (``product.template.attribute.line``) and Product Template Attribute Values
  (``product.template.attribute.value``)

* _"Product Attribute Creation"_: Allow to create Attributes (``product.attribute``) and Values (``product.attribute.value``).

* _"Product Tag Creation"_: Allow to create Tags (``product.tag``)

* _"Product Category Creation"_: Allow to create Categories (``product.category``)

* _"Unit of Measure Creation"_: Allow to create Unit of Measures (``uom.uom``) and Unit of Measures Categories (``uom.category``)


bealdav (approved):
awesome

hparfr (commented):
Thanks for this module.

I think the name does not tell us that much about the subjet. May be something like product_menu or product_app .

It looks a lot like this module : https://github.com/OCA/odoo-pim/tree/14.0/pim

bealdav:
yes product_app could be nice, thanks

rvalyi:
> Thanks for this module.
> 
> I think the name does not tell us that much about the subjet. May be something like product_menu or product_app .
> 
> It looks a lot like this module : https://github.com/OCA/odoo-pim/tree/14.0/pim

Eventually the pim module could start depending on this new module if it makes things simpler and more integrated... I could also be named product_management. Not sure which is the best. "catalog" could also be used.

bealdav:
After Contacts app, Catalog sounds nice

legalsylvain:
Hi.

Thanks for your review and answer. About the name, the rational are the following : 
Odoo "product" module is incomplete. If you install it, you have no menu entry to manage the products. (that's not very good !), and you can not also manage product.category, product.tag, etc... because there is no menu entry AND because there are no according access rights.

So ``product_usability`` for me, is a correct name, because it provides "missing things" that are expected when you install product module. The objective is to add that missings things, and the list of missing things can change over the versions.

If we name the module "product_app" or "product_menu", and if version 18 (or +) introduces an entry menu, the name will not be OK, but the module still is needed, because access rights are not present or other things.

So i prefer that generic name, as it will become more resistant over time.

**Note :** 
A similar module exists in account-financial-tools repo, named "account_usability", that adds missing entry menus, AND other things. See : https://github.com/OCA/account-financial-tools/tree/16.0/account_usability#account---missing-menus--saxon-accounting
Previously, the name of the module was "account_menu", but it has been renamed to "account_usability" for the exact same reason when migrating in V15 see : https://github.com/OCA/account-financial-tools/pull/1275#pullrequestreview-821772826

**Note 2:**

About ``pim`` module. (https://github.com/OCA/odoo-pim/tree/14.0/pim). 
- the last accepted version is 14.0, that is quite old.
- indeed menu entries are similar.
- security is [less modular](https://github.com/OCA/odoo-pim/blob/14.0/pim/security/pim_security.xml) than in the design of product_usability and allows less things.

So, i propose to maintain this one, and deprecate ``pim``, and add a rename of ``pim`` to ``product_usability`` in OpenUpgrade.

Thanks !

pedrobaeza:
For me, `usability` says nothing. `product_management` fits better.

bealdav:
@legalsylvain  `catalog` also embrace more concepts than just `product_any_string` and is an attractive name that can be extended `catalog_other_feature` better than `product_management_other_feature` IMO

legalsylvain:
> For me, usability says nothing. product_management fits better.

Hi. Here is my rational behind the choice of "usability". 

In the history of OCA modules: 

### "usability"

Looks to be used to improve Odoo Core module / Odoo Core features. (account / product_packaging / etc...) **without adding changes** in database layout : for exemple 
    - `product_packaging_usability` by akretion, ([code](https://github.com/OCA/stock-logistics-warehouse/tree/16.0/product_packaging_usability)). 
    - `account_invoice_period_usability` by akretion ([code)](https://github.com/OCA/account-invoicing/tree/8.0/account_invoice_period_usability).
    -  `stock_packaging_usability` (akretion, [code](https://github.com/OCA/stock-logistics-tracking/tree/14.0/stock_packaging_usability))
    - `account_usability` (by Akretion and GRAP, [code](https://github.com/OCA/account-financial-tools/tree/17.0/account_usability))
The existing modules just add views, action or button.

image

### "management"

looks to be used to add **new features** (or to implement features in another way than Odoo does it, for exemple, for assets). 

image

[fieldservice_change_management](https://github.com/OCA/field-service/tree/14.0/fieldservice_change_management), [account_asset_management](https://github.com/OCA/account-financial-tools/tree/17.0/account_asset_management/), [mrp_subcontracting_partner_management](https://github.com/OCA/manufacture/tree/16.0/mrp_subcontracting_partner_management) all adds a lot of new fields and new features.


Could we keep the same logic for this module, or could you explain alternative logic ? 

Thanks for your remarks !

pedrobaeza:
Management is management. To manage something. In this case, to manage products. No matter if you need to add 3 views, 4 actions or whatever. Usability is to improve that specific part of something. Going to that subjective way, every OCA module improves the usability of something. I still vote for `product_management`. `catalog` can have other concepts and it's an alien terminology.

legalsylvain:
Thanks for your reponse Pedro. I'm not sure to understand. Do you mean all the existing "usability" OCA modules should be renamed into "management" ? What I really don't understand is that the current proposed module "product_usability" does exactly the same things as the accepted module "account_usability", so I don't understand why to change product_usability and not account_usability.

Thanks for your precision.

pedrobaeza:
Each case is different, and I wasn't there when that modules were decided to have that names. Now I'm on this one, and as said, for this one, I think the valid name is `product_management`.

legalsylvain:
> Each case is different, and I wasn't there when that modules were decided to have that names.

Not sure to understand, Do you mean account_usability is bad named ? and should be renamed ?
Also please mention also why the case is different. I really don't see differencies.

bealdav:
Arf https://medium.com/the-product-marketer/the-art-of-feature-naming-four-survival-tips-76bd86f82cff

1. Keep names simple. A name `doesn’t have to explain everything` about a product. It’s simply a `bookmark` for a concept.
2. Anchor as much as possible to existing knowledge among your customers. ...
3. Say what your product or `feature does` or is. Sounds simple, but there is a lot of art in this. Try to `strike a balance between using technical terms and plain English`. ....
4. Try to use the name. `Try it out in marketing content`, docs and product design mocks. See if it feels right.


The problem with `management` is that you could append it to any concept in ERP because ERP is management. Don't repeat yourself !

Using `management` is like using `tool`. It defines what ?

pedrobaeza:
> Do you mean account_usability is bad named ?

I haven't analyzed it, but in that case, being a gathering of a lot of little "improvements", can be a valid name.

In this case, you are only enabling to manage the products. And that's why I insist with `product_management`, @bealdav, the same as Odoo have `sale_management` for the module that adds the "Sales" menu, while `sale` only adds the model, so we are mimicking (better or worse) the Odoo paradigm.

bealdav:
mimick odoo bad choices is a bad idea IMO

> don't read over your neighbor's shoulder

> no leas por encima del hombro de tu vecino

bealdav:
Then @pedrobaeza think product_usability is a good name ? Nice

pedrobaeza:
> mimick odoo bad choices is a bad idea IMO

Not always. Indeed, Odoo has taken some bad decisions in the past, but there's also some things that you shouldn't do the contrary for consistency. Example: naming m2o with `_id` suffix is a debatable thing, but if you don't follow the guideline, you won't have consistency across code. For module names, we at OCA talk about not putting plurals in module names, but if a Odoo core module is called `barcodes`, we should do the expansions calling them `barcodes_<whatever>` for association.

I think something similar happens here, calling `product_management` for that consistency and extension linking.

bealdav:
You're right not always

pedrobaeza:
> Then @pedrobaeza think product_usability is a good name ? Nice

No, I still think `product_management` should be the name.

bealdav:
It's a matter of taste

bealdav:
whatever we need to merge it

bealdav:
Whatever, this page should inspire OCA for the future I think
https://medium.com/the-product-marketer/the-art-of-feature-naming-four-survival-tips-76bd86f82cff

pedrobaeza:
Well, this is not exactly a feature naming, but a module technical name. It follows another rules. Mine are:

- Start with the name of the main are. Examples: `sale_`, `purchase_`, `product_`.
- Continue with the subarea or directly the feature if no more needed: `_order_` or `_secondary_unit`.
- And finally, if needed, the feature or rest of the explanation: `_qty_change_no_recompute`.

bealdav:
We are ok.
If technical name is near of name it's better I suppose.
Module is a feature or a collection of features, then rules are applyable.
Here prefix is product, then it can be `product_access`, `product_access_right` or other if module in future will get other features.

pedrobaeza:
That's why I insist in the generic name `product_management` for gathering together all the needed things for managing the products directly (accesses, security rules, menus, views...). We have used `_security` suffix for other modules where it's clear that the goal is to adjust the security thing. In this mood, `product_usability` would be better than `product_access`, but as said, I think it doesn't reflect the whole idea IMO.

bealdav:
Enough for me ;-)

bealdav:
product_manager 😁

youring:
My little suggestion:
For all UI/UX or functional additions or improvements that can be reasonably combined together but cannot(not necessarily) be clearly categorized, while splitting cost a lot to maintain, in this situation, can we use the name ***_oca_improved**, * for core module such as account, sale, purchase, product, stock, etc.
In this case, **product_oca_improved**. 
This nomenclature doesn't seem to be fundamentally different from *_management or *_usability, but at least it brings a branding effect for OCA.
For ref:
Other OCA modules contains **improved**, mail_improved_tracking_value, base_name_search_improved

bealdav:
@youring good catch `product_oca` sounds really good.

It's a shortname, then easy to use in a message commit like
 `[IMP] product_oca: many improvements here ...` 

 or for an extended module  
`[IMP] product_oca_my_nice_feature: many really nice improvements ...`

compared to 
`[IMP] product_management_my_nice_feature: many really nice improv ...`

legalsylvain:
hi. sorry but I don't see it is fair to block a PR, just for a matter of taste. When we receive PR, reviewers should follow the existing OCA rules and convention. It is not the place to introduce new rules.

1) As said, this module has exactly the same logic as the other existing merged OCA modules, with the suffix ``_usability``. (see [previous comment](https://github.com/OCA/product-attribute/pull/1657#issuecomment-2272814442)). The modules exist since years. So, at this step, we have to follow this existing habit.

2) In contrast, the 2 other OCA modules, which have a ‘_management’ suffix, have a completely different logic.

Then, if clarrification is required please make PR upstream, on OCA convention repo, to specify what mean the following suffix ``_management``, ``_usability``, ``_oca``, . Once the new convention approved, I'll totally open to rename the current module to follow the new convention. but it should be done in the new release (18.0 for exemple), with the other module renaming, to be consistent. (account_asset_management, account_usability, etc..)

Side remark : @bealdav : I don't know if it is written somewhere, but AFAIU, "_oca" is about OCA modules forking EE / CE features, and introducing possible incompatility with the official modules. Ex : l10n_fr_oca from akretion, spreadsheet_oca, web_pwa_oca

Thanks !

bealdav:
No objection to stay with current name in this version whatever the name in future version

Thanks @legalsylvain to clarified the situation allowing to merge this module.

cc @pedrobaeza

youring:
Hi @legalsylvain FYI, just found in 18.0 there is a group called "Product Creation" id=group_product_manager,  should we be consistent with it from now on?
https://github.com/odoo/odoo/blob/3e4c4516f7d238e76e2e9b7b71a807f7cc7d55b3/addons/product/security/product_security.xml#L19

legalsylvain:
@youring : I don't understand what you propose. could you elaborate ?

youring:
@legalsylvain I mean when upgrading `product_usability` to version 18.0, the group name `Product Creation` will be conflict with [the same group name](https://github.com/odoo/odoo/blob/3e4c4516f7d238e76e2e9b7b71a807f7cc7d55b3/addons/product/security/product_security.xml#L20) newly introduced in Odoo 18.0 `product` module whose external id is `product.group_product_manager`
I'm not proposing, just not sure whether it's good to change the group name now or untill migration to v18.0.

legalsylvain:
OK I get it ! Thanks for your message. (sorry, I was pretty lost, I made this PR 5 monthes ago).

Well, for the time being (v16.0), the design is OK. Nothing to do IMO. I just added a ROADMAP.rst file, to mention what you say. See : https://github.com/OCA/product-attribute/pull/1657/commits/7ca4c4ad029de4884ae816caa719e22df0f27ba9

For the V18 migration, we will : 
- adapt the module product_usability, removing the definition of product_usability.group_product_creation and use the new core group.
- in the meantime, we'll have to add a trivial rename_xml_id in OpenUpgrade/18.0/product script. (product_usability.group_product_creation -> product.group_product_manager). That way all users that are member of product_usability.group_product_creation will be member of product.group_product_manager automatically.

Do you think it's OK ? 

Note : I tried to directly use the futur xml_id in the current module. 
```
<record id="product.group_product_manager" model="res.groups">
```

but I have an exception. I think such check is recent. Before, I think it was allowed.

````Exception: Cannot update missing record 'product.group_product_manager'````

youring:
> Do you think it's OK ?

OK, clear enough.
>Exception: Cannot update missing record 'product.group_product_manager'

Coz the group does not exist in 16.0 or 17.0 CE

PR #1660, opened 2024-06-17
[17.0][MIG] product_main_supplierinfo

adriresu:
Standard migration, reused #1353 (Unfinished migration to v16.0)


adriresu:
Can you review, please?
@aliciagaarzo
@sebalix
@Kev-Roche

dalonsod:
Tested on runboat :+1:

adriresu:
Can you review please?
@bizzappdev

product_main_supplierinfo/models/product_product.py
```diff
@@ -0,0 +1,55 @@
...
+                and (
+                    (s.date_start <= today if s.date_start else True)
+                    and (s.date_end >= today if s.date_end else True)
+                )
+            )
+        )
+        if not sellers:
+            sellers = all_sellers.filtered(lambda s: (s.product_id == self))
+            if not sellers:
+                sellers = sellers = all_sellers.filtered(lambda s: not s.product_id)
```
flotho: this line is weird, could you explain the use of the duplicate assignation on the same variable?

flotho (changes requested):

product_main_supplierinfo/models/product_product.py
dalonsod: This line belongs to previous implementation of this addon, added for `14.0` in #1077 , we could ask to @TDu

TDu: This looks like a typo, should be IMO
```suggestion
                sellers = all_sellers.filtered(lambda s: not s.product_id)
```

TDu:
Also the directory `setup` should be removed. Packaging of addons has changed (cf https://github.com/OCA/maintainer-tools/issues/582) in v17

Anxo82:
Done! Review please @TDu 
Thanks

FernandoRomera (approved):
LGTM

flotho (approved):
LGTM, thanks

xaviedoanhduy (approved):

sebalix (approved):

product_main_supplierinfo/models/product_product.py
```diff
@@ -0,0 +1,55 @@
...
+        "seller_ids.date_start",
+        "seller_ids.date_end",
+    )
+    @api.depends_context("company")
+    def _compute_main_seller_id(self):
+        for product in self:
+            sellers = product._get_sellers()
+            product.main_seller_id = fields.first(sellers)
+
+    def _get_sellers(self):
```
rousseldenis, line 32: @adriresu @sebalix My question arrives late but why not using standard method `_get_filtered_sellers()` or `_select_seller()` to avoid code duplication ?

sebalix: Good question, had to re-check the original v14 PR to see if I mentioned something there, but nothing!
The only difference that remembers me a thing is that `_select_seller()` is sorting the sellers based on `price` before returning one (which is OK when encoding a PO). Here the price is not a criteria to become the "main vendor" of a product, but sequence and dates matter.

sebalix: Just saw that starting from Odoo v17, `_select_seller` method accepts a new `ordered_by` parameter, which could be used to force the use of `sequence`, if that works we could drop this method :thinking:

sebalix: It's not that neither, we are sorting on the price at the end of this method too... I don't remember then why I duplicated this code, for sure there is a reason.

sebalix: Find out, so we could not use `_select_seller` in v14 because of this filter:
- https://github.com/odoo/odoo/blame/14.0/addons/product/models/product.py#L637-L638

But starting from v15, which introduced the submethod `_get_filtered_sellers` you were talking about, it should definitely work.

sebalix: Trying this here: https://github.com/OCA/product-attribute/pull/1812
One test is failing, to check.

PR #1664, opened 2024-06-19
[17.0][MIG] product_seasonality

astirpe:
Porting from https://github.com/OCA/product-attribute/pull/1061


product_seasonality/models/res_company.py
```diff
@@ -0,0 +1,40 @@
...
+
+    default_seasonal_config_id = fields.Many2one(
+        string="Default product seasonal configuration", comodel_name="seasonal.config"
+    )
+
+    def _create_default_seasonal_conf(self):
+        self.ensure_one()
+        if not self.default_seasonal_config_id:
+            self.default_seasonal_config_id = self.env["seasonal.config"].create(
+                {"name": _("Default product seasonal configuration: %s") % self.name}
```
rousseldenis:
```suggestion
                {"name": _("Default product seasonal configuration: %(name)s", name=self.name)}
```

product_seasonality/models/res_company.py
```diff
@@ -0,0 +1,40 @@
...
+            )
+
+    @api.model
```
rousseldenis: Using multi decorator is better

product_seasonality/models/seasonal_config_line.py
```diff
@@ -0,0 +1,86 @@
...
+
+        weekday = date.strftime("%A").lower()
+        # check if we are on the correct day of the week
+        if not self[weekday]:
+            return False
+
+        date_end = self.date_end or date
+        return self.date_start <= date <= date_end
+
+    # TODO: shall we use the product's company to retrieve the default config?
```
rousseldenis, line 61: Still applicable ?

product_seasonality/models/res_company.py
astirpe: Right, thanks!

product_seasonality/models/seasonal_config_line.py
astirpe: "seasonal.config" doesn't have a company_id, so I don't think is applicable. Or maybe I don't understand the sense of this TODO. Any idea?

astirpe:
Rebased

PR #1672, opened 2024-06-27
[16.0][ADD] product_packaging_dimension_float

ChrisOForgeFlow:


AlexPForgeFlow (approved):

AaronHForgeFlow:
ok to me, but why don't making the changes in the base module instead? This seems to be a fix on that module.

AaronHForgeFlow:
Actually that fix is a kind of a feature, to what was discussed in here: https://github.com/OCA/product-attribute/pull/957

I am wondering if the original reasons in v14 are still valid in newer versions.

PR #1678, opened 2024-07-04
[16.0][IMP] product_pricelist_direct_print_xlsx: printing a pricelist and selecting an product category, include its child categories.

chaule97:
-  option `Print child categories` in `XLSX Export Options`, if it's enabled, includes products from child categories when generating price list reports.

    - option `Print child categories`
    
     image
    - result:
    
     image


legalsylvain:
Hi @chaule97. I don't understand why not put that option in ``product_pricelist_direct_print`` module .

chaule97:
> Hi @chaule97. I don't understand why not put that option in `product_pricelist_direct_print` module .

Thanks for contributing, I forgot to focus on that

PR #1679, opened 2024-07-05
[16.0][IMP] product_pricelist_direct_print: printing a pricelist and selecting an product category, include its child categories.

chaule97:
-  option `Print child categories` in Product Options, if it's enabled, includes products from child categories when generating price list reports.

    - option `Print child categories`
     
     image

    - result:
    
     image

### DIFF
--- a/product_pricelist_direct_print/wizards/product_pricelist_print.py
+++ b/product_pricelist_direct_print/wizards/product_pricelist_print.py
@@ -66,6 +66,7 @@ class ProductPricelistPrint(models.TransientModel):
         " the last X ordered products will be obtained for the report."
     )
     summary = fields.Text()
+    print_child_categories = fields.Boolean()
     max_categ_level = fields.Integer(
         string="Max category level",
         help="If this field is not 0, products are grouped at max level "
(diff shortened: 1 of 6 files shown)


product_pricelist_direct_print/wizards/product_pricelist_print.py
```diff
@@ -66,6 +66,7 @@ class ProductPricelistPrint(models.TransientModel):
         " the last X ordered products will be obtained for the report."
     )
     summary = fields.Text()
+    print_child_categories = fields.Boolean()
```
legalsylvain, line 69:
```suggestion
    print_child_categories = fields.Boolean(default=True)
```

I guess, if we select category / categories we expect childs categories to be selected IMO.

chaule97: yes, but I wrote 

```
@api.onchange("categ_ids")
    def _onchange_categ_ids(self):
        self.print_child_categories = len(self.categ_ids) > 0
```
        
so, when we select category / categories, print_child_categories will be True first

chaule97: and, If print_child_categories=True first. When we don't select category / categories, we get wrong domain in
```
if self.print_child_categories:
            domain = expression.AND(
                [domain, [("categ_id", "child_of", self.categ_ids.ids)]]
            )
elif self.categ_ids:
            domain = expression.AND([domain, [("categ_id", "in", self.categ_ids.ids)]])
 ```

legalsylvain (approved):
Thanks

cyrilmanuel (approved):
LGTM

PR #1687, opened 2024-07-17
[17.0][IMP] product_supplierinfo_for_customer: add search view

JordiMForgeFlow:


JasminSForgeFlow (approved):

JoanSForgeFlow (approved):
LGMT!

JordiBForgeFlow (approved):

Anxo82:
I've review code and it's OK, but I cant find view search. Could you say me where it is?

JordiMForgeFlow:
@Anxo82 is in Sales > Configuration > Prices for customers. Now you can perform search in that menu:

image

Anxo82:
@JordiMForgeFlow Sorry, I can't find it. I've review Sales > Configuration but it's not there.
imagen
Thanks

JordiMForgeFlow:
@Anxo82 keep in mind you need https://github.com/OCA/sale-workflow/tree/16.0/product_supplierinfo_for_customer_sale

Anxo82:
Ok, but It would be better add it in product_supplierinfo_for_customer_sale?, because if you dont have it  installed, you cant used search view.
Thanks

JordiMForgeFlow:
@Anxo82 I added the view here as the view by itself does not depend on having the sale module, so it made sense to me to centralize it with the rest of the views in the core module. Also, if someone adds another module to add the view in a different menu outside sale, then the search view would need to be added again in the other module.

Nonetheless, this will probably not happen, so for me it wouldn't be a problem to move it to the other module.

Anxo82 (approved):
LGTM! @JordiMForgeFlow 
Thanks

PR #1696, opened 2024-07-30
[14.0] [IMP] product_packaging_dimension: Volume writable

raumschmiede-joshuaL:
We have products that we send only via Dropshipment. The supplier gives us only the volume for the products, not the dimensions. For this case we want to manually set the volume.

I had to fix the tests because with `.new()` the inverse is not called and other things didn't work either. Took some time to find the problem.
Do you want the test fixes in a separate commit?

### DIFF
--- a/product_packaging_dimension/models/product_packaging.py
+++ b/product_packaging_dimension/models/product_packaging.py
@@ -1,4 +1,5 @@
 # Copyright 2019-2021 Camptocamp SA
+# Copyright 2024 Raumschmiede GmbH
 # License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl)
 from odoo import api, fields, models
 
@@ -65,8 +66,8 @@ class ProductPackaging(models.Model):
         "Volume",
         digits=(8, 4),
         compute="_compute_volume",
-        readonly=True,
-        store=False,
+        readonly=False,
+        store=True,
         help="The Packaging volume",
     )
 
--- a/product_packaging_dimension/tests/test_packaging_volume.py
+++ b/product_packaging_dimension/tests/test_packaging_volume.py
@@ -1,4 +1,5 @@
 # Copyright (C) 2021 Akretion (<http://www.akretion.com>).
+# Copyright (C) 2024 Raumschmiede GmbH.
 # License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
 
 from odoo.tests.common import TransactionCase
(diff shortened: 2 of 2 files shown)


raumschmiede-joshuaL:
@jbaudoux

product_packaging_dimension/tests/test_packaging_volume.py
```diff
@@ -103,5 +96,36 @@ def test_output_uom(self):
         self.packaging3.width = 10.0
         self.packaging3.length_uom_id = self.uom_m
         self.packaging3.volume_uom_id = self.uom_ft3
-        self.packaging3._compute_volume()
         self.assertAlmostEqual(35314.7248, self.packaging3.volume)
+
+    def test_manual_volume(self):
+        # Test whether volume can be set manually
```
jbaudoux:
```suggestion
        """Test whether volume can be set manually"""
```

jbaudoux (approved):
LGTM

On the product, the volume is editable. So it makes sense to have a similar behavior on the product packaging. Thanks

product_packaging_dimension/models/product_packaging.py
mt-software-de, line 1: In not dropshipment cases i think it is not good to remove the dimension and store the the new volume in the same field. 
I would rather use a specific manual volume field.
For example. If the dimension get set via shopfloor reception and later someone changes the volume manually, it would set the dimensions to 0 and on the next reception the dimension needs to be updated again.

mt-software-de (changes requested):
Thx. I don't know if it is really good to remove the dimensions.

product_packaging_dimension/models/product_packaging.py
jbaudoux: I initially thought it would be good to have aligned volume and dimensions but if the packaging is not cubic, you may want on purpose a different volume.
Then let's drop that inverse method and put `readonly=False`. Then it will also be aligned with the product behavior where setting the volume does not erase the dimensions.

mt-software-de: Just to point it out. it is nearly the same as the product behavior. :-)
Because the product volume changes only if the user change the dimensions via the form view, because a onchange is used. https://github.com/OCA/product-attribute/blob/3101e0b2440bfa81dd8f6d44b908d41b4808aa43/product_dimension/models/product.py#L14
Here a compute is used. But i think it is fine if something is changing the dimension again, that the volume gets recomputed.

raumschmiede-joshuaL: I removed the inverse and added the code suggestion

jbaudoux (approved):

product_packaging_dimension/models/product_packaging.py
mt-software-de: Maybe we need a migration script for computing the volume via sql. I will test this today with a big amount of data.

mt-software-de: I tested the computation of the new stored field with 53174 packagings. It just took seconds to finish.

mt-software-de (approved):
LGTM

hparfr:
~~/ocabot merge major~~

Oops, I forget I do not have right on this repo :)

raumschmiede-joshuaL:
@jbaudoux, can this PR be merged?

jbaudoux:
@OCA/product-maintainers Could you merge this PR? Thanks

PR #1703, opened 2024-08-07
[16.0][FIX] sale_product_template_tags: missing dependency

alessandro-fiorino:
The module `sale_product_template_tags` references the `product.product_tag` action, but the `product` module isn't referenced in the depends list.

### DIFF
--- a/sale_product_template_tags/__manifest__.py
+++ b/sale_product_template_tags/__manifest__.py
@@ -9,7 +9,7 @@
     "license": "AGPL-3",
     "author": "Camptocamp SA, Odoo Community Association (OCA)",
     "website": "https://github.com/OCA/product-attribute",
-    "depends": ["sale"],
+    "depends": ["sale", "product"],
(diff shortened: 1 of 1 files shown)


alessandro-fiorino:
@ivantodorovich could you review it please ?

ivantodorovich:
The `sale` module depends on `account`, which depends on `product`

So, this is redundant.

What lead you to this change? Perhaps it's another error

alessandro-fiorino:
Didn't noticed the indirect reference to product.
Dev builds on odoo.sh were failing complaining about the missing reference
to the  product.product_tag action.



Il giorno gio 8 ago 2024 alle ore 00:01 Iván Todorovich <
***@***.***> ha scritto:

ivantodorovich:
I'm not sure how dev builds are managed in Odoo.sh, but it sounds like the record might've been removed. Perhaps an update on the `product` module would solve it.

PR #1713, opened 2024-09-03
[14.0][FIX]product_assortment: whitelist preserves 'ALL' domain

PicchiSeba:
Before this PR, having records in the whitelist and no domain specified resulted in the filter being applied only on the items in the whitelist

Before
Screenshot from 2024-09-03 10-09-36

After
Screenshot from 2024-09-03 10-09-56


PicchiSeba:
TODO: add flag to include/exclude domain results

PR #1723, opened 2024-09-12
[16.0][IMP] product_abc_classification_sale_stock: prevent ambiguous column name

JordiMForgeFlow:
When using this module with product_variant_configurator, which adds a `product_id` field in the product_product model due to the inheritance (https://github.com/OCA/product-variant/blob/c133b59be157d48f6a65133432ffde52a1290f27/product_variant_configurator/models/product_product.py#L11), it causes an ambiguous name error.


pedrobaeza (approved):

MiquelRForgeFlow (approved):

LoisRForgeFlow (approved):
:+1:

AaronHForgeFlow (approved):
:+1:

PR #1725, opened 2024-09-19
[17.0][FIX] product_supplierinfo_for_customer: fix missing products in comodel search

xmglord:
When searching for products in a comodel, the results are obtained via the `_name_search` method over `product.product`. Currently, this search is not retrieving all matching results.

This issue occurs because the parameters for the `_name_search` method changed in Odoo v17, and during the migration, this was not adapted. As a result, when the method is called, it only retrieves 100 results due to the default limit not being handled correctly.

This commit updates the overridden `_name_search` method to match the new method signature in Odoo v17 and properly handle the `limit` parameter. This ensures that all matching products are returned during the search.

### DIFF
--- a/product_supplierinfo_for_customer/__manifest__.py
+++ b/product_supplierinfo_for_customer/__manifest__.py
@@ -12,6 +12,7 @@
     "website": "https://github.com/OCA/product-attribute",
     "category": "Sales Management",
     "license": "AGPL-3",
+    "maintainers": ["luisg123v"],
     "depends": ["product"],
     "data": [
         "security/ir.model.access.csv",
--- a/product_supplierinfo_for_customer/models/product_product.py
+++ b/product_supplierinfo_for_customer/models/product_product.py
@@ -16,9 +16,7 @@ class ProductProduct(models.Model):
         )._compute_display_name()
 
     @api.model
-    def _name_search(
-        self, name="", domain=None, operator="ilike", limit=100, order=None
-    ):
+    def _name_search(self, name, domain=None, operator="ilike", limit=None, order=None):
         res = super()._name_search(
             name, domain=domain, operator=operator, limit=limit, order=order
         )


xmglord:
@luisg123v

I explored the possibility of using the `_rec_names_search` property to avoid overriding the `_name_search` method. However, after analysis, I found that it's not feasible because a single product can have multiple `product.customerinfo` records, one for each customer.

While we can use dot notation and the `@property` decorator to reference related fields, Odoo's `_rec_names_search` property is not designed to handle searches over One2many relationships where multiple records need to be considered. Specifically, we cannot efficiently search across all `product.customerinfo` records associated with a product to match a search term in fields like `product_code` or `product_name`.

Given this limitation, the most effective solution is to fix the parameters of the overridden _name_search method to align with Odoo 17's changes.

xmglord:
@pedrobaeza @luisg123v please review

pedrobaeza (commented):
I don't get this change, as the signature is the previous one:

https://github.com/odoo/odoo/blob/4401d4f9af7698ee79aee82bda3b240ef322bf43/odoo/models.py#L1714

xmglord:
> I don't get this change, as the signature is the previous one:
> 
> https://github.com/odoo/odoo/blob/4401d4f9af7698ee79aee82bda3b240ef322bf43/odoo/models.py#L1714

The inheritance is over this one:

https://github.com/odoo/odoo/blob/4401d4f9af7698ee79aee82bda3b240ef322bf43/odoo/models.py#L1750

And the change is because in the replanishment (or any other model with a field related to product.product) if you search for a product it will only take into account the first hundred products which is not expected.

pedrobaeza (approved):
OK, sorry, I lost the initial `_`.

luisg123v:
Hi @xmglord,

LGTM.

Could we take this chance to add myself as maintainer, please?

Regards,

xmglord:
> Hi @xmglord,
> 
> LGTM.
> 
> Could we take this chance to add myself as maintainer, please?
> 
> Regards,

Done, also I test again the dot notation and the error was a typo, but it do not work using `_rec_name_search` since it used in `_name_search` in [here](https://github.com/odoo/odoo/blob/17.0/odoo/models.py#L1758) but there's an inheritance in odoo that do not use the model property in [here](https://github.com/odoo/odoo/blob/17.0/addons/product/models/product_product.py#L530) so it will not be possible to use.

luisg123v (approved):
LGTM :+1: 

CC @desdelinux @moylop260

marielejeune (approved):
LGTM.

Could we merge this so I can take it into the 18.0 migration I'm working on?

xmglord:
@pedrobaeza this is ready for merge

desdelinux (approved):

PR #1728, opened 2024-09-24
[17.0][ADD] product_identification.

Murtaza-OSI:


product_identification/models/productid_category.py
```diff
@@ -0,0 +1,8 @@
+from odoo import fields, models
+
+
+class ProductIdCategory(models.Model):
+    _name = "product.product.id_category"
```
legalsylvain: except an historical error in odoo, there is no "_" in model name.
maybe "product.identification.category" is enought. don't you think ? (and product.identification instead of product.product.id_number)

product_identification/models/product_template.py
```diff
@@ -0,0 +1,9 @@
+from odoo import fields, models
+
+
+class ProductTemplate(models.Model):
+    _inherit = "product.template"
+
+    id_numbers = fields.One2many(
+        "product.product.id_number", "product_id", string="Identification Numbers"
```
legalsylvain: it doesn't looks consistent.
``    id_numbers = fields.One2many() ``
in product.template
but in the other model, it is in product.product

``    product_id = fields.Many2one("product.product", string="Product")``

I think you should choose between product and template.

product_identification/models/product_template.py
```diff
@@ -0,0 +1,9 @@
+from odoo import fields, models
+
+
+class ProductTemplate(models.Model):
+    _inherit = "product.template"
+
+    id_numbers = fields.One2many(
```
legalsylvain:
```suggestion
    identification_ids = fields.One2many(
```

product_identification/models/productid_number.py
```diff
@@ -0,0 +1,50 @@
...
+
+    name = fields.Char(string="ID Number", required=True)
+    category_id = fields.Many2one(
+        "product.product.id_category", string="Category", required=True
+    )
+    issued_by = fields.Many2one("res.partner")
+    date_issued = fields.Date()
+    expiry_date = fields.Date()
+    place_of_issue = fields.Char(string="Place of Issue")
+    product_id = fields.Many2one("product.product", string="Product")
```
legalsylvain:
```suggestion
    product_id = fields.Many2one("product.product", string="Product", required=True)
```
I guess it doesn't make sense to have an inentification without product. Or did i missed something ?

legalsylvain (changes requested):
Hi. Thanks for contributing ! 

some remarks inline, one blocking.

otherwise, LGTM. I think that maybe this feature is present in OCA, but I don't remember which module...

product_identification/models/productid_category.py
Murtaza-SerpentCS: Improved

PR #1730, opened 2024-09-27
[14.0][ADD] product_pricelist_overcharge_by_amount

christian-ramos-tecnativa:
This module enables adding or subtracting amounts from a final pricelist item, applying rules directly to the price.


product_pricelist_overcharge_by_amount/models/product_pricelist.py
```diff
@@ -0,0 +1,160 @@
...
+        "product.pricelist.overcharge.item",
+        "item_id",
+        "Overcharge Items",
+        help="Only match prices from the selected supplier",
+        copy=True,
+    )
+
+
+class ProductPricelistOverchargeItem(models.Model):
+    _name = "product.pricelist.overcharge.item"
```
rousseldenis, line 46: @Christian-RB Please split one model per file. Thanks

christian-ramos-tecnativa: @rousseldenis Thanks for the review and apologies for the delayed response. AFAIK if the model is used as a `One2many` field in the main model, it was acceptable to include it in the same file. Has this approach changed?

product_pricelist_overcharge_by_amount/models/product_pricelist.py
```diff
@@ -0,0 +1,160 @@
...
+
+    sequence = fields.Integer(string="Sequence", default=1)
+    item_id = fields.Many2one(
+        "product.pricelist.item",
+        "Pricelist Item",
+        ondelete="cascade",
+    )
+    applied_on = fields.Selection(
+        [
+            ("allways", "Allways"),
```
rrebollo, line 56:
```suggestion
            ("always", "Always"),
```
There is a typo

product_pricelist_overcharge_by_amount/models/product_pricelist.py
```diff
@@ -0,0 +1,160 @@
...
+            ("equal", "Equal than"),
+            ("smaller", "Smaller/Equal than"),
+            ("bigger", "Bigger/Equal than"),
+            ("between", "Between"),
+        ],
+        "Apply",
+        default="allways",
```
rrebollo, line 63:
```suggestion
        default="always",
```

product_pricelist_overcharge_by_amount/models/product_pricelist.py
```diff
@@ -0,0 +1,160 @@
...
+            )
+        return True
+
+    def can_be_apply(self, price):
+        """Check if the overcharge is applicable"""
+        self.ensure_one()
+        if hasattr(self, "%s_can_be_apply" % self.applied_on):
+            return getattr(self, "%s_can_be_apply" % self.applied_on)(price)
+
+    def allways_can_be_apply(self, price):
```
rrebollo, line 109:
```suggestion
    def always_can_be_apply(self, price):
```

product_pricelist_overcharge_by_amount/readme/DESCRIPTION.rst
rrebollo, line 1: Thanks a lot for the effort you’ve already put into documenting the addon — really appreciate it!

Since you’re building on top of the standard pricelist functionalities, and given that Odoo Community already handles both surcharge and discount (percentage-based) at the pricelist level, it would really help if the documentation could explicitly highlight **what’s missing in Community that this addon comes to extend or complete**.

Maybe a simple comparison — showing where native stops and where your module steps in — could clarify the functional gap this addresses. I think that would make it much easier for others to understand the specific value this addon brings and in which use cases it would be the right fit.

rrebollo (changes requested):
Code Review: Great work! The code looks good to me (LGTM). Thanks so much for your contribution!  

I’ve left a few mandatory requests — the typo in “allways” is probably the most obvious one — along with some additional suggestions for your consideration. Feel free to address them as you see fit.

Also, runboat is missing.

product_pricelist_overcharge_by_amount/__manifest__.py
```diff
@@ -0,0 +1,17 @@
+# Copyright 2024 Binhex - Christian Ramos
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+{
+    "name": "Amount overcharge in sales pricelists",
+    "summary": "Allows to add/substract amount to the final priceslists price",
+    "version": "14.0.0.1.0",
```
edescalona, line 6: Hi @Christian-RB, wouldn't the version be 14.0.1.0.0? Thanks for your contribution.

edescalona (changes requested):

PR #1737, opened 2024-10-01
[17.0][FIX] product_print_category: Fix user has no group

bosd:
on rule: print category rule user

```
WARNING odoo odoo.addons.base.models.ir_model: Rule Print Category Rule User has no group, this is a deprecated feature. Every access-granting rule should specify a group. 
```

To silenece warning in odoo logger.


bosd:
@rf-ife Can you please review? :pray:

legalsylvain (approved):
Make sense.

rf-ife (approved):

PR #1739, opened 2024-10-03
[16.0][FIX] product_cost_security: apply the same logic on the label

etobella:
@luisDIXMIT


sergio-teruel (approved):

pedrobaeza (approved):
/ocabot merge minor

yajo:
I'm testing this in the runboat previous to the merge. This was the case with permission:

image

And this, without permission:

image

So, although the PR is theoretically correct because it hides the label together with the field... that was already happening before, right? So what is this actually doing?

PR #1745, opened 2024-10-10
[18.0] [MIG] product_pricelist_supplierinfo

marielejeune:


JordiBForgeFlow:
Hi @pedrobaeza @marielejeune We have proposed to rename 'margin' to 'markup'. 
https://github.com/acsone/product-attribute/pull/9

Markup =  (sales price - cost) / cost 
Margin = (sales price - cost) / sales price

See https://www.investopedia.com/ask/answers/102714/whats-difference-between-profit-margin-and-markup.asp

pedrobaeza:
Well, as here we are doing sales pricelists, we are not talking about costs, but the sales margin you want. It's a very common nomenclature to call it "margin over cost".

JordiBForgeFlow:
Pedro, the problem is that the spanish equivalent to 'Markup' is 'Margen
sobre coste'. But you cannot translate this spanish term to english
literally, but as Markup.

JordiBForgeFlow:
Besides, it's important to name it correctly to avoid confusion, as the
calculation being made is the Markup, not the margin

product_pricelist_supplierinfo/models/product_supplierinfo.py
```diff
@@ -0,0 +1,21 @@
...
+        default=0,
+        digits=(16, 2),
+        help="Margin to apply on price to obtain sale price",
+    )
+
+    def _get_supplierinfo_pricelist_price(self):
+        self.ensure_one()
+        sale_price = self.price
+        if self.sale_margin:
+            sale_price = (self.price + (self.price * (self.sale_margin / 100))) or 0.0
```
florian-dacosta, line 20: Since v17, there is a discount field on supplierinfo and a `price_discounted`
If you want to sale a product depending on the price you buy it, this is the price_discounted that should be used, right ?
Or at least, if not by default, it should be an option in the module ?
What do you think ? @pedrobaeza @JordiBForgeFlow @marielejeune

pedrobaeza: Yeah, I think it should be included in this module.

CLaurelB: Hello, and sorry to jump in, but maybe a field named like `no_supplierinfo_discount` could work, similar to the existing field `no_supplierinfo_min_quantity`: https://github.com/OCA/product-attribute/blob/5c4e5f6c/product_pricelist_supplierinfo/models/product_pricelist_item.py#L15.

The idea would be to use this field to bypass the discount in the supplier info, passing its value here as a parameter: https://github.com/OCA/product-attribute/blob/5c4e5f6c/product_pricelist_supplierinfo/models/product_template.py#L44

so that you can select between `self.price` and `self.price_discounted` using `no_supplierinfo_discount` as a validation here: https://github.com/OCA/product-attribute/blob/5c4e5f6c/product_pricelist_supplierinfo/models/product_supplierinfo.py#L16

This could be implemented in version 17.0.

Regards.

product_pricelist_supplierinfo/views/product_supplierinfo_view.xml
```diff
@@ -0,0 +1,38 @@
...
+
+    <record id="product_supplierinfo_tree_view" model="ir.ui.view">
+        <field name="name">product.supplierinfo.tree.view.pricelist</field>
+        <field name="model">product.supplierinfo</field>
+        <field name="inherit_id" ref="product.product_supplierinfo_tree_view" />
+        <field name="arch" type="xml">
+            <field name="price" position="after">
+                <field
+                    name="sale_margin"
+                    string="Sale margin"
```
CLaurelB, line 12: Maybe "Sale Margin (%)" to make it clearer, similar to the discount field: https://github.com/odoo/odoo/blob/75571a66/addons/product/models/product_supplierinfo.py#L65

marielejeune:
Superseeded by https://github.com/OCA/product-attribute/pull/1837

PR #1753, opened 2024-10-29
[18.0][MIG] product_medical: Migration to 18.0

LucasTran380381:
Changes:

- remove string redundant in fields
- remove migrations directory 
- change tree view to list view


ivantodorovich (approved):

TDu (approved):

sebalix (approved):

rousseldenis (approved):
Code review.

Even if some fields in product template should maybe include 'medical' in their names to avoid possible conflicts

LucasTran380381:
> Code review.
> 
> Even if some fields in product template should maybe include 'medical' in their names to avoid possible conflicts

@rousseldenis could you explain possible conflict ?

sebalix (commented):
Getting back on my first review: there are two pre-commit commits, and the 2nd one is bumping the version number wrongly, then it's reverted in the migration commit afterwards. This has to be fixed.

Regarding a "medical" prefix for fields that could conflicts with other non-medical topics, this is a good improvement indeed. Can you handle this with the proper migration script? :pray:

LucasTran380381:
> Getting back on my first review: there are two pre-commit commits, and the 2nd one is bumping the version number wrongly, then it's reverted in the migration commit afterwards. This has to be fixed.
> 
> Regarding a "medical" prefix for fields that could conflicts with other non-medical topics, this is a good improvement indeed. Can you handle this with the proper migration script? 🙏

@sebalix thanks for your suggestion. I added migrations script and fixed commit

product_medical/models/product_template.py
```diff
@@ -7,34 +7,36 @@
...
     )
-    doc_lot_related = fields.Boolean(string="Lot Related", default=False)
-    doc_validity_date = fields.Date(string="Validity Date")
-    ce_certificate_medical_class_ids = fields.Many2many(
+    medical_doc_lot_related = fields.Boolean(string="Lot Related", default=False)
+    medical_doc_validity_date = fields.Date(string="Validity Date")
+    medical_ce_certificate_medical_class_ids = fields.Many2many(
         "ir.attachment",
-        relation="product_ce_certificate_medical_class_rel",
+        relation="product_medical_ce_certificate_medical_class_rel",
```
simahawk: These changes of relation tables are not migrated AFAIS.

product_medical/migrations/18.0.1.0.0/pre-migrate.py
```diff
@@ -0,0 +1,42 @@
...
+}
+
+
+def _rename_fields(env):
+    openupgrade.rename_fields(
+        env,
+        [
+            (
+                model_name,
+                model_name.replace(".", "_"),
```
sebalix: This conversion is not always true, to get the name of the table, that would be better to use `env[model_name]._table`, but at the same time:
- using the registry in a pre script is sometimes not a good idea
- no issue here for `product.template` data model, that will work

So not blocking :man_shrugging:

product_medical/models/product_template.py
```diff
@@ -7,34 +7,36 @@
...
         column1="product_template_id",
         column2="attachment_id",
         string="Declaration of Conformity",
     )
-    doc_lot_related = fields.Boolean(string="Lot Related", default=False)
-    doc_validity_date = fields.Date(string="Validity Date")
-    ce_certificate_medical_class_ids = fields.Many2many(
+    medical_doc_lot_related = fields.Boolean(string="Lot Related", default=False)
+    medical_doc_validity_date = fields.Date(string="Validity Date")
+    medical_ce_certificate_medical_class_ids = fields.Many2many(
```
sebalix: We could name it `medical_ce_certificate_class_ids` I guess, removing the second `medical` occurrence.

product_medical/models/product_template.py
sebalix: `product_medical_ce_certificate_class_rel` based on my previous comment

product_medical/views/product_template.xml
```diff
@@ -19,22 +19,22 @@
                         <group name="conformity_left" string="Categories">
                             <field name="medical_class_id" />
                             <field name="medicine_category_id" />
-                            <field name="ppe_category_id" />
```
sebalix: We should rename the `conformity` page to `medical_conformity` still to avoid conflicts with other modules

sebalix (approved):
Tested on runboat, LG.
Regarding migration script, I could not test it but if something doesn't work we'll fix it ASAP anyway.

PR #1757, opened 2024-11-04
[16.0][FW] [14.0][FIX] product_stock_state : product_template_view

henrybackman:
Port of #1066 from 14.0 to 16.0.

PR #1760, opened 2024-11-06
[FIX] uom_category_active: CI is broken. because onchange doesn't raise an error

legalsylvain:
Swith from an onchange logic to an create / write logic to be sure to handle all the cases.

For the time being, several 16.0 are blocked because CI is broken. 

https://github.com/OCA/product-attribute/pull/1757
https://github.com/OCA/product-attribute/pull/1657


simahawk (approved):
LG

henrybackman (approved):

PR #1754, opened 2024-10-31
[16.0][TEST] Do not merge. Test uom_category_active CI

legalsylvain:
Hi. This is a Test PR.

Context : I have a weird error in my PR : https://github.com/OCA/product-attribute/pull/1657

The error is 
```
2024-10-31 13:13:22,733 485 ERROR odoo odoo.addons.uom_category_active.tests.test_uom_category_active: FAIL: TestProductCategoryActive.test_archive_reference_uom
Traceback (most recent call last):
  File "/__w/product-attribute/product-attribute/uom_category_active/tests/test_uom_category_active.py", line 33, in test_archive_reference_uom
    with self.assertRaises(UserError):
  File "/usr/lib/python3.10/contextlib.py", line 142, in __exit__
    next(self.gen)
  File "/opt/odoo/odoo/tests/common.py", line 382, in _assertRaises
    with ExitStack() as inner:
  File "/usr/lib/python3.10/contextlib.py", line 576, in __exit__
    raise exc_details[1]
  File "/usr/lib/python3.10/contextlib.py", line 561, in __exit__
    if cb(*exc_details):
AssertionError: UserError not raised
```

PR #1761, opened 2024-11-06
[16.0] uom_category_active: disable broken test

simahawk:
See https://github.com/OCA/product-attribute/pull/1754


simahawk:
It has been reworked here https://github.com/OCA/product-attribute/pull/1760

PR #1683, opened 2024-07-09
[16.0][ADD] product_code_unique_multicompany

kevinkhao:

PR #1561, opened 2024-03-21
[ADD] product_pricelist_margin

nguyenminhchien:
This module shows the product's cost and margin from the pricelists.

PR #1763, opened 2024-11-07
[16.0][FIX] add addClassCleanup in testcase

chaule97:
The test class in the `product_abc_classification_sale_stock` module has the same name as a test class in `product_abc_classification`, which is causing an error in GitHub CI in https://github.com/OCA/product-attribute/pull/1561, https://github.com/OCA/product-attribute/pull/1683

### DIFF
--- a/base_product_mass_addition/tests/test_product_mass_addition.py
+++ b/base_product_mass_addition/tests/test_product_mass_addition.py
@@ -20,6 +20,7 @@ class TestProductMassAddition(TransactionCase):
         # Setup Fake Models
         cls.loader = FakeModelLoader(cls.env, cls.__module__)
         cls.loader.backup_registry()
+        cls.addClassCleanup(cls.loader.restore_registry)
         from .models.order import ModelOrder, ModelOrderLine
 
         cls.loader.update_registry(
--- a/product_packaging_container_deposit/tests/test_container_deposit_order_mixin.py
+++ b/product_packaging_container_deposit/tests/test_container_deposit_order_mixin.py
@@ -15,6 +15,7 @@ class TestProductPackagingContainerDepositMixin(Common):
         super().setUpClass()
         cls.loader = FakeModelLoader(cls.env, cls.__module__)
         cls.loader.backup_registry()
+        cls.addClassCleanup(cls.loader.restore_registry)
         from .fake_models import (
             ContainerDepositOrderLineTest,
             ContainerDepositOrderTest,
--- a/product_secondary_unit/tests/test_secondary_unit_mixin.py
+++ b/product_secondary_unit/tests/test_secondary_unit_mixin.py
@@ -11,6 +11,7 @@ class TestProductSecondaryUnitMixin(TransactionCase, FakeModelLoader):
         super().setUpClass()
         cls.loader = FakeModelLoader(cls.env, cls.__module__)
         cls.loader.backup_registry()
+        cls.addClassCleanup(cls.loader.restore_registry)
         from .models import SecondaryUnitFake
 
         cls.loader.update_registry((SecondaryUnitFake,))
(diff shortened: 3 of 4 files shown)


rousseldenis (approved):

rousseldenis:
> The test class in the `product_abc_classification_sale_stock` module has the same name as a test class in `product_abc_classification`, which is causing an error in GitHub CI in #1561, #1683

@chaule97 I agreed on name change because it is duplicated from base module.

But the errors you point to are not related to that.

Please highlight the real problem you want to solve

lmignon (changes requested):
Thank you for the contribution but the name of a test class should never cause problem when running tests... IMO the diagnostic is not right. Moreover when you change the name of a class into a module named as the call, the module must also be renamed....

rousseldenis:
@chaule97 https://github.com/OCA/product-attribute/blob/16.0/base_product_mass_addition/tests/test_product_mass_addition.py#L21

This is the real error.

Missing a teardown(). See https://github.com/OCA/odoo-test-helper/blob/master/README.rst

You do the fix ?

rousseldenis:
Or @legalsylvain ?

chaule97:
> @chaule97 https://github.com/OCA/product-attribute/blob/16.0/base_product_mass_addition/tests/test_product_mass_addition.py#L21
> 
> This is the real error.
> 
> Missing a teardown(). See https://github.com/OCA/odoo-test-helper/blob/master/README.rst
> 
> You do the fix ?

Thanks. I will continue it

legalsylvain:
Feel free to add the teardown in this PR.

chaule97:
Thanks everyone, I have added teardown

legalsylvain:
thanks @chaule97 !

legalsylvain (approved):

base_product_mass_addition/tests/test_product_mass_addition.py
```diff
@@ -36,6 +36,11 @@ def setUpClass(cls):
             **cls.quick_ctx
         )
 
+    @classmethod
+    def tearDownClass(cls):
+        cls.loader.restore_registry()
+        return super().tearDownClass()
```
lmignon: It's better to register the call to the `restore_registry` method in the list of cleanup methods to execute when all the tests into the class are executed just after the fake registry is backed up
```python
cls.loader.backup_registry()
cls.addClassCleanup(cls.loader.restore_registry)
``` 

In this way, even if an exception occurs into your setup, the registry will be cleaned up

lmignon (changes requested):

lmignon (changes requested):
see my last comment

base_product_mass_addition/tests/test_product_mass_addition.py
rousseldenis: @lmignon Isn't it supported by test class ?

If not, odoo-test-helper documentation should be adapted too.

chaule97: @lmignon I have changed it

product_abc_classification_sale_stock/tests/test_abc_classification_profile.py
```diff
@@ -6,7 +6,7 @@
 from odoo.tests.common import TransactionCase
 
 
-class TestABCClassificationProfile(TransactionCase):
+class TestABCClassificationProfileSaleStock(TransactionCase):
```
lmignon: @chaule97 This change is useless but if you change the classname you must change the filename.

lmignon (changes requested):

product_abc_classification_sale_stock/tests/test_abc_classification_profile.py
chaule97: I have renamed

lmignon: IMO this change is useless and we must avoid useless changes. Is-it really a problem for you?

chaule97: okay, I have removed this commit. I remember Odoo 17 have issue when it have duplicate testcase class name. I think that it was fixed in version 18.0

https://github.com/OCA/pos/pull/1232

lmignon (approved):
Thank you  for this fix!

PR #1776, opened 2024-11-20
[16.0][FIX] product_supplierinfo_stock_picking_type: call _compute_price_unit_and_date_planned_and_name on change PO picking type

AlexPForgeFlow:
Core Odoo method has changed behaviour, now on onchange_product_id prices are not recomputed, so we need to call _compute_price_unit_and_date_planned_and_name instead to ensure that when we change the picking type, prices will be recomputed.


AlexPForgeFlow:
Closing. That was not an issue

PR #1777, opened 2024-11-21
[16.0][FIX] product_supplierinfo_stock_picking_type: rename on change picking type method

AlexPForgeFlow:
If module mrp_subcontracting_dropshipping, form Odoo, is installed in our system, there are conflicts between onchange methods, mrp_subcontracting_dropshipping method https://github.com/odoo/odoo/blob/16.0/addons/mrp_subcontracting_dropshipping/models/purchase.py#L23 is overriding the onchange method previously defined in this module. So I suggest renaming the onchange method.

### DIFF
--- a/product_supplierinfo_stock_picking_type/models/purchase_order.py
+++ b/product_supplierinfo_stock_picking_type/models/purchase_order.py
@@ -7,6 +7,8 @@ class PurchaseOrder(models.Model):
     _inherit = "purchase.order"
 
     @api.onchange("picking_type_id")
-    def onchange_picking_type_id(self):
+    def onchange_picking_type_id_onchange_product(self):
+        # Method name is to avoid conflicts with mrp_subcontracting_dropshipping
+        # that have an onchange method named onchange_picking_type_id
         for line in self.order_line:
             line.onchange_product_id()
--- a/product_supplierinfo_stock_picking_type/tests/test_product_supplierinfo_stock_picking_type.py
+++ b/product_supplierinfo_stock_picking_type/tests/test_product_supplierinfo_stock_picking_type.py
@@ -102,5 +102,5 @@ class TestProductSupplierinfoStockPickingType(BaseCommon):
         po = self._create_purchase_order(self.picking_in_c)
         self.assertEqual(po.order_line.price_unit, 20)
         po.picking_type_id = self.picking_in_a
-        po.onchange_picking_type_id()
+        po.onchange_picking_type_id_onchange_product()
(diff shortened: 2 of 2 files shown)


GuillemCForgeFlow (approved):
makes sense 👍🏿

JordiMForgeFlow (approved):

product_supplierinfo_stock_picking_type/models/purchase_order.py
```diff
@@ -7,6 +7,6 @@ class PurchaseOrder(models.Model):
     _inherit = "purchase.order"
 
     @api.onchange("picking_type_id")
-    def onchange_picking_type_id(self):
```
MiquelRForgeFlow, line 10: I would add a commented note here saying that this name is to avoid conflict with the one in mrp_subcontracting_dropshipping

MiquelRForgeFlow (approved):

JoanSForgeFlow (approved):

product_supplierinfo_stock_picking_type/models/purchase_order.py
```diff
@@ -7,6 +7,8 @@ class PurchaseOrder(models.Model):
...
+    def onchange_picking_type_id_onchange_product(self):
+        # Method name is to avoid conflicts with mrp_subcontracting_dropshipping
+        # that have an onchange method named onchange_picking_type_id
```
victoralmau, line 12: I think you can keep the same method but call the super if the method exists.

```
if hasattr(super(), "onchange_picking_type_id"):
    getattr(super(), "onchange_picking_type_id)
```

victoralmau (changes requested):

product_supplierinfo_stock_picking_type/models/purchase_order.py
AlexPForgeFlow: Hi @victoralmau !
No, as I already mentioned, the Odoo function is overriding the method, so in the module hierarchy, product_supplierinfo_stock_picking_type comes before mrp_subcontracting_dropshipping. For this reason, the method is being overridden. To implement your solution, we would need to add it to Odoo's method.

victoralmau (approved):

sebalix (approved):

zamberjo (approved):

AlexPForgeFlow:
@victoralmau Can we get it merged?

victoralmau:
Ping @pedrobaeza

pedrobaeza (approved):
/ocabot merge patch

PR #1779, opened 2024-11-22
[FIX] product_uom_measure_type: Fixed the issue, that the module cannot be installed #1778

### LINKED ISSUE #1778: Cannot install product_uom_measure_type on migrated databases
Cannot install product_uom_measure_type on migrated databases

## Module
product_uom_measure_type

## Describe the bug
When installing this module on migrated databases (in my case coming from version 13.0, where the measure_type was a odoo core field) we get an error message, that column "meaure_type" already exists. 
This happens, because in the init hook we actively want to add a column "meaure_type" in different tables, without checking, if this column may already exists.

## To Reproduce
Try to install product_uom_measure_type on a migrated database (migrated from 13.0 to some higher version)

1. Migrate database from 13.0 to any higher version
2. Install product_uom_measure_type
3.

**Expected behavior**
The module can be installed also on migrated databases, where the column already exists.

baguenth:
[FIX] product_uom_measure_type: Fixed the issue, that the module cannot be installed on migrated databases which already have a column 'measure_type'

This pr fixed https://github.com/OCA/product-attribute/issues/1778

@dreispt @legalsylvain @rousseldenis 

Kindly ask to check this pr :)

### DIFF
--- a/product_uom_measure_type/hooks.py
+++ b/product_uom_measure_type/hooks.py
@@ -17,7 +17,7 @@ def pre_init_hook(cr):
     cr.execute(
         """
         ALTER TABLE uom_category
-        ADD column measure_type character varying;
+        ADD column IF NOT EXISTS measure_type character varying;
         """
     )
 
@@ -50,7 +50,7 @@ def pre_init_hook(cr):
     cr.execute(
         """
         ALTER TABLE uom_uom
-        ADD column measure_type character varying;
+        ADD column IF NOT EXISTS measure_type character varying;
         """
     )
 
@@ -67,7 +67,7 @@ def pre_init_hook(cr):
     cr.execute(
         """
         ALTER TABLE product_template
-        ADD column uom_measure_type character varying;
+        ADD column IF NOT EXISTS uom_measure_type character varying;
         """
     )
 


legalsylvain (commented):
Hi. thanks for fixing that issue ! 

note : The implementation is correct, but I wonder if it's not better to test if the column exist, and if yes, return and doesn't execute code. if avoids all the UPDATE records that are, I guess useless if the column still exists.
What do you think ? 

Kind regards.

baguenth:
Hi @legalsylvain, 

thank you for the quick feedback! 

I think we still need to run the update queries. The reason is, that after 13.0 at leaset two new uom categories got introduced: uom_categ_wtime and uom_categ_surface

All three update functions are actually build on top of each other

1) Update measure_type in uom_category --> Here we at least need to set measure_type for uom_categ_wtime and uom_categ_surface
2) Update measure_type in uom_uom, based on uom_category --> Due to 1), there could be a change here
3) Update measure_type in product_template, based on uom_uom --> Due to 2), there could be a change here

Technically we could reduce the update to only handle the new uom categories, in case the column measure_type already exists. This would increase the complexity of the hook a lot, though. Since the update functions do not really hurt and are also uncritical regarding performance, I would rather keep it simple.

legalsylvain (approved):
Thanks a lot for your explanation ! 


/ocabot merge patch

PR #1782, opened 2024-11-25
[16.0][IMP] product_pricelist_direct_print add hooks

mmequignon:
This pull requests add hooks to the following modules:
 - product_pricelist_direct_print
 - product_pricelist_direct_print_xlsx

This is meant to make overrides of the reports easier.

### DIFF
--- a/product_pricelist_direct_print/wizards/product_pricelist_print.py
+++ b/product_pricelist_direct_print/wizards/product_pricelist_print.py
@@ -93,6 +93,7 @@ class ProductPricelistPrint(models.TransientModel):
 
     @api.depends_context("product")
     def _compute_product_price(self):
+        # TODO drop this computed field
         product = self.env.context["product"]
         price = self.get_pricelist_to_print()._get_product_price(
             product, 1, date=self.date
@@ -108,6 +109,20 @@ class ProductPricelistPrint(models.TransientModel):
             )
         else:
             self.product_price = float_round(price, precision)
+
+    def get_price_for_pricelist(self, pricelist, product):
+        # TODO enable this instead
+        # pricelist.ensure_one()
+        # product.ensure_one()
+        # price = pricelist._get_product_price(
+        # product, 1, date=self.date
+        # )
+        # if self.vat_mode == "vat_excl":
+        # return product.taxes_id.compute_all(price)["total_excluded"]
+        # elif self.vat_mode == "vat_incl":
+        # return product.taxes_id.compute_all(price)["total_included"]
+        # return price
+        return self.with_context(product=product).product_price
 
     @api.depends("partner_ids")
     def _compute_partner_count(self):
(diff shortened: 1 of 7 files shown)


sebalix (approved):
LG, always nice to get additional tests

TDu (approved):

santostelmo (approved):

mmequignon:
@legalsylvain @CarlosRoca13 
Hi! Could I please get your insights on this?
Thanks!

product_pricelist_direct_print/wizards/product_pricelist_print.py
```diff
@@ -95,6 +96,20 @@ def _compute_product_price(self):
         else:
             self.product_price = price
 
+    def get_price_for_pricelist(self, pricelist, product):
+        # TODO enable this instead
```
legalsylvain, line 114: Is it wip ?

mmequignon: More like an open question for reviewers.
Is there a benefit for using a computed field depending on the context rather than using a method?

legalsylvain: I think your right. I'm not sure to understand the current design. 
however, changing the code could break possible module overloading this module. Not sure what to think here.
In all cases, if you want to introduce questions, please add better a note in the ROADMAP.rst file that mention "replace computed function xxx by yyy when porting this module". i think it's better than adding commented code with TODO.
What do you think ?

CarlosRoca13: I agree with @legalsylvain, adding these changes to the module at this stage could cause inheritance issues in other modules that rely on them.

CarlosRoca13 (changes requested):
The definition of the methods is being changed, which will cause errors in modules that use this one.

mmequignon:
> The definition of the methods is being changed, which will cause errors in modules that use this one.

Well, I even added tests for that only module depending on this one, and using those `_add_header` methods.
I think the state of this module with those changes is better.

santostelmo:
@mmequignon can you please rebase ?

cyrilmanuel:
could we reopen this PR and try to rebase the code ?

cyrilmanuel:
@legalsylvain  thanks a lot :) 

@CarlosRoca13  hi , could you check if the modifications seems good for you ? we want to merge this PR

cyrilmanuel:
@legalsylvain  is it possible to merge this PR ? i've inculded the fix on branch 16.0 about template and rebase the code. seems to be good as @CarlosRoca13 didn't add any update on that topic :s

CarlosRoca13:
My comment still beeing valid... If I have a private module that is extending this one this changes will break my module :sweat_smile:

product_pricelist_direct_print/wizards/product_pricelist_print.py
sergio-teruel: I agree Too

sergio-teruel (changes requested):

mmequignon:
> My comment still beeing valid... If I have a private module that is extending this one this changes will break my module 😅

Yeah well, this PR is all about making this module easier to extend.
IMO those signatures weren't optimal, and this version is better.
I took care of the updating the references in the OCA.

Honestly this has been opened for 1 year and half…
Can someone take a decision please…

@OCA/product-maintainers

cyrilmanuel:
@OCA/product-maintainers  please can we merge this pr or status and close this ?

PR #1787, opened 2024-11-27
[18.0][MIG] product_net_weight

alexis-via:


rvalyi (approved):

luc-adhoc (approved):
functional review

product_net_weight/models/product_product.py
```diff
@@ -4,23 +4,35 @@
...
+            if (
+                not float_is_zero(product.weight, precision_digits=prec)
+                and float_compare(
+                    product.net_weight, product.weight, precision_digits=prec
+                )
+                > 0
+            ):
                 raise ValidationError(
-                    _("The net weight of product must be lower than gross weight.")
+                    _("The net weight of product '%s' must be lower than gross weight.")
```
rousseldenis, line 36: You should use translation function arguments instead.

rousseldenis (changes requested):
Little change

product_net_weight/models/product_product.py
alexis-via: It is not required when you have a single variable term

rvalyi: indeed, there are no warning. Merging this...

PR #1752, opened 2024-10-25
[18.0][MIG] product_packaging_dimension: Migration to 18.0

LucasTran380381:
depend: PR 
- https://github.com/OCA/product-attribute/pull/1751

PR #1802, opened 2024-12-05
[18.0][MIG] product_packaging_type_pallet: Migration to 18.0

trisdoan:
### Note
- depends on: 
   - https://github.com/OCA/product-attribute/pull/1751
   - https://github.com/OCA/product-attribute/pull/1752


product_packaging_type_pallet/views/product_packaging_type.xml
```diff
@@ -1,16 +1,13 @@
...
 <odoo>
     <record id="view_product_packaging_type_form_inherit" model="ir.ui.view">
-        <field name="name">product.packaging.type.form.inherit</field>
-        <field name="model">product.packaging.type</field>
-        <field
-            name="inherit_id"
-            ref="product_packaging_type.view_product_packaging_type_form"
-        />
+        <field name="name">stock.package.type.form.inherit</field>
+        <field name="model">stock.package.type</field>
```
sebalix, line 5: I would rename the file `stock_package_type.xml` accordingly

product_packaging_type_pallet/models/product_packaging.py
```diff
@@ -3,15 +3,15 @@
 from odoo import fields, models
 
 
-class ProductPackagingType(models.Model):
-    _inherit = "product.packaging.type"
```
sebalix: Because of this, should we rename the module `product_packaging_type_pallet` to `stock_package_type_pallet`?
We also have already some `stock_package_type_*` modules (in v16) hosted in OCA/stock-logistics-warehouse. Should it be moved there?

@jbaudoux @rousseldenis what do you think?

jbaudoux: @sebalix This module is obsolete. It was a dependency of the reception screen but this is being replaced by shopfloor_reception. At least, that's the plan. And we need to make this configurable on the package type https://github.com/OCA/wms/blob/14.0/shopfloor_reception_packaging_dimension/services/reception.py#L26
so that you can configure on pallet that you only want the height without width/length

trisdoan:
Took recommendation from @jbaudoux, this PR should be closed

PR #1803, opened 2024-12-05
[18.0][MOV] product_packaging_type_required becomes product_packaging_level_required

trisdoan:
### Note
- depends on: https://github.com/OCA/product-attribute/pull/1806
### This change
- This module's feature was extracted from the original [product_packaging_type_required](https://github.com/OCA/product-attribute/tree/14.0/product_packaging_type_required) module.
- Now, packaging level is the one that should be marked as required


product_packaging_type_required/models/product_packaging_type.py
```diff
@@ -0,0 +1,54 @@
...
+
+import logging
+
+from odoo import api, fields, models
+
+_logger = logging.getLogger(__name__)
+
+
+class PackageType(models.Model):
+    _inherit = "stock.package.type"
```
sebalix: With this change:

- I would rename the file `stock_package_type.py`
- Should we also rename the module from `product_packaging_type_required` to `stock_package_type_required`, and move it in OCA/stock-logistics-warehouse alongside other `stock_package_type_*` modules (some in v16)?

cc @jbaudoux @rousseldenis

product_packaging_type_required/views/product_packaging_type_view.xml
```diff
@@ -1,29 +1,23 @@
 <?xml version="1.0" encoding="UTF-8" ?>
```
sebalix: This file can be renamed `stock_package_type.xml`

jbaudoux (changes requested):
It's now package level. This is the one that needs to be marked as required.
Need to rename module to `product_packaging_level_required` and change models and filenames.

In v16, odoo integrated `stock.package.storage.type` as `stock.package.type`. And `product.packaging.type` has been renamed to `stock.packaging.level`.

product_packaging_type_required/models/product_packaging_type.py
jbaudoux: no, see my comment on the PR

trisdoan:
Hello @jbaudoux , @sebalix, I just updated the code, please help to take a look :pray:

product_packaging_level_required/data/cron.xml
```diff
@@ -2,17 +2,12 @@
 <odoo noupdate="1">
     <record forcecreate="True" id="ir_cron_packaging_required" model="ir.cron">
         <field name="name">Packaging Required</field>
-        <field
-            ref="product_packaging_type.model_product_packaging_type"
-            name="model_id"
-        />
+        <field ref="stock.model_stock_package_type" name="model_id" />
```
jbaudoux: Should be level here also I guess

product_packaging_level_required/models/product_packaging_level.py
```diff
@@ -8,32 +8,30 @@
...
 
     @api.model
-    def cron_check_create_required_packaging(self):
-        """Create required packaging for each consumable product if missing."""
-        existing_products = self.env["product.product"].search(
-            [("type", "in", ("product", "consu"))]
-        )
-        required_packaging_types = self.search([("required", "=", True)])
+    def cron_check_create_required_packaging_level(self):
+        """Create required packaging level for each consumable product if missing."""
```
jbaudoux: It's product packing you are creating

product_packaging_level_required/models/product_packaging_level.py
```diff
@@ -8,32 +8,30 @@
 _logger = logging.getLogger(__name__)
 
 
-class ProductPackagingType(models.Model):
-    _inherit = "product.packaging.type"
+class ProductPackagingLevel(models.Model):
+    _inherit = "product.packaging.level"
 
     required = fields.Boolean()
```
jbaudoux: auto_create_packaging
Help: a cron will create a packaging for each product missing a packaging with this level

jbaudoux (changes requested):
I'm not fan of the module name. What about product_packaging_auto_create?

jbaudoux (approved):
LG. Can you reword commit with new module name `product_packaging_auto_create` as well as PR title

jbaudoux (changes requested):
The design with the cron cannot work like that anymore because there is now a default=1 on the product packaging qty and a constraint that qty>0

I would park this PR for now as this cron is anyway not used anymore in prod.

trisdoan:
DRAFT: checking if any features can be added

simahawk:
@trisdoan @jbaudoux ping

jbaudoux:
In my opinion this is obsolete. For sales, it is up to the product manager to create properly the packaging. Otherwise it will be defined during reception (shopfloor_reception must handle that)

jbaudoux:
* https://github.com/OCA/wms/issues/1018

PR #1662, opened 2024-06-18
[17.0] [MIG] product_packaging_level: Migration to 17.0

bizzappdev:

PR #1806, opened 2024-12-09
[18.0][MIG] product_packaging_level: Migration to 18.0

trisdoan:
### Note
- based on: https://github.com/OCA/product-attribute/pull/1662

### Changes in 18.0

1. get became private since [here](https://github.com/odoo/odoo/pull/118701/commits/2b93736d5cd2a8bcba9cb612a4e680b6cd7f3854) 
2. _lang_get_xxx  were merged into one api _get_data which is simpler and easy to extend since [here](https://github.com/odoo/odoo/commit/9f5470f247526fe54ff86e931cc835ef8942132a) 

### This changes
1. Took this occasion to fix an issue
[[FIX] product_packaging_level: cannot input name with user_defined policy](https://github.com/OCA/product-attribute/pull/1806/commits/6ec3f39429a833529a1d274258ac825149e79cca)
    - Without this commit, name of packaging is read-only even though the name policy is user_defined
    - Another fix: `_active_languages` is not available in `product.packaging.level`, which is also an issue in 16.0 when it's introduced https://github.com/OCA/product-attribute/pull/1527


sebalix (approved):
LG, I made a comment on the v17 migration PR about a change that should land there, and not in the v18 migration commit IMO: https://github.com/OCA/product-attribute/pull/1662

But not blocking.

product_packaging_level/models/product_packaging_level.py
```diff
@@ -0,0 +1,77 @@
...
+
+class ProductPackagingLevel(models.Model):
+    _name = "product.packaging.level"
+    _description = "Level management for product.packaging"
+    _order = "sequence, code"
+
+    def _default_language(self):
+        lang_code = self.env["ir.default"]._get("res.partner", "lang")
+        def_lang_id = self.env["res.lang"]._get_data(code=lang_code).id
+        return def_lang_id or self._active_languages()[0]
```
thienvh332, line 15: `_active_languages` ​​does not exist in `product.packaging.level`.

trisdoan:
~~DRAFT: inspect weird bug: _active_languages does not exist~~

jbaudoux (approved):
LGTM

product_packaging_level/models/product_packaging.py
```diff
@@ -0,0 +1,152 @@
...
+            rec.name = rec._get_name_from_policy()
+
+    # Keep this method to respect translations on level name
+    @api.depends("product_id", "packaging_level_id", "name_policy")
+    def _compute_display_name(self):
+        for record in self:
+            if record.product_id and record.packaging_level_id:
+                record.display_name = record._get_name_from_policy(lang=self.env.lang)
+            else:
+                return super()._compute_display_name()
```
ivantodorovich: This isn't correct.

it's not properly handling the recordset. The return shouldn't be within the loop.

trisdoan: Hello @ivantodorovich, indeed, I updated the code, will include it in the backport as well: https://github.com/OCA/product-attribute/pull/1809

ivantodorovich: Thanks!

ivantodorovich (approved):

Maurice67205:
Why is this module linked by default to openupgradelib?

jbaudoux:
> Why is this module linked by default to openupgradelib?

Probably because there was in v16 version a migration script to convert to new data model. Likely the migration script has been removed because not relevant anymore, but the dependency removal has been forgotten.

PR #1812, opened 2024-12-10
[17.0][IMP] product_main_supplierinfo: replace `_get_sellers` by std `_get_filtered_sellers` method

sebalix:
Starting from v15, Odoo proposes the `_get_filtered_sellers` method that could be re-used to get sellers.

### DIFF
--- a/product_main_supplierinfo/models/product_product.py
+++ b/product_main_supplierinfo/models/product_product.py
@@ -25,31 +25,19 @@ class ProductProduct(models.Model):
     )
     @api.depends_context("company")
     def _compute_main_seller_id(self):
-        for product in self:
-            sellers = product._get_sellers()
-            product.main_seller_id = fields.first(sellers)
-
-    def _get_sellers(self):
-        """Returns all available sellers of a product based on some constraints.
-
-        They are ordered and filtered like it is done in the standard 'product' addon.
-        """
-        self.ensure_one()
-        all_sellers = self._prepare_sellers(False).filtered(
-            lambda s: not s.company_id or s.company_id.id == self.env.company.id
-        )
-        today = fields.Date.context_today(self)
-        sellers = all_sellers.filtered(
-            lambda s: (
-                (s.product_id == self or not s.product_id)
-                and (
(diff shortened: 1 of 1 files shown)


xaviedoanhduy:
hi @sebalix, I have create this [PR](https://github.com/xaviedoanhduy/product-attribute/pull/3) to improve your changes.

sebalix:
Hi @xaviedoanhduy , thanks for the followup! I took your commit to replace mine here (to preserve co-authorship).

product_main_supplierinfo/models/product_product.py
```diff
@@ -26,30 +26,18 @@ class ProductProduct(models.Model):
...
-                (s.product_id == self or not s.product_id)
-                and (
-                    (s.date_start <= today if s.date_start else True)
-                    and (s.date_end >= today if s.date_end else True)
-                )
+    def _get_filtered_sellers(
+        self, partner_id=False, quantity=0.0, date=None, uom_id=False, params=False
+    ):
+        res = super()._get_filtered_sellers(partner_id, quantity, date, uom_id, params)
+        if not res:
```
sebalix: @xaviedoanhduy the issue here is that we are modifying the behavior of Odoo std regarding other business flows.
We should check a context key here to not break anything:
```suggestion
        if not res and self.env.context.get("compute_main_seller"):
```

product_main_supplierinfo/models/product_product.py
```diff
@@ -26,30 +26,18 @@ class ProductProduct(models.Model):
     @api.depends_context("company")
     def _compute_main_seller_id(self):
         for product in self:
```
sebalix:
```suggestion
        for product in self.with_context(compute_main_seller=True):
```

rousseldenis:
Seems promising!

xaviedoanhduy:
hi @sebalix, changed according to your suggestion.
thanks

sebalix:
@xaviedoanhduy thank you for the update, it looks good. I cannot review my own PR but I approve your changes.

rousseldenis (approved):

PR #1813, opened 2024-12-10
[17.0][MIG] product_abc_classification_sale_stock: Migration to 17.0

kobros-tech:


product_abc_classification_sale_stock/tests/test_abc_classification_profile.py
```diff
@@ -0,0 +1,289 @@
...
+        # test computed classification and check that the classification is
+        # also set on the product_templale
+        self.stock_profile._compute_abc_classification()
+        self._assertLevelIs(self.product1, "a")
+        self._assertLevelIs(self.product3, "a")
+        self._assertLevelIs(self.product5, "a")
+        self._assertLevelIs(self.product6, "a")
+        self._assertLevelIs(self.product2, "a")
+        self._assertLevelIs(self.product4, "a")
+        self._assertLevelIs(self.product_new, "b")
```
anothingguy: these profiles are different from v16 and they're wrong levels. Probably `pick._action_done()` doesn't work.

anothingguy (changes requested):

requirements.txt
```diff
@@ -0,0 +1,2 @@
+# generated from manifests external_dependencies
+freezegun
```
anothingguy, line 2: already added in Odoo requirements

product_abc_classification_sale_stock/tests/test_abc_classification_profile.py
kobros-tech: yes, I was fixing an issue.

now everything is fine and you can update your migration to 18.0 based on what I just improved.

I am going to review your migration too and tell my feedback.

Thanks

kobros-tech:
@lmignon 

Thanks for approving the fix on 16.0, here is the clean migration for 18.0 based on the fix.

PR #1823, opened 2024-12-18
[18.0] [MIG] product_assortment: Migration to 18.0

PieterPaulussen:


bosd (changes requested):
Thanks for your work on this.
Functional review. Im getting different results on older versions of this module.

Create a assortment.
Whitelist 1 specific product. 

Behaviour on V16 and lower.
Notice that the smartbutton displays 1 product.
image

On this PR:
The smartbuttons shows 44 products in this assortment:
image

bosd:
@PieterPaulussen Can you please fix the functionality and add an additional test? :pray:

PieterPaulussen:
@bosd Thanks for the feedback, but I think the original functionality was broken in the previous module.
The domains of the assortment are additive, not the other way around.

Example 1:
- You add an Allowed product domain based on 1 specific product category of which the result is 25 products.
- Then you also include 1 additional product in the 'Products to include' list.
The result in this case is 26 products.

Example 2:
- You add an Allowed product domain based on 1 specific product category of which the result is 25 products.
- Then you also include 1 additional product in the 'Products to include' list, but it is already inside the category.
The result in this case is 25 products.

Example 3:
- You add the following domain: [('id', '=', 0)] which excludes all products from the assortment.
- Then you set 1 additional product in the 'Products to include' list.
The result in this case is 1 products. 

Is that not exactly what the assortment is supposed to do? There is also an inverse domain which you can enable using the boolean flag `Apply restricted product domain` on the form.
This domain has the subtractive effect on the result.

bosd:
@PieterPaulussen 
Actually the behavioir of the domain filter and orm has changed. (I think in V17).

Your examples are correct. Only the problem arises when the domain is empty.

Example:
Product domain: `[]`
and 1 whitelisted product should result in 1 matched product.

But in this PR it results in 44 products.
While in previous versions the result was as expected 1 product.
Which is the desired behaviour.

It does'nt make sense that one can define a list of `whitelisted product` which are not evaluated, because by default all products are added.

I adapted the code of this module and (depending modules) to accomodate for these changes in the orm.
This one can be closed now as it is superseeded by #1864

PieterPaulussen:
@bosd 
You are incorrect. This is not the desired behavior of the module. An open ended domain `[]` will return all products by default which is the standard behaviour of a search operation in the ORM.

I don't believe your PR should superseed mine as this version has been used in production for a couple of my clients, so the logic stands.

I would like to resolve the issue before either of the PR's are merged. Thanks

bosd:
> @bosd You are incorrect. This is not the desired behavior of the module. An open ended domain `[]` will return all products by default which is the standard behaviour of a search operation in the ORM.

Yes, This is exactly my point.

In the beginning I mis understood, the point that you want to "fix" behaviour of this module.

> I don't believe your PR should superseed mine as this version has been used in production for a couple of my clients, so the logic stands.

Well, With all respect, I disagree. We have it in production as well. 
IMO it is bad practice to change the behaviour of the module in a migration.
I just checked backwards until version 13. (Which you migrated by the way, Thanks).
The behaviour is all the same.
image

Depending on your configuration / use case, you will need to make a migration script for exsisting clients to accomodate for the changes in behaviour. (Might as well go for other users as well).

But, honestly I'm advocating against such a big functional change in a migration.

So, open-source is going to do what it is supposed to do.
We have 2 competing PR's.
Hopefully soon someone is gonna join the discussion.
cc: @pedrobaeza 👀 🙏 

> I would like to resolve the issue before either of the PR's are merged. Thanks

Seems, the best thing to do 👍

pedrobaeza:
Without knowing the specific problem, a migration shouldn't change the expected behavior in a module, unless an agreement is made (previously or in between, but advertising the change clearly).

@PieterPaulussen, if you think the previous behavior is incorrect, you have to raise an issue saying so, for getting reasoning behind the current one, and agree a possible change in the current versions or in the migration, but not to hide that behavior change inside the migration itself. I don't even see an extra commit with that change and a detailed explanation of it in the commit message.

Thus, this PR shouldn't be valid as is. You can remove the change in the behavior, and then your PR will have priority over @bosd's one, or if you prefer, let their one.

About your customer current situation, I'm sorry about it, but it's a risk you have assumed not having this proposal even approved by one of the users/contributors/maintainers. You can put the change of behavior of top of the standard one if this is finally not changed.

I know it seems that contributing here is worst than keep things on your own, but think that these protections are for the good of all. Imagine that others make similar changes on their own on other OCA modules you are using, and when you update, you get the bad surprise. Holding this and at least planning the change, avoid all these problems.

I ping other contributors of this module to join on the change itself: @rousseldenis @CarlosRoca13

PieterPaulussen:
@bosd @pedrobaeza 
Thanks both for you contribution to the discussion. Below some clarifications on my side.

> Without knowing the specific problem, a migration shouldn't change the expected behavior in a module, unless an agreement is made (previously or in between, but advertising the change clearly).

I completely agree. However, I did not change any behavior of the module itself (which is why I did not add any commit). The change is apparently in the ORM on the behaviour of `ir.filters`.

> @PieterPaulussen, if you think the previous behavior is incorrect, you have to raise an issue saying so, for getting reasoning behind the current one, and agree a possible change in the current versions or in the migration, but not to hide that behavior change inside the migration itself. I don't even see an extra commit with that change and a detailed explanation of it in the commit message.

Don't worry, I hold the OCA in the highest esteem and do not wish to be subversive in PR's at all. Nor was this my intention. As far as my customers situation goes, they will not encounter the issue since they use the functionality in a different way.

I think the central point of the discussion should be altered to: "Because the ORM now behaves differently, should the new migrated version of the module also reflect that behaviour or not?".

If you look at it from a UI perspective, it becomes confusing.

**Example 1 (Odoo v16):**
1. The 'Allowed product domain' states `Match all records` and just below `59 records`.
Screenshot 2025-02-25 at 12 24 56
2. The products to include lists 1 specific product. 
3. The product counter (and the products in the resulting action) are limited to that 1 specific product in the 'Products to Include'.

**Example 2 (Odoo v18):**
1. The 'Allowed product domain' states `Match all records` and just below `49 records`.
Screenshot 2025-02-25 at 12 32 17
2. The products to include lists 1 specific products. 
3. The product counter (and the products in the resulting action) are limited to the entire result of all domains in the assortment.

**Example 3 (Odoo v16):**
Screenshot 2025-02-25 at 12 35 06
1. The 'Allowed product domain' is set to a specific product category.
2. The products to include lists 1 specific products.
3. The product counter (and the products in the resulting action) are limited to the entire result of all domains in the assortment, in this case 4 products.

**Example 4 (Odoo v18):**
This test is the same as Example 3, but then in Odoo v18. The result of the assortment is the same as v16. Meaning, the "Allowed product domain" is always applied.

**To Conclude:**
I would like to argue that, because the ORM now changed its implementation, the actual result in more correct than in the previous versions. I did not have to make any changes to the code to correct this.
If we need to alter this module to reflect the OLD behaviour, additional code should be added to do so. (Which is what this PR https://github.com/OCA/product-attribute/pull/1864#issuecomment-2681467012 does.)

product_assortment/tests/test_product_assortment.py
```diff
@@ -91,7 +91,24 @@ def test_product_assortment_view(self):
         included_product = self.env.ref("product.product_product_7")
         self.assortment.write({"whitelist_product_ids": [(4, included_product.id)]})
         res = self.assortment.show_products()
-        self.assertEqual(res["domain"], [("id", "in", [included_product.id])])
```
bosd, line 94: This line has been changed to make the tests pass with the new orm behaviour.

bosd: See comment https://github.com/OCA/product-attribute/pull/1864#issuecomment-2684047562 Why I think that, PR is more "complete".

PieterPaulussen:
@bosd Regarding your comment [in your PR here](https://github.com/OCA/product-attribute/pull/1864#issuecomment-2684047562)

I personally find it confusing, after discussing it with peers, is the visual representation of the filter results for the "Allowed Product Domain". It clearly states "Match all records" when empty yet the resulting products are not all records in the older version (or your migrated one). What is your opinion on that?

bosd:
> I personally find it confusing, after discussing it with peers, is the visual representation of the filter results for the "Allowed Product Domain".

True, But I'm actually used to that. Matching all products means **no** filter.
So it will still return **all** products.
This is still true for the "all products" test.
(See other pr with screenshot)

_It is actually debatable to have an "all" filter in a real use case, because it litterally does filter nothing._

My main issue is that the "point and click" a.k.a whitelisted product list is broken in this pr.
Matching All products + 2 results in .... all products.
My expected result would be, when no filter is defined, only 2 products should returned.

If I would like to do the use case of my client with this version. I would have to do a workaround.
Create a "bogus" filter, which matches to nothing. Then the whitelisted products will be evaluated correctly.

image
To me, the workaround is a "Dirty" solution to achieve the desired result.

product_assortment/tests/test_product_assortment.py
bosd: @PieterPaulussen Can you agree, on the fact that unit tests should not be altered on a functional level between versions to make them pass?

Additional unit test, as are added in both pr's are obviously ok.

PieterPaulussen:
What if we compare it to the behaviour of "Products to Include" to that of the "Products to Exclude"? If we used your implementation, we would get "No products - 1 product to exclude". While in my implementation, you would get "All products - 1 product to exclude". 

Does my implementation not make more sense in that case?

bosd:
> While in my implementation, you would get "All products - 1 product to exclude".

Good Catch!
Indeed that would make more sense.

My proposal would be to adopt "my" pr. So the product to exclude filter is working.
As well as add some unit tests for this scenario.

So that we would have a "working" "filter for both "Products to include" and "Products to exclude".

PieterPaulussen:
> My proposal would be to adopt "my" pr. So the product to exclude filter is working. As well as add some unit tests for this scenario.

Would you not agree that this PR already covers the correct functionality? I don't see the point in using yours.

I'll look into adding new unittests to cover the missing cases.

bosd:
> If we used your implementation, we would get "No products - 1 product to exclude".

Oopsie, I had written my previous reply without testing the functionality.
This statement is incorrect. This is actually handled by test case.
`def test_assortment_with_black_list_product_domain(self):`

Which is passing on both PR's.
https://github.com/OCA/product-attribute/pull/1823/files#diff-07c79df5af22b005fcdef6b8a76e647d8ddb72ee378fb61cfd0d3a37bbba56f3R137

image

pedrobaeza:
@carlosdauden you may have an opinion here as well.

bosd:
> Would you not agree that this PR already covers the correct functionality? I don't see the point in using yours.

Actually I don't agree with this statement.
Sadly the diff is not working. So would have to do a manual compare.
But this one is lacking some commits, like the one with demo data.
And it is missing some fixes and typo correction from previous pr.

I agree, The missing commits could easily be cherry-picked / forward ported.

My main objection is to have lost the "whitelistproduct" functionality.
Where the unit tests in this migration PR are adjusted to make them pass.
Like mentioned [here](https://github.com/OCA/product-attribute/pull/1823#pullrequestreview-2643211759)
Which invalidates the use case of the test.

TBH, I'm seeing some lost functionality in this pr. As opposed to the competing PR.
With all respect, do you have some missing functionality there? or a reason to block other than "it looks wierd / (confusing)"?

PieterPaulussen:
> My main objection is to have lost the "whitelistproduct" functionality. Where the unit tests in this migration PR are adjusted to make them pass. Like mentioned [here](https://github.com/OCA/product-attribute/pull/1823#pullrequestreview-2643211759) Which invalidates the use case of the test.
> 

**You have not lost this feature as it is still available, your configuration just needs to change.**
And the only thing the unit tests warranted a change for was to comply to Core Odoo changes.

> TBH, I'm seeing some lost functionality in this pr. As opposed to the competing PR. With all respect, do you have some missing functionality there? or a reason to block other than "it looks wierd / (confusing)"?

I still beg to differ. A UI perspective matters the same as the technical perspective because when someone (a basic user) installs this module and sees messages like "Match all records" and "46 products" in the UI and then suddenly is confronted with not having 46 products but just one **while those messages are still visible**. That is not confusing to you? Keep in mind that still most basic users do not read code nor understand the core features.

Given my counterpart "exclusion" example, it proves that the functionality of this module is correct and that it aligns with what a user would expect being: "You have all products and then choose to exclude 1 of them".

I would also like to stop chatting about this as the effort to respond has become higher than the gain for the community.

carlosdauden:
Perhaps the clearest option for the user is to add a check similar to `apply_black_list_product_domain` so that the domain is displayed and applied.

carlosdauden:
Originally the intention of this module was to add the products of the whitelist to the products returned by the domain, I don't know if this functionality has been changed in some migration, but it is true that it is not very understandable that if we have a filter with `domain=[]` this returns all products and if we add a product to the whitelist to that filter it only returns the product of the whitelist...

carlosdauden:
In previous versions the behaviour was not as expected (tested in v15):
```
res = []
result_domain = [('id', 'in', [12429])]
```
the result of this OR operator:
https://github.com/OCA/product-attribute/blob/0082384b6b5e92abf9d9c4903c0b7ef5d06c0136/product_assortment/models/ir_filters.py#L116
was returning:
`[[('id', 'in', [12429])]`

This commit seems to fix that problem and introduces the debate:
https://github.com/odoo/odoo/commit/450f5c9307a71d3ccd7b1ba2ce6d5d90f8a352f3

PieterPaulussen:
@carlosdauden Thank you for your contribution to the discussion. 

Are you implying that the original functionality in previous versions is currently broken? If so, it would be appreciated to have a third voice in this matter to tip the discussion in favour of any of the proposed solutions?

carlosdauden:
Yes, according to the original idea and looking at the code, it was always intended that the result of an OR expression was an aggregation of the results of both domains.
The thing is that we probably didn't realise it because most users use domain or domain + whitelist, but not the whitelist option leaving the domain as all `[]` records.

PieterPaulussen:
> Yes, according to the original idea and looking at the code, it was always intended that the result of an OR expression was an aggregation of the results of both domains. The thing is that we probably didn't realise it because most users use domain or domain + whitelist, but not the whitelist option leaving the domain as all `[]` records.

Then this merits an additional number of unit tests to solidify the result.

@bosd What do you think?

carlosdauden:
It is possible to create a migration script that detects such cases (whitelist option used and the `domain=[]`) and replaces the `domain=[]` with `domain=[(0, '=', 1)]`.

PieterPaulussen:
> It is possible to create a migration script that detects such cases (whitelist option used and the `domain=[]`) and replaces the `domain=[]` with `domain=[(0, '=', 1)]`.

That can be added easily yes.

carlosdauden:
Why was this PR closed?

PieterPaulussen:
@carlosdauden I applied some fixes to this branch and then force pushed it. I think one of the OCA repository rules forces this PR to be closed because of that. It cannot be reopened because the other PR from @bosd is now first in line instead of this original one.

Perhaps someone with more rights can re-open this PR, and I'll gladly push the update to this branch instead. For now I've re-created the PR here: https://github.com/OCA/product-attribute/pull/1904

PieterPaulussen:
@carlosdauden I applied some fixes to this branch and then force pushed it. I think one of the OCA repository rules forces this PR to be closed because of that. It cannot be reopened because the other PR from @bosd is now first in line instead of this original one.

Perhaps someone with more rights can re-open this PR, and I'll gladly push the update to this branch instead. For now I've re-created the PR here: https://github.com/OCA/product-attribute/pull/1904

PieterPaulussen:
@carlosdauden I applied some fixes to this branch and then force pushed it. I think one of the OCA repository rules forces this PR to be closed because of that. It cannot be reopened because the other PR from @bosd is now first in line instead of this original one.

Perhaps someone with more rights than me can reinstate this PR, and I'll gladly push the update to this branch instead.

You can find the update here: https://github.com/The-O-Team/product-attribute/tree/18.0-mig-product_assortment

pedrobaeza:
It should be a git manipulation mangling on your part, probably a rebase over an incorrect branch with no common base with the target branch of the PR. Thus, it can't be reopened.

PieterPaulussen:
@pedrobaeza I performed the same steps to do the migration, why can create a new PR for the same branch then?

pedrobaeza:
I was just explaining the possible problem why this got auto-closed. It's not a problem to open a new PR. It's just the communication history is split across several threads.

PieterPaulussen:
I see, do you know any way to resolve the issue? Or should I move this to a new PR?

pedrobaeza:
Not sure if you do another rebase and forced push again will enable the reopening, but the best change is to open a new PR.

bosd:
> @bosd What do you think?

From experience some non technical users get nervous when seeing a domain field.
(Just looking at the UI maked their jaws drop.)
They just want to have a very simple user interface, Where they can select a partner and add (whitelist) products.
 
Adding some "bogus" domain as a workaround will solve the functional level.
The proposal of adding...
`domain=[] with domain=[(0, '=', 1)]`

But it is even more confusing for the non technical user.

What do you think about adding a checkbox `apply_advanced_product_domain`.
When checking that, it will show the domain field and widget.
(Which we could hide be default).
When that box is checked, the domain is evaluated.
(No matter if it is `[]` or `[(0, '=', 1)]`)

Likely we can do the same for the partner fields.
In such way we can cover with this module a "Simple" assortment definition use-case.
And well as advanced use-cases for the more technical / power user which used domains or a combination of domain and white/black list.

PR #1827, opened 2024-12-27
[18.0][MIG] product_document_domain: Migration to 18.0

BhaveshHeliconia:


product_document_domain/models/product_template.py
```diff
@@ -0,0 +1,30 @@
...
+
+class ProductTemplate(models.Model):
+    _inherit = "product.template"
+
+    def _compute_product_document_count(self):
+        for template in self:
+            res = super()._compute_product_document_count()
+            template.product_document_count -= template.env[
+                "product.document"
+            ].search_count(
```
rousseldenis: IMHO, this could be improved using `read_group()` function before the loop.

BhaveshHeliconia: Done. Thanks for the review!

tonygalmiche (approved):
Tested on a local Odoo 18.0 instance: on a variant, the Documents count/list now correctly includes the documents of the related template in addition to its own, while the template count no longer double-counts variant documents. 
Matches the description. 
Works as expected.

Issue #1750, opened 2024-10-24
product_category_code_unique: Select different restrictions


manuelregidor:
In the current version of the module product_category_code_unique (https://github.com/OCA/product-attribute/tree/14.0/product_category_code_unique), the only restrictions available is checking that the code is unique within the whole system. However, some people might have other needs in terms of restrictions of codes. That is why I suggest improving the module in order to be able to select between different types of restrictions:

- Restriction 1: Code field must be unique. If this restriction is selected, it will not be possible to have a duplicated code in the system (this is the currently available restriction).
- Restriction 2: Code field cannot be duplicated in parent nodes or son nodes.
- Restriction 3: Code field must be unique for product categories with the same parent category.

Only one of theseoptions could be selected and right after selecting a new option it will be checked whether the restriction can be applied or not.

If I add this feature to the module, do you think it will be interesting enough?

Thank you

manuelregidor:
@HaraldPanten @rousseldenis @legalsylvain Could you have a look at my proposal, please? Thank you.

HaraldPanten:
@manuelregidor I think it's interesting, yes.

I have a couple of questions:

1- The restrictions would be optional, right?
2- In which version are you planning to make these changes?

manuelregidor:
@HaraldPanten 
1.- Yes, the module could be used with no restrictions at all.
2.- I woul make it for v16, but it could be easily migrated.

HaraldPanten:
@manuelregidor OK! go then 👍

PR #1829, opened 2025-01-02
[16.0][MIG+IMP] product_category_code_unique: restriciton options

manuelregidor:
Based on https://github.com/OCA/product-attribute/issues/1750

T-6836


luis-ron (approved):
Functional Review: LGTM 👍🏻

product_category_code_unique/models/product_category.py
```diff
@@ -1,16 +1,67 @@
...
-        ("uniq_code", "unique(code)", "The category code must be unique!"),
-    ]
+    def _get_parents(self):
+        self.ensure_one()
+        parent = self.parent_id
+        parents = self.env["product.category"]
+        while parent:
+            parents += parent
+            parent = parent.parent_id
+        return parents
```
Tisho99: Why you don't use parent_of operator?

product_category_code_unique/models/res_config_settings.py
```diff
@@ -0,0 +1,37 @@
...
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+
+from odoo import api, fields, models
+
+
+class ResConfigSettings(models.TransientModel):
+
+    _inherit = "res.config.settings"
+
+    product_cat_code_unique_restriction = fields.Selection(
```
Tisho99, line 12: The v15 module only had 1 option: "system". I think that option should be the default one

product_category_code_unique/models/product_category.py
```diff
@@ -1,16 +1,67 @@
...
+                    domain = [("code", "=", cat.code), ("id", "in", to_check.ids)]
+                    within = _("category hierarchy")
+                if self.search_count(domain) > 1:
+                    raise ValidationError(
+                        _("The category code must be unique within {within}!").format(
+                            within=within
+                        )
+                    )
+
+    @api.constrains("code")
```
Tisho99: What if we have 2 categories with the same code but different parent, and then we edit the parent of one of the categories to be the same?

I think parent_id and child_ids fields should be added to constrains

Tisho99 (commented):
Technical review.

Some comments

product_category_code_unique/models/res_config_settings.py
manuelregidor: A default option has not been set so it can be decided after installing the module.

product_category_code_unique/models/product_category.py
manuelregidor: You're right. Changed.

product_category_code_unique/models/product_category.py
manuelregidor: Changed.

manuelregidor:
@Tisho99 Could you reviw again, please?

product_category_code_unique/models/res_config_settings.py
Tisho99: But the module shoud behave the same way as before without changing any setting, with this improvement the default behaviour will change

manuelregidor: The default behaviour has changed because the module has changed. I don't think we should set a default restriction in a module that accepts different types of restrictions, as we don't know which restrictions the user wants to use. A default restriction could throw exceptions that the user does not want.

Tisho99 (approved):
LGTM

rousseldenis:
@manuelregidor Could you maybe split the migration changes from improvements in commits ?

product_category_code_unique/models/product_category.py
```diff
@@ -1,26 +1,73 @@
...
 
 
 class ProductCategory(models.Model):
 
     _inherit = "product.category"
 
-    _sql_constraints = [
-        ("uniq_code", "unique(code)", "The category code must be unique!"),
-    ]
+    def _get_parents(self):
```
rousseldenis, line 13: `parent_id` field do the same and you can have just one parent as this is a many2one.

product_category_code_unique/models/product_category.py
```diff
@@ -1,26 +1,73 @@
...
+
+    def _get_children(self):
+        self.ensure_one()
+        return self.search([("parent_id", "child_of", self.id)])
+
+    def _get_hierarchy_cats(self):
+        self.ensure_one()
+        return self._get_parents() + self._get_children()
+
+    def _code_restriction(self, restriction=False):
```
rousseldenis, line 25: IMHO, the `restriction` attribute is not useful.

product_category_code_unique/models/res_config_settings.py
```diff
@@ -0,0 +1,37 @@
...
+            "- Parent-Children: Parent and Children Product Category Codes of the"
+            " same category cannot be duplicated.\n"
+            "- Category Hierarchy: Product Category Codes cannot be duplicated "
+            "within the same category hierarchy.\n"
+        ),
+    )
+
+    @api.constrains("product_cat_code_unique_restriction")
+    def _check_product_cat_code_unique_restriction(self):
+        for sel in self.filtered("product_cat_code_unique_restriction"):
```
rousseldenis: This is not useful as there should only have one configuration record...

product_category_code_unique/models/product_category.py
```diff
@@ -1,26 +1,73 @@
...
-    ]
+    def _get_parents(self):
+        self.ensure_one()
+        return self.search([("parent_id", "parent_of", self.id)])
+
+    def _get_children(self):
+        self.ensure_one()
+        return self.search([("parent_id", "child_of", self.id)])
+
+    def _get_hierarchy_cats(self):
```
rousseldenis: Prefer explicit naming instead. And maybe `_get_hierarchy_categories_for_code_unique()` to avoid naming collisions.

product_category_code_unique/models/product_category.py
```diff
@@ -1,26 +1,73 @@
...
+                        "&",
+                        ("code", "=", cat.code),
+                        ("id", "in", to_check.ids),
+                    ]
+                    within = _("parent and children")
+                elif restriction == "hierarchy":
+                    to_check = cat._get_hierarchy_cats()
+                    domain = [("code", "=", cat.code), ("id", "in", to_check.ids)]
+                    within = _("category hierarchy")
+                if self.search_count(domain) > 1:
```
rousseldenis: IMHO, this should be avoided for performance reasons. 

A query to get all the categories should be done outside the loop. Then, compare the recordset values.

rousseldenis (changes requested):

product_category_code_unique/models/product_category.py
rousseldenis: No, ok.

product_category_code_unique/models/product_category.py
manuelregidor: The restriction attribute has been added so it can be passed when the method is called from _check_product_cat_code_unique_restriction in res.config.settings model. _check_product_cat_code_unique_restriction  in res.config.settings is executed before the system parameter is modified changed, so we need to pass the selected restriction to the _code_restriction method in product.category to check whether the restriction option can be changed.

product_category_code_unique/models/product_category.py
manuelregidor: Sorry, but I don't know exactly what you mean here.

manuelregidor:
@rousseldenis Thanks for your feedback. I've already split the changes into 2 commits.

product_category_code_unique/models/product_category.py
manuelregidor: @rousseldenis I've checked your commet again but I still don't know exactly what changes you are suggesting here. Please, could you give me a more detailed explanation so I can do the changes? Thank you

manuelregidor:
@HaraldPanten Could you review, please?

manuelregidor:
@rousseldenis Have you had the chance to take a look at my last comments? Thank you.

product_category_code_unique/models/product_category.py
rousseldenis: self = recordset. So, the loop you do will launch an SQL query for each `search_count` instance.

For performances reasons, this should be avoided. You should find a way to get all results before the loop. Then do the logic with the result.

manuelregidor: I've applied some changes. Could you have a look? Thank you.

HaraldPanten:
@rousseldenis If it's OK for you, we'll merge this one. All your comments have been considered.

HaraldPanten (approved):

HaraldPanten:
Merging, as all comments have been considered.

/ocabot merge nobump

PR #1830, opened 2025-01-02
[16.0][ADD] product_merge

sbejaoui:
This module allows users to efficiently merge multiple product templates into one. This merge process ensures that attributes and variants from all selected products are consolidated into the primary product template, without creating any new variants. This approach is particularly important for maintaining data integrity and avoiding unnecessary database load.

By not creating new variants during the merge, the module helps to prevent heavy updates on existing tables, making it ideal for large-scale databases.


sale_product_merge/wizards/product_merge_wizard.py
```diff
@@ -0,0 +1,25 @@
...
+            ]
+        )
+        pricelist_items.write(
+            {
+                "applied_on": "0_product_variant",
+                "product_id": product_variant.id,
+                "product_tmpl_id": self.product_tmpl_id.id,
+            }
+        )
+        return super()._move_variant_to_template(product_variant, attribute_values)
```
chienandalu: This could go in the main module, as that model is in the `product` module. Or is there anything else?

product_merge/wizards/product_merge_wizard.py
```diff
@@ -0,0 +1,171 @@
...
+        )
+        other_templates = (
+            self.line_ids.product_id.product_tmpl_id - self.product_tmpl_id
+        )
+        for line in self.line_ids:
+            product_variant = line.product_id
+            self._update_supplier_info(product_variant)
+            self._move_variant_to_template(product_variant, line.attribute_value_ids)
+
+        other_templates.write({"active": False})
```
lmignon, line 129: @sbejaoui IMO something is missing. You should use a generic method to update into the database all the columns where the other_templates.ids are referenced.... The [`merge_records`](https://github.com/OCA/openupgradelib/blob/master/openupgradelib/openupgrade_merge_records.py#L979) from the openupgrade lib could be a good candidate :thinking:

lmignon (changes requested):

StefanRijnhart:
Possible duplicate of https://github.com/OCA/stock-logistics-warehouse/tree/16.0/base_product_merge?

StefanRijnhart:
This module emits error logs during the tests. Can you mute the logger when testing the exception?
```
2025-01-07T23:29:23.6985267Z 2025-01-07 23:29:23,697 500 [1;32m[1;49mINFO[0m odoo odoo.addons.product_merge.tests.test_product_merge: Starting TestProductMerge.test_action_merge_products_same_attribute_value ... 
2025-01-07T23:29:23.7717242Z 2025-01-07 23:29:23,770 500 [1;31m[1;49mERROR[0m odoo odoo.sql_db: bad query: UPDATE "product_product" SET "combination_indices" = '20', "write_date" = '2025-01-07T23:29:23.397594'::timestamp, "write_uid" = 1 WHERE id IN (78)
2025-01-07T23:29:23.7718580Z ERROR: duplicate key value violates unique constraint "product_product_combination_unique"
2025-01-07T23:29:23.7719177Z DETAIL:  Key (product_tmpl_id, combination_indices)=(63, 20) already exists.
2025-01-07T23:29:23.7719532Z

**Review by StefanRijnhart — COMMENTED** [review]
See comments above.

**Comment by lmignon** [comment]
> Possible duplicate of https://github.com/OCA/stock-logistics-warehouse/tree/16.0/base_product_merge?

The goal is not the same. This PR propose to merge 2 product templates into one product with 2 variants.

**Comment by StefanRijnhart** [comment]
From what I can see, `base_product_merge` will not merge product variants if there are product attributes at play, in which case the result is also one product with the combined set of variants. Can you be more precise about the difference in outcome?

**Comment by lmignon** [comment]
> From what I can see, `base_product_merge` will not merge product variants if there are product attributes at play, in which case the result is also one product with the combined set of variants. Can you be more precise about the difference in outcome?

I didn't know that the `base_product_merge` module existed so I didn't really test it. 
A typical use case is, for example, when you have two TShirt products in DB, one for TShirt M and one for TShirt L, which are not variants of each other. The wizard will allow you to group the products together under the same template while completing the attribute values possible at template level. I don't get the impression from reading the code of the `base_product_merge` module that this is the use case supported.
Moreover we avoid modifying the variant identifiers. We're trying to minimize the impact on the DB as much as possible by keeping the current variant identifiers and only replacing the values of the template that has been orphaned by its variant. The merge must be possible on a running instance with a lot of sales and logistics activities.

**Comment by StefanRijnhart** [comment]
Sounds like there is a strong case to try and improve `base_product_merge` for this scenario, also given that it already uses OpenUpgradelib's `merge_records`.

File: product_merge/wizards/product_merge_wizard.py [hunk 3]
```diff
@@ -0,0 +1,207 @@
...
+            self._update_pricelist_item(product_variant)
+            self._update_supplier_info(product_variant)
+            self._move_variant_to_template(product_variant, line.attribute_value_ids)
+        merge_records(
+            self.env,
+            self.product_tmpl_id._name,
+            other_templates.ids,
+            self.product_tmpl_id.id,
+            field_spec=self._get_merge_field_spec(),
+            delete=False,
```
chienandalu, line 127: Out of curiosity. Why to keep the old template?

product_merge/wizards/product_merge_wizard.py
```diff
@@ -0,0 +1,207 @@
...
+
+class ProductMergeWizard(models.TransientModel):
+    _name = "product.merge.wizard"
+    _description = "Merge Products Wizard"
+
+    product_tmpl_id = fields.Many2one(
+        comodel_name="product.template",
+        string="Product Model",
+        domain="[('id', 'in', product_ids)]",
+        required=True,
```
rousseldenis, line 18:
```suggestion
        required=True,
        ondelete="cascade"
```

product_merge/wizards/product_merge_wizard_line.py
```diff
@@ -0,0 +1,28 @@
...
+
+from odoo import api, fields, models
+
+
+class ProductMergeWizardLine(models.TransientModel):
+
+    _name = "product.merge.wizard.line"
+    _description = "Merge Products Wizard Line"
+
+    wizard_id = fields.Many2one("product.merge.wizard", string="Wizard", required=True)
```
rousseldenis:
```suggestion
    wizard_id = fields.Many2one("product.merge.wizard", string="Wizard", required=True, ondelete="cascade")
```

product_merge/wizards/product_merge_wizard_line.py
```diff
@@ -0,0 +1,28 @@
...
+    product_id = fields.Many2one(
+        comodel_name="product.product", string="Product", required=True
```
rousseldenis:
```suggestion
        comodel_name="product.product", string="Product", required=True, ondelete="cascade"
```

product_merge/wizards/product_merge_wizard.py
```diff
@@ -0,0 +1,207 @@
...
+        store=True,
+        readonly=False,
+    )
+
+    def _check_minimum_products(self):
+        """
+        Ensure that at least two products are selected for merging.
+        """
+        for wizard in self:
+            if len(wizard.product_ids) < 2:
```
rousseldenis, line 41: An alert should be great on wizard form in order to warn user before clicking on merge.

rousseldenis (commented):
Seems great.

product_merge/wizards/product_merge_wizard.py
sbejaoui: As a merge strategy, we chose to keep the target values. This behavior can be modified through inheritance of _get_merge_field_spec. However, after the merge, the user may need to check one of the merged products to see, for example, the previous price or any other useful data.

That’s why we archive the merged products and leave the cleanup to the user if they wish to do so.

product_merge/wizards/product_merge_wizard.py
sbejaoui: This check can't be done before the user completes their input
otherwise, they would receive a warning after adding the first product
IMO, it makes sense to trigger it when they click on the button

PieterPaulussen:
@sbejaoui Any chance that you could create another PR for V18?

nihelgabsi-acsone (approved):
Functional Test  + Code review

giarve (approved):

PR #1834, opened 2025-01-06
[16.0][FIX] product_category_active: harmonization of category display.

legalsylvain:
In odoo, archived items are not displayed by default. (see users view, products, etc...) The product_category_active module doesn't follow this convention because it is changing the context of the action of product.category. As a result, all categories are displayed, polluting visibility with obsolete items.

- Remove the active_test in the context of the action.
- create a migration script to remove the active_test on existing instances.
- add filters to display archived and 'archived or active' categories, in case you want to see all categories

### DIFF
--- a/product_category_active/README.rst
+++ b/product_category_active/README.rst
@@ -78,6 +78,7 @@ Authors
 ~~~~~~~
 
 * Tecnativa
+* GRAP
 
 Contributors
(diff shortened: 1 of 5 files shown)


rousseldenis (approved):

simahawk:
LG

/ocabot merge nobump

PR #1835, opened 2025-01-06
[16.0] write several products without forcing the sequence on non-serialized products

emiliesoutiras:
On data where some serialized products do not yet have a sequence, writing on all or part of the products should not assign a sequence to all modified products regardless of their tracking status, but only to those whose tracking is not 'none'.

(recreate from PR https://github.com/OCA/product-attribute/pull/1614)

### DIFF
--- a/product_lot_sequence/models/product.py
+++ b/product_lot_sequence/models/product.py
@@ -86,24 +86,46 @@ class ProductTemplate(models.Model):
 
     def write(self, vals):
         seq_policy = self.env["stock.lot"]._get_sequence_policy()
-        if seq_policy == "product":
-            for template in self:
+        if seq_policy == "global":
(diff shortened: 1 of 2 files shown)


emiliesoutiras:
cc @Kev-Roche, @metaminux

metaminux (approved):

Kev-Roche (approved):
LGTM tested

dreispt:
Con you squash commits? That will cause less noise on the history and easier to fw/backport.

emiliesoutiras:
> Con you squash commits? That will cause less noise on the history and easier to fw/backport.

It's done ;)

metaminux:
Hi @dreispt 

Can you look at this PR, please ?

PR #1846, opened 2025-01-17
[16.0][FIX] product_packaging_level: Removed usage of undefined function

### LINKED ISSUE #1845: product_packaging_level: AttributeError: 'product.packaging.level' object has no attribute '_active_languages'
Module product_packaging_level is raising an error message, that the function _active_languages is not defined.

## Module
product_packaging_level

## Describe the bug
Module product_packaging_level is raising an error message, that the function _active_languages is not defined. This function indeed does not exist, so we need to remove the function call.

## To Reproduce
**Affected versions**: 16.0

**Expected behavior**
Only functions are used, which exist.

baguenth:
Fixes #1845

### DIFF
--- a/product_packaging_level/models/product_packaging_level.py
+++ b/product_packaging_level/models/product_packaging_level.py
@@ -12,7 +12,7 @@ class ProductPackagingLevel(models.Model):
     def _default_language(self):
         lang_code = self.env["ir.default"].get("res.partner", "lang")
         def_lang_id = self.env["res.lang"]._lang_get_id(lang_code)
-        return def_lang_id or self._active_languages()[0]
+        return def_lang_id or self.env["res.lang"].search([]).ids[0]
 
     name = fields.Char(required=True, translate=True)
     code = fields.Char(required=True)


baguenth:
@simahawk @ValentinVinagre @HaraldPanten @luis-ron Kindly asking for a review

product_packaging_level/models/product_packaging_level.py
```diff
@@ -12,7 +12,7 @@ class ProductPackagingLevel(models.Model):
     def _default_language(self):
         lang_code = self.env["ir.default"].get("res.partner", "lang")
         def_lang_id = self.env["res.lang"]._lang_get_id(lang_code)
-        return def_lang_id or self._active_languages()[0]
+        return def_lang_id or self.env["res.lang"].search([]).ids[0]
```
ValentinVinagre, line 15: Exist in this module:
https://github.com/odoo/odoo/blob/ef6f0b0f5eaba19c92cfa8046b5ab4a60bfacdb3/addons/website/models/website.py#L65

baguenth: That's true. But the model "product.packaging.level" doesn't inherit from website model, so this function is not member of "product.packaging.level".

ValentinVinagre: why don't use by default the lang of the user?

baguenth: We can also use the default lang of the user. I just kept what was originally intended by the author. The author wanted to use _active_languages, so I replaced the function call with what is inside _active_languages (website).

ValentinVinagre: We do not use this module but the change is correct if you want to keep that functionality. It is best to have it reviewed by a repository PSC.

ValentinVinagre (approved):
LGTM 👍🏻

HaraldPanten (approved):

PR #1848, opened 2025-01-17
17.0 mig product_usability

arthru:
I just changed the version in manifest file.

The module installs and the menu and views are working


legalsylvain:
Thanks

/ocabot merge nobump

PR #1853, opened 2025-01-21
[16.0][FIX] product_category_active: in odoo, disabled items are muted by default.

legalsylvain:
this patch makes also the current module compatible with product_category_type. (same repo)


quentinDupont (approved):
trivial and great :heavy_check_mark:

PR #1857, opened 2025-01-23
[18.0][IMP] product_packaging_level: remove renaming hook

trisdoan:
### This change

- Module was renamed to product_packaging_level from 16.0
    - A hook is added to handle the smooth transition: https://github.com/OCA/product-attribute/blob/93e4b9cefe0a62564de6c7b8302f3442125e7337/product_packaging_level/hooks.py#L8-L25
    - The hook has been improved in https://github.com/OCA/product-attribute/pull/1809

- We believe we should not carry on this hook forever
   - Keeping this hook is unnecessary as openupgrade would have run it at the 16.0 step
   - the same as we don't keep migrations across versions


trisdoan:
Hello @rousseldenis , @ChrisOForgeFlow, do you agree with this?

sebalix (approved):
I agree with this (even if it's not the more convenient for those who migrate with Odoo E.), we have to be consistent and maintain the code only where the migration occurs.

PR #1860, opened 2025-01-25
[18.0][MIG] product_pricelist_direct_print

bosd:
depends on: #1864


bosd:
This option is not working yet. It adds all items from the db.
image


Edit: Fixed now :green_heart:

Dranyel-Bosd (approved):
LGTM :1st_place_medal:

bosd:
@legalsylvain Can you please review?

bosd:
@CarlosRoca13 Can you please review this one?

CarlosRoca13:
I'm not sure this reference really convinces me; it could be confusing. Couldn't we avoid adding a reference when there isn't one?
image

bosd:
@CarlosRoca13 I'm sorry but I don't understand your point.
The reference has not been changed by this pr.
When there is no reference given. link in example below for the `Office Combo` it will not be shown.
image

CarlosRoca13:
It is happening when the option "Show variants" is selected:
image

image

bosd:
Yes, I got that. When you set an default_code on the product_template and generate the pricelist without the variants option. Then the default_code of the product_template is shown on the pricelist. (If the field is avalable)
When you check the box `Show variants`  then the report is listing the `default_code` set on all the product variants.

When there are multiple product.products the field is not available.
eg. customizable desk:
image

With show variants:
image

That is exactly what is supposed to be happening. right?

CarlosRoca13:
> That is exactly what is supposed to be happening. right?

I believe so, or at least it should appear as it did in previous versions, since that reference that shows up when it's not set is a bit annoying to me...

Example from previous versions:
image

victoralmau:
In addition to what my colleague comments on https://github.com/OCA/product-attribute/pull/1860#issuecomment-2808825321

Can you unify all commits in the migration commit?
ejemplo

pedrobaeza (approved):
I have rebased + squash the PR, reducing significantly the number of commits, so we can proceed with the merging:

/ocabot merge nobump

PR #1747, opened 2024-10-10
[16.0][FIX] product_pricelist_alternative: No alternative price in waterfall pricelists

santostelmo:
Initial design of this module expected that:
When one pricelist inherits the prices of another pricelist in standard waterfall (compute_price == "formula" and base == "pricelist"), do not inherit the alternative prices on parent pricelist.

PR #1863, opened 2025-01-28
[18.0][MIG] product_pricelist_alternative: Migration to 18.0

vvrossem:
This PR adds the context key `force_price_recomputation`, used in `sale`'s `sale.order::_recompute_prices()`, triggered by `action_update_prices()` button that recomputes the unit price of all products based on the new pricelist.

This is done in order to correctly select the alternative pricelist when changing the pricelist on a SO and updating the prices.

TODO: 

- [ ] include https://github.com/OCA/product-attribute/pull/1747 once merged


product_pricelist_alternative/views/product_pricelist_view.xml
```diff
@@ -34,13 +34,7 @@
                 </group>
             </group>
             <xpath
-                expr="//page[@name='pricelist_rules']/field[@name='item_ids']/tree/field[@name='product_id']"
-                position="after"
-            >
-                <field name="alternative_pricelist_policy" />
-            </xpath>
-            <xpath
-                expr="//page[@name='pricelist_rules']/field[@name='item_ids']/tree[@groups='product.group_sale_pricelist']/field[@name='name']"
```
vvrossem: note: `groups="product.group_sale_pricelist"` is [removed in 18.0](https://github.com/odoo/odoo/blob/18.0/addons/product/views/product_pricelist_views.xml#L74-L87).
Only a single `<list>` remain whereas [before two <tree> elements existed](https://github.com/odoo/odoo/blob/16.0/addons/product/views/product_pricelist_views.xml#L75-L107)

vvrossem:
@santostelmo would you mind reviewing this PR?

santostelmo (approved):
LG

vvrossem:
@grindtildeath would you mind reviewing this PR? :pray:

product_pricelist_alternative/__manifest__.py
```diff
@@ -3,14 +3,15 @@
 
 {
     "name": "Product Pricelist Alternative",
-    "version": "18.0.1.0.0",
+    "version": "18.0.1.1.0",
```
rousseldenis: @vvrossem If you don't provide migration scripts, this is not needed as updated by the bot.

product_pricelist_alternative/__manifest__.py
```diff
@@ -3,14 +3,15 @@
...
     "development_status": "Beta",
     "category": "Product",
     "summary": "Calculate product price based on alternative pricelists",
     "author": "Camptocamp, Odoo Community Association (OCA)",
     "website": "https://github.com/OCA/product-attribute",
     "license": "AGPL-3",
     "depends": [
+        "sale",
```
rousseldenis: IMHO, this is not correct. The module should not depends on sale for such changes.

You should do a glue module.

rousseldenis (changes requested):
Don't agree on `sale` dependency addition

santostelmo:
> Don't agree on `sale` dependency addition

At first glance, product dependency is enough

product_pricelist_alternative/__manifest__.py
santostelmo: 16.0 version depends of product
https://github.com/OCA/product-attribute/blob/ec24dffa276cbbf92a316e110a7c929c736e2809/product_pricelist_alternative/__manifest__.py#L14

santostelmo (changes requested):
Please make it depend of product

product_pricelist_alternative/__manifest__.py
vvrossem: I'll do a glue module `sale_product_pricelist_alternative` and submit it to OCA/sale-workflow, do we agree?

vvrossem: or should the glue module be added to OCA/product-attribute? :thinking:

product_pricelist_alternative/models/product_pricelist.py
```diff
@@ -67,6 +67,9 @@ def _compute_price_rule(
         compute_price=True,
         **kwargs,
     ):
+        if self.env.context.get("force_price_recomputation"):
+            compute_price = True
```
grindtildeath: IMO you don't need to depend on `sale` for adding such code based on a context key.

grindtildeath (commented):
Code looks good, drop the dependency and keep your extra code, no need for a glue module

vvrossem:
- `sale` dependency has been removed.
- context key is kept [as suggested by grindtildeath](https://github.com/OCA/product-attribute/pull/1863#pullrequestreview-2598582630)

santostelmo (approved):

sebduch73:
Hello @rousseldenis 
Can you validate the PR please if it's all good ?
thx

grindtildeath (approved):

gurneyalex (approved):

PR #1864, opened 2025-02-01
[18.0][MIG] product assortment

bosd:
Superseeding #1823

Forward ported missing commits.
Adjusted code to new api behavior.
Fixed the missing assortments from being displayed when coming from the smartbutton on the partner form.


Dranyel-Bosd (approved):
LGTM :100:

BT-jcolmeiro (approved):
LGTM :+1:  Functional & Code review. Module does what is expected and issue of the original migration PR is fixed now.

bosd:
@rafaelbn Can you please merge?

rafaelbn:
Hello @bosd , sorry. I'm not PSC in this repo 😢

bosd:
@JordiBForgeFlow or @rousseldenis Can you merge? 🙏

PieterPaulussen:
I would like to raise the discussion which has been added in this PR: https://github.com/OCA/product-attribute/pull/1823

The statements @bosd made are incorrect and the discussion should be resolved first.

PieterPaulussen (changes requested):
Not all discussions have been resolved. Please see https://github.com/OCA/product-attribute/pull/1823

bosd:
For users, skipping in. We have 2 competing PR's now.
This one retaining functionality / behaviour since V12.
And the "Fixed"version in #1823 (edit: should be different behaviour one.)

> The statements @bosd made are incorrect and the discussion should be resolved first.

FWIW, My statements are correct. We just have different use case / opinion of how this module should function from V17 and onwards. (details are in the other PR).

PieterPaulussen:
> FWIW, My statements are correct. We just have different use case / opinion of how this module should function from V17 and onwards. (details are in the other PR).

@bosd  Sorry about that. We indeed have a different perspective on this module. No insult intended of course.

bosd:
@PieterPaulussen Thanks for your constructive stance on this.
Let's iron out the differences between the PR's and ideally come to a quick resolution.
So this and the depending modules can be merged.

Ideally, all of these cases would already have been covered by unit tests. So this discussion about functionality would not be necessary.
(I remember adding some additional assortment unit tests, but it could as well be in depending modules).

I have replicated your example no3.
Now with products available on both runboats.
There is no difference between the two PR's.
image
Result is the same as on V16.

I[ tried to make a diff](https://github.com/bosd/product-attribute/compare/18.0-mig-product_assortment2...The-O-Team:product-attribute:18.0-mig-product_assortment) between the two PR's. But did not succeed in getting an usable result.
In this PR's I have fixed some typo's.
From memory I included some missing commits. (Like the demo data)
This PR's was also based on your pr. (The change in unit tests to make them pass with the new orm behaviour is reverted.)
The additional unit test you made is included here. (Sorry, authorship has not been retained, because it is squashed into the migration commit.)


I will indicate with some comments, the changes in how the expression is built.
So, from my point of view, this PR, is more "complete".

@PieterPaulussen I'd be happy to include some modifications here if needed.
Are there use cases which cannot be fulfilled with this version?

product_assortment/models/ir_filters.py
```diff
@@ -83,17 +81,17 @@ def get_partner_domain_fields(self):
...
         return field_set
 
     @api.depends("partner_ids", "partner_domain")
     def _compute_all_partner_ids(self):
         """Summarize selected partners and partners from partner domain field"""
-        for ir_filter in self:
+        for ir_filter in self.sudo():
             if not ir_filter.is_assortment:
                 ir_filter.all_partner_ids = False
-            elif ir_filter.partner_domain != "[]":
```
bosd: Likely this is the big differentiator between the PR's.
`elif ir_filter.partner_domain != "[]":`
would never evaluatue to true. Because by default the value assigned is `[]`
Meaning comparison should be `ir_filter.partner_domain != []:`

PieterPaulussen:
> @PieterPaulussen I'd be happy to include some modifications here if needed. Are there use cases which cannot be fulfilled with this version?

I have no objection other than my comment [on my PR here](https://github.com/OCA/product-attribute/pull/1823#issuecomment-2684211108), but the fact of the matter remains the same. 

We are both opinionated on the outcome of this discussion so I think we should have a referee or something ;-). Let's have the other peers of the community shine their light on this.

bosd:
> We are both opinionated on the outcome of this discussion so I think we should have a referee or something ;-). Let's have the other peers of the community shine their light on this.

Well, I rather try to get consensus among us. Because we are also blocking each other PR's.
And we clearly are the people with the most knowledge on the topic at hand.

Here is the "All test" among the PR's.
image

Here is the "All except storage box"side by side.
image


**Here is the difference between pr's**
The Point and click test.
(Use case of my client)
image
There is a "workaround" to get the desired result. But it's quite dirty.

product_assortment/models/ir_filters.py
bosd: @PieterPaulussen Can you agree, that the evaluation in this code is wrong?
(Has been there for quite some versions, but undeteced by unit tests / use cases)

PieterPaulussen:
Aren't we going around in circles now? I still think the default behaviour of the functionality should always support the ORM way of working. If you need to add functionality to a migrated module to restore the behaviour which relied on a different ORM result, is that not counter intuitive when you want to use this module and you start from scratch with the configuration? Would you not always desire the module to behave the same as the ORM?

bosd:
> Aren't we going around in circles now?

Not really, IMO, we are ironing out the desired behaviour of this module.
When there is consensus, we can adopt the code accordingly and add unit tests to increase coverage and insure functionality among versions.

It's quite a tricky thing. As there where some (undiscovered) bugs in the code of this module.
And the unit tests of the module where a bit lacking for all the different scenario's.
( Althoug it has a mature label )


>  I still think the default behaviour of the functionality should always support the ORM way of working. If you need to add functionality to a migrated module to restore the behaviour which relied on a different ORM result, is that not counter intuitive when you want to use this module and you start from scratch with the configuration? Would you not always desire the module to behave the same as the ORM?

Yeah, how the ORM behaves is a factor. But also how we build the expression to pass to the orm.
image

PR #1499, opened 2024-01-19
[15.0][IMP] product_pricelist_direct_print: Allow printing the products sold to a client from x date

CarlosRoca13:
cc @Tecnativa TT46996

ping @pedrobaeza @carlosdauden

PR #1866, opened 2025-02-02
[16.0][FW] [15.0][IMP] product_pricelist_direct_print: Allow printing the products sold to a client from x date

bosd:
Port from 15.0 to 16.0:
- #1499

### DIFF
--- a/product_template_tags/i18n/it.po
+++ b/product_template_tags/i18n/it.po
@@ -7,7 +7,7 @@ msgstr ""
 "Project-Id-Version: Odoo Server 16.0+e\n"
 "Report-Msgid-Bugs-To: \n"
 "POT-Creation-Date: 2023-04-10 12:58+0000\n"
-"PO-Revision-Date: 2024-07-29 16:58+0000\n"
+"PO-Revision-Date: 2025-04-02 09:06+0000\n"
 "Last-Translator: mymage <stefano.consolaro@mymage.it>\n"
 "Language-Team: \n"
 "Language: it\n"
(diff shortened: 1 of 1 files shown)


product_pricelist_direct_print/wizards/product_pricelist_print.py
```diff
@@ -268,10 +272,15 @@ def force_pricelist_send(self):
...
     def _get_sale_order_domain(self, partner):
-        return [
+        domain = [
             ("state", "not in", ["draft", "sent", "cancel"]),
             ("partner_id", "child_of", partner.id),
         ]
+        if self.product_selling_date_threshold:
+            domain = expression.AND(
+                [domain, [("date_order", ">=", self.product_selling_date_threshold)]]
+            )
```
carlosdauden:
```suggestion
            domain.append(("date_order", ">=", self.product_selling_date_threshold))
```
I know I review the PR in the previous version and I didn't put this suggestion, but seeing that it's a very simple domain we can save ourselves the use of expression.AND.

carlosdauden (approved):

product_pricelist_direct_print/wizards/product_pricelist_print.py
legalsylvain: expression.AND is more explicit. And that is exactly the objective of this function.

bosd:
@CarlosRoca13 Can you please review?

bosd:
hmm. Why is this one automatically closed?

pedrobaeza:
Because you may rebase over an incorrect branch which is not related to current commit history.

PR #1868, opened 2025-02-03
[18.0][FIX] product_medical: fix pre-migrate.py

sebalix:
Odoo upgrade is complaining about the `migrate` signature, even if decorated with `openupgrade.migrate`. Fixed by switching to the std signature.


sebalix:
Was fixed in `openupgradelib` (to install from `master`):
- https://github.com/OCA/openupgradelib/pull/396

Closing

PR #1879, opened 2025-02-06
[18.0][MIG] product_variant_route_mto: Migration to 18.0

chaule97:
- move from `OCA/stock-logistics-workflow` to `OCA/product-attribute` , rename it `product_variant_route_mto`
- I format-patch from https://github.com/OCA/stock-logistics-workflow/pull/1461
- I updated `route_ids` `stored=True` because [this code](https://github.com/odoo/odoo/blob/bfab889e4d1578092e03e50a893773d549c78648/addons/stock/models/product.py#L667C9-L667C98) made error `Non-stored field product.product.route_ids cannot be searched. `


jbaudoux (approved):

jbaudoux:
cc @mmequignon @grindtildeath

product_variant_route_mto/models/product_product.py
```diff
@@ -0,0 +1,53 @@
...
+        store=True,
+        readonly=False,
+        help=IS_MTO_HELP,
+    )
+
+    route_ids = fields.Many2many(
+        "stock.route",
+        compute="_compute_route_ids",
+        domain="[('product_selectable', '=', True)]",
+        store=True,
```
grindtildeath: IMO we do not want to store this field and must define a search function instead.

The goal with this field is to use the `route_ids` from the template - plus/minus the MTO route depending on the setting of the variant - when reading/searching on the product variant. By storing this, the route_ids from the template will not be used anymore, with the possibility to have inconsistencies between product and template as the product variant routes will not reflect anymore the routes from the template.

grindtildeath (changes requested):

product_variant_route_mto/models/product_product.py
jbaudoux: Please point the open PR in v14 fixing this so that it can be ported once finished

jbaudoux (changes requested):
Waiting fix on v14

product_variant_route_mto/models/product_product.py
simahawk: I agree.
@chaule97 can you please drop the store attribute and add a search function (if necessary)?
Just here, not on v14 for now.

chaule97: I have updated it, please review my new updates

stock_product_variant_mto/models/product_product.py
```diff
@@ -23,30 +23,50 @@ class ProductProduct(models.Model):
...
+                    template_routes += mto_route
+                elif not product.is_mto and mto_route in template_routes:
+                    template_routes -= mto_route
+
+            product.route_ids = template_routes
+
+    def _search_route_ids(self, operator, value):
+        mto_route = self.env.ref("stock.route_warehouse0_mto", raise_if_not_found=False)
+        if not mto_route:
+            return [(0, "=", 1)] if operator in ("=", "in") else [(0, "=", 0)]
```
grindtildeath: :warning: you need to check the `value` before returning such a domain.

If I search for something like `[('route_ids`, `in`, buy_route.id)]` I would expect to get the products where the buy route is defined and the MTO being archived or not should not have any influence.

grindtildeath (commented):
I guess we need a test on the search of route_ids :grimacing:

jbaudoux (changes requested):
There are several issues, i take it over here https://github.com/OCA/product-attribute/pull/1970

PR #1882, opened 2025-02-10
[16.0][ADD] product_expiry_alert

rousseldenis:


lmignon (approved):

product_expiry_alert/__manifest__.py
```diff
@@ -0,0 +1,19 @@
+# Copyright 2024 ACSONE SA/NV
```
FrancoMaxime, line 1:
```suggestion
# Copyright 2025 ACSONE SA/NV
```

If i read correctly, every copyright year's date are 2025. So i think this one should be 2025 too, no?

FrancoMaxime (approved):
LGTM: code review

PR #1887, opened 2025-02-13
[16.0][IMP] product_pricelist_direct_print: Add decimal precision to price fields

CarlosRoca13:
With this change we are sure that the prices shown on the report has the correct precision.

cc @Tecnativa TT55097

ping @carlosdauden @pedrobaeza


pedrobaeza (approved):

legalsylvain (approved):
Thanks !

/ocabot merge patch

PR #1892, opened 2025-02-19
[18.0][ADD] product_attribute_value_avoid_auto_fill: New module to avoid add all values of attribute when selecting it

CarlosRoca13:
cc @Tecnativa TT55171


product_attribute_value_auto_fill_option/models/product_template_attribute_line.py
```diff
@@ -0,0 +1,28 @@
...
+            )
+        elif (
+            not self.attribute_id.fill_all_values
+            and self.attribute_id.create_variant == "no_variant"
+        ):
+            self.value_ids = self.value_ids.filtered(
+                lambda pav: pav.attribute_id == self.attribute_id
+            )
+        else:
+            return super()._onchange_attribute_id()
```
carlosdauden: I think we could simplify it like this:
```suggestion
            return super()._onchange_attribute_id()
        self.value_ids = False
```

carlosdauden (changes requested):

CarlosRoca13:
Changes done @carlosdauden

carlosdauden (approved):

sergio-teruel (approved):

PR #1893, opened 2025-02-19
[FIX] [16.0] product_sticker: Allow to select which models can use Stickers

Shide:
This change is introduced because when creating submodules that uses Stickers, all Stickers (that matches category, attributes, etc) are chosen to be displayed in all models.

Real Example: 
The [Risk Insurance Product Sticker](https://github.com/OCA/credit-control/tree/16.0/partner_risk_insurance_product_sticker_invoice_report) module introduces a new field which pretends to show only in invoices what Invoice has been risk secured.
If you create a Sticker with the proper configuration, the Sticker shows in the [Picking reports](https://github.com/OCA/stock-logistics-reporting/tree/16.0/stock_picking_report_product_sticker), and should not appear in this reports.

With this change, you can select on which models you want to apply the sticker.
If no model is selected, will be available for all models (what is happening right now).

MT-9075 @moduon @rafaelbn @yajo @fcvalgar @EmilioPascual please review if you want 😄

### DIFF
--- a/product_sticker/models/product_product.py
+++ b/product_sticker/models/product_product.py
@@ -11,21 +11,11 @@ class ProductProduct(models.Model):
         action["domain"] = [("id", "in", stickers.ids)]
         return action
 
-    def _get_sticker_arguments(self):
-        pavs = self.product_template_variant_value_ids.product_attribute_value_id
-        return {
-            "categories": self.categ_id,
-            "attributes": pavs.attribute_id,
-            "attribute_values": pavs,
-        }
-
     @api.returns("product.sticker")
-    def get_product_stickers(self):
-        """Product Stickers related to this Product Variant and its Template"""
-        # Product Template: Common stickers
-        pt_stickers = self.product_tmpl_id.get_product_stickers()
-        # Product Product: Specific stickers
-        pp_stickers = self.env["product.sticker"]._get_stickers(
-            **self._get_sticker_arguments()
+    def get_product_stickers(self, extra_domain=None):
+        """Product Stickers related to this Product Variant and
+        its Template for certain models"""
+        return self.env["product.sticker"]._get_stickers(
+            self,
+            extra_domain=extra_domain,
         )
-        return pt_stickers | pp_stickers
--- a/product_sticker/models/product_sticker.py
+++ b/product_sticker/models/product_sticker.py
@@ -22,6 +22,11 @@ class ProductSticker(models.Model):
         max_height=64,
         store=True,
     )
+    available_model_ids = fields.Many2many(
+        comodel_name="ir.model",
+        string="Available Models",
+        help="Models where this sticker is available. Empty means all models.",
+    )
     product_category_id = fields.Many2one(
         comodel_name="product.category",
         string="Category",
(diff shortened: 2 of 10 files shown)


Shide:
Other PRs are being created to match this new way to work:
- https://github.com/OCA/account-invoice-reporting/pull/353
- https://github.com/OCA/stock-logistics-reporting/pull/370

product_sticker/models/product_product.py
```diff
@@ -20,12 +20,14 @@ def _get_sticker_arguments(self):
         }
 
     @api.returns("product.sticker")
-    def get_product_stickers(self):
-        """Product Stickers related to this Product Variant and its Template"""
+    def get_product_stickers(self, model_names=None):
```
yajo: suggestion: if you use compatible data types, reasoning about them becomes simpler. Some suggestions follow the same idea. Here, for example, is easier to work with `()` as a default:

```suggestion
    def get_product_stickers(self, model_names=()):
```

product_sticker/models/product_sticker.py
```diff
@@ -91,6 +96,27 @@ def _build_sticker_domain_company(self):
...
+                        "in",
+                        self.env["ir.model"]
+                        .sudo()
+                        .search([("model", "in", model_names)])
+                        .ids,
+                    )
+                ],
+                [("available_model_ids", "=", False)],
+            ]
+        )
```
yajo: IIUC this should be equivalent:

```suggestion
        return [
            "|",
            ("available_model_ids.model", "in", model_names),
            ("available_model_ids", "=", False),
        ]
```

product_sticker/models/product_sticker.py
```diff
@@ -161,23 +187,27 @@ def _build_sticker_domain_attributes(self, attributes=None, attribute_values=Non
         return attribute_domain
 
     def _get_sticker_domains(
-        self, categories=None, attributes=None, attribute_values=None
+        self, categories=None, attributes=None, attribute_values=None, model_names=None
```
yajo:
```suggestion
        self, categories=None, attributes=None, attribute_values=None, model_names=()
```

product_sticker/models/product_sticker.py
```diff
@@ -161,23 +187,27 @@ def _build_sticker_domain_attributes(self, attributes=None, attribute_values=Non
...
         )
-        return [company_domain, category_domain, attribute_domain]
+        model_domain = self._build_sticker_domain_model(model_names)
+        return [company_domain, category_domain, attribute_domain, model_domain]
 
     @api.model
-    def _get_stickers(self, categories=None, attributes=None, attribute_values=None):
-        """Get stickers for given categories, attributes and attribute values"""
+    def _get_stickers(
+        self, categories=None, attributes=None, attribute_values=None, model_names=None
```
yajo:
```suggestion
        self, categories=None, attributes=None, attribute_values=None, model_names=()
```

product_sticker/models/product_template.py
```diff
@@ -22,8 +22,10 @@ def _get_sticker_arguments(self):
         }
 
     @api.returns("product.sticker")
-    def get_product_stickers(self):
-        """Attribute Stickers related to this Product Template and its variants"""
+    def get_product_stickers(self, model_names=None):
```
yajo:
```suggestion
    def get_product_stickers(self, model_names=()):
```

product_sticker/tests/test_product.py
```diff
@@ -2,6 +2,23 @@
 
 
 class TestStickersOnProducts(ProductStickerCommon):
+    def _test_model_availability(self, product):
+        random_models = self.env["ir.model"].search([], limit=2)
```
yajo: thought: I understand that this *should not fail*. However, it opens the door for uncontrolled inputs, and thus non-reproducible tests... Couldn't you just use some pre-controlled set of models? Just to make sure tests are reproducible. For example:

```suggestion
        random_models = self.env.ref("base.model_ir_model") + self.env.ref("base.model_ir_ui_view")
```

fcvalgar (approved):
LGTM, thank you @Shide

rafaelbn (approved):
I think this is an improvement 👍🏼 😄

yajo (approved):
Some non-blocking suggestions.

Shide:
I've added a **kwargs to allow easy inherintance when proceeded.

Shide:
@rousseldenis do you have any questions or doubts about this change? Before merging...

Shide:
I've simplified A LOT all the logic and method inheritance...

PR #1902, opened 2025-02-26
[IMP] [16.0] product_sticker: Allow to choose size

Shide:
Allows to choose the size of the sticker

MT-8836 @moduon @rafaelbn @yajo @fcvalgar please review if you want 😄

### DIFF
--- a/product_sticker/views/product_sticker_views.xml
+++ b/product_sticker/views/product_sticker_views.xml
@@ -47,6 +47,7 @@
                                 options="{'no_create': True}"
                                 groups="product.group_product_variant"
                             />
+                            <field name="image_size" />
                             <field name="show_sticker_note" widget="boolean_toggle" />
                             <field
                                 name="note"
@@ -88,6 +89,7 @@
                     name="product_attribute_value_id"
                     groups="product.group_product_variant"
                 />
+                <field name="image_size" optional="hide" />
                 <field
                     name="company_id"
                     groups="base.group_multi_company"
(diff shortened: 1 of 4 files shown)


product_sticker/views/product_sticker_views.xml
```diff
@@ -289,4 +292,30 @@
             </search>
         </field>
     </record>
+
+    <template id="report_sticker_template">
+        <!-- Required parameters: stickers, position -->
+        <t t-if="stickers">
+            <t t-set="extra_classes" t-value="stickers._get_extra_classes(position)" />
+            <div t-attf-class="col-12 #{extra_classes}">
```
yajo: suggestion: I don't think we should call a method in the model to get some classes. We'd be putting view logic outside of the view, so it's harder to understand the relationships between the classes and the outcome.

Instead, you can do this here:

```suggestion
            <t t-set="position_classes" t-value="'text-end mt-1 mb-1' if position == 'top_right' else 'text-start mt-5 mb-1' if position == 'bottom_left' else ''" />
            <div t-attf-class="col-12 #{position_classes} #{extra_classes}">
```

By leaving `extra_classes` without definition, IIUC, you allow the template caller to set it like this:

```xml
<t t-call="product_sticker.report_sticker_template>
  <t t-set="extra_classes">a b c</t>
</t>
```

This still enables sensible extensibility, and avoids relying on model methods just for view logic.

yajo (approved):

product_sticker/views/product_sticker_views.xml
Shide: Done ❤️

fcvalgar (approved):
Functional review.

LGTM, thank you @Shide

PR #1904, opened 2025-02-28
[18.0][MIG] product_assortment: Migration to 18.0

PieterPaulussen:


PieterPaulussen:
@bosd I've updated the migration to cover the suggested changes and added a migration script to cover your the use cases for your customers. Feel free to review and provide feedback.

bosd:
Crossposting my feedback from the old pr:

From experience some non technical users get nervous when seeing a domain field on the UI.
(Just looking at the UI maked their jaws drop.)

> Do you want me to work with that?

Yes, It is a matter of eductation..
But they just want to have a very simple user interface, Where they can select a partner and add (whitelist) products.

Adding some "bogus" domain as a workaround will solve the functional level.
The proposal of adding...
domain=[] with domain=[(0, '=', 1)]

But it is even more confusing for the non technical user.


What do you think about adding a checkbox apply_advanced_product_domain.
When checking that, it will show the domain field and widget.
(Which we could hide be default). So the UI is very nice, simple and clean.
When that box is checked, the domain is displayed and evaluated.
(No matter if it is [] or [(0, '=', 1)])

Likely we can do the same for the partner fields.
(So useability is the same)
In such way we can cover with this module a "Simple" assortment definition use-case.
And well as advanced use-cases for the more technical / power user which used domains or a combination of domain and white/black list.

@PieterPaulussen What do you think?

PieterPaulussen:
> Crossposting my feedback from the old pr:
> 
> From experience some non technical users get nervous when seeing a domain field on the UI. (Just looking at the UI maked their jaws drop.)
> 
> > Do you want me to work with that?
> 
> Yes, It is a matter of eductation.. But they just want to have a very simple user interface, Where they can select a partner and add (whitelist) products.
> 
> Adding some "bogus" domain as a workaround will solve the functional level. The proposal of adding... domain=[] with domain=[(0, '=', 1)]
> 
> But it is even more confusing for the non technical user.
> 

But is that not what the new filter editor is all about? To make it more accessible for non-technical users? Please also keep in mind the form explicitly states 'All xxx products' when the domain is empty.

> What do you think about adding a checkbox apply_advanced_product_domain. When checking that, it will show the domain field and widget. (Which we could hide be default). So the UI is very nice, simple and clean. When that box is checked, the domain is displayed and evaluated. (No matter if it is [] or [(0, '=', 1)])

Will this not make the `ir.filter` behaviour even more complicated to understand? Because the `domain` field (which is the main domain for the products) is the core feature of the `ir.filter` model? Otherwise it would be more efficient to re-define a new model `product.assortment` and start the functionality from scratch?

What about inverting the logic of that flag? We could add a flag 'No product at all' (or some other label), and when checked, we hide the `domain`field and set the void domain `[(0, '=', 1)]` automatically? Would that work for you?

bosd:
> But is that not what the new filter editor is all about? To make it more accessible for non-technical users?
@PieterPaulussen 
I agree, But yes, from experience, domain is still too technical for some people.

> Please also keep in mind the form explicitly states 'All xxx products' when the domain is empty.

Yes, but with the proposed checkbox. It would be a intentional action of the user to add all products. By activating the advanced "features". With my proposal, the domain widget would be hidden by default. Thus not showing the "all xx products" text.


> Otherwise it would be more efficient to re-define a new model product.assortment and start the functionality from scratch?

IMO, a single module. Which provides a 'simple' scenario and a 'advanced' one, is prefferred. Over having two modules with very similar / overlapping functionality.
( Also for an upgrade path. First let them use the simple whitelist/blacklist features. When they have more experience, they can evolve to use the "scary" advanced domain widget. )

 > What about inverting the logic of that flag? We could add a flag 'No product at all' (or some other label), and when checked, we hide the domainfield and set the void domain [(0, '=', 1)] automatically? Would that work for you?

Yes, technically this will work. But still would prefer the other way around. An aditional checkbox to "show advanced features" (the domain field). Would IMO make more sense from users point of view. Instead of a button / checkbox "Show Simple version".

Even when inverting. We have the same problem as before. An unclear user experience.
The user will check the box  'No product at all' (or some other label).
Then will whitelist products...

Resulting in a
"No product at all" + the list of whitelisted products..

PieterPaulussen:
> Even when inverting. We have the same problem as before. An unclear user experience. The user will check the box 'No product at all' (or some other label). Then will whitelist products...
> 
> Resulting in a "No product at all" + the list of whitelisted products..

I disagree that my proposal has an unclear user experience. (You seem to keep forgetting the filter widget displays a text 'All xxx products' which clearly states what is included. Even wit a link to the products.) I believe we need to align it more with other functionality in Odoo. By comparison the `ir.rule` model provides similar functionality where 'all' has priority over 'limitations'.

I stand by my suggestion to meet you half way by adding a flag to explicitly disable that open domain and exclude all products.

bosd:
>  (You seem to keep forgetting the filter widget displays a text 'All xxx products' which clearly states what is included. Even wit a link to the products.)

I did not forget that. That's why I propose to hide that widget when unused. Did you catch that?
I think it thicks all the boxes mentioned here so far:
🟢 Simple and advanced use case support
🟢 Clear and easy to understand UI
🟢 No bogus `0 = 1`  evaluation which scares non tech users
🟢 Aligned with odoo functionality (Domain evaluation behaves the same as in core.)


Again if we go for a workaround it is very hard to explain to a non-technical user why there should be a  `[(0, '=', 1)]` for something he is not using.
Remember we are not all programmers, who are using this module. Assortments can be created at C-level or sales.

Even when the `0=1`  would be filled by a migration script / button. You run into this confusing situation for those users.

PieterPaulussen:
> > (You seem to keep forgetting the filter widget displays a text 'All xxx products' which clearly states what is included. Even wit a link to the products.)
> 
> I did not forget that. That's why I propose to hide that widget when unused. Did you catch that? I think it thicks all the boxes mentioned here so far: 🟢 Simple and advanced use case support 🟢 Clear and easy to understand UI 🟢 No bogus `0 = 1` evaluation which scares non tech users 🟢 Aligned with odoo functionality (Domain evaluation behaves the same as in core.)

🔴 The UI issue would already be resolved by adding the flag to set the void domain.
🔴 Your solution does not respect the ORM behaviour, otherwise we would not be having this discussion.

> Again if we go for a workaround it is very hard to explain to a non-technical user why there should be a `[(0, '=', 1)]` for something he is not using. Remember we are not all programmers, who are using this module. Assortments can be created at C-level or sales.

This is where my 'meet you halfway' solution comes in handy. The flag would automatically set the correct domain and hide the widget in the UI. Which tackles the 'technical' complexity of setting the void domain.

> Even when the `0=1` would be filled by a migration script / button. You run into this confusing situation for those users.
This is only to accomodate the pre-existing configurations you have for your specific customers and was suggested by @carlosdauden.

I believe we are getting back to the stalemate discussion between yours and mine.
I've made the following concession for you: Added a specific migration script to resolve the pre-existing situations.

I'm willing to add another concession to 'simplify' the correct behaviour: Add a new UI flag + set the void domain automatically.

It would be nice if you would meet me half way on this... Don't you think? Otherwise we will end up in the same stalemate.

pedrobaeza:
Please unblock this and remember that this is a migration. Future improvements can be done in later PRs.

PieterPaulussen:
@pedrobaeza I understand, but if we need to get back to the core issue: The ORM behaviour of `ir.filter` has changed. Should the new module respect the new change or should broken functionality be kept?

@carlosdauden has pointed out that the originally intended behaviour has changed in a previous version and remained undetected until now.

If you have an example of a precedent which could resolve the matter, please feel free to point it out.

pedrobaeza:
Yeah, I have followed the conversation and I agree that part should be changed. I was talking about all the extra flags and UI simplifications. @bosd I think current status is enough for validating the migration. Extra things can be performed later. Do you agree?

bosd:
Ok, we are not coming to an agreement on the user interace of this module.

> 🔴 Your solution does not respect the ORM behaviour, otherwise we would not be having this discussion.

Well, I was meeting you half way. The logic can respect the ORM behaviour. You just have to activate it. For the ones who wants to use it.

>This is where my 'meet you halfway' solution comes in handy. The flag would automatically set the correct domain and hide the widget in the UI. Which tackles the 'technical' complexity of setting the void domain.

I did'nt catch that yet.  The runboat is unavailable to test / review this.

> I think current status is enough for validating the migration. Extra things can be performed later. Do you agree?

@pedrobaeza  I agree that we should get things moving again. 
So the migration flow of depending modules can continue. 
We can forward port the missing commits and improvements later.

Ideally, I would like to have functional tested this one on a runboat, before moving forward.

rlimaeco (approved):

pedrobaeza:
@bosd have you performed your last functional review?

bosd:
@pedrobaeza Not yet, Will have to do it locally, b/c no runboat.
Will do it today!

bosd:
I'm sorry to say that this does'nt do it for me. :scream:
While, I actually wish to move forward and have at least one version merged on V18. :+1: 
I really appreciate everyones effort to work things out with this module.
Let me clarify..

It looks like we are talking about different functionality. :pleading_face:
In this PR a new section has been added. The _Apply restricted product domain_. (Marked in purple in screenshot).
Which is actually a "don't care / neutral" for me and our clients.
Although I understand that this functionality has been added "to meet me half way".
Anyway, we have it.. So let's review..
image

My assumption, is (wrong :dizzy_face: ) that one can set the filter above like they used to and aditionally define the restricted domain.
In the case of the screenshot, an open filter matching all products.
Then create a _restricted product domain_ which contains also all products.
From the test in the screenshot. I'm expecting to first include all products. Then exclude all the products with the usage of the restricted domain. Resulting in an assortment of 0 products. 
Instead, in the screenshot it shows `274` products.

So please educate me, how this new _restricted product domain_ is supposed to work.

I badly want to approve this or any PR to get things moving. But now I'm actually more inclined to disapprove.
Because the functionality is not clear, to me and adding to the UI. (instead of simplifying) My preference is to rather have the version without this. But it's a **don't care**.. so will be neutral / non-blocking. 

## If someone understands how to use it and want to have it. Give an approving review :+1: and Let's move.. :trophy:


### Simple use case
Sorry, for repeating or sounding like broken record. But it's by seeing this implementation I'm getting the impression that my use-case / explanation was not clear enough.

Best to explain it with an image. The client wants to ignore the `SCARY` domain fields inside the blue marked area from screenshot below.
They just want to **simply** point and click the clients and products (from the pink area). Get the correct result of 2 products in the green area. (like in old versions of this module)
Adding a checkbox and what not to achieve this with the modified orm behaviour is ok.
IMO this al could very easily resolved with a simple checkbox and an `and` function.
Would even approve the complicated workaround by filling in a bogus domain evaluation in the database as proposed earlier. 
But please hide the technical hack, so nothing "Complicated" , "Technical" or scary is shown to the end user.
image


I hope this use case is clear now. 
Yes we can implement it at a later stage/different pr.

Again, I have good intentions, I will not be the blocking one this. But will sustain (for now) on approving new functionality I don't need or understand. (albeit specially added for me :heart: )

PieterPaulussen:
> ... Resulting in an assortment of 0 products. Instead, in the screenshot it shows `274` products.

Your case in invalid. The domain operator is wrong. What you would want to use is the `child_of` operator instead of the `in` operator. This results in the result set:
Screenshot 2025-03-06 at 16 10 43

The result of the domain in your example is void since there are no products which are directly part of the 'All' category.

PieterPaulussen:
By the way when done properly, the restricted product domain is applied as expected and results in the expect result set of products.
Screenshot 2025-03-06 at 16 23 41

In case you did not know default operator of the search view domain on `product.product` and `product.template` is also `child_of`, not `in`.
Screenshot 2025-03-06 at 16 28 02

bosd:
Thanks, That's clear now. With this addition we can even cover more advanced use cases. :+1:

PieterPaulussen:
@rlimaeco Thanks for the additional changes to my branch. Merged to get the tests green!

bosd (approved):
:trophy: :boom: :checkered_flag:

PieterPaulussen:
> Thanks, That's clear now. With this addition we can even cover more advanced use cases. 👍

Great news! Thanks for the reviews.

pedrobaeza:
Bravo for the collaboration and understanding.

/ocabot merge nobump

PR #1905, opened 2025-02-28
[16.0][IMP] product_supplierinfo_import_by_barcode: To product_supplierinfo_import

juancarlosonate-tecnativa:
- [x] Add supplier in templates.
- [x] Allow matching at the template level by product.template char fields
- [x] Add the option to update only existing supplierinfos without creating new ones.  
- [x] Add option at the template level, which groups records in the view by updated and non-updated after the import process.

@Tecnativa TT55210
@pedrobaeza @chienandalu


juancarlosonate-tecnativa:
ping @pedrobaeza @chienandalu can you take a look? thx!

pedrobaeza:
You shouldn't rewrite the whole commit history for renaming it. That should be done when migrating to upper versions. Now, you just need to add the new feature on top in one commit, respecting the name AFAIK (not sure what you agreed with @chienandalu though, but for sure renaming this way not).

Please perform that change ASAP.

juancarlosonate-tecnativa:
> You shouldn't rewrite the whole commit history for renaming it. That should be done when migrating to upper versions. Now, you just need to add the new feature on top in one commit, respecting the name AFAIK (not sure what you agreed with @chienandalu though, but for sure renaming this way not).
> 
> Please perform that change ASAP.

Alright, i thought this was the way to do it, anyway, i will wait for @chienandalu comments before proceeding, thanks!

chienandalu:
> > You shouldn't rewrite the whole commit history for renaming it. That should be done when migrating to upper versions. Now, you just need to add the new feature on top in one commit, respecting the name AFAIK (not sure what you agreed with @chienandalu though, but for sure renaming this way not).
> > Please perform that change ASAP.
> 
> Alright, i thought this was the way to do it, anyway, i will wait for @chienandalu comments before proceeding, thanks!

I think we agreed to avoid renaming the module in this version and do so in the future migration as the less problematic way to proceed.

product_supplierinfo_import/models/product_supplierinfo_import_template.py
```diff
@@ -8,13 +8,30 @@ class ProductSupplierInfoImportTemplate(models.Model):
...
     header_offset = fields.Integer(required=True, default=0)
-    barcode_header_name = fields.Char(required=True)
+    search_header_name = fields.Char(required=True)
     template_line_ids = fields.One2many(
         comodel_name="product.supplierinfo.import.template.line",
         inverse_name="template_id",
         string="Headers mapping",
     )
+    only_update_existing = fields.Boolean(
+        string="Just Update Existing",
```
chienandalu: The field name generates a field label automatically, so you don't need to define it twice here. And as they're different in this case, they'd generate two different terms, which isn't the best.
```suggestion
```

product_supplierinfo_import/models/product_supplierinfo_import_template.py
```diff
@@ -8,13 +8,30 @@ class ProductSupplierInfoImportTemplate(models.Model):
...
+        help="If enabled, only existing supplierinfo records will be updated. "
+        "New ones will not be created.",
+        default=False,
+    )
+    show_unupdated_rates = fields.Boolean(
+        string="Unupdated Rates",
```
chienandalu: same as above
```suggestion
```

product_supplierinfo_import/models/product_supplierinfo_import_template.py
```diff
@@ -8,13 +8,30 @@ class ProductSupplierInfoImportTemplate(models.Model):
...
         inverse_name="template_id",
         string="Headers mapping",
     )
+    only_update_existing = fields.Boolean(
+        string="Just Update Existing",
+        help="If enabled, only existing supplierinfo records will be updated. "
+        "New ones will not be created.",
+        default=False,
+    )
+    show_unupdated_rates = fields.Boolean(
```
chienandalu: *unupdated* sounds a little bit weird
```suggestion
    show_not_updated_rates = fields.Boolean(

File: product_supplierinfo_import/models/product_supplierinfo_import_template.py [hunk 4]
```diff
@@ -8,13 +8,30 @@ class ProductSupplierInfoImportTemplate(models.Model):
...
+        string="Unupdated Rates",
+        help="If checked, at the end of the import,"
+        " it will display the supplier's rate records that were not updated.",
+    )
+    fields_product_template = fields.Many2one(
```
chienandalu:
```suggestion
    import_criteria_field_id = fields.Many2one(
```

product_supplierinfo_import/models/product_supplierinfo_import_template.py
```diff
@@ -8,13 +8,30 @@ class ProductSupplierInfoImportTemplate(models.Model):
     _description = "Templates to be detected when importing vendor pricelist"
 
     name = fields.Char(required=True)
+    supplier_id = fields.Many2one(comodel_name="res.partner")
     header_offset = fields.Integer(required=True, default=0)
-    barcode_header_name = fields.Char(required=True)
+    search_header_name = fields.Char(required=True)
```
chienandalu: This will requiere a migration script

product_supplierinfo_import/wizards/product_supplierinfo_import.py
```diff
@@ -16,17 +17,42 @@ class ProductSupplierInfoImport(models.TransientModel):
     _name = "product.supplierinfo.import"
     _description = "Import supplier info records"
 
-    supplier_id = fields.Many2one(comodel_name="res.partner", required=True)
+    supplier_id = fields.Many2one(
+        "res.partner",
```
chienandalu: This is a formal preference, but we prefer to be explicit for better readeability
```suggestion
        comodel_name="res.partner",

File: product_supplierinfo_import/wizards/product_supplierinfo_import.py [hunk 7]
```diff
@@ -16,17 +17,42 @@ class ProductSupplierInfoImport(models.TransientModel):
...
+        string="Supplier",
+        related="template_id.supplier_id",
```
chienandalu: This can't be related, as it will change the previous logic. Better use a compute storable field that depends on the template

product_supplierinfo_import/wizards/product_supplierinfo_import.py
```diff
@@ -16,17 +17,42 @@ class ProductSupplierInfoImport(models.TransientModel):
...
-        help="If a product isn't found by its barcode, it will be created with the "
+        help="If a product isn't found by its Search Field, it will be created with the "
         "provided data",
         default=True,
     )
     supplierinfo_file = fields.Binary(required=True)
     supplierinfo_filename = fields.Char()
-    template_id = fields.Many2one(comodel_name="product.supplierinfo.import.template")
+    template_id = fields.Many2one(
+        "product.supplierinfo.import.template",
```
chienandalu:
```suggestion
        comodel_name="product.supplierinfo.import.template",

File: product_supplierinfo_import/wizards/product_supplierinfo_import.py [hunk 9]
```diff
@@ -91,60 +123,35 @@ def _prepare_supplierinfo_values(self, row_data):
     def _update_create_supplierinfo_data(self, parsed_data):
         """Create or import vendor list for the parsed data"""
         supplier_infos = self.env["product.supplierinfo"]
+        all_supplier_infos = self.env["product.supplierinfo"].search(
+            [("partner_id", "=", self.supplier_id.id)]
+        )
+        search_field = self.template_id.fields_product_template.name
+        if not search_field:
+            raise UserError(_("No search field has been selected in the template."))
```
chienandalu: This isn't needed as that field should be required in the template
```suggestion
```

product_supplierinfo_import/wizards/product_supplierinfo_import.py
```diff
@@ -91,60 +123,35 @@ def _prepare_supplierinfo_values(self, row_data):
...
         for row_data in parsed_data:
-            # Repeating headers...
```
chienandalu: Why to delete this comment?

product_supplierinfo_import/wizards/product_supplierinfo_import.py
```diff
@@ -155,39 +162,113 @@ def _update_create_supplierinfo_data(self, parsed_data):
...
+            product = self.env["product.template"].create(product_data)
+            product.message_post(
+                body=_("Created from supplier price list import"),
+                type="note",
+            )
+            return product
+        except ValidationError:
+            _logger.warning(
+                f"The product {vendor_product_name}"
+                " could not be created due to a barcode collision."
```
chienandalu: Now it could be a barcode collision or something else as the field criteria is flexible

product_supplierinfo_import/wizards/product_supplierinfo_import.py
```diff
@@ -155,39 +162,113 @@ def _update_create_supplierinfo_data(self, parsed_data):
...
+        product_data = {
+            "created_from_supplierinfo_import": True,
+            "name": vendor_product_name,
+            "purchase_ok": False,
+            "sale_ok": False,
+            **(
+                {search_field: search_value}
+                if search_field in self.env["product.template"]._fields
+                else {}
+            ),
```
chienandalu: Why is this check needed? The field is selected in the template config, so this should fail

product_supplierinfo_import/wizards/product_supplierinfo_import.py
```diff
@@ -155,39 +162,113 @@ def _update_create_supplierinfo_data(self, parsed_data):
...
+    def _update_existing_supplierinfo(self, supplierinfo, row_data):
+        """Updates an existing supplier price list record."""
+        try:
+            supplierinfo.write(self._prepare_supplierinfo_values(row_data))
+            supplierinfo.import_status = "new_or_updated"
+            return supplierinfo
+        except Exception:
+            raise UserError(
+                _("The row format seems to be incorrect: %(row)s", row=str(row_data))
+            ) from Exception
```
chienandalu: This shouldn't fail either...

product_supplierinfo_import/models/product_supplier_info.py
```diff
@@ -0,0 +1,11 @@
+from odoo import fields, models
+
+
+class ProductSupplierInfo(models.Model):
+    _inherit = "product.supplierinfo"
+
+    import_status = fields.Selection(
```
chienandalu: This could be just a boolean

product_supplierinfo_import/wizards/product_supplierinfo_import.py
```diff
@@ -155,39 +162,113 @@ def _update_create_supplierinfo_data(self, parsed_data):
...
+                    else {}
+                ),
+            }
+        )
+        try:
+            product_supplierinfo.write(self._prepare_supplierinfo_values(row_data))
+        except Exception:
+            raise UserError(
+                _("The row format seems to be incorrect: %(row)s", row=str(row_data))
+            ) from Exception
```
chienandalu: This shouldn't fail either

product_supplierinfo_import/tests/test_product_supplierinfo_import.py
```diff
@@ -213,14 +230,13 @@ def test_simple_import(self):
...
 
-    @freeze_time("2024-07-01")
+    @freeze_time("2025-03-01")
     def test_complex_file_import(self):
         """This file has some traps, like header offsets, category rows, repeating
         headers..."""
         # 1. Import the test case file
         self._import_supplierinfo_file(
             "test_supplier_drinks_complex_sheet.xlsx",
-            self.supplier_2,
```
chienandalu: Overall, you shouldn't change the logic of existing tests, as they should be working along with the new features. To test the new stuff, add new test cases

product_supplierinfo_import/readme/DESCRIPTION.md
```diff
@@ -1,2 +1 @@
-This module allows to import vendor price lists from excel files matching the products
-from their barcode.
+This module allows to import vendor price lists from excel.
```
chienandalu: Don't drop the whole matching thing. Just change to state that the matching can be done by any template char field

product_supplierinfo_import/readme/CONFIGURE.md
```diff
@@ -1,12 +1,18 @@
...
+4.  **Assign a default supplier.** (Set a default supplier that will be associated with this template)
+5.  **In the Header Offset, determine the line where the headers are located.** (Specify the line number where the column headers are found in your sheet file)
+6.  **In the Search Header Name, set the name of the column you want to filter by.** (Indicate the column name used for filtering the data)
+7.  **In the Search Field, set the data you want to filter by.** (Specify the field used for filtering)
+8.  Set a headers mapping with the rest header names and which fields we should update. (Map the remaining header names to the corresponding fields that should be updated in the system)
+9. A header mapping could have no related field, but is important to have it if we want to detect the sheet data when we are importing.
+10. The order isn't important. (The order of the header mappings does not affect the import process)
+11. **If 'Only Update Existing' is checked, only existing pricelists will be updated; no new ones will be created.**
+12. **If 'Show Unupdated Rates' is enabled, the unupdated rates will be shown grouped at the end of the import.**
```
chienandalu: Why setting in bold the whole sentences?

product_supplierinfo_import/models/product_supplierinfo_import_template.py
```diff
@@ -8,13 +8,30 @@ class ProductSupplierInfoImportTemplate(models.Model):
...
+    )
+    show_unupdated_rates = fields.Boolean(
+        string="Unupdated Rates",
+        help="If checked, at the end of the import,"
+        " it will display the supplier's rate records that were not updated.",
+    )
+    fields_product_template = fields.Many2one(
+        comodel_name="ir.model.fields",
+        domain=[("model", "=", "product.template"), ("ttype", "=", "char")],
+        string="Search Field",
```
chienandalu:
```suggestion
        help="The field that will be used to match the products from the excel file with the ones in the Odoo database",

File: product_supplierinfo_import/models/product_supplierinfo_import_template.py [hunk 20]
```diff
@@ -8,13 +8,30 @@ class ProductSupplierInfoImportTemplate(models.Model):
...
+        default=False,
+    )
+    show_unupdated_rates = fields.Boolean(
+        string="Unupdated Rates",
+        help="If checked, at the end of the import,"
+        " it will display the supplier's rate records that were not updated.",
+    )
+    fields_product_template = fields.Many2one(
+        comodel_name="ir.model.fields",
+        domain=[("model", "=", "product.template"), ("ttype", "=", "char")],
```
chienandalu: So transtion is easier:
```suggestion
        domain=[("model", "=", "product.template"), ("ttype", "=", "char")],
        default=lambda self: self.env.ref("product.field_product_template__barcode"),

**Review by chienandalu — CHANGES_REQUESTED** [review]
Overall look in the good path :D

Some changes needed though:

File: product_supplierinfo_import/wizards/product_supplierinfo_import.py (hunk 12 above)
Comment (juancarlosonate-tecnativa) [reply product_supplierinfo_import/wizards/product_supplierinfo_import.py]: Mistake, forgot to remove that when i was doing the first version

File: product_supplierinfo_import/models/product_supplier_info.py (hunk 14 above)
Comment (juancarlosonate-tecnativa) [reply product_supplierinfo_import/models/product_supplier_info.py]: was just an aesthetic thing for the group by but okay

Comment (chienandalu) [reply product_supplierinfo_import/models/product_supplier_info.py]: Mmm. You're right. Keep it then

File: product_supplierinfo_import/models/product_supplierinfo_import_template.py (hunk 19 above)
Comment (juancarlosonate-tecnativa) [reply product_supplierinfo_import/models/product_supplierinfo_import_template.py]: If i remove this string, then "fields_product_template" will appear, maybe it wont be understood correctly. Should rename the variable or add string in the view field?

**Comment by juancarlosonate-tecnativa** [comment]
ping @chienandalu

File: product_supplierinfo_import_by_barcode/__manifest__.py [hunk 21]
```diff
@@ -1,8 +1,8 @@
 # Copyright 2024 Tecnativa - David Vidal
 # License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
 {
-    "name": "Import supplier pricelists by barcode",
-    "summary": "Import supplier pricelists by barcode",
+    "name": "Import supplier pricelists",
+    "summary": "Import supplier pricelists",
     "version": "16.0.1.0.0",
```
chienandalu: Bump the version manually to ensure the migration script. Agree @pedrobaeza ?

product_supplierinfo_import_by_barcode/README.rst
```diff
@@ -39,20 +39,38 @@ matching the products from their barcode.
...
 
-1. Go to *Purchase > Configuration > Vendor pricelist import templates*.
-2. Edit or create a new one.
-3. Set the name of the header to match by barcode.
-4. Set a headers mapping with the rest header names and which fields we
-   should update.
-5. A header mapping could have no related field, but is important to
-   have it if we want to detect the sheet data when we are importing.
-6. The order isn't important.
+1.  Go to *Purchase > Configuration > Vendor pricelist import
```
chienandalu, line 48: There's this extra space that is making the diff larger than needed. Why? Something about your editor?
```suggestion
1. Go to *Purchase > Configuration > Vendor pricelist import
```

product_supplierinfo_import/models/product_supplierinfo_import_template.py
chienandalu: Yes, that's why I suggested rename it to `import_criteria_field_id` which is more descriptive

product_supplierinfo_import_by_barcode/tests/test_product_supplierinfo_import_by_barcode.py
```diff
@@ -11,15 +11,28 @@
...
+            {
+                "name": "Olé Drinks S.A.",
+                "supplier_rank": 1,
+            }
+        )
         cls.import_template_1 = cls.env["product.supplierinfo.import.template"].create(
             {
                 "name": "Spanish Drinks S.A.",
-                "barcode_header_name": "bar code",
+                "supplier_id": cls.supplier_1.id,
```
chienandalu: The existing tests should work without setting this field. It should be optional for not breaking the existing behavior.

You can force the supplier in the specific test case and check if the supplier is correctly detected

product_supplierinfo_import_by_barcode/tests/test_product_supplierinfo_import_by_barcode.py
```diff
@@ -265,3 +262,68 @@ def test_complex_file_import(self):
...
+                    "price": 13.04,
+                    "date_start": fields.Date.from_string("2026-03-01"),
+                    "date_end": False,
+                },
+            }
+        )
+        self.assertEqual(len(self.prod_horchata.seller_ids), 1)
+
+
+class TestProductSupplierinfoImportByProductCode(TestProductSupplierinfoImportCommon):
```
chienandalu: why creating a new class? the common one was created to reuse the declared data in an extra module. You can simply use the test class. Otherwise, you're redeclaring everything again...

product_supplierinfo_import_by_barcode/wizards/product_supplierinfo_import.py
```diff
@@ -44,7 +75,13 @@ def _detect_template(self, sheet):
                 self.template_id = template
                 return header_values
         if not self.template_id:
-            raise UserError(_("We couldn't find a suitable template for your file"))
+            raise UserError(
+                _(
+                    "We couldn't find a suitable template for your file. "
+                    "Number of columns found in header: %d"
+                )
+                % len(header_values)
```
chienandalu: Is this info relevant to the user? :thinking:

product_supplierinfo_import_by_barcode/readme/CONFIGURE.md
```diff
@@ -1,12 +1,18 @@
-Vendor pricelists sheet files could come in a variety of column formats. We can configure
-templates to match them to know what data we need to store in our pricelist.
+Vendor Pricelist Import Templates
+
```
chienandalu:
```suggestion
```

chienandalu (changes requested):

product_supplierinfo_import_by_barcode/__manifest__.py
pedrobaeza: Yes

product_supplierinfo_import_by_barcode/wizards/product_supplierinfo_import.py
juancarlosonate-tecnativa: I added it because, in some cases, there might be hidden columns in the files, which could lead to configuration errors, originally used it as a warning for myself, but i left it because found it useful. If you want, i can remove it.

product_supplierinfo_import_by_barcode/tests/test_product_supplierinfo_import_by_barcode.py
juancarlosonate-tecnativa: check again please

product_supplierinfo_import_by_barcode/wizards/product_supplierinfo_import.py
chienandalu: In that case, I think it would be more complete to show the column list so we user can compare to the supossed template. So the message would be something like:

No matching template for these header columns: <columns list> Total header columns: <columns length>

product_supplierinfo_import_by_barcode/readme/DESCRIPTION.md
```diff
@@ -1,2 +1 @@
-This module allows to import vendor price lists from excel files matching the products
-from their barcode.
+This module allows importing vendor price lists from Excel files, matching products using any product template char field.
```
chienandalu, line 1: Add a roadmap section with a todo to rename the module

chienandalu (commented):
I'm testing in runboat with the complex spreadsheet. It is detected buy when imported, the following error ocurrs:

  File "/mnt/data/odoo-addons-dir/product_supplierinfo_import_by_barcode/wizards/product_supplierinfo_import.py", line 247, in action_import_file
    ) = self._update_create_supplierinfo_data(parsed_data)
  File "/mnt/data/odoo-addons-dir/product_supplierinfo_import_by_barcode/wizards/product_supplierinfo_import.py", line 175, in _update_create_supplierinfo_data
    supplier_infos += self._create_new_supplierinfo(
  File "/mnt/data/odoo-addons-dir/product_supplierinfo_import_by_barcode/wizards/product_supplierinfo_import.py", line 236, in _create_new_supplierinfo
    product_supplierinfo.write(self._prepare_supplierinfo_values(row_data))
  File "/mnt/data/odoo-addons-dir/product_supplierinfo_import_by_barcode/wizards/product_supplierinfo_import.py", line 121, in _prepare_supplierinfo_values
    {
  File "/mnt/data/odoo-addons-dir/product_supplierinfo_import_by_barcode/wizards/product_supplierinfo_import.py", line 122, in <dictcomp>
    tl.field_id.name: row_data[tl.header_name]
KeyError: 'Final\n'

juancarlosonate-tecnativa:
@chienandalu it seems to be due to these blank lines 
image

chienandalu:
> @chienandalu it seems to be due to these blank lines

Yes, it's deliberated. As when the user copy/pastes from the excel headers the could be pasting those line feeds as well it it was like that in the original.

That case was working before, if a don't remember wrong (and indeed the template is detected), do you know why it isn't now?

juancarlosonate-tecnativa:
> > @chienandalu it seems to be due to these blank lines
> 
> Yes, it's deliberated. As when the user copy/pastes from the excel headers the could be pasting those line feeds as well it it was like that in the original.
> 
> That case was working before, if a don't remember wrong (and indeed the template is detected), do you know why it isn't now?

It seems that one of the try blocks we removed was holding this error in version 16, but i'm investigating further. In the end, values.update is not being parsed in any way. The _parse_header function, both now and before, was only used in the template detector. I'm looking for a solution. 
image

product_supplierinfo_import_by_barcode/wizards/product_supplierinfo_import.py
```diff
@@ -155,39 +169,98 @@ def _update_create_supplierinfo_data(self, parsed_data):
...
+        return supplier_infos, unchanged_supplier_infos
+
+    def _create_new_product(self, search_value, row_data, search_field):
+        """Creates a new product if allowed."""
+        vendor_product_name_header = self.template_id.template_line_ids.filtered(
+            lambda x: x.field_id.name == "product_name"
+        ).header_name
+        vendor_product_name = row_data.get(
+            vendor_product_name_header,
+            _("%(code)s ((product imported))", code=search_value),
```
chienandalu:
```suggestion
            _("%(code)s (product imported)", code=search_value),
```

chienandalu (commented):
Let's try to avoid importing this rows. Add a method that checks if the columns corresponding to importing fields have any content and ignore the ones that doesn't.

image

product_supplierinfo_import_by_barcode/wizards/product_supplierinfo_import.py
juancarlosonate-tecnativa: take a look at this and let me know what you think, don't like it much, but i think it works well

product_supplierinfo_import_by_barcode/wizards/product_supplierinfo_import.py
```diff
@@ -81,70 +119,51 @@ def _prepare_supplierinfo_values(self, row_data):
...
                 if tl.field_id
             }
         )
-        return values
+        cleaned_values = {
+            key: value for key, value in values.items() if value not in (None, "")
+        }
+        if set(cleaned_values.keys()) <= {"delay", "date_start"}:
+            return None
+        return cleaned_values
```
chienandalu: A simpler implementation
```python
    def _prepare_supplierinfo_values(self, row_data):
        """Overridable hook method so we can inject general wizard values"""
        values = {
            tl.field_id.name: row_data.get(
                self._parse_header([tl.header_name])[0], ""
            )
            for tl in self.template_id.template_line_ids
            if tl.field_id
        }
        # Avoid rows where there's no information given
        all_values_with_content = all([x not in {None, ""} for x in values.values()])
        if not all_values_with_content:
            return
        values.update({
            "delay": self.delay,
            "date_start": self.date_start,
        })
        return values

File: product_supplierinfo_import_by_barcode/wizards/product_supplierinfo_import.py [hunk 30]
```diff
@@ -155,39 +174,103 @@ def _update_create_supplierinfo_data(self, parsed_data):
...
-                    _("The row format seems to be incorrect %(row)s", row=str(row_data))
-                ) from Exception
-        return supplier_infos
+            return product
+        except ValidationError:
+            _logger.warning(
+                f"The product {vendor_product_name}"
+                f" could not be created due to a {search_field} collision."
+            )
+            return None
```
chienandalu: simply `return` is the same as `return None`. Anyway, prefer to return always the same datatype. For example, `return self.env["product.product"]`. It makes it potentially easier to deal with it the method later in different scenarios.

product_supplierinfo_import_by_barcode/wizards/product_supplierinfo_import.py
```diff
@@ -155,39 +174,103 @@ def _update_create_supplierinfo_data(self, parsed_data):
...
+        supplierinfo.import_status = "new_or_updated"
+        return supplierinfo
+
+    def _create_new_supplierinfo(
+        self, previous_supplierinfo, product, row_data, search_field, search_value
+    ):
+        """Creates a new supplier price list record, updating the previous one if needed."""
+        values = self._prepare_supplierinfo_values(row_data)
+        if not values:
+            return None
```
chienandalu: Same here `return self.env["product.supplierinfo"]`

product_supplierinfo_import_by_barcode/wizards/product_supplierinfo_import.py
juancarlosonate-tecnativa: much better, thanks

juancarlosonate-tecnativa:
It wasn't possible to select a sheet, so i added the index to allow selection. Additionally, some price lists have empty columns. To address this without overcomplicating the logic, we can simply add a space in a new field, which will allow detection but won't store any data. What do you think?

product_supplierinfo_import_by_barcode/models/product_supplierinfo_import_template.py
```diff
@@ -8,13 +8,31 @@ class ProductSupplierInfoImportTemplate(models.Model):
     _description = "Templates to be detected when importing vendor pricelist"
 
     name = fields.Char(required=True)
+    supplier_id = fields.Many2one(comodel_name="res.partner")
     header_offset = fields.Integer(required=True, default=0)
-    barcode_header_name = fields.Char(required=True)
+    sheet_index = fields.Integer(required=True, default=0)
```
chienandalu: UX suggestion: for a regular user isn't very intuitive that the first sheet is the number 0. 
```suggestion
    sheet_number = fields.Integer(required=True, default=1, help="The number of the sheet page in the spreadsheet where the pricelist are")
```

product_supplierinfo_import_by_barcode/wizards/product_supplierinfo_import_views.xml
```diff
@@ -8,7 +8,7 @@
                 <field name="supplierinfo_filename" invisible="1" />
                 <field name="template_id" invisible="1" />
                 <group>
-                    <field name="supplier_id" />
+                    <field name="supplier_id" required="1" />
```
pedrobaeza, line 11: Why changing this to required? The auto-detection pattern for some cases should be still applied AFAIK.

juancarlosonate-tecnativa: mm, i think i don't quite understand you, but if the template doesn't have a supplier, shouldn't it be required when importing?

pedrobaeza: Yeah, nevermind. My fault..

pedrobaeza (approved):
/ocabot merge major

PR #1908, opened 2025-03-06
[18.0] migration leftover product_assortment

bosd:
Leftover migration from #1904


bosd:
@carlosdauden Can you take a look here?

product_assortment/models/ir_filters.py
```diff
@@ -90,10 +88,10 @@ def get_partner_domain_fields(self):
     @api.depends("partner_ids", "partner_domain")
     def _compute_all_partner_ids(self):
         """Summarize selected partners and partners from partner domain field"""
-        for ir_filter in self:
+        for ir_filter in self.sudo():
```
bizzappdev: Not sure whether we should use compte_sudo or add self.sudo?
IMO compute_sudo is better

product_assortment/models/ir_rule.py
```diff
@@ -0,0 +1,36 @@
...
+        "xml" not in config["dev_mode"],
+        tools.ormcache(
+            "self.env.uid",
+            "self.env.su",
+            "model_name",
+            "mode",
+            "tuple(self._compute_domain_context_values())",
+        ),
+    )
+    def _compute_domain(self, model_name, mode="read"):
```
bizzappdev: IMO, the proper place will be the search method of ir.filters

product_assortment/models/ir_rule.py
```diff
@@ -0,0 +1,36 @@
...
+        """Inject extra domain for restricting filter (Assortments) when the user
+        has not the group 'Product Assortment Manager'.
+        """
+        res = super()._compute_domain(model_name, mode=mode)
+        user = self.env.user
+        if model_name == "ir.filters" and not self.env.su:
```
bizzappdev:
```suggestion
        if model_name == "ir.filters" and not self.env.is_superuser():
```

bizzappdev (changes requested):

product_assortment/models/ir_filters.py
bosd: "I’ve updated the field to use compute_sudo=True instead of adding .sudo() inside the loop. Since we are performing a search() on res.partner based on a dynamic domain, compute_sudo ensures the computation is consistent and avoids potential AccessErrors for non-admin users, while also being more performant for the Odoo cache."

product_assortment/models/ir_rule.py
bosd: Good catch! refactored it. moved this logic to the _search method of ir.filters instead. This keeps the restriction scoped to the model itself and avoids the overhead of the global rule engine

bosd:
please keep open.

PR #1910, opened 2025-03-10
[16.0][IMP] product_attribute_model link attribute value relation enhancement

ivs-cetmix:
This pr is re-opened because https://github.com/OCA/product-attribute/pull/1566 was closed, although it had "ready to merge" labels/

Before
-------

The field linked_record_ref of type Reference is used to store the information about which record is linked to a particular attribute value. However, Reference fields are stored in db as text fields which renders them impractical for usage in queries.

For example if user wants to get all attribute values that belong to a recordset the following steps should be done:

Get all attributes that belong to the record model using:
attributes = self.env["product.attribute"].search([("linked_model_id", "=", model_id)])

Use mapped together with filtered to get corresponding attribute values:
matching_attribute_values = attributes.mapped("value_ids").filtered(lambda v: v.linked_record_ref == record)

This will generate a lot of queries and may affect the performance badly.

Now
----

The linked_record_ref field has been enhanced by becoming a computed field, dynamically created based on the model and res_id fields, similar to the mail.message model. This change allows for faster search queries with just one line of code:

["product.attribute.value"].search([("model", "=", self._name), ("res_id", "in", self.ids)])

This adjustment significantly reduces the complexity and improves search speed.


Aldeigja (approved):
Functional LGTM

ivs-cetmix:
Hey @OCA/product-maintainers would appreciate someone having a look at this! The original module was merged long ago, so people are using it already. However this patch that significantly improved performance is stuck for more than half a year for some strange reason 🫤

PR #1912, opened 2025-03-10
[18.0][IMP] product_customerinfo: Allow searching product templates by customer info

carlos-lopez-tecnativa:
Now, in the sale order, the product template is displayed by default. This update adds the ability to search for product templates using customer info.

Reference: https://github.com/odoo/odoo/pull/155449

TT54233
@Tecnativa @pedrobaeza @victoralmau could you please review this.

### DIFF
--- a/product_customerinfo/models/product_template.py
+++ b/product_customerinfo/models/product_template.py
@@ -3,7 +3,9 @@
 # Copyright 2015 Tecnativa
 # Copyright 2019 Tecnativa - Pedro M. Baeza
 # License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl.html).
-from odoo import fields, models
+from odoo import api, fields, models
+from odoo.osv import expression
+from odoo.tools.misc import unique
 
 
 class ProductTemplate(models.Model):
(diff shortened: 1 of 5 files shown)


product_customerinfo/models/product_template.py
```diff
@@ -20,3 +22,89 @@ class ProductTemplate(models.Model):
...
+                        )
+                    )
+                # => Feature drop here,
+                # one record can only have one display_name now,
+                # instead separate with `,`
+                product_template.display_name = ", ".join(unique(temp))
+
+    @api.model
+    def name_search(self, name, args=None, operator="ilike", limit=100):
+        res = super().name_search(name, args=args, operator=operator, limit=limit)
```
pedrobaeza, line 90: I think you should control the extra search by a context key activated in the sales order, and give priority to the result of product.customerinfo records. How is this handled in purchase by core?

carlos-lopez-tecnativa: In [purchase.order](https://github.com/odoo/odoo/blob/18edce4d859935bd1425144e4c835acabc5f68f4/addons/purchase/views/purchase_views.xml#L246), the `partner_id` is passed by context, similar to how it is done in [sale.order](https://github.com/odoo/odoo/blob/18edce4d859935bd1425144e4c835acabc5f68f4/addons/sale/views/sale_order_views.xml#L531-L551).

In this module, when searching for `product.product`, the `name_search` does not expect any specific context(just `partner_id`). For `product.template`, it follows the same logic as `product.product`. What exactly needs to be changed? Please provide more details.

`product.product`
https://github.com/OCA/product-attribute/blob/0f082a39b5ea459851acb370f67c05333f557479/product_customerinfo/models/product_product.py#L20-L41

`product.template`
https://github.com/Tecnativa/product-attribute/blob/0b2668128d84949bff76d553e0f4d93fb7e8dfb4/product_customerinfo/models/product_template.py#L95-L96

pedrobaeza: OK, I see the context key. It seems enough.

pedrobaeza (commented):
Can part of this code reused in product.product for not duplicating it?

carlos-lopez-tecnativa:
> Can part of this code reused in product.product for not duplicating it?

Note that both `name_search` implementations are different in terms of the fields used: `product.product` relies on `product_tmpl_id.customer_ids`, while `product.template` uses `customer_ids`. I don’t believe it is necessary to create a new `AbstractModel` here. Please let me know if this solution is sufficient or if the abstract model is needed

pedrobaeza (approved):

product_customerinfo/models/product_template.py
```diff
@@ -20,3 +22,86 @@ class ProductTemplate(models.Model):
...
+        # check access and use superuser
+        self.check_access("read")
+
+        product_template_ids = self.sudo().ids
+        supplier_info_by_template = {}
+        if partner_ids:
+            # prefetch the fields used by the `display_name`
+            domain = [
+                ("product_tmpl_id", "in", product_template_ids),
+                ("product_id", "=", False),
```
victoralmau, line 53: I think this should not be product_id=false or if you define a specific record for a variant it does not work correctly, right?

victoralmau (changes requested):

product_customerinfo/models/product_template.py
carlos-lopez-tecnativa: When creating a record from `product.product` or `product.template`, the `product_id` field is always empty by default. It is only set when the user explicitly selects a variant in this field. In such cases, the customer info should be linked specifically to that variant. Therefore, from my perspective, it is correct that this record is not considered here.

victoralmau (approved):

PR #1919, opened 2025-03-12
[18.0][MIG] product_optional_product_quantity: Migration to 18.0 

BhaveshHeliconia:


BhaveshHeliconia:
Hey @OCA/product-maintainers would be great if someone could have a look at this module.

product_optional_product_quantity/models/product_optional_line.py
```diff
@@ -0,0 +1,62 @@
...
+    product_tmpl_id = fields.Many2one(
+        comodel_name="product.template",
+        domain="[('id', '!=', optional_product_tmpl_id)]",
+        string="Related Product",
+        required=True,
+    )
+    company_id = fields.Many2one(related="product_tmpl_id.company_id")
+    optional_product_tmpl_id = fields.Many2one(
+        comodel_name="product.template",
+        domain=[
```
tonygalmiche: Found a bug in product_optional_line.py: the `domain` on `optional_product_tmpl_id`
is defined as a Python list literal instead of a string:

    domain=[
        ("company_id", "in", [company_id, False]),
        ("id", "!=", product_tmpl_id),
    ],

Since it's not a string, `company_id` and `product_tmpl_id` are resolved at class
definition time as the Field objects themselves, not as dynamic per-record values.
This causes a crash when opening the "Product" field:

    psycopg2.errors.InvalidTextRepresentation: invalid input syntax for type integer:
    "product.optional.line.company_id"

Should be:
    domain="[('company_id', 'in', [company_id, False]), ('id', '!=', product_tmpl_id)]"

tonygalmiche (changes requested):

tonygalmiche:
Note: none of the existing unit tests exercise this domain, since they create
records directly via ORM (bypassing client-side domain evaluation). That's
why CI is green despite this being broken in the UI.

product_optional_product_quantity/models/product_optional_line.py
BhaveshHeliconia: Thanks @tonygalmiche! Good catch. I've updated.

BhaveshHeliconia:
> Note: none of the existing unit tests exercise this domain, since they create records directly via ORM (bypassing client-side domain evaluation). That's why CI is green despite this being broken in the UI.

Ah right, that makes sense! Added a test via Form() to ensure the UI domain evaluation is covered going forward.

tonygalmiche (approved):
Tested on a local Odoo 18.0 instance: optional products and quantity field work correctly as documented (base module, usage flow left to future dependent modules). Works as expected.

Thanks @BhaveshHeliconia for the quick fix!

PR #1926, opened 2025-03-12
[18.0][MIG] product_get_price_helper: migrate to v18

simahawk:
:warning: ATM any module depending on product discount machinery has an hidden dependency on `sale` due to this bug https://github.com/odoo/odoo/issues/202035.

In the migration I had to mock the `_show_discount` method to make it work.
In the context of this module we don't care that much because we only want to ensure a proper behavior based on the value of that flag.


simahawk:
@qgroulard @acsonefho ping :)

product_get_price_helper/demo/pricelist.xml
```diff
@@ -0,0 +1,29 @@
...
+    <!--
+        FORCE ONLY ONE ITEM ON THE PRICE LIST
+        When a price list is created, odoo assign a default price list item/
+        To be sure that only our new item is assigned to the price list
+        we reassign the item_ids....
+    -->
+    <record id="pricelist_1" model="product.pricelist">
+        <field
+            name="item_ids"
+            eval="[(6, 0, [ref('product_get_price_helper.item_1')])]"
```
acsonefho: Please can you use the new syntax with `Command.xxx()`

acsonefho (approved):
Really nice work :muscle: 
Only a minor improvement (but not blocking)

product_get_price_helper/demo/pricelist.xml
simahawk: done

acsonefho (approved):

qgroulard (approved):
LGTM (code review), thanks !

PR #1931, opened 2025-03-17
[16.0][FIX]product_supplierinfo_for_customer: misc changes for search count.

BhaveshHeliconia:


product_supplierinfo_for_customer/models/product_supplierinfo.py
```diff
@@ -8,18 +8,20 @@ class ProductSupplierInfo(models.Model):
 
     @api.model
     def search(self, args, offset=0, limit=None, order=None, count=False):
-        res = super().search(args, offset=offset, limit=limit, order=order, count=count)
+        res = super().search(args, offset=offset, limit=limit, order=order, count=False)
```
rousseldenis: IMHO, this is not correct as you change the passed parameter.

product_supplierinfo_for_customer/models/product_supplierinfo.py
```diff
@@ -8,18 +8,20 @@ class ProductSupplierInfo(models.Model):
...
         if (
             self.env.context.get("customerinfo")
             and self._name == "product.supplierinfo"
         ):
             limit2 = limit - len(res) if limit else limit
             res2 = self.env["product.customerinfo"].search(
-                args, offset=offset, limit=limit2, order=order, count=count
+                args, offset=offset, limit=limit2, order=order, count=False
```
rousseldenis: But `count` argument is `False` by default...

BhaveshHeliconia:
@rousseldenis, The search method is called from the search_count method with the count=True argument in Odoo 16 core (https://github.com/odoo/odoo/blob/16.0/odoo/models.py#L1511). In this case, the result of the super() call will be the number of records (an integer). However, in the code, the len() function is being used on the result of the super() call (https://github.com/OCA/product-attribute/pull/1931/files#diff-6fc0b1283cb7afcdf7bf53b3e8e5e8a343f43ff4a84d4c65f68e81e756f67933L16), which will attempt to call len() on an integer object, ultimately causing an error.

BhaveshHeliconia:
Hey @OCA/product-maintainers  would be great if someone could have a look at this changes.

product_supplierinfo_for_customer/models/product_supplierinfo.py
```diff
@@ -8,18 +8,20 @@ class ProductSupplierInfo(models.Model):
 
     @api.model
     def search(self, args, offset=0, limit=None, order=None, count=False):
-        res = super().search(args, offset=offset, limit=limit, order=order, count=count)
+        res = super().search(args, offset=offset, limit=limit, order=order, count=False)
         if (
             self.env.context.get("customerinfo")
             and self._name == "product.supplierinfo"
```
baguenth, line 14: can we not just add a condition here like 

`and not count`

to bypass the customer info logic, if count is true?

product_supplierinfo_for_customer/models/product_supplierinfo.py
```diff
@@ -8,18 +8,20 @@ class ProductSupplierInfo(models.Model):
...
         ):
             limit2 = limit - len(res) if limit else limit
             res2 = self.env["product.customerinfo"].search(
-                args, offset=offset, limit=limit2, order=order, count=count
+                args, offset=offset, limit=limit2, order=order, count=False
             )
             res2 = res2.read(list(self.env["product.supplierinfo"]._fields.keys()))
```
baguenth: If also in case of count = true we need to consider the customer info, can we not do something like:
```python
if count:
    res += res2
else:
    for result in res2:
        res += self.env["product.supplierinfo"].new(result)
```

baguenth (changes requested):

baguenth (approved):
lgtm

BhaveshHeliconia:
ping @rousseldenis

BhaveshHeliconia:
Hello @OCA/product-maintainers, it would be great if someone could have a look at these changes.

sebinthomasst (approved):

BhaveshHeliconia:
Hey @OCA/product-maintainers Could you merge this?

BhaveshHeliconia:
@pedrobaeza can we merge this?

pedrobaeza (changes requested):
The `if` should be after getting the existing `res2`. See the suggestion.

BhaveshHeliconia:
@pedrobaeza I’m not sure I fully understand. Could you please elaborate or point me to the specific suggestion you’re referring to?

product_supplierinfo_for_customer/models/product_supplierinfo.py
pedrobaeza: @BhaveshHeliconia 

```suggestion
if count:
    res += res2
else:
    res2 = res2.read(list(self.env["product.supplierinfo"]._fields.keys()))
    for result in res2:
    res += self.env["product.supplierinfo"].new(result)
```

BhaveshHeliconia: @pedrobaeza We've already covered this case here: https://github.com/OCA/product-attribute/pull/1931/changes#diff-6fc0b1283cb7afcdf7bf53b3e8e5e8a343f43ff4a84d4c65f68e81e756f67933R16-R21

pedrobaeza: But you are repeating code (the search) that there's no need to repeat.

product_supplierinfo_for_customer/models/product_supplierinfo.py
```diff
@@ -13,13 +13,17 @@ def search(self, args, offset=0, limit=None, order=None, count=False):
             self.env.context.get("customerinfo")
             and self._name == "product.supplierinfo"
         ):
-            limit2 = limit - len(res) if limit else limit
+            res_len = res if count else len(res)
+            limit2 = (limit - res_len) if limit else limit
```
pedrobaeza, line 17: Why changing this? There's no need AFAIK.

BhaveshHeliconia:
@pedrobaeza When count=True, res is an integer. In that case, using limit2 = limit - len(res) if limit else limit will raise an error, since len() cannot be applied to an integer.

product_supplierinfo_for_customer/models/product_supplierinfo.py
pedrobaeza: OK, understood.

pedrobaeza (approved):
Thanks for the patch.

/ocabot merge patch

Now this should be fw-ported.

BhaveshHeliconia:
> Thanks for the patch.
> 
> /ocabot merge patch
> 
> Now this should be fw-ported.

In Odoo 17/18/19, this is no longer needed because we use the search_fetch method here and the count parameter is not included in this method’s arguments.

PR #1932, opened 2025-03-17
[18.0][MOV] stock_packaging_calculator => product_packaging_calculator

sebalix:
And moved from `OCA/stock-logistics-warehouse` where it was already migrated (based on these comments: https://github.com/OCA/stock-logistics-warehouse/pull/2173#issuecomment-2717484358).

As this module doesn't depend on `stock`, there is no need for this module to:
- be hosted in a stock-logistics-* repository
- be named with prefix `stock_`

As `stock_packaging_calculator` is already available in 18.0, it is transformed into a transitional module to install the new `product_packaging_calculator`:
- https://github.com/OCA/stock-logistics-warehouse/pull/2294


jbaudoux (approved):
lgTM

product_packaging_calculator/hooks.py
```diff
@@ -0,0 +1,12 @@
...
+from openupgradelib import openupgrade as ou
+
+
+def pre_init_hook(env):
+    # Migrate from 'stock_packaging_calculator'
+    if ou.is_module_installed(
+        env.cr, "stock_packaging_calculator"
+    ) and not ou.is_module_installed(env.cr, "product_packaging_calculator"):
+        modules = [("stock_packaging_calculator", "product_packaging_calculator")]
+        ou.update_module_names(env.cr, modules, merge_modules=True)
```
sebalix: One issue by doing this:

if we have `stock_packaging_calculator` installed, then we switch to its [new version](https://github.com/OCA/stock-logistics-warehouse/pull/2294) (transitional module), this will install `product_packaging_calculator`, and the hook here will remove `stock_packaging_calculator` entry from `ir_module_module` table. And if we still have other modules depending on this one, they have their dependencies broken (and could be uninstalled by mistake).

sebalix: Will do a try with `update_module_moved_fields` + `update_module_moved_models` instead

sebalix: Working well with `update_module_moved_fields` + `update_module_moved_models` :+1: This keeps the existing `stock_packaging_calculator` module installed. Up to the admin to uninstall it afterwards.

PR #1839, opened 2025-01-09
[ADD] product_attribute_custom_value_variant

SirAionTech:

PR #1934, opened 2025-03-18
[18.0][MIG] product_attribute_custom_variant

rousseldenis:
Based on :

- https://github.com/OCA/product-attribute/pull/1839

See related module : https://github.com/OCA/sale-workflow/pull/3631


rousseldenis:
@SirAionTech This can be your interest.

I've added two improvements you can maybe backport to your PR.

rousseldenis:
@monen17

rousseldenis:
I think the sale part can be splitted.

product_attribute_custom_value_variant/models/sale_order_line.py
```diff
@@ -0,0 +1,95 @@
...
+
+            # If new attribute values have been created,
+            # use them to create a new variant and set it in the line
+            attribute_lines = line.product_template_id.attribute_line_ids
+            custom_combination = attribute_lines.product_template_value_ids.filtered(
+                lambda ptav, new=new_attribute_values: ptav.product_attribute_value_id
+                in new
+            )
+            if custom_combination:
+                self._create_custom_attribute_combination(
```
monen17:
```suggestion
                line._create_custom_attribute_combination(
```

product_attribute_custom_value_variant/models/sale_order_line.py
```diff
@@ -0,0 +1,95 @@
...
+    ) -> tuple[ProductAttributeValue, ProductAttributeCustomValue]:
+        """
+        Retrieve the new custom attribute from custom value and
+        the one to remove
+        """
+        self.ensure_one()
+        # Create new attribute values for each "Create custom variant" custom value
+        values_to_unlink = self.env["product.attribute.custom.value"].browse()
+        new_attribute_values = self.env["product.attribute.value"].browse()
+        # Create new attribute values for each "Create custom variant" custom value
```
monen17:
```suggestion
        # Create new attribute values for each "Create custom variant" custom value
        values_to_unlink = self.env["product.attribute.custom.value"].browse()
        new_attribute_values = self.env["product.attribute.value"].browse()
```
Duplicated comment

product_attribute_custom_value_variant/models/sale_order_line.py
```diff
@@ -0,0 +1,95 @@
...
+            ).value_ids -= attribute_line.value_ids & new_attribute_values
+
+    def _assign_new_custom_variant(self):
+        """Create new variants and assign them to the `self`."""
+        custom_values_to_unlink = self.env["product.attribute.custom.value"].browse()
+        for line in self:
+            new_attribute_values, custom_values_to_unlink = (
+                self._get_new_and_unlink_custom_attribute()
+            )
+            custom_values_to_unlink |= custom_values_to_unlink
```
monen17:
```suggestion
        custom_values_to_unlink = self.env["product.attribute.custom.value"].browse()
        for line in self:
            new_attribute_values, line_custom_values_to_unlink = (
                self._get_new_and_unlink_custom_attribute()
            )
            custom_values_to_unlink |= line_custom_values_to_unlink
```
I'm afraid that without a renaming (suggested above), the new `custom_values_to_unlink` would shadow the existing _out-of-the-loop_ variable creating an unwanted behavior.

product_attribute_custom_value_variant/README.rst
```diff
@@ -0,0 +1,110 @@
...
+custom values become real attribute values. Here is an example:
+
+1. Create an attribute "Length (cm)"
+2. Add some values to the attribute:
+
+   - 5
+   - 10
+   - Custom
+
+3. For the "Custom" value, enable custom and Create custom variant
```
monen17:
```suggestion
3. For the "Custom" value, enable "Free text" and "Create custom variant"
```
The field has been renamed in https://github.com/odoo/odoo/commit/50613922fbe67a3bc7c8ccec5799ac49bed0c23b#diff-8acbc08970352949f307ee98ef3f4004f46e0f588ac1b6e82b3b40eaf91d3ebeL39-R39

product_attribute_custom_value_variant/models/sale_order_line.py
```diff
@@ -1,60 +1,87 @@
...
+)
+from odoo.addons.product.models.product_attribute_value import ProductAttributeValue
+
 
 class SaleOrderLine(models.Model):
     _inherit = "sale.order.line"
 
+    def _get_new_and_unlink_custom_attribute(
+        self,
+    ) -> tuple[ProductAttributeValue, ProductAttributeCustomValue]:
```
monen17: This is interesting, are we using types now in Odoo? Maybe I'm not updated with the latest trends, do you have any reference?

monen17 (changes requested):
Thanks for the PR!
I have tried this module in a local instance and it works great!

I also have reviewed the code and noticed a few thinkgs, see the other comments.

Please edit the title of commit https://github.com/OCA/product-attribute/pull/1934/commits/38ddd110479722d4c144044740e20368fc40b7dd because it's too long according to the guidelines:
> please check if the commit message is cut with ellipsis

(from https://github.com/OCA/odoo-community.org/blob/master/website/Contribution/CONTRIBUTING.rst#71commit-message)

and gets truncated: 
image
Since the module name consumes a lot of space already, you can either shorten it or the commit description with something like:
> [IMP] product_attribute_custom_value_variant: Readability
>
> Split code in functions

One more thought: it might also be a **[REF]** because the module's behavior is not changing.

About the backporting, I prefer to do such things when PRs are merged for two reasons:
- avoid the overhead of keeping multiple PRs synced
- one more reason to merge PRs :smirk:

---
> I think the sale part can be splitted.

I agree that ideally this module would only depend on `product` and we'd have a _bridge_ module to make it work with `sale`, but as far as I know custom attribute values have very little (maybe none?) use without sales so the _bridge_ module would be installed very often making the decoupling useless.
This said, if someone wants to do it it's ok by me.

product_attribute_custom_value_variant/models/sale_order_line.py
rousseldenis: Typing is great to auto document the code (it's Odoo here, but it's valid for other projects) as it's standard python and does not harm the code.

rousseldenis:
> About the backporting, I prefer to do such things when PRs are merged for two reasons:
> 
> * avoid the overhead of keeping multiple PRs synced
> * one more reason to merge PRs 😏

Yes, I approved your 16.0's PR. If you can attract people there, we can merge :-)

rousseldenis:
> I agree that ideally this module would only depend on `product` and we'd have a _bridge_ module to make it work with `sale`, but as far as I know custom attribute values have very little (maybe none?) use without sales so the _bridge_ module would be installed very often making the decoupling useless.
> This said, if someone wants to do it it's ok by me.

I will publish the split here soon. The idea was also to simplify the code to be called on sale side in order to replace the variant by the new one.

The idea was to be able to get new variant from custom attribute in any other flow.

rousseldenis:
> (from https://github.com/OCA/odoo-community.org/blob/master/website/Contribution/CONTRIBUTING.rst#71commit-message)
> 
> and gets truncated:

I agree with short commit message BUT I'm not with non meaning ones. And that's not because github interface truncates it that they should go to something like `[IMP] p_a_c_v: (...)` :sweat_smile:

sbejaoui (approved):

PR #1935, opened 2025-03-19
[16.0][ADD] product_attachment_link

MarinaAForgeFlow:


JordiMForgeFlow (approved):
code review 👍🏼

AaronHForgeFlow (approved):
Code LGTM

JoanSForgeFlow (approved):

rousseldenis:
Hi @MarinaAForgeFlow Thanks for this.

IMHO, module name should be changed as `mgmt` is not meaningful.

And `management` is too broad to explain what this module does.

I'd would go with `product_related_attachment` or `product_all_attachment_link` or ...

MarinaAForgeFlow:
> Hi @MarinaAForgeFlow Thanks for this.
> 
> IMHO, module name should be changed as `mgmt` is not meaningful.
> 
> And `management` is too broad to explain what this module does.
> 
> I'd would go with `product_related_attachment` or `product_all_attachment_link` or ...

Hi @rousseldenis , I agree with you and also think that 'management' is too broad. However, there are two modules in the OCA that do practically the same as this one but for different views and models: '[purchase_order_product_attachment_mgmt'](https://github.com/OCA/purchase-workflow/tree/16.0/purchase_order_product_attachment_mgmt) and '[mrp_attachment_mgmt](https://github.com/OCA/manufacture/tree/16.0/mrp_attachment_mgmt)' . So, I decided that it was best to align the naming with this existing modules.

MarinaAForgeFlow:
I’d like to get some feedback on this.
Do you think the action context should be updated to prevent creating new attachments, since they wouldn’t be linked to any record? Alternatively, should we disable creation in the template action, but adjust the action context in the variant so that any new attachment gets automatically linked to the related product template?

JordiMForgeFlow:
@MarinaAForgeFlow I would disable creation, same as they do for example in the purchase_order_product_attachment_mgmt module

rousseldenis:
> So, I decided that it was best to align the naming with this existing modules.

Not really...

I'd prefer renaming

MarinaAForgeFlow:
> > So, I decided that it was best to align the naming with this existing modules.
> 
> Not really...
> 
> I'd prefer renaming

Alright! 😊 I’ll update it to better describe the feature. Do you think product_attachment_link sounds good? Or would product_variant_template_attachment_link be better? It’s longer, but more descriptive.

rousseldenis:
> product_attachment_link

Could be sufficent as it embed 'all' products :-)

product_attachment_link/models/product_product.py
```diff
@@ -0,0 +1,24 @@
...
+
+    def action_see_product_template_attachments(self):
+        domain = [
+            ("res_model", "=", "product.template"),
+            ("res_id", "in", self.product_tmpl_id.ids),
+        ]
+        action = self.env["ir.actions.actions"]._for_xml_id("base.action_attachment")
+        context = action.get("context", {})
+        if isinstance(context, str):
+            context = safe_eval(context)
```
JordiMForgeFlow:
```suggestion
        context = action.get("context", "{}")
        context = safe_eval(context)
```

JordiMForgeFlow (approved):
LGTM 👍🏼

rousseldenis (approved):
Code and functional review

LauraCForgeFlow (approved):

PR #1940, opened 2025-03-26
[16.0][NEW] uom_alias

renatonlima:
I propose this new module that has the functionality of adding one or more aliases to the UoM. This can be useful, for example, when importing electronic invoices from suppliers. In some cases, the supplier's invoice may have the same UoM but with different names. For example: In Odoo, the product may have the UoM Unit, but in the supplier's electronic invoice it may be UND, Units.


rvalyi:
Hello @renatonlima , I think many European countries fix this using the "UNCE" unified nomenclature for Europe with the uom_unece module https://github.com/OCA/community-data-files/tree/17.0/uom_unece
But as for the rest of the world (like us in Brazil), this seems very useful indeed...

pinging some of the electronic invoicing guys: @alexis-via @simahawk @etobella @pedrobaeza 
cc @marcelsavegnago @antoniospneto

pedrobaeza:
It seems a valid use case for me. I'm just missing a check for not repeating the "code" of other existing UoM (there's only a unique constraint on alias names, but not alias<>UoM cross-check).

hparfr (approved):

uom_alias/models/uom_uom.py
```diff
@@ -0,0 +1,35 @@
...
+    ):
+        args = args or []
+        domain = []
+        if name:
+            domain = ["|", ("name", operator, name), ("alias_ids.code", operator, name)]
+        return self._search(
+            expression.AND([domain, args]),
+            limit=limit,
+            access_rights_uid=name_get_uid,
+        )
```
antoniospneto:
```suggestion
    @property
    def _rec_names_search(self):
        return list(set(super()._rec_names_search + ["alias_ids.code"]))
```

In Odoo 16+, I see many people implementing it this way. Could this be better?

antoniospneto (approved):
Left a minor suggestion to simplify the code using _rec_names_search (optional improvement, not a blocker). Everything else looks solid – tested on Runboat, and the functionality works flawlessly. Great addition!

rvalyi:
> Left a minor suggestion to simplify the code using _rec_names_search (optional improvement, not a blocker). Everything else looks solid – tested on Runboat, and the functionality works flawlessly. Great addition!

The idea is to cut through the fat of the l10n_br_fiscal module (largest OCA module by far) with something more standard and depend on uom_alias only at the level of l10n_br_fiscal_edi if possible to make the basic things simpler (easier to grasp and migrate etc....). While the alias uom idea was brought by a contribution from Kmee a few years ago, the code here has not a single line in common, it's really a clean rewrite and this was not yet properly ported to v16 in l10n_br_fiscal BTW.  cc @marcelsavegnago @mbcosta

renatonlima:
Hi @pedrobaeza,

I agree with you, actually the uom.alias.code should be unique to avoid having two uom.uom with the same uom.alias.code

renatonlima:
@antoniospneto

Thanks for your review, I implemented your suggestion, but with a slight modification because _rec_names_search returns None and generated an error so I added an or [] in the concatenation

antoniospneto (approved):

uom_alias/models/uom_alias.py
```diff
@@ -0,0 +1,22 @@
...
+    code = fields.Char()
+
+    uom_id = fields.Many2one(comodel_name="uom.uom")
+
+    _sql_constraints = [
+        (
+            "uom_alias_unique",
+            "UNIQUE(code)",
+            "You can note repeat the alias name",
+        )
```
rvalyi, line 21: s/you can note/you cannot/

rvalyi (changes requested):

rvalyi (approved):
OK for me too now, tested in Runboat, works great

PR #1856, opened 2025-01-23
[17.0][FIX] product_state: fix issue with access rule and missing dependency

mb-ife:

PR #1738, opened 2024-10-01
[17][FIX] product_state: Fix rule has no group

bosd:
on rule: product state public

```
2024-10-01 10:46:34,453 261 WARNING odoo odoo.addons.base.models.ir_model: Rule product.state.public has no group, this is a deprecated feature. Every access-granting rule should specify a group

```

To silenece warning in odoo logger.

PR #1944, opened 2025-04-02
[17.0][IMP][PORT] product state: access rule and dependency like 18.0

CRogos:
continues: https://github.com/OCA/product-attribute/pull/1856
https://github.com/OCA/product-attribute/pull/1681
https://github.com/OCA/product-attribute/pull/1738


CRogos:
@mb-ife @simahawk could you have a look and merge?

CRogos:
@simahawk I totally agree that the v18 version of product_state is a lot cleaner, but I am not sure if we should do so many changes in v17?

CRogos:
@pedrobaeza what do you think about these kind of changes in an stable version?

pedrobaeza:
For me changing module topology (dependencies, splits, renamings, and so on), shouldn't be done in stable, but on the following migration.

CRogos:
@simahawk than we should merge https://github.com/OCA/product-attribute/pull/1856, even though this is not the optimal solution and closes this and the other PRs.

But I would really like to fix the bugs in this module.

simahawk:
The point here is exactly the contrary IMO.
#1856 is adding another useless dependency. A mistake over a mistake.
Here we make sure we don't need it.

This change is basically harmless. No matter which option we choose, the module will have to be upgraded (and the other 2 modules have `auto_install=True`). The worst thing that can happen is that you'll see a menu entry that you don't expect in case you've disabled the old one. Or, if someone is referencing the menu items from a module (maybe to turn them off) it might be broken but the fix is very easy: if you really want to improve something is just to add small migration of the bunch of records that have been moved (menu items, views) and you could simply preserve the old xids to keep ext ref intact. 

I'd leave the decision to @emagdalenaC2i which is the maintainer of product_state though.

My $0.02.

CRogos:
@simahawk thanks for your feedback.
I totally agree that the solution in this PR (backported from v18) is technically much better and clean. 

On the other hand, the other PR adds a missing dependency which must be already installed. It does not make anything worse, because it was already wrong before and the added dependency must already be installed.

I can live with both solutions, but I really would like apply the bugfixes which are also included in this PRs.

CRogos:
@flotho, @yvaucher could you review this improved version of https://github.com/OCA/product-attribute/pull/1738 you've already reviewed?

MohamedOsman7 (approved):
Functional & Code review LGTM

CRogos:
@simahawk could you add a review for this so we can merge?

simahawk (approved):

CRogos:
@pedrobaeza can we merge?

PR #1947, opened 2025-04-07
[16.0][IMP] product_print_category - report wizard launch with print button

quentinDupont:
In Form and List view, `Print Products` is in Print button instead of Action

image


legalsylvain (approved):
Makes sense ! 

thanks !

PR #1949, opened 2025-04-09
[17.0][FIX] product_supplierinfo_for_customer: ensure partner context.

Qlasta:
partner context passed from view not always present.

**Description**
If sale order Pricelist type `partner`,  `price_unit` of sale order line is calculated using partner context passed in the view (https://github.com/odoo/odoo/blob/17.0/addons/sale/views/sale_order_views.xml#L520-L526).
However context is lost if price computation is triggered again from backend (e.g. when updating quantities) and it sets `list price`, but not `product.customerinfo` price.

**Current Behaviour**
When pricelist type is "Partner Prices on the product form", unit price in sale order line sometimes is set from list price not from customer info. Example by updating product quantity a few times in sale order line.

**Expected Behavior**
When pricelist type is "Partner Prices on the product form", unit price in sale order line is always set from customer info.

**Proposed implementation**
extend `_get_product_price_context` and update context by adding `partner_id` from sale order.

### DIFF
--- a/product_supplierinfo_for_customer/__manifest__.py
+++ b/product_supplierinfo_for_customer/__manifest__.py
@@ -13,7 +13,7 @@
     "category": "Sales Management",
     "license": "AGPL-3",
     "maintainers": ["luisg123v"],
-    "depends": ["product"],
+    "depends": ["sale"],
     "data": [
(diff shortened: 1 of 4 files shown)


Qlasta:
wrong module

PR #1952, opened 2025-04-17
[FIX] [18.0] product_state: inefficient post_install_hook

PieterPaulussen:
With databases containing a large set of products, the post install hook which populates the values in the database for existing products was inefficient. The ORM could not handle this dataset. Instead, we should use direct SQL to set the appropriate values directly.


bosd:
Can you please fix pre-commit?

PR #1953, opened 2025-04-21
[17.0][ADD] product_attribute_auto_add: module add

Bearnard21:
This module saves time by automatically managing product attribute values. 
When you add a new value to an attribute, the system automatically adds a new attribute value to all relevant products.

How it works:
1. You add a new value to an existing attribute
2. The system automatically adds this value to all products using this attribute


GabbasovDinar (approved):

geomer198 (approved):
LGTM

product_attribute_auto_add/models/product_attribute.py
```diff
@@ -0,0 +1,17 @@
...
+
+
+from odoo import fields, models
+
+
+class ProductAttribute(models.Model):
+    _inherit = "product.attribute"
+
+    attribute_line_auto_add = fields.Boolean(
+        string="Auto-Create Variants",
```
rousseldenis: I think this string is not correct as it won't create variant automatically (as it depends on `create_variant` configuration - https://github.com/odoo/odoo/blob/17.0/addons/product/models/product_attribute.py#L23)

product_attribute_auto_add/models/product_attribute_value.py
```diff
@@ -0,0 +1,41 @@
...
+            recordset: The newly created attribute values.
+        """
+        results = super().create(vals_list)
+        attribute_line_obj = self.env["product.template.attribute.line"]
+        auto_add_map = defaultdict(results.browse)
+        for record in results.filtered("attribute_id.attribute_line_auto_add"):
+            auto_add_map[record.attribute_id] |= record
+
+        for attribute, values in auto_add_map.items():
+            lines = attribute_line_obj.search(
```
rousseldenis: This should be changed. You should do the search in another manner to do a global search and then in the iteration, filter the result recordset to improve performances.

rousseldenis (commented):
Some comments.

I have a concern about this. How do you manage products that does not have a particular attribute value ?

Aldeigja (approved):
Functional LGTM

Bearnard21:
@rousseldenis The code has been updated, please take a look.

Bearnard21:
@JavierGalera8
@rvalyi
@rousseldenis

Could you please take a look at the PR

rousseldenis:
> I have a concern about this. How do you manage products that does not have a particular attribute value ?

@mike-cetmix You didn't answered this. :sweat_smile:

GabbasovDinar:
@rousseldenis Hello,

I've added a new setting to the product template to allow disabling the automatic addition of attribute values.
The documentation has also been updated to reflect this change.
Please have a look when you have a moment.

ivs-cetmix:
Hello @rousseldenis could you please check the latest updates?

rousseldenis (approved):

PR #1954, opened 2025-04-22
[16.0][IMP] product_get_price_helper: improve performance

thienvh332:
### Result
Test results before and after change:
- Step 1: get the server action trick ready and time a call with a list of about 100 products
  - Time duration: **0.5092663764953613** seconds
  - sql_count: **462**
  - entry_count: **499**
- Step 2: fix the implementation of _get_price() to have it run efficiently on multiple products (**This PR**)
- Step 3: use the server action trick again to time the call on the same set of products as in step 1 and report back the improvement
  -  Time duration: **0.04854106903076172** seconds
  - sql_count: **26**
  - entry_count: **33**

==> **Execution time is expected to improve by ~10x with this change.**

### Notes:
- The test data is newly created based on the local  source. In practice it will take longer to execute due to the complexity of the data.

### DIFF
--- a/product_get_price_helper/models/product_product.py
+++ b/product_get_price_helper/models/product_product.py
@@ -25,7 +25,7 @@ class ProductProduct(models.Model):
         :param company:     Optional.
         :param date:        Optional.
 
-        :returns: dict with the following keys:
+        :returns: A dictionary where keys are product IDs and values are dictionaries with:
 
             <value>                 The product unitary price
             <tax_included>          True if product taxes are included in <price>.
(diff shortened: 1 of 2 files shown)


santostelmo (approved):

simahawk:
pls rewrite the commit msg to include the module name. The explanation of what you have done (eg: compute in batches) goes into the description of the commit.

thienvh332:
> pls rewrite the commit msg to include the module name. The explanation of what you have done (eg: compute in batches) goes into the description of the commit.

My bad, thanks for the reminder.
I've updated it.

simahawk (approved):

PR #1956, opened 2025-04-24
[FIX] product_dimension: product volume conversion support freedom units

gabe-grinspan:
I encountered an issue with the following flow when both the product_dimension and product_logistics_uom modules are installed:
1. Set system-wide dimensions to imperial UOMs (ft, ft^3, lbs)
2. Go to a product and set dimensions
3. The volume is locked as a meters^3 value regardless of the volume set

To solve this I am multiplying the value from the product lengths, converting to meters, then convert to m^3 and then back to the UOM set on the product volume. This should support all configurations, especially those with American leanings.

### DIFF
--- a/product_dimension/models/product_product.py
+++ b/product_dimension/models/product_product.py
@@ -16,7 +16,9 @@ class ProductProduct(models.Model):
         "Dimensional UoM",
         domain=lambda self: self._get_dimension_uom_domain(),
         help="UoM for length, height, width",
-        default=lambda self: self.env.ref("uom.product_uom_meter"),
+        default=lambda self: self.env[
+            "product.template"
+        ]._get_length_uom_id_from_ir_config_parameter(),
     )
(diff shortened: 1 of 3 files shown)


gabriel-grinspan:
@rousseldenis can you or someone else review?

gabe-grinspan:
@rousseldenis @pedrobaeza
I discovered a bug in the code for `product_logistics_uom` that seems to have been there since v16. I went ahead and updated my commit to also fix that issue, as part of the purpose of this was to enhance compatibility between the two modules. Please have someone review.

Also the tests failing are not related to my changes.

gabe-grinspan:
@hparfr @simahawk @ValentinVinagre @HaraldPanten @luis-ron
Can someone please review this?

rousseldenis:
@gabe-grinspan Could you add a test to highlight the problem ? Thanks

PR #1961, opened 2025-04-25
Explain what the module is not

zeroheure:
"code" content is not a shortcut


rousseldenis:
@zeroheure You should explain a little bit more.

As I see, a search has been added in views: https://github.com/OCA/product-attribute/blob/18.0/product_category_code/views/product_category.xml#L22

PR #1965, opened 2025-04-30
[FIX] 18.0 product_status: re-align template and variant form view

PieterPaulussen:
This is the template view prior to the changes:
image
You can clearly see that the information groups for e-commerce are split and that the product status group has been glued to the left side of the form.

This is a screenshot of the result:
image
The product status group has been moved to its own page and now presents itself consistently without breaking the other information groups.


PieterPaulussen:
@pedrobaeza a review would be nice. Or a merge even better🙂

pedrobaeza:
I'm not using this module, sorry.

PR #1969, opened 2025-05-07
[18.0][MIG] product_route_mto

jbaudoux:
Depends on:
* https://github.com/OCA/stock-logistics-warehouse/pull/2330

cc @sebalix @rousseldenis

PR #1970, opened 2025-05-07
[18.0][MIG] product_variant_route_mto (was stock_product_variant_mto)

jbaudoux:
[MOD] I forbid to set a variant as non MTO when the product is MTO as this is confusing. You can only set a variant as MTO when the product is not MTO ( cc @Highcooley )

Depends on:
* https://github.com/OCA/stock-logistics-warehouse/pull/2330
* https://github.com/OCA/product-attribute/pull/1969

Replaces: (I moved fixes from the migration commit into a separate commit) 
* https://github.com/OCA/product-attribute/pull/1879

cc @sebalix @simahawk @grindtildeath


product_variant_route_mto/models/product_template.py
```diff
@@ -7,33 +8,46 @@
 class ProductTemplate(models.Model):
     _inherit = "product.template"
 
-    def write(self, values):
-        if "route_ids" not in values:
+    def dis_write(self, values):
```
jbaudoux: FIXME: test this can be dropped

product_variant_route_mto/__manifest__.py
```diff
@@ -9,10 +10,10 @@
     "category": "Inventory",
     "website": "https://github.com/OCA/product-attribute",
     "author": "Camptocamp SA, Odoo Community Association (OCA)",
-    "maintainers": ["mmequignon"],
+    "maintainers": ["mmequignon", "jbaudoux"],
     "license": "AGPL-3",
     "installable": True,
     "auto_install": False,
-    "depends": ["stock"],
+    "depends": ["product_route_mto"],
```
sebalix, line 17: Nitpick: your commit message should be `rely on product_route_mto` (which rely itself on `stock_route_mto`)

jbaudoux: updated commit msg

sebalix (approved):

simahawk (approved):

PR #1979, opened 2025-05-16
[18.0][ADD] product_category_name_translatable

JordiMForgeFlow:


MarinaAForgeFlow (approved):
LGTM

LauraCForgeFlow (approved):

AaronHForgeFlow (approved):
LGTM!

AndreuOForgeFlow (approved):

rvalyi (approved):

rousseldenis:
@JordiMForgeFlow Could you explain how you manage that case : https://github.com/odoo/odoo/pull/36717

(the reason Odoo removed the translated field)

Thanks!

JordiMForgeFlow:
@rousseldenis indeed I did not know about this limitation. Nonetheless, it may not be a problem if there is no important usage of complete_name and the customer is aware of it. What we could do is to update the README to put a warning about this, WDYT?

rousseldenis:
> @rousseldenis indeed I did not know about this limitation. Nonetheless, it may not be a problem if there is no important usage of complete_name and the customer is aware of it. What we could do is to update the README to put a warning about this, WDYT?

Yes, should be great

PR #1985, opened 2025-05-26
[16.0][IMP] product_pricelist_margin computation - allow pricelist item margin simulation

santostelmo:
The following refactoring is proposed:
- The price is based on the computation applied from the current pricelist item. This allows the user to make price simulations while editing the pricelist item.

### DIFF
--- a/product_pricelist_margin/models/product_pricelist_item.py
+++ b/product_pricelist_margin/models/product_pricelist_item.py
@@ -2,7 +2,7 @@
 # License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
 
 from odoo import api, fields, models
-from odoo.tools.float_utils import float_is_zero
+from odoo.tools.float_utils import float_is_zero, float_round
 
 
 class ProductPricelistItem(models.Model):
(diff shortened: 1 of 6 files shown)


legalsylvain:
Hi. Sorry i dont see the relation between your description (par + message commit) and the changes in the code.

santostelmo:
> Hi. Sorry i dont see the relation between your description (par + message commit) and the changes in the code.

Hi @legalsylvain, sorry I probably didn't clearly express my intention

What I mean is that in the previous development, price was computed assuming all the rules of current pricelist by calling `_compute_price_rule`. This implementation does not allow price simulations based on the current item. That happens for example when you add a new item with fixed price for a product already present in the current pricelist.

Consider the following usecase covered by `test_margin_with_fixed_price_computation`
Before:
1. Set product P1 cost = 20
2. Add pricelist item for P1 with fixed_price = 35 => Margin: 35 - 20 = 15 ✅
3. Add another pricelist item for P1 with fixed price = 30 => Margin: **35**-20 = 15 ❌
This is wrong because `_compute_price_rule` always consider [the rule from point 2 ](https://github.com/odoo/odoo/blob/2aa58419605b112f46570a4f7d1f533415937c40/addons/product/models/product_pricelist.py#L154)with a fixed price of 35.

With this implementation
1. Product P1 cost = 20
2. Add pricelist item for P1 with fixed price = 35 => Margin: 35 - 20 = 15 ✅
3. Add another pricelist item for P1 with fixed price = 30 => **Margin: **30**-20 = 10** ✅
Now the pricelist item margin simulation is correct

santostelmo:
@nguyenminhchien could you please have a look into this PR ?

nguyenminhchien (approved):
Lgtm

gurneyalex (approved):

product_pricelist_margin/models/product_pricelist_item.py
```diff
@@ -70,6 +74,7 @@ def _compute_margin(self):
             )
 
             item.margin = price_vat_excl - cost
-            item.margin_percent = (
-                price_vat_excl and (item.margin / price_vat_excl) * 100
+            item.margin_percent = float_round(
+                price_vat_excl and (item.margin / price_vat_excl) * 100,
+                self.env["decimal.precision"].precision_get("Product Unit of Measure"),
```
legalsylvain, line 79: Please dont use precision in computation. It is not overloadable. 

(Odoo core doesn't use such precision for Margin. 

https://github.com/odoo/odoo/blob/18.0/addons/sale_margin/models/sale_order_line.py)

legalsylvain (commented):
Sorry i forgot to confirm a review.

PR #1995, opened 2025-06-06
adding product cost and logic

shimemuleymges:


product_abc_classification_sale_stock/models/abc_classification_profile.py
```diff
@@ -129,13 +215,8 @@ def _get_data(self, from_date=None):
         customer_location_ids = (
             self.env["stock.location"].search([("usage", "=", "customer")]).ids
         )
-        # Collect all the product linked to the profile to be sure to provide
```
lmignon, line 132: Why do you remove these useful comments?

lmignon (changes requested):
Thank you for the proposal. IMO this new profile must got into a dedicated addon to ease the maintenance.

rousseldenis:
Hi @shimemuleymges. Thanks for this.

IMHO, too much changes here.

Don't rename metthods attributes, no value added.

For me, cost behavior should be done in another module.

shimemuleymges:
Ok I will create custom module and inherite

PR #1686, opened 2024-07-16
[17.0][MIG] product_abc_classification: Migration to 17.0

fgaviiria:

PR #2007, opened 2025-06-27
[17.0][MIG] product_abc_classification: Migration to 17.0

anothingguy:
supersedes https://github.com/OCA/product-attribute/pull/1686


giarve:
When duplicating a classification profile I get:

"The operation cannot be completed: Profile name must be unique"

maybe is also for lower versions. If you could create a separate commit overriding "def copy" for abc.profile that adds (copy) to the previous name, it could be backported or already fixed in this version

anothingguy:
added copy function and re-run pre-commit in separate commit

PR #2020, opened 2025-07-10
[18.0][MIG] product_supplierinfo_import_by_barcode > product_supplierinfo_import: Migration to 18.0

victoralmau:
Migration to 18.0

Please @carlos-lopez-tecnativa and @juancarlosonate-tecnativa can you review it?

@Tecnativa TT54470


product_supplierinfo_import_by_barcode/models/product_supplierinfo_import_template.py
```diff
@@ -0,0 +1,71 @@
...
+                for h in self.template_line_ids.mapped("header_name")
+            ]
+            + [self.search_header_name]
+        )
+
+    @api.constrains("sheet_number")
+    def _check_integer_value(self):
+        for record in self:
+            if record.sheet_number < 1:
+                raise ValidationError(_("The Sheet Number cannot be less than 1."))
```
carlos-lopez-tecnativa:
```suggestion
                raise ValidationError(self.env._("The Sheet Number cannot be less than 1."))
```

product_supplierinfo_import_by_barcode/models/product_supplierinfo_import_template.py
```diff
@@ -0,0 +1,71 @@
...
+class ProductSupplierInfoImportTemplateLine(models.Model):
+    _name = "product.supplierinfo.import.template.line"
+    _description = "Vendor import columns and fields"
+
+    template_id = fields.Many2one(comodel_name="product.supplierinfo.import.template")
+    header_name = fields.Text(
+        help="Copy it directly from the origin cell so it matches right"
+    )
+    field_id = fields.Many2one(
+        comodel_name="ir.model.fields", domain=[("model", "=", "product.supplierinfo")]
```
carlos-lopez-tecnativa: I think this should only display storable fields. What do you think?

carlos-lopez-tecnativa: These two fields must be marked as required (either at the field level or in the view) to prevent the user from creating a line with these fields empty and later getting errors like this:
image

product_supplierinfo_import_by_barcode/readme/ROADMAP.md
```diff
@@ -0,0 +1 @@
+Rename the module in future migrations to product_supplierinfo_import.
```
carlos-lopez-tecnativa: What about this rename? Can it be done in this migration?

product_supplierinfo_import_by_barcode/tests/test_product_supplierinfo_import_by_barcode.py
```diff
@@ -0,0 +1,317 @@
...
+    def _import_supplierinfo_file(
+        self, file_path, supplier, delay=0, create_new_products=True
+    ):
+        """ "Helper: import an excel file for testing"""
+        return self.env["product.supplierinfo.import"].create(
+            {
+                "supplierinfo_filename": file_path,
+                "supplierinfo_file": self._data_file(file_path),
+                "supplier_id": supplier.id,
+                "date_start": fields.Date.today(),
```
carlos-lopez-tecnativa: In business code, I prefer to use `context_today` to prevent mistakes with time zones. In test-related code, it's not very important, but it can cause tests to fail at certain hours if you use it in any assert. If you can update it to my suggestion, that would be great, but it's not blocking.
```suggestion
                "date_start": fields.Date.context_today(),
```

product_supplierinfo_import_by_barcode/wizards/product_supplierinfo_import.py
```diff
@@ -0,0 +1,288 @@
...
+    _description = "Import supplier info records"
+
+    supplier_id = fields.Many2one(
+        comodel_name="res.partner",
+        string="Supplier",
+        compute="_compute_supplier_id",
+        store=True,
+        readonly=False,
+    )
+    date_start = fields.Date(string="Validity", required=True)
```
carlos-lopez-tecnativa: I don't understand this field. In the sample template used in the test, I see a column for `Start Date`.
image
I created a field mapping to use this column with the `Start Date` field, but it’s ignored, `date_start` is used in this case.
image
So, does that mean I can never use `Start Date` in the field mapping?
I expected it to use the value from the file, and if no mapping field exists, then use this field.
What do you think?

product_supplierinfo_import_by_barcode/wizards/product_supplierinfo_import.py
```diff
@@ -0,0 +1,288 @@
...
+            lambda x: x.sudo().field_id.name == "product_name"
+        ).header_name
+        vendor_product_name = row_data.get(
+            vendor_product_name_header,
+            _("%(code)s (product imported)", code=search_value),
+        )
+        product_data = {
+            "created_from_supplierinfo_import": True,
+            "name": vendor_product_name,
+            "purchase_ok": False,
```
carlos-lopez-tecnativa: Just to clarify, if I import data and new products are created, why are these new products not marked as `purchase_ok=True`?
Isn't it more likely that these products will be used in the purchase flow?
I expected this field to be set to True.
image

product_supplierinfo_import_by_barcode/wizards/product_supplierinfo_import.py
```diff
@@ -0,0 +1,288 @@
...
+            "sale_ok": False,
+            search_field: search_value,
+        }
+        try:
+            product = self.env["product.template"].create(product_data)
+            product.message_post(
+                body=_("Created from supplier price list import"),
```
carlos-lopez-tecnativa:
```suggestion
                body=self.env._("Created from supplier price list import"),
```

product_supplierinfo_import_by_barcode/wizards/product_supplierinfo_import.py
```diff
@@ -0,0 +1,288 @@
...
+            context = {}
+        context.update({"visible_product_tmpl_id": False})
+        return {
+            "name": "Imported supplier infos",
+            "type": "ir.actions.act_window",
+            "view_mode": "list",
+            "res_model": "product.supplierinfo",
+            "domain": domain,
+            "context": context,
+            "help": _(
```
carlos-lopez-tecnativa:
```suggestion
            "help": self.env._(
```

carlos-lopez-tecnativa (changes requested):
Additionally, could you add some templates and field mappings as demo data (using the Excel files used in the test) to facilitate manual testing and save time?

pedrobaeza:
/ocabot migration product_supplierinfo_import_by_barcode

This module should be renamed to `product_supplierinfo_import`.

product_supplierinfo_import_by_barcode/wizards/product_supplierinfo_import.py
victoralmau: IMO this is an error because the code does not indicate that it is intentional, therefore, I have changed it.

product_supplierinfo_import_by_barcode/models/product_supplierinfo_import_template.py
victoralmau: The field that must be required is `header_name` (already changed). The field `field_id` is optional.

product_supplierinfo_import_by_barcode/wizards/product_supplierinfo_import.py
victoralmau: IMO this is somewhat intentional because the date (`date_start`) from the wizard is used (in the template if you notice the field is not mapped), maybe a good idea would be to not allow to select that field to avoid confusion.

product_supplierinfo_import/security/ir.model.access.csv
```diff
@@ -1,4 +1,4 @@
 id,name,model_id:id,group_id:id,perm_read,perm_write,perm_create,perm_unlink
-product_supplierinfo_import_by_barcode.access_product_supplierinfo_import_template,access_product_supplierinfo_import_template,product_supplierinfo_import_by_barcode.model_product_supplierinfo_import_template,base.group_user,1,1,1,1
-product_supplierinfo_import_by_barcode.access_product_supplierinfo_import_template_line,access_product_supplierinfo_import_template_line,product_supplierinfo_import_by_barcode.model_product_supplierinfo_import_template_line,base.group_user,1,1,1,1
-product_supplierinfo_import_by_barcode.access_product_supplierinfo_import,access_product_supplierinfo_import,product_supplierinfo_import_by_barcode.model_product_supplierinfo_import,base.group_user,1,1,1,1
+product_supplierinfo_import.access_product_supplierinfo_import_template,access_product_supplierinfo_import_template,product_supplierinfo_import.model_product_supplierinfo_import_template,base.group_user,1,1,1,1
```
pedrobaeza: You don't need to put the full XML-ID here.

product_supplierinfo_import/tests/test_product_supplierinfo_import.py
```diff
@@ -28,72 +27,7 @@ def setUpClass(cls):
...
-                            "header_name": "Final",
-                            "field_id": cls.env.ref(
-                                "product.field_product_supplierinfo__price"
-                            ).id,
-                        }
-                    ),
-                ],
-            }
-        )
+        cls.import_template_2 = cls.env.ref("product_supplierinfo_import.template_2")
```
pedrobaeza: Don't use demo data in tests. It's OK to have both things, but each one should be independent.

pedrobaeza (changes requested):

product_supplierinfo_import/tests/test_product_supplierinfo_import.py
victoralmau: This suggestion again? We have already discussed it other times, and we each have totally opposite opinions. 
Is there any guideline that explicitly forbids this? Is this comment just a suggestion or something blocking?

I'm not going to debate any more about this, what you decide will be what will be done from now on in all PRs, modules and reviews. 

I will **NOT** duplicate code or maintain duplicate code, therefore, if this is finally necessary, I will remove the demo data and leave only the data in the test setup.

pedrobaeza: Yes, it's a guideline:

https://github.com/OCA/odoo-community.org/blob/master/website/Contribution/CONTRIBUTING.rst#avoid-relying-on-demo-data

victoralmau: I do not read there that this is forbidden, it indicates that it is recommended to create the data in the tests (and is related to not using demo data from core modules). In any case there is no problem, if you prefer I will delete the demo data and problem solved. 

I will take this suggestion into account from now on for all OCA modules.

pedrobaeza: @victoralmau demo data are not duplicated code against test code.

Both are intended for different things. Mixing both, in one hand, it makes you to touch in 2 places when something changes across versions, just to make tests to work.

In the other, it makes that if you are running tests locally, and you have modified such demo data, tests will fail and drives you crazy to discover the problem.

That's why it's very advisable to separate both. The guideline is not strictly forbidding it, as most of the guidelines are recommendations.

Check for example this one: https://github.com/OCA/odoo-community.org/blob/master/website/Contribution/CONTRIBUTING.rst#614avoid-dynamic-dates. That one is something that you wouldn't do ever, but it's redacted also with the words "avoid".

In this specific module, I would put demo data if we accompany them with a sample Excel, and mentioning this in the README, but one trouble is that the Excel should be accessible in runboat. The goal of the demo data is to provide easy setup to test things by users.

victoralmau: Okay, I'm not going to argue my case for using demo data defined in the module itself, because I've done so before. "_Any change_" made to the data will most likely cause almost any test in any module to fail, so that argument has always seemed very weak to me.

That said, no problem. I'll keep in mind from now on not to use demo data in tests to avoid the same change being suggested over and over again, and also not to create demo data to avoid "defining the same data" twice.

pedrobaeza (approved):

product_supplierinfo_import/wizards/product_supplierinfo_import.py
```diff
@@ -0,0 +1,288 @@
...
+                )
+            sheet = workbook.sheet_by_index(template.sheet_number - 1)
+            header_values = self._parse_header(sheet.row_values(template.header_offset))
+            if set(header_values) == header:
+                self.template_id = template
+                return header_values
+        if not self.template_id:
+            raise UserError(
+                self.env._(
+                    f"No matching template for these header columns.\n"
```
carlos-lopez-tecnativa, line 89: No format string is necessary in this line.
```suggestion
                    "No matching template for these header columns.\n"
```

carlos-lopez-tecnativa (approved):
LGTM

PR #2023, opened 2025-07-11
[FIX][18.0] product_sticker: Show stickers on one attribute value products with create variant

Shide:
If you configure a product template with an attribute with one value that creates variants, the product.product won't have attribute value data since it's the only one created.

With this fix, those products will show stickers properly.

MT-10848 @moduon @chienandalu @yajo @Andrii9090 @u0f please review if you want 😄


product_sticker/tests/common.py
```diff
@@ -5,6 +5,8 @@
 
 from odoo.tests import common, tagged
 
+# TODO: Hacer los tests para los stickers que solo tienen un atributo
+
```
chienandalu: It's done, isn't it? :slightly_smiling_face: 

```suggestion
```

product_sticker/models/product_sticker.py
```diff
@@ -206,7 +206,12 @@ def _get_stickers(self, products, extra_domain=None):
         no_variant_attribute_lines = product_templates.attribute_line_ids.filtered(
             lambda al: al.attribute_id.create_variant == "no_variant"
         )
-        pp_pavs = products.product_template_variant_value_ids.product_attribute_value_id
+        pp_pavs = self.env["product.attribute.value"].browse()
+        for product in products:
+            if product.product_template_variant_value_ids:
+                pp_pavs |= product.product_template_variant_value_ids.product_attribute_value_id  # noqa: E501
```
chienandalu, line 213: Binary union tends to be less efficient than regular agreggation in my experience
```suggestion
                pp_pavs += product.product_template_variant_value_ids.product_attribute_value_id  # noqa: E501
```

product_sticker/models/product_sticker.py
```diff
@@ -206,7 +206,12 @@ def _get_stickers(self, products, extra_domain=None):
...
+            else:
+                pp_pavs |= product.product_template_attribute_value_ids.attribute_line_id.value_ids  # noqa: E501
```
chienandalu, line 215:
```suggestion
                pp_pavs += product.product_template_attribute_value_ids.attribute_line_id.value_ids  # noqa: E501
```

chienandalu (changes requested):
Code review. Just some tiny changes. Also: missing commit message and task

product_sticker/models/product_sticker.py
yajo: suggestion: actually you can do all in one shot:


```suggestion
        products_with_variants = products.filtered("product_template_variant_value_ids")
        pp_pavs = products_with_variants.product_template_variant_value_ids.product_attribute_value_id | (products - products_with_variants).product_template_attribute_value_ids.attribute_line_id.value_ids
```

product_sticker/models/product_sticker.py
yajo: However, in this case, if another product has the same attribute value, sum it would create duplicates, right? Thus I think it's better as it is, or with my other suggestion below.

yajo (approved):

product_sticker/models/product_sticker.py
Shide: As @yajo said, I don't want duplicates. Just a set of recordsets.

product_sticker/models/product_sticker.py
Shide: I think is less confusing if I use an if statement. I've added a comment to describe the first part of the IF

product_sticker/models/product_sticker.py
chienandalu: Ok, makes sense. Just be aware the larger the recordset (thoundans of records) the larger the performance drawback of the union, as an `OrderedSet` is being made every time

chienandalu (commented):
There's no commit message with the changes

Shide:
> There's no commit message with the changes

What changes? The only relevant commit is the "Show stickers on one attribute value products with create variant"

chienandalu (approved):
> What changes? The only relevant commit is the "Show stickers on one attribute value products with create variant"

Simply the description in the PR would suffice

chienandalu (approved):
:smile: :+1:

PR #2030, opened 2025-07-21
[16.0][IMP] product_pricelist_margin original vs final price

santostelmo:
With this development, when the user adds an item to the price list, the original price can be compared with the final price after the item has been applied.

Original price: The price of the product before applying this item.
Final price: The price of the product after applying this item.

### DIFF
--- a/product_pricelist_margin/models/product_pricelist_item.py
+++ b/product_pricelist_margin/models/product_pricelist_item.py
@@ -20,6 +20,14 @@ class ProductPricelistItem(models.Model):
         string="Margin (%)",
         compute="_compute_margin",
     )
+    original_price = fields.Float(
+        compute="_compute_margin",
+        help="Original price before applying this pricelist item.",
+    )
+    final_price = fields.Float(
+        compute="_compute_margin",
+        help="Final price after applying this pricelist item.",
+    )
 
     @api.depends(
         "compute_price",
@@ -43,6 +51,8 @@ class ProductPricelistItem(models.Model):
             ):
                 item.margin = 0
                 item.margin_percent = 0
+                item.original_price = 0
+                item.final_price = 0
                 continue
 
             price = item._compute_price(
@@ -55,6 +65,8 @@ class ProductPricelistItem(models.Model):
             if float_is_zero(price, precision_digits=item.currency_id.rounding):
                 item.margin = 0
                 item.margin_percent = 0
+                item.original_price = 0
+                item.final_price = 0
                 continue
 
             current_company = self.env.company
@@ -66,13 +78,19 @@ class ProductPricelistItem(models.Model):
                 item.currency_id,
                 product=item.product_tmpl_id,
             )
-
             price_vat_excl = res["total_excluded"]
 
             cost = self.env.user.company_id.currency_id.compute(
                 item.cost, item.currency_id
             )
-
+            pricelist = item.pricelist_id._origin or item.pricelist_id
+            item.original_price = pricelist._get_product_price(
+                item.product_tmpl_id,
+                item.min_quantity,
+                item.product_tmpl_id.uom_id,
+                fields.Datetime.now(),
+            )
+            item.final_price = price_vat_excl
             item.margin = price_vat_excl - cost
             item.margin_percent = float_round(
                 price_vat_excl and (item.margin / price_vat_excl) * 100,
--- a/product_pricelist_margin/views/product_pricelist_item_views.xml
+++ b/product_pricelist_margin/views/product_pricelist_item_views.xml
@@ -9,6 +9,8 @@
                 <field name="cost" invisible="1" />
                 <field name="margin" optional="show" />
                 <field name="margin_percent" optional="show" />
+                <field name="original_price" optional="hide" />
+                <field name="final_price" optional="hide" />
             </field>
         </field>
     </record>
@@ -41,6 +43,8 @@
                     <field name="cost" invisible="1" />
                     <field name="margin" />
                     <field name="margin_percent" />
+                    <field name="original_price" />
+                    <field name="final_price" />
                 </group>
             </xpath>
         </field>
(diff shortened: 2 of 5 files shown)


cyrilmanuel (approved):

wouitmil (approved):

leemannd:
Hello @OCA/product-maintainers , Do you have time to have a look at it and maybe merge it?

legalsylvain:
could you rebase ? I'd like to test on runboat.

product_pricelist_margin/views/product_pricelist_item_views.xml
```diff
@@ -41,6 +41,8 @@
                     <field name="cost" invisible="1" />
                     <field name="margin" />
                     <field name="margin_percent" />
+                    <field name="original_price" />
+                    <field name="final_price" />
```
legalsylvain, line 47: maybe should be added also in the tree view, with optional = show.

santostelmo: I added it on the view but with optional = hide to keep the fields that are currently shown in the tree view

santostelmo:
> could you rebase ? I'd like to test on runboat.

Rebase done

product_pricelist_margin/models/product_pricelist_item.py
```diff
@@ -66,13 +78,19 @@ def _compute_margin(self):
...
             )
-
+            pricelist = item.pricelist_id._origin or item.pricelist_id
+            item.original_price = pricelist._get_product_price(
+                item.product_tmpl_id,
+                item.min_quantity,
+                item.product_tmpl_id.uom_id,
+                fields.Datetime.now(),
+            )
+            item.final_price = price_vat_excl
```
legalsylvain, line 93: Hum. I guess this computation will be wrong if vat are included.

_get_product_price will return vat incl price. 
final_price will be vat excl.

legalsylvain (changes requested):

product_pricelist_margin/models/product_pricelist_item.py
santostelmo: @legalsylvain  `pricelist._get_product_price()` returns the price before taxes (VAT excluded). 
Pricelist prices are stored and computed without taxes. Taxes are applied separately afterward when creating sale order lines, invoices, etc.

In this case margin is always calculated based on tax excluded amounts.
Let me know if that sounds correct for you

cyrilmanuel: HI @legalsylvain,  can we have your input on this topic :)

legalsylvain: I'll do test. Thanks !

legalsylvain:
Hi. 
I don't get it. I made test, and the results are wrong. 

I create 2 identical products : 
- Cost : 50 $ vat excl 
- sale price 100 $ (= 83.33 + 16.67 of VAT).
- I configure one with a VAT included, and another, with VAT excluded.

Then I create a new pricelist, I create two pricelist item, with the two products, and I put a fixed price of 60 $.
The margin is so : 60 $ - 50 $ = 10 $ 

### Vat Excl

image

### Vat incl

image

### New pricelist


### Current Result 

the margin are different

image

### Expected result

Same product should have same margin.

Did I missed something ?

santostelmo:
> Hi. I don't get it. I made test, and the results are wrong.
> 
> I create 2 identical products :
> 
> * Cost : 50 $ vat excl
> * sale price 100 $ (= 83.33 + 16.67 of VAT).
> * I configure one with a VAT included, and another, with VAT excluded.
> 
> Then I create a new pricelist, I create two pricelist item, with the two products, and I put a fixed price of 60 $. The margin is so : 60 $ - 50 $ = 10 $
> 
> ### Vat Excl
> image
> ### Vat incl
> image
> ### New pricelist
> ### Current Result
> the margin are different
> 
> image
> ### Expected result
> Same product should have same margin.
> 
> Did I missed something ?

For me it looks correct. I think it's a matter of interpretation.
Using your use case:

Product (VAT excluded) => Sale net amount = 60
Margin = 60-50 = 10

Product (VAT included) => Sale net amount = 60-60*20%=48 
Margin = 48 - 50 = -2 (no margin)

legalsylvain:
Hi @santostelmo. Please take a look again. The value you used in your last message are not the same !
Regards.

Note : no, margin is not a matter of interpretation ;-)

santostelmo:
> Hi @santostelmo. Please take a look again. The value you used in your last message are not the same ! Regards.
> 
> Note : no, margin is not a matter of interpretation ;-)

I don't get why you say the values aren't the same? (We can make a quick to discuss the subject briefly 😉) 
On the VAT excluded product, we the net amount is 60 to calculate the margin.
On the VAT included product ,we deduct 20% VAT before calculate the margin. The net amount is 48
If the final margin is negative then we assume that there's no margin.

To double check with AI prompt:
> In odoo I have a Product A and Product B.
> - Product A configured with VAT excluded at 20%. On the pricelist the fixed price is 60.
> - Product B configured with VAT included at 20%. On the pricelist the fixed price is 60.
> - The cost of product A and product B is 50
> What is the margin of product A and product B ?

Answer:
> In Odoo, the Margin on a sales order line is calculated as the difference between the Tax-Excluded Sales Price and the Cost.
The key difference here lies in how Odoo treats the "Fixed Price" in the pricelist based on whether the tax is marked as "Included in Price" or not.
Calculation Breakdown
image

legalsylvain:
I'm sorry. I made a mistake in my review. It looks like the values are correct.

Thanks for the explanation.

/ocabot merge minor

santostelmo:
> I'm sorry. I made a mistake in my review. It looks like the values are correct.
> 
> Thanks for the explanation.
> 
> /ocabot merge minor

Thanks a lot for the time spent on testing that

cyrilmanuel (approved):

cyrilmanuel:
hi, @legalsylvain , @pedrobaeza  merge the fix with update template https://github.com/OCA/product-attribute/pull/2226 , i rebase the code, it shall be merge if you redo the command :) 

thanks for your help

PR #2031, opened 2025-07-21
[16.0][FIX] product_catalog: open catalog from sales order

kobros-tech:
image

image


kobros-tech:
@pedrobaeza 

I am sorry if I made entire change, we can discuss.

I just need to have catalog kanban view from sales order form view, I tried your version and it didn't work for me.

thanks!

pedrobaeza:
Isn't this https://github.com/OCA/product-attribute/tree/16.0/product_catalog_sale ?

kobros-tech:
> Isn't this https://github.com/OCA/product-attribute/tree/16.0/product_catalog_sale ?

yes it is and I took it from odoo-pim to here.

the merged version of it is buggy with me when I tey to use in sales orders as in higher versions.

pedrobaeza:
I don't understand what you mean, but this is working in runboat, and any change you make for sales order should be done in the other module, not the base one.

kobros-tech:
> I don't understand what you mean, but this is working in runboat, and any change you make for sales order should be done in the other module, not the base one.

all right, I to understand how does it work in odoo and I want to review maybe I know only one use case in sales orders but other cases I am not aware of.

thanks

PR #2039, opened 2025-08-12
[18.0][ADD] product_merge

PieterPaulussen:


PieterPaulussen:
Re-tested this module in V18 (after making some changes to its design to support V18). But it proved a useless attempt. Merging products with this module has proved impossible / unstable.

PR #2040, opened 2025-08-13
[16.0][FIX] product_uom_po_domain inheritance with other module qd

quentinDupont:
Related to this issue : https://github.com/OCA/product-attribute/issues/2041


legalsylvain:
hi @quentinDupont. still draft ?

quentinDupont:
> hi @quentinDupont. still draft ?

Yep because I think that's not the proper way of fixing the issue I described, what do you think ? https://github.com/OCA/product-attribute/issues/2041

PR #2042, opened 2025-08-14
[18.0][FIX] product_set: fix product set search behavior

MohamedOsman7:
This PR fixes the user experience when selecting product sets in forms by enhancing the _search_display_name behavior. Product sets are now easier to find through the standard search field.

Before:
Searching for a product set by keywords did not always return relevant results unless "Search more..." was used.
image

After:
The _search_display_name method is now extended to fix matching, making relevant product sets appear directly in the autocomplete suggestions.
image

### DIFF
--- a/product_set/models/product_set.py
+++ b/product_set/models/product_set.py
@@ -32,7 +32,9 @@ class ProductSet(models.Model):
         "it's going to be available for all of them.",
     )
 
-    display_name = fields.Char(compute="_compute_display_name")
+    display_name = fields.Char(
+        compute="_compute_display_name", search="_search_display_name"
+    )
 
     def _inverse_active(self):
         """Set the active field on the set lines."""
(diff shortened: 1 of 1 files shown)


product_set/models/product_set.py
```diff
@@ -52,3 +54,15 @@ def _compute_display_name(self):
...
+    def _search_display_name(self, operator, value):
+        domain = [
+            "|",
+            "|",
+            ("name", operator, value),
+            ("ref", operator, value),
+            ("partner_id.name", operator, value),
+        ]
+
+        return domain
```
pedrobaeza: You can compact the code:

```suggestion
        return [
            "|",
            "|",
            ("name", operator, value),
            ("ref", operator, value),
            ("partner_id.name", operator, value),
        ]
```

pedrobaeza (approved):

CRogos (approved):
LGTM

MohamedOsman7:
@pedrobaeza Could you please merge?

PR #2043, opened 2025-08-17
[18.0][FIX] product_assortment: TypeError:  An incorrect keyword argument (vals_list) was passed 

bosd:
Fixes:
TypeError

An incorrect keyword argument (vals_list) was passed to a superclass's create method.

Fix:
Modify the create method in product_assortment to pass vals_list as a positional argument.


pedrobaeza (approved):
Yes, let's fast-track it.

/ocabot merge patch

PR #2045, opened 2025-08-20
[FIX] product_packaging_level: Allow to input packaging name depending on name policy

rousseldenis:
On product form, the user was not allowed to fill in the packaging name even if the name policy is set on 'user_defined'.

### DIFF
--- a/product_packaging_level/views/product_packaging_view.xml
+++ b/product_packaging_level/views/product_packaging_view.xml
@@ -16,7 +16,7 @@
                 <attribute name="required">barcode_required_for_gtin</attribute>
             </xpath>
             <field name="name" position="attributes">
-                <attribute name="column_invisible">1</attribute>
+                <attribute name="readonly">name_policy != 'user_defined'</attribute>
             </field>
         </field>
     </record>


rousseldenis:
@trisdoan @jbaudoux

product_packaging_level/views/product_packaging_view.xml
```diff
@@ -16,7 +16,7 @@
                 <attribute name="required">barcode_required_for_gtin</attribute>
             </xpath>
             <field name="name" position="attributes">
-                <attribute name="column_invisible">1</attribute>
+                <attribute name="readonly">name_policy != 'user_defined'</attribute>
```
jbaudoux, line 19: it's better to not show the name if it's meaningless and likely not right
```suggestion
                <attribute name="invisible">name_policy != 'user_defined'</attribute>
```

jbaudoux (changes requested):

jbaudoux:
Similar fix but on the form view https://github.com/OCA/product-attribute/pull/2102

jbaudoux:
ping @rousseldenis Can you finalize this?

PR #2060, opened 2025-09-03
[18.0][FIX] product_secondary_unit: don't show inactive UoMs

gdgellatly:
Because of the active test context in the one2many on product, inactive UoM's would show when this is not the intention


gdgellatly:
Test failure seems unrelated and actually occurs after this modules tests

sergio-teruel (approved):
Oops! Really? !!!

kielm (approved):
Functional review LGTM

PR #2067, opened 2025-09-12
[18.0][MIG] product_form_pricelist

bosd:
Standard migration using odoo-module-migrator /ocabot migration product_form_pricelist .


bealdav:
Hi @bosd thanks for this work.

I have noticed this exception in logs:
```bash
odoo.tools.convert.ParseError: while parsing /__w/product-attribute/product-attribute/product_form_pricelist/views/res_config_settings_view.xml:3
Error while parsing or validating view:

Element '<setting id="auth_signup_documents">' cannot be located in parent view

View error context:
Error: Errors detected in log.
{'file': '/__w/product-attribute/product-attribute/product_form_pricelist/views/res_config_settings_view.xml',
 'line': 1,
 'name': 'res.config.settings form',
 'view': ir.ui.view(2411,),
 'view.model': 'res.config.settings',
 'view.parent': ir.ui.view(1921,),
 'xmlid': 'res_config_settings_view_form'}
```

EDIT:

I suppose that the cause is 
```bash
addons/website_sale/views/res_config_settings_views.xml
12:            <setting id="auth_signup_documents" position="replace"/>
```

To avoid this kind of problem you might place your field inside
```xml
<block title="Pricing" id="pricing_setting_container">
```
Could you have a look ?

Thanks a lot

bealdav:
I also noted that there is a `.crush` directory with files inside. Do you know where does it comes from ?
Don't you think we could remove them ?

product_form_pricelist/__manifest__.py
```diff
@@ -0,0 +1,29 @@
+# Copyright 2020 Akretion (https://www.akretion.com).
+# @author Sébastien BEAU <sebastien.beau@akretion.com>
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+{
+    "name": "product_form_pricelist",
```
bealdav:
```suggestion
    "name": "Product Form Pricelist",
```

product_form_pricelist/__manifest__.py
```diff
@@ -0,0 +1,29 @@
...
+        "sale",
+    ],
+    "data": [
+        "security/res_groups.xml",
+        "views/product_template_view.xml",
+        "views/product_view.xml",
+        "views/res_config_settings_view.xml",
+    ],
+    "demo": [],
+    "qweb": [],
```
bealdav:
```suggestion

```

bealdav: no more used key

product_form_pricelist/__manifest__.py
bosd: Great Spot! :+1:

bosd:
> I also noted that there is a `.crush` directory with files inside. Do you know where does it comes from ? Don't you think we could remove them ?

Could be coming from crush cli, which I only tried once, but never used. Will remove it.

bealdav:
Thanks a lot for your reactivity. Only 2 things now:

- auth_signup_documents anchor
- crush files

bealdav (approved):
LGTM, thanks
Not mandatory but I suppose it would be better if you squash the 2 last commits to avoid to have crush files added and removed.
The rule to separate precommit and migration can be escaped for this case, thanks

bealdav:
Hi @bosd if you can deal with  https://github.com/OCA/product-attribute/pull/2067#pullrequestreview-3533477322 I'll manage to get your module merged.

Deal ?

bosd:
@bealdav  Deal :smile: :fireworks:  Can you please merge this one as well? https://github.com/OCA/brand/pull/277

PaulGoubert (approved):

bealdav:
Please @rvalyi could you merge, I deeply checked this PR. Thanks.

bealdav:
@bosd job done by @sebastienbeau

Issue #2068, opened 2025-09-15
[16.0] Compatibility Issue: ValidationError between product_logistics_uom and product_net_weight


xAdrianCif:
## Module
product_net_weight (when used together with product_logistics_uom)

## Describe the bug
When both `product_logistics_uom` and `product_net_weight` modules are installed simultaneously, a ValidationError occurs when trying to save products with custom weight Units of Measure (UoM). The validation incorrectly compares net weight and gross weight values that are stored in different UoM, leading to false validation errors.

**Error message:**
```
ValidationError: The net weight of product must be lower than gross weight.
```

**Root cause:** The validation in `product_net_weight` compares `net_weight` (in grams, as entered by user) with `weight` (converted to kg by `product_logistics_uom`), causing mathematically incorrect comparisons like `800g > 1kg` evaluating as `800 > 1 = True`.

## To Reproduce
**Affected versions**: 16.0

**Prerequisites:**
- Ensure both modules are installed: `product_logistics_uom` and `product_net_weight`

1. Go to runboat v16.0 and verify both modules are installed
2. Navigate to `Settings > Inventory > Units of Measure` and confirm system UoM is set to **kg** (kilograms)
3. Go to `Inventory ` and select any product variant or create a new one
4. In the product form, go to the **Inventory** tab
5. Change the **Weight UoM** field from `kg` to `g` (grams)
6. Set **Product Weight** (gross weight) to `1000` (1000 grams)
7. Set **Net Weight** to `800` (800 grams, which should be valid since 800g < 1000g)
8. Click **Save**

**Expected behavior**
The product should save successfully since the net weight (800g) is correctly lower than the gross weight (1000g).

**Actual behavior**
A ValidationError is raised: "The net weight of product must be lower than gross weight" even though 800g < 1000g is mathematically correct.

**Additional context**
- **Technical details**: The issue occurs because `product_logistics_uom` stores the weight value converted to system UoM (1kg) while `product_net_weight` validation compares the unconverted net_weight value (800g) with the converted weight (1kg), essentially comparing `800 > 1`.
- **Impact**: Users cannot properly configure products with custom weight UoM when both modules are installed
- **Modules interaction**: This is a compatibility issue between two otherwise functional modules
- **OCA Repository**: product-attribute
- **Environment**: The issue can be reproduced in any Odoo 16.0 environment with both modules installed

xAdrianCif:
I have opened this pull request that fixes the encountered issue.
[...]

PR #2069, opened 2025-09-15
[16.0][ADD] Product_logistics_uom_net_weight

xAdrianCif:
**Why we need this module:**
- User sets product UoM to grams, enters 1000g gross weight and 800g net weight
- `product_logistics_uom` stores gross weight as 1kg (converted to system UoM)
- `product_net_weight` compares 800g > 1kg → validation fails incorrectly

**Related issue**:
- https://github.com/OCA/product-attribute/issues/2068

## Solution
Created an integration module that:

✅ **Auto-installs** when both modules are present  
✅ **Adds UoM-aware net weight fields** that handle proper conversions  
✅ **Fixes validation logic** to compare weights in the same UoM  
✅ **Preserves existing data** with a safe migration hook  

## Technical Approach

**New fields:**
- `product_net_weight`: Displays net weight in product's UoM
- Computed fields handle conversion between system and product UoM
- Validation now compares 800g vs 1000g (both in grams) ✓

**Migration hook:**
- Direct copy of existing `net_weight` data (no conversion needed)
- Safe because existing data is guaranteed to be in system UoM
- Much faster than computed field recalculation

**Why it's safe:**
Before this module, it was impossible to have `net_weight` in different UoM than system UoM due to the validation bug itself.

## Result
Users can now:
- Set custom weight UoM per product (grams, kg, etc.)
- Enter net weight values that validate correctly
- Enjoy seamless compatibility between both modules

The integration is completely transparent - it just works when both modules are installed.


xAdrianCif:
Pinging the maintainers of both modules, @hparfr and @legalsylvain , in case you have any input or just for visibility.

legalsylvain:
What a huge glue module ! 

I'm not using product_logistics_uom module. I let @hparfr validate this PR.

thanks !

OscarIndias:
LGTM!

OscarIndias (approved):

hparfr (commented):
Did you try to remove the precision on net_weight ?

xAdrianCif:
> Did you try to remove the precision on net_weight ?

Hi @hparfr 

I have tried different strategies, but none of them worked, nothing related to precisions either, because the issue is that the field shown on the product form is product_weight, which is always converted to the system’s base UoM before being stored in the weight field. This weight field is the one used for the comparison performed in the product_net_weight module. That module doesn’t know about the product_weight field, so there’s no built-in way to reconcile this problem.

Maybe I’m not understanding your approach. If you’ve tried something that works and doesn’t trigger the error I mentioned in the issue, I’d be happy to read about it!

Thanks.

imoralesc (approved):

AdrianaSaiz (approved):
LGTM

Good solution for the problem

hparfr:
I think precision should be set to `False` from net_weight as it's done for weight or volume in product_logistics_uom.

```py
    # remove rounding from volume and weight
    # this is needed to avoid rounding errors when converting between units
    # and is safe since we display the volume and weight in the product's
    # volume and weight UOM. In the same time, we need to keep the volume
    # we ensure that no information is lost by storing the volume and weight
    # without rounding.
    volume = fields.Float(digits=False)
    weight = fields.Float(digits=False)
```

product_logistics_uom_net_weight/views/product_product.xml
```diff
@@ -0,0 +1,36 @@
...
+
+    <record id="view_product_product_form" model="ir.ui.view">
+        <field name="model">product.product</field>
+        <field name="inherit_id" ref="product_net_weight.view_product_product_form" />
+        <field name="priority">99</field>
+        <field name="arch" type="xml">
+            <label for="net_weight" position="replace">
+                <label for="product_net_weight" />
+            </label>
+            <field name="net_weight" position="replace">
```
hparfr: Possible to hide net_weight  instead of `replace`  ?

hparfr (changes requested):

xAdrianCif:
Hi @hparfr !

First of all, thank you very much for the review and the suggestions.

I realized that I didn’t fully understand your first message about digits=False. I’ve checked the original module and applied the changes you suggested to that field.

On the other hand, I completely agree about hiding the fields instead of replacing them. I’ve also done the same with the labels, since they might be referenced in downstream modules and I want to avoid breaking any inheritance.

I’ve uploaded all the changes. Any additional suggestions are welcome, thanks!

hparfr (approved):
Thanks

/ocabot merge nobump

xAdrianCif:
Hi @pedrobaeza, could you please help with merging this addon?
Additionally, should I add someone else as a maintainer of the module?
Thank you very much!

PR #2089, opened 2025-10-01
[19.0][MIG] product_dimension: Migration to 19.0.

hieulucky111:


product_dimension/models/product_product.py
```diff
@@ -0,0 +1,37 @@
...
+    _inherit = "product.product"
+
+    product_length = fields.Float("length")
+    product_height = fields.Float("height")
+    product_width = fields.Float("width")
+    dimensional_uom_id = fields.Many2one(
+        "uom.uom",
+        "Dimensional UoM",
+        help="UoM for length, height, width",
+        default=lambda self: self.env.ref("uom.product_uom_meter"),
```
ferran-S73: @hieulucky111  I think this default isn't working as intended since it's overriding the unit I choose when creating a product manually. I'm unsure if this is the case for previous versions though.

Also, this is more of a personal doubt. Since UoM categories no longer exist in v19 should we try and filter the available uoms for this field somehow? Odoo's approach in some settings fields is just to hardcode a couple of UoM for each category though I don't quite like it.
https://github.com/odoo/odoo/blob/9900375bc0bac2754150fd8cd6a1e45ecad8da83/addons/product/models/res_config_settings.py#L14

hieulucky111: Hello @ferran-S73 
I’ve rechecked the behavior in both versions 18 and 19. When creating a new Product Template and selecting a Dimension UoM, the selected UoM disappears after saving. (Same issue in both versions)

Regarding the UoM filter, we could use a System Parameter to store the list of available UoM names. However, I don’t think that’s necessary for now.

Vang-NguyenPhu (approved):
LGTM

Reyes4711-S73 (approved):
LGTM

Reyes4711-S73:
@OCA/product-maintainers Please, can you merge this PR?

ferran-S73 (approved):

product_dimension/models/product_product.py
StefanRijnhart: It's likely that this issue can be solved by aligning with how Odoo does this for `weight`, by storing the UoM on the product template (https://github.com/odoo/odoo/blob/27b3d2fc055c586f4d53c92f5a7a93301d7d7d17/addons/product/models/product_template.py#L110-L113).

mav-adhoc:
LGTM

BhaveshHeliconia (approved):
LGTM!

JordiMForgeFlow:
@rousseldenis could you merge this one? :)

StefanRijnhart:
Is there no interest in fixing https://github.com/OCA/product-attribute/pull/2089#pullrequestreview-3343355783?

Camille0907 (approved):
Migration LGTM, thank you 🙂

Camille0907:
@OCA/product-maintainers Hello, is there anything missing, before we can get the PR merged ? :slightly_smiling_face:

product_dimension/models/product_product.py
hieulucky111: hello @StefanRijnhart, do you see this as a critical blocker, or can we treat it as a future enhancement?
Personally, I think we can address this in a later.

StefanRijnhart: I'd say it's pretty serious if this module provides a field that is not stored when creating a new record even if the user has selected a value for it. The product.template model is now more prominent than back when this (old) module was first introduced, so the impact of the bug has gotten bigger even if the bug has existed in past versions.

hieulucky111: Ok, i will improve this soon.

hieulucky111: hello @StefanRijnhart, i have applied your suggestion. I also move the default UoM from product.product to product.template. (Because variant creation will set again the value of product template).
Thank you.

StefanRijnhart: Thanks for the update! Checking runboat, I still seem to be able to reproduce the original issue. Did it work for you while testing locally?

hieulucky111: Yes, it works for me before pushing the updated code. Maybe we need to upgrade again module?

Jortolsa-S73 (approved):

JordiMForgeFlow:
@rousseldenis could this one be merged?

product_dimension/models/product_product.py
StefanRijnhart: Weird. The failure to store the dimensional uom when creating product templates is easily reproced in a test, it turns out. The solution is also easy: apparently, a similar issue occurs with the other fields in this module, and this is fixed with the override on `_prepare_variant_values`. Fix proposed onto your branch in https://github.com/komit-consulting/product-attribute/pull/1. You can squash it into your own commit.

StefanRijnhart:
@JordiMForgeFlow Let's wait until https://github.com/OCA/product-attribute/pull/2089#pullrequestreview-3343355783 is fixed. We found the solution now.

product_dimension/models/product_product.py
hieulucky111: Hello @StefanRijnhart, thanks for your fixed.
I have merged your fixed and squashed the commits

StefanRijnhart (approved):
OK, let's go!

/ocabot merge nobump

PR #2095, opened 2025-10-02
[14.0][IMP] product_pricelist_revision: add rounding_digits field to …

mathieudelva:
…allow rounding the "fixed_price" with the

 number of digits corresponding to the value of the new field


product_pricelist_revision/wizards/pricelist_duplicate_wizard.py
```diff
@@ -13,6 +13,7 @@ class ProductPricelistItemDuplicateWizard(models.TransientModel):
     date_start = fields.Datetime(required=True)
     date_end = fields.Datetime()
     variation_percent = fields.Float(digits="Product Price", string="Variation %")
+    rounding_digits = fields.Integer()
```
Kev-Roche: It will be nice to have the product.decimal_price as default value here.

Kev-Roche (changes requested):

rousseldenis (approved):

PR #2109, opened 2025-10-22
[16.0][IMP] product_lot_sequence: product sequence exceptions

chienandalu:
There might be cases where we don't want to assign specific sequences to our
product and keep the core behavior. For that we've created a flag that allows
to do so.

- The flag *Use specific sequence* is on by default.
- Set it off and the product will behave as in Odoo core.
- Once you do so, you can remove the product's lot sequence as well.

(TODO: remove this vídeo, as it's an old version of the change)
https://www.loom.com/share/4edde17a26ef4d2f9d909e59ea538f7b?sid=86bbaefb-926b-46a8-982f-c9b1a800e364

cc @moduon MT-12305

please review @rafaelbn @fcvalgar @Shide

### DIFF
--- a/product_lot_sequence/models/product.py
+++ b/product_lot_sequence/models/product.py
@@ -8,6 +8,10 @@ from odoo import api, fields, models
 class ProductTemplate(models.Model):
     _inherit = "product.template"
 
+    use_specific_lot_sequence = fields.Boolean(
+        help="Use an specific lot sequence for this product",
+        default=True,
+    )
     lot_sequence_id = fields.Many2one(
         "ir.sequence",
         string="Entry Sequence",
@@ -33,11 +37,15 @@ class ProductTemplate(models.Model):
         compute="_compute_display_lot_sequence_fields"
     )
 
-    @api.depends("tracking")  # For products being created (before saved).
+    @api.depends("tracking", "use_specific_lot_sequence")
     def _compute_display_lot_sequence_fields(self):
-        self.display_lot_sequence_fields = (
+        product_sequence_policy = (
             self.env["stock.lot"]._get_sequence_policy() == "product"
         )
+        for product in self:
+            product.display_lot_sequence_fields = (
+                product_sequence_policy and product.use_specific_lot_sequence
+            )
 
     @api.model
     def _create_lot_sequence(self, vals):
@@ -59,6 +67,18 @@ class ProductTemplate(models.Model):
             "lot_sequence_number_next", 1
         )
         return seq
+
+    def create_lot_sequence(self):
+        self.ensure_one()
+        if self.lot_sequence_id:
+            return
+        self.lot_sequence_id = self._create_lot_sequence({})
+
+    def remove_lot_sequence(self):
+        self.ensure_one()
+        if not self.lot_sequence_id:
+            return
+        self.lot_sequence_id.unlink()
 
     # do not depend on 'lot_sequence_id.date_range_ids', because
     # lot_sequence_id._get_current_sequence() may invalidate it!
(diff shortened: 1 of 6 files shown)


product_lot_sequence/models/product.py
```diff
@@ -8,6 +8,9 @@
 class ProductTemplate(models.Model):
     _inherit = "product.template"
 
+    disallow_lot_sequence = fields.Boolean(
+        help="Don't use an specific lot sequence for this product"
+    )
```
Shide: Would be great to have this variable inversed in meaning with a migration script.
Seems more readable and less confusing

```suggestion
    use_specific_lot_sequence = fields.Boolean(
        help="Use an specific lot sequence for this product",
        default=True,
    )
```

yajo (changes requested):
IMHO the fact that checking (or unchecking, as per https://github.com/OCA/product-attribute/pull/2109/files#r2451880343) a checkbox removes or creates a record is a bit surprising.

I think that we can do a better UI: 

Image

When removing, add a `confirm="Are you sure?"` or similar to the button.

This way, the effect and usability is nearly the same, but we avoid surprises.

product_lot_sequence/models/product.py
chienandalu: Done

chienandalu:
@Shide @yajo changes done. Take a look

product_lot_sequence/models/product.py
```diff
@@ -8,6 +8,10 @@
 class ProductTemplate(models.Model):
     _inherit = "product.template"
 
+    use_specific_lot_sequence = fields.Boolean(
+        help="Use an specific lot sequence for this product",
+        default=True,
+    )
```
yajo, line 14: This field can be removed now, and all other uses of it.

Since the creation or removal is done by buttons, then the presence or absence of a sequence is enough to make the module work.

product_lot_sequence/models/product.py
```diff
@@ -33,11 +37,15 @@ class ProductTemplate(models.Model):
...
+    @api.depends("tracking", "use_specific_lot_sequence")
     def _compute_display_lot_sequence_fields(self):
-        self.display_lot_sequence_fields = (
+        product_sequence_policy = (
             self.env["stock.lot"]._get_sequence_policy() == "product"
         )
+        for product in self:
+            product.display_lot_sequence_fields = (
+                product_sequence_policy and product.use_specific_lot_sequence
+            )
```
yajo, line 48: This method should start by checking that if `tracking == "none"`, then `product.display_lot_sequence_fields = False`.

Otherwise, just use the default value from config, since the checkbox should be removed.

product_lot_sequence/models/product.py
```diff
@@ -33,11 +37,15 @@ class ProductTemplate(models.Model):
...
 
     @api.model
```
yajo, line 50:
```suggestion
```

fcvalgar (approved):
Functional Review,

LGTM, thank you @chienandalu

yajo (changes requested):
When creating a new product, I think this is inconsistent:

Image

Then, if you do this:

1. Select "by lots"
2. Create sequence
3. Uncheck "use specific lot sequence".

You get another inconsistency:

Image

I think the problem is that the check box is now unnecessary. See comments below. If we remove the checkbox, it's gonna be less code and simpler UX.

Thanks!

product_lot_sequence/models/product.py
chienandalu: No, because there's an automation that creates the sequence on every write if it isn't one and we want to avoid that (and I don't want to change that behavior that might affect other users of the module)

yajo: Even in that case, the behavior is inconsistent if we have 2 ways to do the same.

However, if that's the goal, there's an easier way to achieve it. Just add support for a new system parameter that disables the automatic creation of sequences, so if the user wants one, they just have to click on the create button.

rafaelbn:
WIP

PR #2112, opened 2025-10-23
[16.0][FIX] product_assortment: All assortments are applied to original partner when partner is duplicated

sergiobstoj:
FW from https://github.com/OCA/product-attribute/pull/1563


danielduqma (approved):

pedrobaeza (approved):
/ocabot merge patch

Please forward port it to upper versions cc @carlosdauden

PR #2114, opened 2025-10-24
[18.0][FIX] product_sequence: test disable

eduezerouali-tecnativa:
Since odoo 16.0 the template is copied instead and its first variant is returned.
https://github.com/OCA/OCB/blob/9fe44f03ba29e98e4bbc103983125e1393e45358/addons/product/models/product_product.py#L460

So that part should be move to `product.template`  and that is the reason why test were failing in the repository.

This should be backported?

Please could you review @juancarlosonate-tecnativa @pedrobaeza @victoralmau

### DIFF
--- a/product_sequence/tests/test_product_sequence.py
+++ b/product_sequence/tests/test_product_sequence.py
@@ -34,7 +34,7 @@ class TestProductSequence(TransactionCase):
         product_1 = self.product_product.create(dict(name="Orange", default_code="/"))
         self.assertRegex(str(product_1.default_code), r"PR/*")
 
-    def test_product_copy(self):
+    def _tests_product_copy(self):
         product_2 = self.product_template.create(
             dict(name="Apple", default_code="PROD02")
         )
@@ -127,7 +127,7 @@ class TestProductSequence(TransactionCase):
         self.assertEqual(product_claudia.default_code[:3], "PAR")
         self.assertEqual(product_claudia.product_tmpl_id.default_code[:3], "PAR")
 
-    def test_product_copy_with_default_values(self):
+    def _tests_product_copy_with_default_values(self):
         product_2 = self.product_template.create(
             dict(name="Apple", default_code="PROD02")
         )


pedrobaeza (commented):
`copy_data` should be implemented at `product.product` level, not `product.template`.

juancarlosonate-tecnativa:
I think the issue is not related to the copy, the problem lies in the create. If you look at the test error, when the product is created with the default_code, it’s set as PROD02 in the test, but during the copy, the sequence uses /, completely ignoring the product’s original default_code. Because of that, it treats / as the default value and generates the next sequence, which causes the copied product not to match the expected default_code or the copy logic. I believe the default_code from the template should be taken and assigned as the value to make it work correctly.

eduezerouali-tecnativa:
@pedrobaeza are you 100% sure to change `copy` to `copy_data` ? https://github.com/odoo/odoo/pull/154132/files#diff-35eeb9b2264b76eca30fcb0a923757dde3ca14c636ce6faf66ef31a6b649799bL443

I do agree with @juancarlosonate-tecnativa. I found that the problem is when it tries to generate the first variant when coping the template.

pedrobaeza:
OK, I guessed that the problem was that the method was not intervening on the proper place, but the aim is to fix the tests, so go ahead with any proposal to fix them.

pedrobaeza:
@eduezerouali-tecnativa please finish this to get green CI. If not able to fix it, just disable current tests.

eduezerouali-tecnativa:
ping @pedrobaeza

product_sequence/models/product_product.py
```diff
@@ -38,6 +38,7 @@ def create(self, vals_list):
             else:
                 vals_list_updated.append(vals)
         res = super().create(vals_list_updated)
+
```
pedrobaeza: Remove this extra line.

pedrobaeza (commented):
Indicate in the extended description of the commit message that they are making the branch to fail, and no easy solution has been found, so they are disables for now.

pedrobaeza (approved):
/ocabot merge patch

PR #2118, opened 2025-10-29
[18.0][FIX] product_pricelist_direct_print: show only defined products is not showing the correct products

CarlosRoca13:
The ANDs on the domains is excluding products that are defined but aren't on the categories added to the pricelist, so this is not correct. Set ORs to wor as is was working in last versions.

cc @Tecnativa TT58949

ping @pedrobaeza @carlosdauden

### DIFF
--- a/product_pricelist_direct_print/wizards/product_pricelist_print.py
+++ b/product_pricelist_direct_print/wizards/product_pricelist_print.py
@@ -312,38 +312,38 @@ class ProductPricelistPrint(models.TransientModel):
     def get_products_domain(self):
         domain = [("sale_ok", "=", True)]
         if self.show_only_defined_products:
-            aux_domain = []
+            aux_domain = [(0, "=", 1)]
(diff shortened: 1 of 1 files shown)


pedrobaeza (approved):
Indeed: https://github.com/OCA/product-attribute/blob/970fb4799f239f0c46a2fabb72f0b5d882e559ee/product_pricelist_direct_print/wizards/product_pricelist_print.py#L333

legalsylvain (approved):

legalsylvain:
Could you add a test ?

pedrobaeza:
You can rebase for green CI.

CarlosRoca13:
There were already tests for this feature. Since the tests were failing, they changed the ORs to ANDs in the migration so that the test wouldn’t fail, instead of addressing the root cause of the problem.

CarlosRoca13:
Done @pedrobaeza

PR #2121, opened 2025-10-31
[18.0][IMP] product_cost_security : add explaination in the README about th…

cvinh:
…e behavior of this module when we use Point of Sale


cvinh:
misunderstood how pos_product_cost_security worked in v16

PR #2126, opened 2025-11-06
[18.0][ADD] product_download_feed: New module

eduezerouali-tecnativa:
cc @Tecnativa TT58933
ping @sergio-teruel @carlosdauden @juancarlosonate-tecnativa 

This module provides an HTTP endpoint that allows authenticated Odoo users to export the product catalog as a CSV file.
The export automatically applies the user’s assigned pricelist, ensuring that product prices in the CSV match those shown in sales quotations or the customer portal.


pedrobaeza:
Please check CI.

The word `catalog` is used by Odoo for other thing (the product picker), so I don't think it should be used here.

eduezerouali-tecnativa:
> Please check CI.
> 
> The word `catalog` is used by Odoo for other thing (the product picker), so I don't think it should be used here.

And what about `feed`, i think it is more accurate than catalog. What do you think?

pedrobaeza:
It seems good. `product_download_feed` is a good candidate. @sergio-teruel @carlosdauden do you have any other idea?

eduezerouali-tecnativa:
ping @carlosdauden @sergio-teruel

AungKoKoLin1997 (commented):
I think the README should include a `Usage` section. At the moment, I had to read the code to find the export endpoint and understand the available parameters. 

From the implementation, it seems the export is intended to be triggered by calling the endpoint directly. The controller already supports selecting a pricelist and choosing exported fields via query parameters, but I don’t see any UI entry point (e.g., a wizard) for functional users to configure and download the CSV.

In my opinion, a simple export wizard (pricelist + fields) would make this easier to use from the UI. Am I missing anything regarding the intended usage of this module?

pedrobaeza:
@AungKoKoLin1997 in fact, this doesn't have any UI, because it provides a feed (endpoint) for machines to get that information. Any proposal to improve the documentation?

PR #2127, opened 2025-11-06
[19.0][MIG] product_sequence: Migration to 19.0

fd-oerp:


fd-oerp:
@rousseldenis Could you please take this PR into consideration and let me know if there’s anything that needs to be addressed?
I noticed that two checks are failing, but I’m unable to identify the exact cause.

Thank you

bosd (approved):
Thanks for your contribution. LGTM :+1:

product_sequence/tests/test_product_sequence.py
```diff
@@ -0,0 +1,132 @@
...
+        product_1 = self.product_product.create(dict(name="Orange", default_code="/"))
+        self.assertRegex(str(product_1.default_code), r"PR/*")
+
+    def _tests_product_copy(self):
+        product_2 = self.product_template.create(
+            dict(name="Apple", default_code="PROD02")
+        )
+        product_2.flush_recordset()
+        copy_product_2 = product_2.product_variant_id.copy()
+        self.assertEqual(copy_product_2.default_code, "PROD02-copy")
```
NICO-SOLUTIONS, line 43: It seems like the previous test assumed that copying a product would always set default_code to "<original_code>-copy", which caused test failures.  Odoo actually generates a new default_code using a sequence.

```suggestion
    def tests_product_copy(self):
        product_2 = self.product_template.create(
            dict(name="Apple", default_code="PROD02")
        )
        product_2.flush_recordset()
        copy_product_2 = product_2.product_variant_id.copy()
        self.assertTrue(copy_product_2.default_code, "Copy should have a default_code")
        self.assertNotEqual(
            copy_product_2.default_code,
            product_2.default_code,
            "Copy should not reuse original default_code",
        )
```

product_sequence/tests/test_product_sequence.py
```diff
@@ -0,0 +1,132 @@
...
+
+    def _tests_product_copy_with_default_values(self):
+        product_2 = self.product_template.create(
+            dict(name="Apple", default_code="PROD02")
+        )
+        product_2.flush_recordset()
+        copy_product_2 = product_2.product_variant_id.copy(
+            {"default_code": "product test sequence"}
+        )
+        self.assertEqual(copy_product_2.default_code, "product test sequence")
```
NICO-SOLUTIONS, line 132: same here...

```suggestion
    def tests_product_copy_with_default_values(self):
        product_2 = self.product_template.create(
            dict(name="Apple", default_code="PROD02")
        )
        product_2.flush_recordset()
        self.assertTrue(copy_product_2.default_code, "Copy should have a default_code")
        self.assertNotEqual(
            copy_product_2.default_code,
            product_2.default_code,
            "Copy should not reuse original default_code",
        )
```

NICO-SOLUTIONS (changes requested):
Just a quick look at the tests:
With these changes, the tests now check for the guaranteed behavior in Odoo (that a new default_code is generated when copying a product) rather than assuming a fixed -copy suffix.
This makes the tests stable and they will no longer fail due to sequence-generated codes.

product_sequence/tests/test_product_sequence.py
fd-oerp: @NICO-SOLUTIONS  Should I update this test case to verify that the copied product generates a new sequence, or is there a better way to improve this test?

fd-oerp:
> Just a quick look at the tests: With these changes, the tests now check for the guaranteed behavior in Odoo (that a new default_code is generated when copying a product) rather than assuming a fixed -copy suffix. This makes the tests stable and they will no longer fail due to sequence-generated codes.

@NICO-SOLUTIONS , thank you for the review. Should I update these test case to verify that the copied product generates a new sequence, or is there a better way to improve this test?

NICO-SOLUTIONS:
> > Just a quick look at the tests: With these changes, the tests now check for the guaranteed behavior in Odoo (that a new default_code is generated when copying a product) rather than assuming a fixed -copy suffix. This makes the tests stable and they will no longer fail due to sequence-generated codes.
> 
> @NICO-SOLUTIONS , thank you for the review. Should I update these test case to verify that the copied product generates a new sequence, or is there a better way to improve this test?

Hey @fd-oerp … uffff… quite a while ago… I’m trying to remember...

I think the updated tests focus on the behavior that really matters: a copied product has a default_code and doesn’t reuse the original one, and they could be a reasonable approach. But I can’t exclude that I might be wrong on some details — these suggestions came mostly from the gut, from a quick look.
For the test with a specific default_code, we could check that the value is exactly copied, but I think it’s better not to tie the tests too closely to how Odoo generates or handles the default_code, since that can vary across modules.

These changes are just suggestions to make the tests more robust while still covering the intended behavior — and I’m always open to discussing other approaches, as that can only improve them!

fd-oerp:
> > > Just a quick look at the tests: With these changes, the tests now check for the guaranteed behavior in Odoo (that a new default_code is generated when copying a product) rather than assuming a fixed -copy suffix. This makes the tests stable and they will no longer fail due to sequence-generated codes.
> > 
> > 
> > @NICO-SOLUTIONS , thank you for the review. Should I update these test case to verify that the copied product generates a new sequence, or is there a better way to improve this test?
> 
> Hey @fd-oerp … uffff… quite a while ago… I’m trying to remember...
> 
> I think the updated tests focus on the behavior that really matters: a copied product has a default_code and doesn’t reuse the original one, and they could be a reasonable approach. But I can’t exclude that I might be wrong on some details — these suggestions came mostly from the gut, from a quick look. For the test with a specific default_code, we could check that the value is exactly copied, but I think it’s better not to tie the tests too closely to how Odoo generates or handles the default_code, since that can vary across modules.
> 
> These changes are just suggestions to make the tests more robust while still covering the intended behavior — and I’m always open to discussing other approaches, as that can only improve them!

@NICO-SOLUTIONS , Apologies for the long delay on this.

I believe that if the test case should include the copied record scenario, we should update it to verify that the next sequence/default_code is generated correctly from the source record. Alternatively, we could completely eliminate the copied product scenario from the test.

What do you think?

NICO-SOLUTIONS:
@fd-oerp 
Don’t worry about the delay – there’s no need to apologize! I just need to dig into my memory a bit 😄. I really appreciate your time contributing and reviewing PRs, and it’s great when we help each other improve the tests. Ideally, this can help strengthen collaboration for future reviews and progress.

In general, I think keeping the copied product scenario but only checking that the copied product has a default_code and doesn’t reuse the original one is sufficient. I’ll take another close look at the module in the coming days to make sure I am not missing any important points.

AEstLo (approved):
LGTM - code review

Vicent-S73:
LGTM

NICO-SOLUTIONS:
@fd-oerp 
Hey, sorry for getting back so late.

From my point of view, updating the tests to not rely solely on a fixed "-copy" assumption makes them more robust, since default_code can now be generated either via a category sequence or with a "-copy" suffix depending on the scenario. Both cases should be covered in the tests.

I think the current copy() method works for simple cases with manually set default_code, but it might behave inconsistently when sequences are involved. We could keep it as-is for manual codes, or consider adjusting the copy() method to use the sequence for copied products as well, to ensure consistent default_code values.

This is just my impression — I haven’t tested it in detail yet.

Additional note: Apart from the approach taken, to get all checks green and improve coverage, we would need more test cases covering all paths: sequence-generated codes, -copy suffix for manual codes, and explicitly passed default_code values.

stferraro:
There is an unexpected behavior in the current implementation. According to the module documentation, the product internal reference should be generated based on the category prefix (e.g., lamp0001 for the "lamp" category). However, in some cases, the system falls back to the default product sequence (e.g., PR0008) instead of using the category-based sequence.

This behavior does not align with the expected functionality described in the module documentation.

NICO-SOLUTIONS:
@stferraro 
Could you please clarify in which specific cases this behavior occurs? It would be important to understand when and why the default PR sequence is used instead of the category-based sequence.

From the implementation, the PR sequence is used whenever no category-based sequence can be resolved. This happens if:
- no category is set,
- the category has no sequence_id,
- no parent category has a sequence_id (when parent lookup is enabled),
- or parent lookup is disabled and the direct category has no sequence_id.

In all these cases, the system falls back to the default PR sequence as intended.

stferraro:
@NICO-SOLUTIONS Apologies for the confusion. I initially didn’t consider the configuration option that controls whether parent category prefixes are used. This likely explains the fallback to the default PR sequence. I’ll recheck the configuration and validate the behavior accordingly.

product_sequence/tests/test_product_sequence.py
FrancoMaxime: Hi @fd-oerp, thanks for the PR. I think you can just update the test case. In fact it just verifies that the default code is not the same and that it is set.

FrancoMaxime (commented):
LGTM: code review

stferraro (approved):
lgtm !

Reyes4711-S73 (approved):
LGTM

Karla-Villegas (approved):
LGTM!

BhaveshHeliconia (approved):
Code review LGTM!

Yusuke1998 (approved):
LGTM!

bosd:
some commits should be squashed:

image


image

stferraro:
Hi @NICO-SOLUTIONS check and approved please !

NICO-SOLUTIONS:
@stferraro 

Functionally, it looks good to me. Regarding the tests, I’ve already shared my thoughts earlier and would aim for proper code coverage. There are already sufficient approvals, so the final decision is up to the maintainers.

Additionally, I agree with @bosd  that the commits should be squashed before merging.

stferraro:
@fd-oerp please squash commits

fd-oerp:
> @fd-oerp please squash commits

@stferraro , The last two commits made by me have been squashed.

stferraro:
@NICO-SOLUTIONS The last two commits made by me have been squashed.

NICO-SOLUTIONS:
@fd-oerp 
@stferraro 
Squash merging was not applied correctly. To keep a clean commit history, it should look like this instead:

BhaveshHeliconia:
explained here: https://github.com/OCA/maintainer-tools/wiki/Merge-commits-in-pull-requests#mergesquash-the-commits-generated-by-bots-or-weblate

stferraro:
Hi, @fd-oerp apply @NICO-SOLUTIONS and @BhaveshHeliconia suggestions please!

fd-oerp:
> @fd-oerp @stferraro Squash merging was not applied correctly. To keep a clean commit history, it should look like this instead: 

@NICO-SOLUTIONS Thanks!, It is done!

fd-oerp:
> Hi, @fd-oerp apply @NICO-SOLUTIONS and @BhaveshHeliconia suggestions please!

Done! @stferraro

stferraro (approved):
LGTM!

BhaveshHeliconia (changes requested):
@fd-oerp I can still see some bot/Weblate commits that need to be squashed according to the contribution guidelines.

NICO-SOLUTIONS:
@fd-oerp 
Hey,
Now the commits that should have been squashed are no longer available.
When squashing commits, the commits being merged should still be represented within the target commit that acts as the container. After performing the squash, the target commit should contain the combined changes from all squashed commits.

If you're using PyCharm, you can use the “Interactively Rebase from Here...” feature, which is quite helpful for this workflow. The result should look something like this.


image


image


image


In general, I think it should be sufficient to squash commits down to the last migration commit. If there are older commits before that which should have been squashed, they should have been handled by the person who created the migration to the appropriate version.

If you're up for it, it would also be fine to squash everything down to the very first commit — but from my point of view, this is not a strict requirement for your PR (just my opinion; others may see it differently).

Hope this helps!

raulovallet:
do you need any help here?

bosd:
Can we move forward with this one?
Let's get the approved label...

@BhaveshHeliconia I se eyour last comment LGTM, but also that gh is blocking this as requested changes by you. So can you check, is this one ok for you now?

@NICO-SOLUTIONS Your'e the other blocker. Is this one ok for you as well??

BhaveshHeliconia (approved):
Yes, Squashing the administrative commits is not a blocking issue, so we can proceed.

NICO-SOLUTIONS:
@bosd 
I'm not blocking this PR. There are already enough reviews for that. I just wanted to make sure the commit history is preserved properly, as squashing should retain the full development history for documentation purposes. In this case, some commits seem to have been lost. If this PR is merged as it stands, it will be much harder to recover them.

That being said, it's not my decision whether to merge it in its current state. I just wanted to raise this concern so it can be considered.

BhaveshHeliconia:
@bosd you can supresed this PR if you want, as the original author no longer appears to be active.

bosd:
Hi @fd-oerp, I have opened #2369 which supersedes this PR.

It redoes the migration from the current `18.0` with `git format-patch | git am --keep`, squashing only *runs of consecutive administrative commits* instead of folding them into the functional ones, so every `[MIG]`/`[IMP]` commit keeps its own author and message. Your 19.0 migration commit is cherry-picked there with `-x` and you are credited in `readme/CONTRIBUTORS.md` 

the history rework simply could not be expressed as a review suggestion here. It also adds the new OCA icon and demo data.

Feel free to review #2369, and to close this one if you agree.

PR #2133, opened 2025-11-13
[18.0][ADD] product_packaging_template: New module. Allow to configure packaging at product.template level

marielejeune:
Product packaging are defined at product variant level. It means that if
a product template has 25 variants, and all of them are sold in the same
packaging, one must define all the product.packaging one by one on the
variants.

The aim of this module is to be able to define the product.packaging at
template level. It will then create the product.packaging on all
variants automatically.

PR #2135, opened 2025-11-13
[18.0][ADD] product_packaging_template_purchase: New module

marielejeune:
Depends on https://github.com/OCA/product-attribute/pull/2133

This module extends the product_packaging_template that adds
packaging on template level.

The goal of this module is being able to set the "Purchase" configuration
(determining if a packaging is usable in purchase orders) on the packaging
template level.


acsonefho (approved):
Code LGTM :muscle:

PR #2150, opened 2025-12-02
[18.0][MIG] base product mass addition

bealdav:
This module is not needed anymore because you can use Catalog view from one2many containing product_id field and Catalog widget


base_product_mass_addition/tests/test_product_mass_addition.py
```diff
@@ -0,0 +1,92 @@
...
+        """Using quick magic fields shouldn't write on product's metadata"""
+        # Monkey patch the now method for our testing purpose
+        # other the now method of cr would return the date of the Transaction
+        # which is unique for the whole test
+        self.env.cr.now = now
+        base_date = self.product.write_date
+        # Case 1: Updating qty_to_process shouldn't write on products
+        self.product.qty_to_process = 4.0
+        self.env["product.product"].flush_model()
+        self.assertEqual(base_date, self.product.write_date)
```
bealdav, line 79: Hi @ivantodorovich I can't understand how to make it works.
I tried some things, by no way.
Do you think you can have look here ?
I'm not familiar with cache management even if I dived in for the circumstance.

Thanks a lot.

ivantodorovich: Hello @bealdav !

Ouch, this was a nightmare..  😓 
Essentially this is what's implemented in the `modified` override, and what's described in the comment there:

```
        # OVERRIDE to supress LOG_ACCESS_COLUMNS writes if we're only writing on quick
        # magic fields, as they could lead to concurrency issues.
        #
        # Moreover, from a functional perspective, these magic fields aren't really
        # modifying the product's data so it doesn't make sense to update its metadata.
        #
        # Basically, if all we're modifying are quick magic fields, and we don't have
        # any other column to flush besides the LOG_ACCESS_COLUMNS, clear it.
```

I don't see any immediate reason why this would stop working 18.0.

Since this is low-level, I would also doubt the unit tests, so maybe perform a manual test: adding a product from the sale quick view, and verifying that the product's write_date didn't change.

You can also set a breakpoint in the modified override and inspect if there other fnames besides the `qty_to_process`, `quick_uom_id` or any of `LOG_ACCESS_COLUMNS` when using the quick addition. Normally there shouldn't be any

bealdav: Hi @ivantodorovich thanks for your answer.

I 've done these checks. I really understand it should be a nightmare ;-)

I ask myself if can only remove `write_date` from vals ,overriding `write` when no other fields than  qty_to_process, quick_uom_id or any of LOG_ACCESS_COLUMNS are present ?
Do you approve this strategy ?

ivantodorovich: 🤔 I'm not sure, cause these LOG_ACCESS_COLUMNS are also set at the very last step, during `_write`: https://github.com/odoo/odoo/blob/39b969de4cae74a3298b415f6cad9693922b3386/odoo/models.py#L4875

So even if you're able to catch the ones set by `write` (https://github.com/odoo/odoo/blob/39b969de4cae74a3298b415f6cad9693922b3386/odoo/models.py#L4730-L4732), it's likely not enough

---

Maybe a completely different approach would be to implement the sale order update mechanism in a write override that doesn't call super when only these fields are written -- then we wouldn't need to deal with cache or anything

rousseldenis:
@bealdav What's the status of this ?

bealdav:
I'll close this PR because feature is already ensure by Catalog view from product lines in o2m.

Moreover this module is difficult to maintain because it depends on lower orm layer.

What do you think ?  Please I would like your opinion here @rousseldenis @ivantodorovich

ivantodorovich:
Indeed, I think this module is not needed anymore

PR #2151, opened 2025-12-03
[18.0] [MIG] product_supplierinfo_group

bealdav:
Migration oca-port from 14.0

- [x] fix conflict with other modules
- [x] adapt behavior with new field on 18, such uom, discount
- [x] include https://github.com/OCA/product-attribute/pull/1914
- [X] include https://github.com/OCA/product-attribute/pull/1334

## Before
before

## After
after_1

after_2


product_supplierinfo_group/models/product_supplierinfo.py
```diff
@@ -22,17 +22,29 @@ class ProductSupplierinfo(models.Model):
...
     )
-    company_id = fields.Many2one(related="group_id.company_id", store=True)
-    product_tmpl_id = fields.Many2one(related="group_id.product_tmpl_id", store=True)
-    name = fields.Many2one(related="group_id.partner_id", store=True, required=False)
-    product_id = fields.Many2one(related="group_id.product_id", store=True)
-    product_name = fields.Char(related="group_id.product_name", store=True)
-    product_code = fields.Char(related="group_id.product_code", store=True)
-    sequence = fields.Integer(related="group_id.sequence", store=True)
+    company_id = fields.Many2one(
+        related="group_id.company_id", store=True, default=None
```
florian-dacosta: You don't need to put the store=True anymore for related fields, it is by default.
About the default I am not sure why you need to add it everywhere ?

product_supplierinfo_group/security/supplierinfo_group_security.xml
```diff
@@ -1,13 +1,9 @@
...
-                name="domain_force"
-            >['|', ('company_id', '=', False), ('company_id', 'in', company_ids)]</field>
-        </record>
-    </data>
+<odoo noupdate="1">
+    <record id="product_supplierinfo_group_multi_company_rule" model="ir.rule">
+        <field name="name">Product Supplierinfo Group multi-company</field>
+        <field ref="model_product_supplierinfo_group" name="model_id" />
+        <field eval="True" name="global" />
+        <field name="domain_force">[('company_id', 'in', company_ids + [False])]</field>
```
florian-dacosta, line 7: I think it requires a migration script, because of the noupdate, no?

product_supplierinfo_group/views/supplierinfo_group.xml
```diff
@@ -31,35 +30,34 @@
...
+                            </list>
                         </field>
                     </group>
                 </sheet>
             </form>
         </field>
     </record>
-    <record id="supplierinfo_group_view_tree" model="ir.ui.view">
-        <field name="name">supplierinfo_group_view_tree</field>
+    <record id="supplierinfo_group_view_list" model="ir.ui.view">
```
florian-dacosta, line 47: It is not very recommanded to change the xml_id, because all dependent modules will fail.

If you want to do it anyway, because this is a module with none of few dependencies, you should add a pre-migrate script to rename it (openupgrade.rename_xmlids...)

product_supplierinfo_group/models/product_supplierinfo.py
florian-dacosta: My mistake, store=Trues is not by default at all...

product_supplierinfo_group/views/supplierinfo_group.xml
bealdav: Done

product_supplierinfo_group/security/supplierinfo_group_security.xml
bealdav: Done

florian-dacosta:
Thanks for the fixes @bealdav 
The tests are failing because of warnings : Redundant default on product.supplierinfo.sequence 
I think we need to try adding default=None or default=False for these related fields.

I did some tests and I have found issues.

1) On any product (template), add a second supplierinfo group, and then try to re-organize the sequence (from the handle widget)? Odoo crashes
image


2) On a variant (from product variant menu) When I add a second group, I don't see the variant field, and when I save the new supplier group disappear.
image
image

PR #2152, opened 2025-12-04
[18.0][MIG] product_usability

bealdav:
cc @legalsylvain


product_usability/security/ir.model.access.csv
```diff
@@ -1,10 +1,10 @@
 id,name,model_id:id,group_id:id,perm_read,perm_write,perm_create,perm_unlink
-access_product_template_manager,product.template manager,product.model_product_template,group_product_creation,1,1,1,1
-access_product_product_manager,product.product manager,product.model_product_product,group_product_creation,1,1,1,1
+access_product_template_manager,product.template manager,product.model_product_template,product.group_product_manager,1,1,1,1
+access_product_product_manager,product.product manager,product.model_product_product,product.group_product_manager,1,1,1,1
```
legalsylvain: Why adding this both lines ? it is the same as in ./addons/product/security/ir.model.access.csv:

access_product_template_manager,product.template.manager,model_product_template,group_product_manager,1,1,1,1

legalsylvain: In fact, if you use product.group_product_manager it will give access to all the models. So we should not change this file. But we should create inheritance between product.group_product_manager and product_usability.group_product_manager (and other groups).

product_usability/security/res_groups.xml
```diff
@@ -5,8 +5,13 @@
...
 -->
 <odoo>
+    <!-- We first need to rename this group because of a naming conflict with group_product_creation -->
+    <record id="product.group_product_manager" model="res.groups">
+        <field name="name">Product Manager</field>
+    </record>
+
     <record id="group_product_creation" model="res.groups">
-        <field name="name">Product Creation</field>
+        <field name="name">Product Creation </field>
```
legalsylvain: I guess this change is not necessary.

legalsylvain (approved):
A minor remark inline.
Diff review. LGTM. 
thanks !

product_usability/security/res_groups.xml
bealdav: As you can see in last failed commit

```bash
DETAIL:  Key (category_id, name)=(93, {"en_US": "Product Creation"}) already exists.
```

I propose to remove last commit and approve

bealdav: cc @legalsylvain

product_usability/security/res_groups.xml
```diff
@@ -0,0 +1,51 @@
...
+<!--
+    Copyright (C) 2024-Today GRAP (http://www.grap.coop)
+    @author: Sylvain LE GAL (https://twitter.com/legalsylvain)
+    License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+-->
+<odoo>
+    <!-- We first need to rename this group because of a naming conflict with group_product_creation -->
+    <record id="product.group_product_manager" model="res.groups">
+        <field name="name">Product Manager</field>
+    </record>
```
legalsylvain: CI is red, because of conflict name. I don't understand why. maybe you could try to put this change in a first xml file, and the rest in another ?

bealdav: Conflict beacause name is multi lang / json ? even is only one language installed ?

bealdav: no more sucess.

Hook ?

bealdav:
it's OK with hook @legalsylvain

legalsylvain:
thansk ! could you squash ?

product_usability/security/res_groups.xml
```diff
@@ -35,4 +35,12 @@
         <field name="category_id" ref="base.module_category_usability" />
         <field name="implied_ids" eval="[Command.link(ref('uom.group_uom'))]" />
     </record>
+
+    <!-- pylint:disable=xml-duplicate-record-id -->
```
legalsylvain: I think this line is now useless.

bealdav: Caramba encore raté

legalsylvain (commented):
thanks ! 

/ocabot merge nobump

PR #2157, opened 2025-12-07
[18.0][MIG] product_pricelist_margin

rvalyi:
migration from 16.0

NOTE: in the pricelist form view, in v16 there were two modes: normal and "advanced one" (when user belongs to the product.group_sale_pricelist group). So the module was extending two tree views for the items. Now in v18 the form view is similar to the old advanced mode and only one extension is needed.


santostelmo (approved):
LG

legalsylvain (commented):
/ocabot merge nobump

legalsylvain:
hi @rvalyi . 

Sorry for the noice. new OCA rules are annoying. CI is green, but it is not possible to merge ...

Let me know when you fixed the problem.

rvalyi:
@legalsylvain I fixed the markdown file, let's see if the merge will work now.

/ocabot merge nobump

PR #2166, opened 2025-12-13
[18.0][IMP] product_secondary_unit: add template/variant sync helpers

yostashiro:
Add helper methods for syncing secondary UoM fields between product.template and product.product:
- _compute_template_secondary_uom_field
- _inverse_template_secondary_uom_field

Can be used by dependent modules like stock_secondary_unit.

Example for ProductTemplate in stock_secondary_unit:

```.py
    stock_secondary_uom_id = fields.Many2one(
        comodel_name="product.secondary.unit",
        domain="[('product_tmpl_id', '=', id), ('product_id', '=', False)]",
        string="Second unit for inventory",
        compute="_compute_stock_secondary_uom_id",
        inverse="_inverse_stock_secondary_uom_id",
        store=True,
        readonly=False,
    )

    @api.depends("product_variant_ids.stock_secondary_uom_id")
    def _compute_stock_secondary_uom_id(self):
        self._compute_template_secondary_uom_field("stock_secondary_uom_id")

    def _inverse_stock_secondary_uom_id(self):
        self._inverse_template_secondary_uom_field("stock_secondary_uom_id")
```

Adding no tests here as it's tricky to do so in this module.

@qrtl QT6192

### DIFF
--- a/product_secondary_unit/models/product_template.py
+++ b/product_secondary_unit/models/product_template.py
@@ -21,3 +21,39 @@ class ProductTemplate(models.Model):
             and self.secondary_uom_ids[0]
             or self.secondary_uom_ids
         )
+
+    def _compute_template_secondary_uom_field(self, fname):
+        """Helper to sync template secondary UoM field from variant.
+
+        Can be used in modules depending on product_secondary_unit to implement
+        compute methods for secondary UoM fields on product.template.
+
+        - Single variant: Sync the variant's value to the template
+        - Multiple variants: Keep the existing template value
+
+        :param str fname: name of the secondary UoM field to compute
+        """
+        for template in self:
+            if len(template.product_variant_ids) == 1:
+                template[fname] = template.product_variant_ids[fname]
+            # Keep existing value when multiple variants (don't clear)
+
+    def _inverse_template_secondary_uom_field(self, fname):
+        """Helper to propagate template secondary UoM field to variants.
+
+        Can be used in modules depending on product_secondary_unit to implement
+        inverse methods for secondary UoM fields on product.template.
+
+        - Single variant: Always sync the template's value to the variant
+        - Multiple variants: Only update variants without their own setting
+
+        :param str fname: name of the secondary UoM field to propagate
+        """
+        for template in self:
+            if len(template.product_variant_ids) == 1:
+                variants_to_update = template.product_variant_ids
+            else:
+                variants_to_update = template.product_variant_ids.filtered(
+                    lambda v, f=fname: not v[f]
+                )
+            variants_to_update[fname] = template[fname]
--- a/product_secondary_unit/models/product_template.py
+++ b/product_secondary_unit/models/product_template.py
@@ -23,37 +23,88 @@ class ProductTemplate(models.Model):
         )
 
     def _compute_template_secondary_uom_field(self, fname):
-        """Helper to sync template secondary UoM field from variant.
+        """Helper to sync a template secondary UoM field from its variants.
 
         Can be used in modules depending on product_secondary_unit to implement
         compute methods for secondary UoM fields on product.template.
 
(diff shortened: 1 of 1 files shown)


product_secondary_unit/models/product_template.py
```diff
@@ -21,3 +21,39 @@ def _get_default_secondary_uom(self):
...
+        :param str fname: name of the secondary UoM field to propagate
+        """
+        for template in self:
+            if len(template.product_variant_ids) == 1:
+                variants_to_update = template.product_variant_ids
+            else:
+                variants_to_update = template.product_variant_ids.filtered(
+                    lambda v, f=fname: not v[f]
+                )
+            variants_to_update[fname] = template[fname]
```
yostashiro, line 59: I just realized that sale_order_secondary_unit already implements sync methods slightly differently (there are currently no equivalent implementations for purchase_order_secondary_unit or stock_secondary_unit):

https://github.com/OCA/sale-workflow/blob/51cbf6cc48b10bd86fcdb3c6ace33b7eeb3358d8/sale_order_secondary_unit/models/product_template.py#L20-L40

Each option has its pros and cons, but I personally prefer the logic of these helper methods, as the approach used in sale_order_secondary_unit forces blank values at the template level when more than one secondary UoM is involved among variants, which can be problematic in some situations, e.g., you lose visibility of secondary UoM quantities for the on-hand quantities of some templates.

That said, I can adjust the logic of these helper methods to match sale_order_secondary_unit if what I propose here is considered "breaking". Let me know your thoughts!

hoangtrann: Yeah I do think leaving it as is for visibility is better, which signal special handling if needed

hoangtrann (approved):

AungKoKoLin1997 (approved):
Code Review: LGTM

nobuQuartile (approved):
Code Review: LGTM

BhaveshHeliconia (approved):

AungKoKoLin1997:
@yostashiro 
I pushed a commit changing the helpers to be a plain mirror of the variants: the template shows the variants' value when they all share the same one, and is blank otherwise.

This is consistent with the existing logic in `sale_order_secondary_unit` and with `purchase_order_secondary_unit` in https://github.com/OCA/purchase-workflow/pull/3162.

Regarding the visibility concern raised above, only the value displayed on the template is lost; the values on the variants remain unchanged. Also, templates with multiple variants that rely on the secondary UoM shown on the template for on-hand quantities are relatively uncommon, so I think this trade-off is acceptable.

This behavior should also be fine for existing users of this module, as they have already been using this logic for some time. Keeping the original logic here would instead make the behavior inconsistent with `sale_order_secondary_unit` and purchase_order_secondary_unit, and could prevent us from preserving their current behavior.

Please let me know if you have a different opinion.

Example for ProductTemplate in stock_secondary_unit:

 ```
   stock_secondary_uom_id = fields.Many2one(
        comodel_name="product.secondary.unit",
        domain="[('product_tmpl_id', '=', id), ('product_id', '=', False)]",
        string="Second unit for inventory",
        compute="_compute_stock_secondary_uom_id",
        inverse="_inverse_stock_secondary_uom_id",
        store=True,
        readonly=False,
    )

    @api.depends("product_variant_ids", "product_variant_ids.stock_secondary_uom_id")
    def _compute_stock_secondary_uom_id(self):
        self._compute_template_secondary_uom_field("stock_secondary_uom_id")

    def _inverse_stock_secondary_uom_id(self):
        self._inverse_template_secondary_uom_field("stock_secondary_uom_id")

    @api.onchange("stock_secondary_uom_id")
    def _onchange_stock_secondary_uom_id(self):
        return self._onchange_template_secondary_uom_field("stock_secondary_uom_id")
```

PR #2169, opened 2025-12-17
[FIX] product_print_category: changes on product impact only this pro…

mihien:
…duct

Currently when changing a relevant field of a product variant, all the products variants of this category will be set as obsolete.

This fixes it so that only the modified variant will have to_print = True. 
And adds a few tests to make sure it's valid.

### DIFF
--- a/product_print_category/tests/test_product_print_category.py
+++ b/product_print_category/tests/test_product_print_category.py
@@ -114,7 +114,7 @@ class TestProductPrintCategory(TransactionCase):
             "Print all products should propose 3 products",
         )
 
-    def test_21_onchange(self):
+    def test_20_onchange(self):
         product = self.ProductProduct.create(
             {
                 "name": "Demo Product Product Name",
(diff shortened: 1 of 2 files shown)


product_print_category/tests/test_product_print_category.py
```diff
@@ -136,3 +136,66 @@ def test_21_onchange(self):
...
+        self.assertTrue(template.to_print)
+
+    def test_31_test_wizard_on_change_products_products(self):
+        # Add a new product
+        product = self.ProductProduct.create(
+            {
+                "name": "Demo Product Name",
+                "print_category_id": self.print_category_1.id,
+            }
+        )
```
legalsylvain, line 171: this should be great to create a new variant related to the same template.

product_print_category/tests/test_product_print_category.py
```diff
@@ -136,3 +136,66 @@ def test_21_onchange(self):
...
+        # Set all products as printed
+        wizard = self.ProductPrintWizard.with_context(
+            active_model="product.print.category",
+            active_ids=[self.print_category_1.id],
+        ).create({})
+        wizard.print_report()
+        self.env.ref("product_print_category.pricetag")._render_qweb_pdf(
+            "product_print_category.report_pricetag",
+            wizard.line_ids.ids,
+        )
```
legalsylvain:
```suggestion
        self.print_category_1.mapped("product_to_print_ids").write({"to_print": False})
```
less verbose.

legalsylvain (commented):
thanks for this patch. 

some remarks inline.

otherwise, lgtm.

mihien:
@legalsylvain Thanks for the review, I've added your suggestions

product_print_category/tests/test_product_print_category.py
```diff
@@ -136,3 +136,60 @@ def test_21_onchange(self):
...
+            }
+        )
+
+        # Set all products as printed
+        self.print_category_1.mapped("product_to_print_ids").write({"to_print": False})
+        self.assertEqual(product.to_print, False)
+        self.assertEqual(second_product.to_print, False)
+
+        # Change name of product, shared with second_product
+        product.name = "Demo Product New Name"
```
legalsylvain: Hi. thanks for all the changes. just a question. why not better used some field NOT shared between many products of the same template. 
if so, we can verify that if we change the information of a variant, it doesn't change the information of the other variants of the same template.

i think you can use variant price field.

mihien: I actually had the same idea, but it turns out after a quick test that the price is shared with the template (at least it is in 16 without playing in the settings). I'll have a look if there's another field that isn't shared

legalsylvain: I think you can use ``default_code`` or ``barcode`` that are different for each variant by design.

mihien: Yes thanks for the help, using barcodes did the trick

product_print_category/tests/test_product_print_category.py
```diff
@@ -136,3 +136,64 @@ def test_21_onchange(self):
...
+        self.print_category_1.mapped("product_to_print_ids").write({"to_print": False})
+        self.assertEqual(product.to_print, False)
+        self.assertEqual(second_product.to_print, False)
+
+        # Change barcode of product, not shared with second_product
+        barcode_field = self.env["ir.model.fields"].search(
+            [("model", "=", "product.product"), ("name", "=", "barcode")]
+        )
+        self.print_category_1.write({"field_ids": [(4, barcode_field.id, 0)]})
+        product.barcode = "1234567890128"
```
legalsylvain, line 190: great product : https://www.digit-eyes.com/upcCode/1234567890128.html

legalsylvain:
Thanks ! 

/ocabot merge patch

PR #2171, opened 2025-12-22
[19.0][ADD] product_supplierinfo_indirect_cost

AaronHForgeFlow:
Base module to calculate estimated costs in purchases.


product_supplierinfo_indirect_cost/models/product_supplierinfo.py
```diff
@@ -0,0 +1,20 @@
...
+
+
+class ProductSupplierinfo(models.Model):
+    _inherit = "product.supplierinfo"
+
+    indirect_cost_percent = fields.Float(string="Indirect Cost %", digits=(16, 2))
+    indirect_cost_amount = fields.Float(
+        digits=(16, 2),
+        compute="_compute_indirect_cost_amount",
+    )
```
JordiMForgeFlow, line 12: code LGTM, I wonder if the indirect cost should be actually stored to make it easier for searching/sorting in the UI? WDYT?

product_supplierinfo_indirect_cost/readme/DESCRIPTION.md
```diff
@@ -0,0 +1,6 @@
+This module adds indirect cost information to supplierinfo.
+
+It introduces an indirect cost percentage and a computed indirect cost amount
+based on the supplier price.
+
+This is a base module to calculate estimated costs in purchases.
```
legalsylvain: could you write functional exemples of what is an "estimated costs" and an "indirect cost".

thanks !

AaronHForgeFlow: Indirect cost are all cost related to the purchased products, mainly transportation costs, but also handling, special taxes, insurance, etc.

Estimated costs is the estimated indirect costs based on experience.

In many cases the best companies can do is an estimate based on past years information, because they don't get an invoice for a specific picking, which is what Odoo suggest for the landing cost creation. Those use cases are too complex to put them in Odoo and also they will require a lot of manual work to ensure the data is accurate. In this case the company have a rough estimate for the product and the vendor.

The idea is to automate the creation of landed costs based on these estimates. At the end of the fiscal period they can compare the total estimated costs created versus real cost and do the proper financial adjustments, and maybe even adjust the estimated costs for the next year.

product_supplierinfo_indirect_cost/models/product_supplierinfo.py
AaronHForgeFlow: I agree with that!

JordiMForgeFlow (approved):

product_supplierinfo_indirect_cost/readme/DESCRIPTION.md
legalsylvain: Thanks a lot ! Could you write that description in the DESCRIPTION.md file, that is very usefull to be sure to understand that module.

AaronHForgeFlow: Good idea! I did it. Thank you.

legalsylvain (approved):

PR #2172, opened 2025-12-22
[19.0][IMP] product_net_weight: adopt native style of 19.0

alexis-via:
I just aligned the CSS class/style used by product_net_weight to the CSS class/style used by the product module for the "weight" field.

Before:

net_weight_before

After:

net_weight_after


legalsylvain:
Thanks !

/ocabot merge patch

PR #2176, opened 2026-01-04
[19.0][MIG] product_cost_security: Migration to 19.0

dreispt:


cormaza (approved):
LGTM, maybe should re-run test to check

ferran-S73:
@dreispt can you re-generate the runboat build please?

dreispt:
Fixed test error, related to field check access.

https://github.com/OCA/product-attribute/pull/2176/changes/587d1cb6cf8dd60f8aa7d8db8d46b441fbf19053#diff-e8472398c25c784f588de4084c0151df3f443c5c052eb228fea3d494e110ddd1L56-R61

BhaveshHeliconia (approved):
Functional review LGTM!

product_cost_security/security/product_cost_security.xml
```diff
@@ -0,0 +1,29 @@
...
+
+    <!--
+    Only users with this group can view the fields with this security group.
+    Add group to user_root to avoid errors when other modules install and
+    add fields to views with this group defined as field attribute.
+    -->
+    <record model="res.groups" id="group_product_cost">
+        <field name="name">Read</field>
+        <field name="privilege_id" ref="module_privilege" />
+        <field name="user_ids" eval="[(4, ref('base.user_root'))]" />
```
carlos-lopez-tecnativa, line 15:
```suggestion
        <field name="user_ids" eval="[Command.link(ref('base.user_root'))]" />
```

product_cost_security/security/product_cost_security.xml
```diff
@@ -0,0 +1,29 @@
...
+    Only users within this group can edit costs.
+    -->
+    <record model="res.groups" id="group_product_edit_cost">
+        <field name="name">Edit</field>
+        <field name="privilege_id" ref="module_privilege" />
+        <field
+            name="implied_ids"
+            eval="[(4, ref('product_cost_security.group_product_cost'))]"
+        />
+        <field name="user_ids" eval="[(4, ref('base.user_root'))]" />
```
carlos-lopez-tecnativa, line 27:
```suggestion
        <field
            name="implied_ids"
            eval="[Command.link(ref('product_cost_security.group_product_cost'))]"
        />
        <field name="user_ids" eval="[Command.link(ref('base.user_root'))]" />
```

carlos-lopez-tecnativa (changes requested):
When I try to open the Product form view, I get this error. I tested it with the demo user without the `Product Costs` group. Please check.
Image

Image

carlos-lopez-tecnativa:
ping @dreispt

carlos-lopez-tecnativa (approved):
I tested this locally, and the module works on its own and with account and sale, but the error I reported previously occurs when `purchase` is installed.

Odoo uses `standard_price` in some places, which causes the error if the field is inaccessible to the user: https://github.com/odoo/odoo/blob/0eecf65f7f20cdb7dc1f98f9fc89bf8c37e4c949/addons/purchase/views/product_views.xml#L62

So, a new glue module is probably necessary. What do you think, @pedrobaeza?

Additionally, another error occurs when `stock` is installed and the user does not belong to any group. If they try to create a new product and save it, an error is displayed.
image


This could probably be handled if this module is migrated to V19: https://github.com/OCA/product-attribute/pull/2232

Regarding this PR, I don't want to block the merge because of cosmetic changes. It would be great to update the code to use the new Command syntax, but anyway, LGTM.

TT64200

pedrobaeza:
I think this module can override `get_views` method of `product.template` and `product.product`, and locate appearances of the field in contexts, and clean them out if you don't have the permissions, so no glue module is needed.

Let's merge this anyway, and the other improvement can be done in a separate PR.

/ocabot merge nobump

carlos-lopez-tecnativa:
ping @pedrobaeza

PR #1357, opened 2023-06-09
[14.0] product_pricelist_supplierinfo: Refactoring

ypapouin:
This module has been migrated without taking odoo base code changes into account.
This PR is a proposal to re-use existing implementation.
All major changes are made in separate commits to clearly understand all edits.

- The main method used to compute price form supplier info has been moved from `product.template` to `product.pricelist.item` and renamed to `_compute_price_from_supplierinfo`
- We now override the built-in `_compute_price` from `product.pricelist.item` to get and compute our "supplierinfo" price.
- As a parameter, you can now pass a single product whatever it is (`product.product`/`product.template`)
- Better date support

Note that the first commit (Add supplier support) already has its own PR https://github.com/OCA/product-attribute/pull/1345

@pedrobaeza @victoralmau , can you review this PR as you were contributors on this module ?

PR #2137, opened 2025-11-17
[19.0][MIG] product_pricelist_supplierinfo Migration to v19.0

AaronHForgeFlow:

PR #1837, opened 2025-01-07
[MIG] product_pricelist_supplierinfo: Migration to 18.0

CLaurelB:
Supersed https://github.com/OCA/product-attribute/pull/1745

PR #2177, opened 2026-01-05
[18.0][FIX][IMP] product_pricelist_supplierinfo: Fixes and refactoring

ypapouin:
Fix and improvements for Odoo 18.0 (this PR is based on a closed PR I previously made for Odoo 14.0 https://github.com/OCA/product-attribute/pull/1357)

1. Use rule context: Instead of trying to get the product context, we must use the `product.pricelist.item` context to get the filtered partner. The current implementation is broken, a new test is added to ensure that it works correctly now.
2. Let Odoo compute the produce price: Odoo's pricelist computation is now correctly split, so we override the base price computation instead of the full price and we let Odoo compute surcharge/discount/etc. This improvement removes a lot of code that was copy/paste from Odoo price computation.
3. Refactoring to match Odoo's pricelist API: To match the new price computation signature from https://github.com/odoo/odoo/commit/57ced812. (Also drop support to get "UoM" and "date" from context)
4. More and improved tests on UoM and Dates.

### DIFF
--- a/product_pricelist_supplierinfo/models/product_pricelist_item.py
+++ b/product_pricelist_supplierinfo/models/product_pricelist_item.py
@@ -37,14 +37,17 @@ class ProductPricelistItem(models.Model):
         return self.env.context.get("force_filter_supplier_id", self.filter_supplier_id)
 
     def _compute_base_price(self, product, quantity, uom, date, currency):
+        """Compute the base price for Odoo that will be used for the full price
+        computation (surcharge/discount/etc.)
+        """
+        price = super()._compute_base_price(product, quantity, uom, date, currency)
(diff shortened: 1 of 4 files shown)


ypapouin:
@CLaurelB can you please take a look on this since you did the 18.0 migration ? (https://github.com/OCA/product-attribute/pull/1837)
@AaronHForgeFlow can you please take a look on this since you made a PR for the 19.0 migration ? (https://github.com/OCA/product-attribute/pull/2137)

AaronHForgeFlow:
Thanks @ypapouin code LGTM. I will check how to incorporate this to v19, a refactor was done for the v19 migration and it is not a straightforward forward port.

PR #2182, opened 2026-01-06
[18.0][FIX] product_secondary_unit: UoM conversion in secondary unit mixin

AungKoKoLin1997:
- Replace _get_factor_line() with uom._compute_quantity() to fix incorrect conversion when line UoM differs from product's base UoM. The previous method failed for certain UoM combinations

Related: https://github.com/OCA/stock-logistics-warehouse/pull/2496

@qrtl QT6285

### DIFF
--- a/product_secondary_unit/models/product_secondary_unit_mixin.py
+++ b/product_secondary_unit/models/product_secondary_unit_mixin.py
@@ -55,6 +55,7 @@ class ProductSecondaryUnitMixin(models.AbstractModel):
     def _get_uom_line(self):
         return self[self._secondary_unit_fields["uom_field"]]
 
+    # TODO: This method is now not used in this module. Deprecate it in future.
     def _get_factor_line(self):
         uom_line = self._get_uom_line()
         return self.secondary_uom_id.factor * (
@@ -72,6 +73,19 @@ class ProductSecondaryUnitMixin(models.AbstractModel):
             return []
         return [self._secondary_unit_fields["qty_field"]]
 
+    @api.model
+    def _convert_qty_to_secondary_uom(self, qty):
+        # Intended to be called from other operations if needed.
+        self.ensure_one()
+        uom_line = self._get_uom_line()
+        uom_product = self.product_id[self._product_uom_field]
+        if uom_line != uom_product:
+            qty = uom_line._compute_quantity(qty, uom_product)
+        return float_round(
+            qty / (self.secondary_uom_id.factor or 1.0),
+            precision_rounding=self.secondary_uom_id.uom_id.rounding,
+        )
+
     @api.depends(lambda x: x._get_secondary_uom_qty_depends())
     def _compute_secondary_uom_qty(self):
         for line in self:
@@ -80,13 +94,8 @@ class ProductSecondaryUnitMixin(models.AbstractModel):
                 continue
             elif line.secondary_uom_id.dependency_type == "independent":
                 continue
-            factor = line._get_factor_line()
             qty_line = line._get_quantity_from_line()
-            qty = float_round(
-                qty_line / (factor or 1.0),
-                precision_rounding=line.secondary_uom_id.uom_id.rounding,
-            )
-            line.secondary_uom_qty = qty
+            line.secondary_uom_qty = line._convert_qty_to_secondary_uom(qty_line)
 
     def _get_default_value_for_qty_field(self):
         return self.default_get([self._secondary_unit_fields["qty_field"]]).get(
@@ -114,12 +123,11 @@ class ProductSecondaryUnitMixin(models.AbstractModel):
             rec.env.remove_to_compute(
                 field=rec._fields["secondary_uom_qty"], records=rec
             )
-            factor = rec._get_factor_line()
-            qty = float_round(
-                rec.secondary_uom_qty * factor,
-                precision_rounding=rec._get_uom_line().rounding,
-            )
-            rec[rec._secondary_unit_fields["qty_field"]] = qty
+            qty_base = rec.secondary_uom_qty * rec.secondary_uom_id.factor
+            uom_line = rec._get_uom_line()
+            uom_product = rec.product_id[rec._product_uom_field]
+            qty_line = uom_product._compute_quantity(qty_base, uom_line)
+            rec[rec._secondary_unit_fields["qty_field"]] = qty_line
 
     def _onchange_helper_product_uom_for_secondary(self):
         """Helper method to be called from onchange method of uom field in
@@ -130,13 +138,8 @@ class ProductSecondaryUnitMixin(models.AbstractModel):
             return
         elif self.secondary_uom_id.dependency_type == "independent":
             return
-        factor = self._get_factor_line()
-        line_qty = self._get_quantity_from_line()
-        qty = float_round(
-            line_qty / (factor or 1.0),
-            precision_rounding=self.secondary_uom_id.uom_id.rounding,
-        )
-        self.secondary_uom_qty = qty
+        qty_line = self._get_quantity_from_line()
+        self.secondary_uom_qty = self._convert_qty_to_secondary_uom(qty_line)
 
     @api.model
     def default_get(self, fields_list):
--- a/product_secondary_unit/tests/test_secondary_unit_mixin.py
+++ b/product_secondary_unit/tests/test_secondary_unit_mixin.py
@@ -15,14 +15,13 @@ class TestProductSecondaryUnitMixin(TransactionCase, FakeModelLoader):
         from .models import SecondaryUnitFake
 
         cls.loader.update_registry((SecondaryUnitFake,))
-        cls.product_uom_kg = cls.env.ref("uom.product_uom_kgm")
         cls.product_uom_unit = cls.env.ref("uom.product_uom_unit")
         cls.product_uom_dozen = cls.env.ref("uom.product_uom_dozen")
         cls.product_template = cls.env["product.template"].create(
             {
                 "name": "test",
-                "uom_id": cls.product_uom_kg.id,
-                "uom_po_id": cls.product_uom_kg.id,
+                "uom_id": cls.product_uom_unit.id,
+                "uom_po_id": cls.product_uom_unit.id,
                 "secondary_uom_ids": [
                     Command.create(
                         {
@@ -151,3 +150,44 @@ class TestProductSecondaryUnitMixin(TransactionCase, FakeModelLoader):
         fake_model.write({"secondary_uom_qty": 4})
         self.assertEqual(fake_model.product_uom_qty, 17)
         self.assertEqual(fake_model.secondary_uom_qty, 4)
+
+    def test_secondary_uom_qty_conversion_with_different_line_uom(self):
+        product_uom_litre = self.env.ref("uom.product_uom_litre")
+        volume_category = self.env.ref("uom.product_uom_categ_vol")
+        product_uom_millilitre = self.env["uom.uom"].create(
+            {
+                "name": "mL",
+                "category_id": volume_category.id,
+                "uom_type": "smaller",
+                "factor": 1000.0,
+                "rounding": 0.01,
+            }
+        )
+        product_template = self.env["product.template"].create(
+            {
+                "name": "Test",
+                "uom_id": product_uom_millilitre.id,
+                "uom_po_id": product_uom_millilitre.id,
+                "secondary_uom_ids": [
+                    Command.create(
+                        {
+                            "name": "Bottle",
+                            "uom_id": product_uom_millilitre.id,
+                            "factor": 1200.0,
+                            "dependency_type": "dependent",
+                        }
+                    )
+                ],
+            }
+        )
+        bottle_secondary_uom = product_template.secondary_uom_ids
+        fake_model = self.env["secondary.unit.fake"].create(
+            {
+                "name": "Secondary unit fake (volume)",
+                "product_id": product_template.product_variant_ids.id,
+                "product_uom_id": product_uom_litre.id,
+                "product_uom_qty": 1.2,
+                "secondary_uom_id": bottle_secondary_uom.id,
+            }
+        )
+        self.assertEqual(fake_model.secondary_uom_qty, 1.0)


yostashiro (approved):
Code review. LGTM.

nobuQuartile (approved):
LGTM Code review

kanda999 (approved):
Code review.

AungKoKoLin1997:
In my last commit, I add the configuration for qweb reports presentation.

product_secondary_unit/models/res_company.py
```diff
@@ -0,0 +1,33 @@
...
+    )
+    # Added for supporting the existing report presentation. We can drop this together
+    # with the second qty column in reports if the community agrees with it.
+    hide_secondary_uom_column = fields.Boolean(
+        string="Hide Secondary UoM Column",
+        default=False,
+        help="When enabled, hides the separate Secondary Qty column in the "
+        "reports. The secondary quantity will still be shown in the main Qty column "
+        "based on the 'Secondary Unit Price Display' setting.",
+    )
```
yostashiro: I think we should allow separate settings for sales and purchasing as the requirements can be different. We can add a method like _get_secondary_unit_report_settings or something similar and let it return appropriate values depending on the type of document.

product_secondary_unit/models/res_company.py
```diff
@@ -0,0 +1,65 @@
...
+    secondary_uom_price_display_sale = fields.Selection(
+        selection=PRICE_DISPLAY_SELECTION,
+        string="Secondary Unit Price Display (Sales)",
+        default="primary",
+        help="Configure how unit prices are displayed in sales reports when "
+        "secondary units are used.\n"
+        "- Primary Unit Price Only: Shows only the primary unit price\n"
+        "- Prioritize Secondary Unit Price: Shows secondary unit price when "
+        "available, falls back to primary otherwise\n"
+        "- Both: Shows both prices (primary and secondary)",
```
yostashiro: We may remove the help here as it doesn't show anywhere, and focus on having a good description in the config view?

AungKoKoLin1997: I removed the help text from the field definitions and didn’t expand the help text in the Settings view. Adding the option explanations there made the UI look cluttered and bloated. The behavior is already documented in CONFIGURE.md, and users can find the details there. These settings aren’t changed often, so the current help should be sufficient.

product_secondary_unit/views/res_config_settings_views.xml
```diff
@@ -0,0 +1,42 @@
...
+                >
+                    <field name="hide_secondary_uom_column_sale" />
+                </setting>
+                <setting
+                    id="hide_secondary_uom_column_purchase"
+                    company_dependent="1"
+                    help="If selected, the 'Second Qty' column will be hidden in purchase orders, quotations, and vendor bill reports."
+                >
+                    <field name="hide_secondary_uom_column_purchase" />
+                </setting>
```
yostashiro: Just to declutter the presentation.
```suggestion
                <setting
                    id="secondary_uom_price_display"
                    company_dependent="1"
                    string="Secondary Unit Price Display"
                    help="Configure how unit prices and quantities are shown in reports when secondary units are used."
                >
                    <div class="row mt8">
                        <label for="secondary_uom_price_display_sale" class="col-lg-3 o_light_label" string="Sales" />
                        <field name="secondary_uom_price_display_sale" widget="selection" required="1" />
                    </div>
                    <div class="row mt8">
                        <label for="secondary_uom_price_display_purchase" class="col-lg-3 o_light_label" string="Purchase" />
                        <field name="secondary_uom_price_display_purchase" widget="selection" required="1" />
                    </div>
                </setting>
                <setting
                    id="hide_secondary_uom_column"
                    company_dependent="1"
                    string="Hide Secondary Qty Column"
                    help="If selected, the 'Second Qty' column will be hidden in reports. Enable this when you choose to show the secondary unit price in 'Secondary Unit Price Display' setting."
                >
                    <div class="d-flex gap-4 mt8">
                        <span>
                            <field name="hide_secondary_uom_column_sale" class="oe_inline" />
                            <label for="hide_secondary_uom_column_sale" string="Sales" />
                        </span>
                        <span>
                            <field name="hide_secondary_uom_column_purchase" class="oe_inline" />
                            <label for="hide_secondary_uom_column_purchase" string="Purchase" />
                        </span>
                    </div>
                </setting>
```
image

product_secondary_unit/models/res_company.py
```diff
@@ -0,0 +1,47 @@
...
+    )
+    hide_secondary_uom_column_purchase = fields.Boolean(
+        string="Hide Secondary UoM Column (Purchase)",
+        default=False,
+    )
+
+    def hide_secondary_uom_column(self, record):
+        """Return whether to hide the 'Second Qty' column for this document type."""
+        if record._name == "sale.order" or (
+            record._name == "account.move" and record.is_inbound()
```
yostashiro:
```suggestion
            record._name == "account.move" and record.move_type in self.env["account.move"].get_sale_types(include_receipts=True)
```

yostashiro (changes requested):
Partial review.

product_secondary_unit/models/res_company.py
AungKoKoLin1997: I just use `is_sale_document(include_receipts=True)` since the condition is the same that is applied in that method.

product_secondary_unit/views/res_config_settings_views.xml
```diff
@@ -0,0 +1,72 @@
...
+                    <div class="row mt8">
+                        <label
+                            for="secondary_uom_price_display_sale"
+                            class="col-lg-3 o_light_label"
+                            string="Sales"
+                        />
+                        <field
+                            name="secondary_uom_price_display_sale"
+                            widget="selection"
+                            required="1"
```
yostashiro: Sorry, can we switch this to be part of the field definition?

nobuQuartile (approved):
Functional review and code review: LGTM

product_secondary_unit/models/product_secondary_unit_mixin.py
```diff
@@ -72,6 +73,17 @@ def _get_secondary_uom_qty_depends(self):
             return []
         return [self._secondary_unit_fields["qty_field"]]
 
+    @api.model
+    def _calc_secondary_uom_qty(self, record, qty):
```
yostashiro: I think this goes better with Odoo's method name convention.
```suggestion
    def _convert_qty_to_secondary_uom(self, record, qty):
```

kanda999 (approved):
LGTM: it works as expected in my local env

product_secondary_unit/models/product_secondary_unit_mixin.py
```diff
@@ -146,3 +147,54 @@ def default_get(self, fields_list):
...
+    def _get_secondary_uom_report_type(self):
+        """Return 'sale', 'purchase', or None."""
+        self.ensure_one()
+        if self._name == "sale.order.line" or (
+            self._name == "account.move.line"
+            and self.move_id.is_sale_document(include_receipts=True)
+        ):
+            return "sale"
+        if self._name == "purchase.order.line" or (
+            self._name == "account.move.line" and self.move_id.is_purchase_document()
```
yostashiro: Any reason not to follow the same pattern as the sales side?
```suggestion
            self._name == "account.move.line"
            and self.move_id.is_purchase_document(include_receipts=True)
```

product_secondary_unit/models/product_secondary_unit_mixin.py
```diff
@@ -146,3 +147,54 @@ def default_get(self, fields_list):
...
+
+    def report_show_price_uom(self, uom_source=None):
+        """Return True if UoM should be shown in price column.
+
+        UoM is shown when the line displays multiple UoMs.
+        """
+        self.ensure_one()
+        if not self.secondary_uom_id:
+            return False
+        display, hide_col = self._get_secondary_uom_report_settings()
```
yostashiro: I think we should split methods in the following manner, to make it easier when the community decides to deprecate the secondary qty column in reports.
```suggestion
        display = self._get_secondary_uom_price_display()
        hide_col = self._get_secondary_uom_hide_col()
```

product_secondary_unit/models/product_secondary_unit_mixin.py
```diff
@@ -146,3 +147,56 @@ def default_get(self, fields_list):
...
+        """Return whether secondary UoM column should be hidden on reports."""
+        self.ensure_one()
+        if not self.secondary_uom_id:
+            return True
+        report_type = self._get_secondary_uom_report_type()
+        if not report_type:
+            return True
+        if report_type == "purchase":
+            return self.company_id.hide_secondary_uom_column_purchase
+        return self.company_id.hide_secondary_uom_column_sale
```
yostashiro: If we consider the possibility of someone adding a report type (which is not likely to happen).
```suggestion
        if report_type == "purchase":
            return self.company_id.hide_secondary_uom_column_purchase
        if report_type == "sales":
            return self.company_id.hide_secondary_uom_column_sale
        return True
```

product_secondary_unit/models/product_secondary_unit_mixin.py
```diff
@@ -146,3 +147,56 @@ def default_get(self, fields_list):
...
+        """Return display mode for secondary UoM price on reports."""
+        self.ensure_one()
+        if not self.secondary_uom_id:
+            return "primary"
+        report_type = self._get_secondary_uom_report_type()
+        if not report_type:
+            return "primary"
+        if report_type == "purchase":
+            return self.company_id.secondary_uom_price_display_purchase
+        return self.company_id.secondary_uom_price_display_sale
```
yostashiro: Same as above.

product_secondary_unit/models/res_company.py
```diff
@@ -0,0 +1,51 @@
...
+            record._name == "account.move"
+            and record.is_sale_document(include_receipts=True)
+        ):
+            return self.hide_secondary_uom_column_sale
+        if record._name == "purchase.order" or (
+            record._name == "account.move"
+            and record.is_purchase_document(include_receipts=True)
+        ):
+            return self.hide_secondary_uom_column_purchase
+        return True
```
yostashiro: Can we move this to the mixin model?

yostashiro (approved):
Code review. LGTM.

AungKoKoLin1997:
@sergio-teruel Can you please review this PR changes?

rousseldenis (approved):

rousseldenis:
@sergio-teruel As you are maintainer of this, please feedback, thanks

mav-adhoc:
@AungKoKoLin1997 Can you update the manfiest? Thanks! 
image

AungKoKoLin1997:
> Can you update the manfiest? Thanks!

@mav-adhoc I bump the minor version.

AungKoKoLin1997:
> I bump the minor version.

@mav-adhoc I reverted the changes because the version bump should be handled by OCA Bot during the merge, and manually bumping the version could cause conflicts if multiple PRs are merged for the same module.

Regarding your issue, I believe upgrading the module from the command line should resolve it.

AungKoKoLin1997:
The failing CI because of `product_pricelist_assortment` is not related with current PR changes.

HviorForgeFlow (approved):
Code review 👍

mav-adhoc:
LGTM

HviorForgeFlow:
@rousseldenis could we prioritize merging this, I merged a PR that is working on the top of this new changes, I didn't realize, and now we are blocking many edge users.

https://github.com/OCA/purchase-workflow/pull/2932

Reyes4711-S73 (approved):
LGTM

HviorForgeFlow:
@pedrobaeza @legalsylvain could you merge this? Appreciate

pedrobaeza (changes requested):
Sorry, but the text putting that the previous code fails for certain conversion is not enough for me. In fact, there are no regression tests for those cases. I only see that in the test, the base UoM is changed, making me suspect that something is altering the result of the previous tests.

@AungKoKoLin1997 can you explain the problems? This is also mixing something that is not a fix and the main PR subject. It's OK having it in a different commit, but both should be discussed separately IMO. Starting with, this is only used for sales/purchase, so they should be in their respective modules. OK to have the common selection options here, but not the fields, as if you install only this module and sale, you will see useless options in the configuration. Or am I misunderstanding anything?

AungKoKoLin1997:
@pedrobaeza 
In my first commit, I added a new test case for the UOM conversion fix.

> OK to have the common selection options here, but not the fields, as if you install only this module and sale, you will see useless options in the configuration.

I added non-stored fields to hide the options if the related modules aren’t installed, preventing user confusion.
Could you please review?

pedrobaeza:
I still don't like that. Please add them directly in `sale_secondary_unit` / `purchase_secondary_unit`, not here. And please split in 2 PRs for discussing both separatetly.

AungKoKoLin1997:
@pedrobaeza 
The reason I’m adding these fields in this module (instead of `sale_order_secondary_unit` and `purchase_order_secondary_unit`) is that I’m planning to introduce an [account_move_secondary_unit](https://github.com/OCA/account-invoicing/pull/2230) module. In that module, I’d like to reuse the Sale/Purchase configurations depending on the document type (customer invoice vs. vendor bill).

If the configuration lives only in the `sale_secondary_unit` / `purchase_secondary_unit`, then `account_move_secondary_unit` would likely need to depend on both modules, which I’d prefer to avoid if possible.

So I’m leaning toward adding these fields in `product_secondary_unit`, so the configuration can be shared and reused across Sale, Purchase, and Invoicing.

pedrobaeza:
OK, please open a separate PR for that feature, and explain in commit message and in comments that intention and the reason for that. I personally don't like that configurations anyway, as if you install the modules, you should see them. I think you should go instead with security groups.

AungKoKoLin1997:
@pedrobaeza @yostashiro @rousseldenis @HviorForgeFlow @Reyes4711-S73 @kanda999 @nobuQuartile 
I made a separate PR for the report presentation settings. I also hide the settings when the related child modules aren’t installed by using security groups.

https://github.com/OCA/product-attribute/pull/2211

product_secondary_unit/models/product_secondary_unit_mixin.py
```diff
@@ -72,6 +73,17 @@ def _get_secondary_uom_qty_depends(self):
             return []
         return [self._secondary_unit_fields["qty_field"]]
 
+    @api.model
+    def _convert_qty_to_secondary_uom(self, record, qty):
```
pedrobaeza: Why using here `record` instead of `self`?

product_secondary_unit/tests/test_secondary_unit_mixin.py
```diff
@@ -15,14 +15,13 @@ def setUpClass(cls):
...
         cls.loader.update_registry((SecondaryUnitFake,))
-        cls.product_uom_kg = cls.env.ref("uom.product_uom_kgm")
         cls.product_uom_unit = cls.env.ref("uom.product_uom_unit")
         cls.product_uom_dozen = cls.env.ref("uom.product_uom_dozen")
         cls.product_template = cls.env["product.template"].create(
             {
                 "name": "test",
-                "uom_id": cls.product_uom_kg.id,
-                "uom_po_id": cls.product_uom_kg.id,
+                "uom_id": cls.product_uom_unit.id,
```
pedrobaeza, line 23: I still don't know why are you changing this test master data and this makes me think the behavior is being changed.

product_secondary_unit/models/product_secondary_unit_mixin.py
AungKoKoLin1997: I pass the target record explicitly instead of using `self` because this helper is meant to be callable from flows where `self` isn’t the target record.
Example: https://github.com/OCA/stock-logistics-warehouse/pull/2496

product_secondary_unit/tests/test_secondary_unit_mixin.py
AungKoKoLin1997: I made these changes because the code uses `uom.uom._compute_quantity()` to get the correct quantity when the order line UoM differs from the product’s `uom_id`, and that conversion only works within the same UoM category. The previous test data used `kg` (Weight) for the product UoM and `Units` for the line UoM—different categories—which caused the test to fail. This wouldn’t happen in standard Odoo due to domain restrictions, but the fake model doesn’t enforce that domain. So I updated the product UoM to `Units` to make the test data realistic.

product_secondary_unit/models/product_secondary_unit_mixin.py
pedrobaeza: You can always do the call with `self`, as it has to inherit from this mixin. If not, the architecture seems not OK, mixing things inheriting from the mixin and other not. On the PR you mention, why can't you do `move._convert_qty_to_secondary_uom(vals["quantity"])` instead of `self._convert_qty_to_secondary_uom(move, vals["quantity"])`?

product_secondary_unit/tests/test_secondary_unit_mixin.py
pedrobaeza: The main whole idea of all of this secondary unit thing is to have different category UoMs between the main and the secondary ones, so the conversion must work without using `uom.uom._compute_quantity()`.

AungKoKoLin1997: I think we may be talking about two different situations.
With my improvement, the behavior is the same, and we can have different category UoMs between the main and the secondary ones. What I mean above is the problem of having one UoM category in the product’s uom_id and a different UoM category on the order line (e.g., a sale order line or purchase order line). I believe that won’t happen in real operations, and Odoo standard order lines have a domain to prevent that. The previous tests were using different UoM categories for the product’s uom_id and the fake model’s product_uom_id. That’s why the tests failed. 
https://github.com/OCA/product-attribute/blob/188c4d511b8f789de01e078fc3fffaec474af9dc/product_secondary_unit/tests/test_secondary_unit_mixin.py#L24
https://github.com/OCA/product-attribute/blob/188c4d511b8f789de01e078fc3fffaec474af9dc/product_secondary_unit/tests/test_secondary_unit_mixin.py#L63

Am I missing your point?

product_secondary_unit/models/product_secondary_unit_mixin.py
AungKoKoLin1997: I updated.

product_secondary_unit/tests/test_secondary_unit_mixin.py
pedrobaeza: OK, understood. Thanks for the explanations.

product_secondary_unit/models/product_secondary_unit_mixin.py
```diff
@@ -72,6 +73,18 @@ def _get_secondary_uom_qty_depends(self):
...
 
+    @api.model
+    def _convert_qty_to_secondary_uom(self, qty):
+        # Intended to be called from other operations if needed.
+        self.ensure_one()
+        uom_line = self._get_uom_line()
+        uom_product = self.product_id[self._product_uom_field]
+        qty_base = uom_line._compute_quantity(qty, uom_product)
+        return float_round(
+            qty_base / (self.secondary_uom_id.factor or 1.0),
```
pedrobaeza: This is removing the condition `if self.product_id[self._product_uom_field] != uom_line` that is present in `_get_factor_line`. I think that one is important.

AungKoKoLin1997: I don't think it is needed anymore because `qty_base = uom_line._compute_quantity(qty, uom_product)` will cover that.

AungKoKoLin1997: I added a condition so we only call `uom_line._compute_quantity(qty, uom_product)` when necessary, which should slightly improve performance.

```
if uom_line != uom_product:
     qty = uom_line._compute_quantity(qty, uom_product)
```

pedrobaeza (approved):
/ocabot merge major

PR #2184, opened 2026-01-08
[FIX] product_code_unique - remove duplicate warning for "Internal Reference"

anthonissen-a:
Remove the standard warnings from Odoo when trying to save a duplicate "Internal Reference", because this module already introduces a constraint avoiding duplicates.

### DIFF
--- a/product_code_unique/models/product_product.py
+++ b/product_code_unique/models/product_product.py
@@ -1,7 +1,7 @@
 # Copyright (C) 2018 - TODAY, Open Source Integrators
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
 
-from odoo import models
+from odoo import api, models
 
 
 class ProductProduct(models.Model):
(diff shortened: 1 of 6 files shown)


rvalyi (commented):
doing it this way feels a bit dangerous as it bypasses the method chain entirely. Imagine some module is doing something important in _onchange_default_code, it may not be called anymore because super is not called...

What about calling super and removing the "warning" key from the super result instead?

product_code_unique/models/product_product.py
```diff
@@ -14,3 +14,7 @@ class ProductProduct(models.Model):
             "Internal Reference must be unique across the database!",
         )
     ]
+
+    @api.onchange()
+    def _onchange_default_code(self):
```
rousseldenis, line 19: You cannot do that as you are not aware of other modules overriding this

rousseldenis: @qgroulard

PR #2186, opened 2026-01-12
[19.0][MIG] product_multi_category: Migration to 19.0

bizzappdev:


BhaveshHeliconia (commented):
No test cases are needed for this module, as it only adds an ORM field.

PR #2187, opened 2026-01-12
[19.0][MIG] product_attribute_value_menu: Migration to 19.0

bizzappdev:


product_attribute_value_menu/tests/test_product_attribute_value.py
```diff
@@ -0,0 +1,76 @@
+from odoo.tests.common import TransactionCase
+
+
+class TestProductAttributeValue(TransactionCase):
```
BhaveshHeliconia, line 4:
```suggestion
class TestProductAttributeValue(Basecommon):
```

product_attribute_value_menu/tests/test_product_attribute_value.py
```diff
@@ -0,0 +1,76 @@
...
+    @classmethod
+    def setUpClass(cls):
+        super().setUpClass()
+        cls.env = cls.env(context=dict(cls.env.context, tracking_disable=True))
```
BhaveshHeliconia, line 8:
```suggestion
```

BhaveshHeliconia (changes requested):
Consider using BaseCommon instead of TransactionCase. In this case, there’s no need to redefine cls.env with tracking_disable=True is already available via BaseCommon. This would help simplify and optimize the test code.

product_attribute_value_menu/models/product_attribute_value.py
```diff
@@ -0,0 +1,37 @@
+# Copyright 2022 ForgeFlow, S.L.
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+from odoo import api, fields, models
+
+
+class ProductAttributeValue(models.Model):
+    _inherit = "product.attribute.value"
+
+    product_count = fields.Integer(string="Product", compute="_compute_product_count")
```
jans23, line 10: `    product_count = fields.Integer(compute="_compute_product_count")`

to avoid warning as in #2216

PR #2188, opened 2026-01-12
[19.0][MIG] product_abc_classification: Migration to 19.0

bizzappdev:


product_abc_classification/tests/common.py
```diff
@@ -0,0 +1,141 @@
+# Copyright 2021 ACSONE SA/NV
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+from odoo.fields import Command
+from odoo.tests.common import TransactionCase
+
+
+class ABCClassificationCase(TransactionCase):
```
BhaveshHeliconia, line 8:
```suggestion
class ABCClassificationCase(Basecommon):
```

product_abc_classification/tests/common.py
```diff
@@ -0,0 +1,141 @@
...
+    @classmethod
+    def setUpClass(cls):
+        super().setUpClass()
+        cls.env = cls.env(context=dict(cls.env.context, tracking_disable=True))
```
BhaveshHeliconia, line 12:
```suggestion
```

BhaveshHeliconia (changes requested):
Please use Command consistently. In some places you’re using Command.set(), while in others you’re still using [(6, 0, ...)] and [(0, 0, ...)]. It would be better to standardize on Command for consistency.

PR #2192, opened 2026-01-16
[18.0][MIG] product_next_reception_date

bealdav:
my_product


product_next_reception_date/models/product_product.py
flotho, line 1: Hi, 
thanks for this usefull module.

IMHO things could be improved.
In the search method you wrote :
* Searching on stock.move will be more efficient as product is directly store on the object. Manipulate picking will trigger the ORM the need of computing. On big recordset you'll face performance issues.
* Using limit  = 1  will return a single record and you *need* to have records for **all** the product.product you're manipulating. using limit=1 don't seems to me correct.
* 
As you need to iterate on all the ids of self (product.product) you're manipulating, I don't think that limiting to one here will provide you the expecting result.

flotho (changes requested):

product_next_reception_date/models/product_product.py
bealdav: Hi @flotho,

Oops you're right why I let 'limit=1' with a 'in' domain ? Thanks

About product_id on stock.picking, I never used this field in my own code (except this migration) because it smells weird. But if there are performance issues on it I ask myself why it exists. It's error prone.

product_next_reception_date/models/product_product.py
```diff
@@ -0,0 +1,32 @@
...
+                (
+                    "state",
+                    "in",
+                    ["waiting", "confirmed", "partially_available", "assigned"],
+                ),
+            ],
+            order="scheduled_date desc",
+        )
+        # because of order by desc the last move by product is the earlier
+        move_dict = {move.product_id: move.scheduled_date for move in moves}
```
flotho: Thanks for the already done corrections.
This structure seems not correct to me.
In `moves` you could have multiple line per product. 
So in the move_dict, you need to match on the product_id otherwise the data won't be correct.

flotho (changes requested):

product_next_reception_date/models/product_product.py
bealdav: Last commits are variations to evaluate date in python. I'll keep only one commit.
Yes I can have several products as result of move query but only moves with lower schedule date must be selected.

flotho (approved):
Hi,

Thanks for the update.
Definitively it will provide the proper value.
Algorithm could maybe be improved but there's no more reason to block.
congrats !

mathieudelva (approved):
LGTM

florian-dacosta (approved):
LGTM
The module could be improved to be more multi warehouse friendly, but that may be improvement for the future

florian-dacosta:
@bguillot Can you merge ?

bguillot:
Thanks

bealdav:
Thanks a lot

PR #2196, opened 2026-01-29
[18.0][ADD] product_variant_search

AungKoKoLin1997:
This module improves product variant searching by matching the product display name across all product selection fields.

@qrtl QT6354


product_variant_search/models/product_product.py
```diff
@@ -0,0 +1,49 @@
...
+
+class ProductProduct(models.Model):
+    _inherit = "product.product"
+
+    variant_search_text = fields.Text(
+        compute="_compute_variant_search_text",
+        store=True,
+        translate=True,
+        index=True,
+    )
```
AungKoKoLin1997, line 16: I didn’t include this field in the search view since users can already search by attribute values.
However, I can add it if the community considers it useful.

Also, I’m unsure whether this field should be set to `readonly=False` to allow manual editing, as it could improve usability.
However, any custom value would be overwritten whenever one of the dependent fields changes, although this may be a rare case.

AungKoKoLin1997:
Tests failing is because of wrong translation for `ja.po` in odoo repo.
I think there is unescaped double quote at the start of the this translation line.
[unescaped double quote found](https://github.com/odoo/odoo/blob/71b178c8467936e4d21c90b6b32a85a931e48273/odoo/addons/base/i18n/ja.po#L3285)

yostashiro:
@AungKoKoLin1997 I think the PR should be created in https://github.com/OCA/product-variant instead.

product_variant_search/models/product_product.py
```diff
@@ -0,0 +1,49 @@
...
+        for product in self:
+            variant = (
+                product.product_template_attribute_value_ids._get_combination_name()
+            )
+            base_name = product.name
+            name = f"{base_name} ({variant})" if variant else base_name
+            if product.default_code:
+                product.variant_search_text = f"[{product.default_code}] {name}"
+            else:
+                product.variant_search_text = name
```
yostashiro, line 33: Any problem with doing this?
```suggestion
            product.variant_search_text = product.display_name
```

yostashiro:
> I think there is unescaped double quote at the start of the this translation line.

Odoo's translation team seems to have taken care of this in Weblate and the fix should land over the coming weekend.

product_variant_search/models/product_product.py
```diff
@@ -0,0 +1,49 @@
...
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+
+from odoo import api, fields, models
+from odoo.osv import expression
+
+
+class ProductProduct(models.Model):
+    _inherit = "product.product"
+
+    variant_search_text = fields.Text(
```
yostashiro, line 11: This may fit better?
```suggestion
    variant_search_name = fields.Char(
```

product_variant_search/models/product_product.py
```diff
@@ -0,0 +1,49 @@
...
+                args,
+                [
+                    "|",
+                    ("variant_search_text", operator, name),
+                    ("name", operator, name),
+                ],
+            ]
+        )
+        records = self.search(domain, limit=limit)
+        return [(rec.id, rec.display_name or "") for rec in records]
```
yostashiro, line 49: Avoid a complete override. Can we just work on args and then return super()?

product_variant_search/__manifest__.py
```diff
@@ -0,0 +1,15 @@
+# Copyright 2026 Quartile (https://www.quartile.co)
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+{
+    "name": "Prduct Variant Search",
```
yostashiro, line 4:
```suggestion
    "name": "Product Variant Search",
```

product_variant_search/models/product_template.py
```diff
@@ -0,0 +1,39 @@
...
+        res = super().write(vals)
+        if "attribute_line_ids" in vals:
+            self._recompute_variant_search_text_all_langs()
+        return res
+
+    def _recompute_variant_search_text_all_langs(self):
+        langs = self.env["res.lang"].search([("active", "=", True)]).mapped("code")
+        variants = self.product_variant_ids
+        for lang in langs:
+            variants.with_context(lang=lang)._compute_variant_search_text()
```
yostashiro, line 29: I think we should move these to product.product. There is currently extra overhead for updating all variants in write().

yostashiro (changes requested):

AungKoKoLin1997:
I will create this module in https://github.com/OCA/product-variant.

product_variant_search/models/product_product.py
AungKoKoLin1997: Instead of compute stored field, I just use normal field with `readonly=False` because having compute method doesn't make sense for translate field and assigning multi-lang values when we create or write using another helper method.

PR #2124, opened 2025-11-05
[19.0][MIG] product_assortment: Migration to 19.0

bizzappdev:

PR #2202, opened 2026-02-05
[19.0][FIX] product_assortment: Fix demo data

lmarion-source:
This PR takes the PR : https://github.com/OCA/product-attribute/pull/2124

but modify the demo data since the ones chosen there no longer exist in Odoo 19.0


Alexgars73:
@lmarion-source can chou check the commits pls, there are some commits frome product_code_unique,product_net_weight ...

PR #2204, opened 2026-02-05
[19.0][ADD] product_uom_packaging: replaces removed product.packaging

mdurepos:
New module product_uom_packaging:

- Introduces product.uom.packaging model to bridge products, UoMs, and package types
- Allows product-specific package dimensions for different UoMs
- Provides package compatibility validation with weight/volume checks (non-blocking)
- Adds warning displays in stock operations when packages are incompatible

## Reason for Existence

In Odoo 19.0, the `product.packaging` model was removed and its functionality was
merged into the UoM (Unit of Measure) system. While this simplifies some workflows,
it creates a problem: **UoMs are shared across products**, but physical packaging
dimensions are product-specific.

**Example:** A "12-pack" UoM is shared, but:
- A 12-pack of bricks needs a 12x12x12 box
- A 12-pack of soda cans needs a 6x12x8 box

Without a way to link product + UoM to a specific package type, users cannot:
- Calculate accurate shipping costs (dimensions matter)
- Plan warehouse storage (pallet configurations vary by product)
- Generate accurate packaging labels

This module restores the product-specific packaging configuration that was lost
in the Odoo 19 migration. This should permit migrating modules such as
product_packaging_level to 19.0 as well as providing the above functionality.

## Further development

This module provides core functionality and should be shortly extended by at least two
other modules to allow the selection of package types on products in the Sales and Purchase 
tabs. This will then permit selecting the package type on sales orders/purchase orders and 
having those choices propagate to stock.pickings and their owned stock moves/stock move
lines once when the procurement group chain is intact.


mdurepos:
Noted some missing base functionality. Will create a new PR when ready.

PR #2206, opened 2026-02-06
[FIX] base_product_mass_addition: Trigger onchange for all new lines

SirPyTech:
Backporting https://github.com/OCA/product-attribute/pull/1604 from `15.0`.
@HaraldPanten @angelgarciadelachica @luis-ron you worked on the linked PR, maybe you are interested in this one too?

I also added another commit with a unit test for the added fix.

### DIFF
--- a/base_product_mass_addition/models/product_mass_addition.py
+++ b/base_product_mass_addition/models/product_mass_addition.py
@@ -67,7 +67,5 @@ class ProductMassAddition(models.AbstractModel):
             line = lines.filtered(lambda x: x.id == vals["id"])
             return line.play_onchanges(update_vals, list(update_vals.keys()))
         else:
-            line = lines
-            if len(lines) > 1:
-                line = lines[0]
+            line = self.env[lines._name]
             return line.play_onchanges(vals, list(vals.keys()))
--- a/base_product_mass_addition/README.rst
+++ b/base_product_mass_addition/README.rst
@@ -82,6 +82,9 @@ Akretion
 * `Camptocamp <https://www.camptocamp.com>`_
 
     * Iván Todorovich <ivan.todorovich@camptocamp.com>
+* `PyTech <https://www.pytech.it>`_:
+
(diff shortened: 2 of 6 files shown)


base_product_mass_addition/models/product_mass_addition.py
```diff
@@ -67,7 +67,5 @@ def _complete_quick_line_vals(self, vals, lines_key=""):
             line = lines.filtered(lambda x: x.id == vals["id"])
             return line.play_onchanges(update_vals, list(update_vals.keys()))
         else:
-            line = lines
-            if len(lines) > 1:
-                line = lines[0]
+            line = self.env[lines._name]
```
legalsylvain, line 70: Hi. I see that the test is correct. However, i can not understand why running play_onchanges in an empty recordset can work. Could you help here ? 
thanks !

SirPyTech: Sure, if `play_onchanges` is executed on an empty recordset, the changed values are the default ones:
```py
        # If self is a record (play onchange on existing record)
        # we take the value of the field
        # If self is an empty record we will have an empty value
        if self:
            self.ensure_one()
            record_values = self._convert_to_write(
                {
                    # use sudo to avoid access right issue
                    field_name: self.sudo()[field_name]
                    for field_name, field in self._fields.items()
                }
            )
        else:
            # We get default values, they may be used in onchange
            record_values = self.default_get(self._fields.keys())
```
(from https://github.com/OCA/server-tools/blob/57b21e9953a09cc2da5186131464110a5cf71df9/onchange_helper/models/base.py#L38-L52)

legalsylvain: oh. Thanks !

legalsylvain (approved):

francesco-ooops (approved):
Functional ok!

francesco-ooops:
@legalsylvain as it's a backporting, could it be merged? thanks!

PR #2210, opened 2026-02-11
[18.0][ADD] product_attribute_line_template: New module

pilarvargas-tecnativa:
@Tecnativa TT56921

@pedrobaeza @juancarlosonate-tecnativa please review


product_attribute_line_template/models/product_template.py
```diff
@@ -0,0 +1,41 @@
+# Copyright 2026 Tecnativa - Pilar Vargas
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+
+from odoo import Command, api, fields, models
+
+
+class ProductTemplate(models.Model):
+    _inherit = "product.template"
+
+    attribute_template_ids = fields.Many2many(
```
pedrobaeza: Why m2m? IMO, it should be an m2o that is emptied after selecting it, applying the template to the lines, in an cumulative way.

product_attribute_line_template/views/res_config_settings_views.xml
```diff
@@ -0,0 +1,31 @@
...
+<odoo>
+    <record
+        id="res_config_settings_view_form_inherit_attribute_templates"
+        model="ir.ui.view"
+    >
+        <field name="name">res.config.settings.inherit.attribute.templates</field>
+        <field name="model">res.config.settings</field>
+        <field name="inherit_id" ref="product.res_config_settings_view_form" />
+        <field name="arch" type="xml">
+            <xpath expr="//block" position="after">
```
pedrobaeza: Why not a menu instead?

pilarvargas-tecnativa: Regarding the menu, in this environment, the product module does not expose its own configuration menu to which we can attach this functionality. The available menus belong to other applications (for example, sale), which would mean adding an additional dependency just to be able to attach access from there.

As for the change from Many2many to Many2one, it was correct. I have updated the implementation to use a Many2one selector that applies the template in the onchange cumulatively and clears itself after being applied.

eduezerouali-tecnativa (approved):
code review + tested on local

eduezerouali-tecnativa:
@pilarvargas-tecnativa I only have one concern about this module, why not adding `value_ids` to `product.attribute.template.line` so when you select the template it propagate the values too and not only the attributes

pedrobaeza (approved):
/ocabot merge nobump

pilarvargas-tecnativa:
> @pilarvargas-tecnativa I only have one concern about this module, why not adding `value_ids` to `product.attribute.template.line` so when you select the template it propagate the values too and not only the attributes

I understand the idea, but I think the scope of the module should be limited to propagating the attributes defined in the template, not the specific values of each attribute.

If a template includes an attribute with, for example, 100 values or more, when applying the template you would then have to check those 100 values one by one to remove those that do not apply to the specific product. In that case, the template would no longer be helpful and could result in more manual work or an increased risk of incorrect values being left in place.

PR #2211, opened 2026-02-16
[18.0][IMP] product_secondary_unit: add configuration for report presentation

AungKoKoLin1997:
Add hidden security groups (`group_sale_secondary_unit`, `group_purchase_secondary_unit`) and company-level configuration fields for secondary unit price display and column visibility.

These groups and settings are defined here in the base module so they can be shared across Sale, Purchase, and Invoicing without forcing `account_move_secondary_unit` to depend on `sale_order_secondary_unit` or` purchase_order_secondary_unit`.

Each child module activates the relevant group(s) via `implied_ids` on install, which controls the visibility of the corresponding settings in the UI.

@qrtl QT6210


yostashiro (approved):
Code review. LGTM.

product_secondary_unit/models/res_company.py
```diff
@@ -0,0 +1,51 @@
...
+    ("primary", "Primary Unit Price Only"),
+    ("secondary", "Prioritize Secondary Unit Price"),
+    ("both", "Both Primary and Secondary Unit Prices"),
+]
+
+
+class ResCompany(models.Model):
+    _inherit = "res.company"
+
+    secondary_uom_price_display_sale = fields.Selection(
```
pedrobaeza, line 16: This configurability level seems a bit insane. Are you going to display UoM price on sale, but not on purchase, but show the UoM? Can't this be fit with a single security group "Show seconday unit prices"?

AungKoKoLin1997: For showing the UoM column option, it is intended only to preserve the existing Sales/Purchase report behavior of `sale_order_secondary_unit` and `purchase_order_secondary_unit` when `Primary Unit Price Only` is selected. For the other price display modes, the secondary UoM is already shown in the `quantity` and `price` columns, so showing an additional UoM column would be redundant. 

I don't think Sales and Purchase should always share the same display style for the secondary unit price and the secondary unit column. Therefore, I prefer separate settings for Sales and Purchase documents, with separate security groups to control each document type.

AungKoKoLin1997: This is the expected report when `Primary Unit Price Only` is selected without hiding the UoM column.
image


This is the expected report when `Prioritize Secondary Unit Price` is selected.
image


This is the expected report when `Both Primary and Secondary Unit Prices` is selected.
image

kanda999 (approved):
Confirmed that the report display changes depending on the option settings in my local

nobuQuartile (approved):
Functional review LGTM:
Check the report behaviour with the related purchase_order_secondary_unit PR code.

product_secondary_unit/models/res_company.py
pedrobaeza: But as said, this is too many configurations, while it's not a usual problem to show both. This only increases the maintenance burden to the main module being a residual feature that most people won't use, being enough with the default option. I think this has to be simplified.

yostashiro: @pedrobaeza To me it makes sense to be able to configure sales and purchasing separately as they have different needs. But we can do the secondary UoM price display thing in a separate module if it's too much to be covered in the main module, which will mean four new additional modules (i.e., product_secondary_unit_price, purchase_order_secondary_unit_price, sale_order_secondary_unit_price and account_move_secondary_unit_price).

What we still want to do in this module, though, is to be able to hide the secondary qty column in reports. For this, we don't need to have separate config for sales and purchasing (as we will most likely disable this for both anyway).

Does this sound like a good approach?

pedrobaeza: @sergio-teruel is the main creator of this module, so I let him to decide at last if he considers OK to have all the configuration tweaks in the main module.

dreispt: I also came across this need: to choose if the secondary unit should be visible on sales document or purchase documents.
It can be the case that the feature helps pick quantities on order lines, but we don't want/need them on the PDFs.
And the decision can be different for sale or purchase.

dreispt: @AungKoKoLin1997 That's a feature change and should go in a separate commit, so that it can be backported. Thanks.

AungKoKoLin1997: @dreispt For easier backporting, I’ll split this into two commits:

- Commit 1: Hide column configuration (hide_secondary_uom_column_* fields + related logic)
- Commit 2: Price display configuration (secondary_uom_price_display_* fields + related logic)

AungKoKoLin1997: @dreispt I split into two commits.

AungKoKoLin1997: > @sergio-teruel is the main creator of this module, so I let him to decide at last if he considers OK to have all the configuration tweaks in the main module.

@sergio-teruel Can you please review the changes in this PR? We’re planning to use this module for our customer and would like to proceed. If you think the current changes aren’t suitable, I can follow up based on @yostashiro suggestion above.

product_secondary_unit/models/res_config_settings.py
```diff
@@ -0,0 +1,17 @@
...
+    _inherit = "res.config.settings"
+
+    hide_secondary_uom_column_sale = fields.Boolean(
+        related="company_id.hide_secondary_uom_column_sale",
+        readonly=False,
+    )
+    hide_secondary_uom_column_purchase = fields.Boolean(
+        related="company_id.hide_secondary_uom_column_purchase",
+        readonly=False,
+    )
```
dreispt, line 17: Clarify me please: 
We might not have Sales or Purchase installed.
Will these options work if that is the case, or should they be moved to the corresponding app extension modules?

AungKoKoLin1997: Yes, these options will work for Accounting even if Sales or Purchase is not installed. These are intended for use in Sales, Purchase, and Accounting. I did not add them to the corresponding extension modules because, if they were moved there, [account_move_secondary_unit](https://github.com/OCA/account-invoicing/pull/2230) would need to depend on those extension modules. I wanted to avoid that.

AungKoKoLin1997:
@sergio-teruel @pedrobaeza @dreispt
This PR has been open for a while and hasn’t been able to proceed. Do you have any concerns about the design of this PR?
If you don’t have any concerns or feedback, I’d like to move forward with it.

yostashiro:
@sergio-teruel @OCA/product-maintainers 

This PR has been pending a decision for some time, and I’d like to ask the maintainer or the PSC to make a decision on whether to accept or reject it.

If the proposal is rejected, we will take an alternative approach and propose a new module, such as product_secondary_unit_report_config. However, this would likely result in additional dependent modules for purchase, sales, accounting, and possibly other areas. I’d prefer to avoid this path, as it would add more maintenance burden to the community.

The following PRs are already affected or blocked due to the unclear status of this PR, and more may be added as the 19.0 migration progresses:

https://github.com/OCA/purchase-workflow/pull/2950
https://github.com/OCA/sale-workflow/pull/4152
https://github.com/OCA/sale-workflow/pull/4175
https://github.com/OCA/account-invoicing/pull/2230
https://github.com/OCA/account-invoice-reporting/pull/410
https://github.com/OCA/sale-workflow/pull/4114
https://github.com/OCA/l10n-japan/pull/105

We are okay with either outcome, but leaving the decision pending without a clear timeline is making it difficult to proceed with the related work.

Thank you very much for your attention.

BhaveshHeliconia (approved):
LGTM!

mav-adhoc (commented):
Review on the `product_secondary_unit` side. Two points are mechanically blocking, two are design notes.

### 1. The PR does not merge, and the cause is the generated files

`mergeable_state` is `dirty`. Merging `917b33b` onto `18.0` (`d06c571`) gives exactly one conflict: `product_secondary_unit/static/description/index.html`.

`README.rst` and `static/description/index.html` are produced by `oca-gen-addon-readme` and rewritten by the bot after merge (any `[BOT] post-merge updates` commit, e.g. `eb35fcd`, touches `README.rst`, `static/description/index.html` and `__manifest__.py`). I dropped both files from the two commits and merged again: 0 conflicts. The hand-written parts you actually need, `readme/CONFIGURE.md` and `readme/CONTRIBUTORS.md`, are already in the PR.

Same for the version: the bot bumps `__manifest__.py`, so nothing to add there.

### 2. No test for the new code

`codecov/patch` and `codecov/project` are the only red checks. The PR adds ~130 lines of Python (`res_company.py`, `res_config_settings.py`, and 53 lines in the mixin) without a test. The module already has the scaffolding: `tests/models.py` defines `secondary.unit.fake` via `FakeModelLoader`, so the defaults and the fallback branches are cheap to cover. The sale/purchase branches are not coverable here, which is the same problem as point 4 below.

### 3. `res.company.hide_secondary_uom_column()`

* No `ensure_one()`, while every other method added in this PR has one. `doc.company_id` is a singleton in the templates of the child PRs, so it does not break today, but this is a public method on `res.company`.
* It re-implements the sale/purchase document detection that `_get_secondary_uom_report_type()` already does in the mixin. Two copies of the same rule, one at document level and one at line level, that have to be kept in sync by hand.

### 4. Model names hardcoded in the base module (follow-up to @dreispt's question)

`_get_secondary_uom_report_type()` and `hide_secondary_uom_column()` hardcode `sale.order`, `sale.order.line`, `purchase.order`, `purchase.order.line` and `account.move(.line)` in a module whose only dependency is `product`.

Making `_get_secondary_uom_report_type()` a hook that returns `None` here and is overridden in `sale_order_secondary_unit`, `purchase_order_secondary_unit` and `account_move_secondary_unit` would:

* keep the settings and the groups shared in the base module, which is the point of the PR;
* drop the knowledge of models the base module does not depend on;
* make the sale and purchase branches testable in the modules that own those models, which is where the coverage gap of point 2 is;
* answer @dreispt's question directly, since the base module stays consistent with neither Sales nor Purchase installed.

### Question on `report_show_price_uom()`

With `secondary_uom_price_display_* = 'secondary'` and `hide_secondary_uom_column_* = True`, which is the combination the help text of `Hide Secondary Qty Column` recommends, `report_show_price_uom()` evaluates `not hide_col or display_mode == 'both'` and returns `False`. In `purchase_uom_price` that drops the `/ <secondary unit>` suffix next to `secondary_uom_price`, and the Second Qty column is hidden as well. Is that intended? I read the code and the templates, I did not render the report, so I may be missing something.

PR #2212, opened 2026-02-17
[17.0][OU-ADD] product_pricelist_supplierinfo: Migration scripts

pedrobaeza:
When coming from previous versions, where the discount at product.supplierinfo were not taken into account to compute the sales price, you would expect the same for this version, so for this smooth transition, let's set the new flag for ignoring the discount in all the existing lines for preserving the previous behavior, as well as setting it as default for new items.

@Tecnativa TT61036

### DIFF
--- a/product_pricelist_supplierinfo/migrations/17.0.2.0.0/post-migration.py
+++ b/product_pricelist_supplierinfo/migrations/17.0.2.0.0/post-migration.py
@@ -1,0 +1,12 @@
+# Copyright 2026 Tecnativa - Pedro M. Baeza
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl-3.0)
+from openupgradelib import openupgrade
+
+
+@openupgrade.migrate()
+def migrate(env, version):
+    openupgrade.logged_query(
+        env.cr,
+        "UPDATE product_pricelist_item SET no_supplierinfo_discount=True",
+    )
+    env["ir.default"].set("product.pricelist.item", "no_supplierinfo_discount", "true")


product_pricelist_supplierinfo/migrations/17.0.2.0.0/post-migration.py
```diff
@@ -0,0 +1,12 @@
...
+from openupgradelib import openupgrade
+
+
+@openupgrade.migrate()
+def migrate(env, version):
+    openupgrade.logged_query(
+        env.cr,
+        "UPDATE product_pricelist_item SET no_supplierinfo_discount=True",
+    )
+    env["ir.default"].set("product.pricelist.item", "no_supplierinfo_discount", "true")
```
luisg123v, line 12: Are you sure about this?

I think this could be a source of functional discrepancies, new lines will start having this default but only in migrated databases that went through this migration script. What do you think about setting as default at field declaration instead?

luisg123v (changes requested):

product_pricelist_supplierinfo/migrations/17.0.2.0.0/post-migration.py
pedrobaeza: Yes, I'm sure, as this is only for migrated DBs. Users of these DBs are used to create pricelist items without having to mark any check for having that behavior, and if this default is not put, they will have differences between the old and new pricelist items. If they are aware, they will ask for removing it, or they will uncheck it when needed.

pedrobaeza: If changing this default now, current users of 17 will have a change of behavior comparable to the migrated users.

pedrobaeza: @luisg123v please update your review.

luisg123v: I don't fully agree with this, but I don't want to block it.

luisg123v (approved):

luisg123v:
Sorry for the delay, I was not set as a reviewer and it went out of my radar.

PR #2223, opened 2026-02-25
[15.0][ADD] product_pricelist_discount_by_range: New module

Andrii9090-tecnativa:
This module allows you to define discount ranges in a pricelist based  on a reference price (e.g., cost).
Instead of applying a single discount percentage to all products, you can configure different discount percentages for specific price intervals.
If the reference price is out of the ranges, the pricelist will not be applied  and Odoo will show a warning.

image

@Tecnativa

@pedrobaeza @christian-ramos-tecnativa can you review, please?
TT61165


product_pricelist_discount_by_range/models/product_pricelist_item.py
```diff
@@ -0,0 +1,36 @@
...
+    discount_range_ids = fields.One2many(
+        "product.pricelist.discount.range",
+        "pricelist_item_id",
+    )
+
+    def _compute_price(self, price, price_uom, product, quantity=1.0, partner=False):
+        computed_price = super()._compute_price(
+            price, price_uom, product, quantity, partner
+        )
+        if self.discount_type == "range" and self.discount_range_ids:
```
pedrobaeza: You want error in this case

```suggestion
        if self.discount_type == "range":
```

product_pricelist_discount_by_range/models/product_pricelist_item.py
```diff
@@ -0,0 +1,36 @@
...
+            discount = 0.0
```
pedrobaeza: A 0 discount is a valid value, so init this to `None` and do that comparison.

product_pricelist_discount_by_range/models/product_pricelist_discount_rage.py
```diff
@@ -0,0 +1,49 @@
...
+        "product.pricelist.item", string="Pricelist Item", ondelete="cascade"
+    )
+    min = fields.Float(required=True)
+    max = fields.Float(required=True)
+    percentage = fields.Float(required=True)
+
+    @api.constrains("min", "max", "percentage", "pricelist_item_id")
+    def _check_range(self):
+        for record in self:
+            if record.min > record.max:
```
pedrobaeza: This is convenient to be done by SQL constraint.

product_pricelist_discount_by_range/models/product_pricelist_discount_rage.py
```diff
@@ -0,0 +1,49 @@
...
+                raise ValidationError(
+                    _("The minimum cannot be greater than the maximum.")
+                )
+            if record.min < 0 or record.max < 0:
+                raise ValidationError(_("Values must be non-negative."))
+            if record.min >= record.max:
+                raise ValidationError(_("Min value must be less than max value."))
```
pedrobaeza: Duplicated constraint and the other may be possible. Negative values can be put for example on product discounts.

pedrobaeza (commented):
- Please remove the label on the one2many field
- Is there any way to have the radio selection values in one line?
- For the name of the child model, it's a good guideline to start with the name of the parent model, so you can call it `product.pricelist.item.discount_range`, and place it inside the same Python file for better discoverability.
- You have to reset `discount` field to 0 if using ranges, or the call to super will produce an incorrect returned price.

pedrobaeza:
I'm also wondering if it's better to do first the discount and then the call to super...

product_pricelist_discount_by_range/models/product_pricelist_discount_rage.py
Andrii9090-tecnativa: Here I am only checking min and max; prices cannot have a negative value.

Andrii9090-tecnativa:
> I'm also wondering if it's better to do first the discount and then the call to super...

I changed it and now it runs `super()` at the end to already have the discounted price by range and then follow the flow.

product_pricelist_discount_by_range/models/product_pricelist_item_discount_range.py
christian-ramos-tecnativa, line 1: The file should be called product_pricelist_item_discount_ra**n**ge.py

christian-ramos-tecnativa (changes requested):

product_pricelist_discount_by_range/models/product_pricelist_item_discount_range.py
```diff
@@ -0,0 +1,50 @@
...
+    )
+    min = fields.Float(required=True)
+    max = fields.Float(required=True)
+    percentage = fields.Float(required=True)
+
+    _sql_constraints = [
+        ("min_gt_max", "CHECK(min < max)", "Min value must be less than max value."),
+        (
+            "negartive_values",
+            "CHECK(min > 0 AND max > 0)",
```
christian-ramos-tecnativa: Can we do this less restrictive for the min?

```suggestion
            "CHECK(min >= 0 AND max > 0)",
```
When I tried to create the range the first thing I did was settign the min to zero and max to a higher value

product_pricelist_discount_by_range/models/product_pricelist_item.py
```diff
@@ -0,0 +1,39 @@
...
+            for discount_range in self.discount_range_ids:
+                if discount_range.min <= price <= discount_range.max:
+                    discount = discount_range.percentage
+                    break
+            if discount is not None:
+                price = price * (1 - discount / 100)
+            else:
+                raise ValidationError(
+                    _("Discount range not found for the given price.")
+                )
```
christian-ramos-tecnativa, line 42: Im not sure about this ValidationError, IMO if the price is out of range we should call the super or make posible for the user to set an infinite:
Setup 1:
- min 0.01 max 5 disc 5%
- min 5.01 max 10 disc 10%

If we create a sale line with this pricelist and a product above 10 the raise will not allow you to confirm the sale

Setup 2:
- min 0.01 max 5 disc 5%
- min 5.01 max 10 disc 10%
- min 10.01 max 100000 disc 0%

With this setup we cover the previus situation but setting a max value this big is not user friendly.
I think we should remove the raise or treate the zeros as infinites and add it to the help of each column:
min: 0 = -infinite
max: 0 = infinite
WDYT

christian-ramos-tecnativa (changes requested):

product_pricelist_discount_by_range/models/product_pricelist_item.py
Andrii9090-tecnativa: The idea here is to notify the user that a price list with discount ranges is being used and that there is no range for this price. 

It might be better to issue a warning that is not blocking, but in this case, call it a super...

Andrii9090-tecnativa:
@pedrobaeza can you review?

product_pricelist_discount_by_range/models/product_pricelist_item.py
```diff
@@ -0,0 +1,39 @@
...
+        [("simple", "Simple"), ("range", "By range")],
+        default="simple",
+    )
+    discount_range_ids = fields.One2many(
+        "product.pricelist.item.discount_range",
+        "pricelist_item_id",
+    )
+
+    @api.onchange("discount_type")
+    def _onchange_discount_type(self):
```
pedrobaeza: This only works on interactive edition, so better to convert the field to computed writable to do it always.

product_pricelist_discount_by_range/models/product_pricelist_item.py
pedrobaeza: No, it should be an explicit error. You want the flow to stop, not to continue with incorrect data, which being prices, is critical.

product_pricelist_discount_by_range/models/product_pricelist_item_discount_range.py
```diff
@@ -0,0 +1,50 @@
...
+        "product.pricelist.item", string="Pricelist Item", ondelete="cascade"
+    )
+    min = fields.Float(required=True)
+    max = fields.Float(required=True)
+    percentage = fields.Float(required=True)
+
+    _sql_constraints = [
+        ("min_gt_max", "CHECK(min < max)", "Min value must be less than max value."),
+        (
+            "negartive_values",
```
pedrobaeza:
```suggestion
            "negative_values",
```

But I insist, the prices can be negative in specific user cases, so remove this constraint, please. Simply don't limit this.

product_pricelist_discount_by_range/models/product_pricelist_item_discount_range.py
```diff
@@ -0,0 +1,50 @@
...
+            "CHECK(min > 0 AND max > 0)",
+            "Values must be non-negative.",
+        ),
+    ]
+
+    @api.constrains("min", "max", "percentage", "pricelist_item_id")
+    def _check_range(self):
```
pedrobaeza: Be more verbose and explicit on the method name:

```suggestion
    def _check_ranges_overlap(self):
```

pedrobaeza (commented):
The label with the help in blue disappear when you select by range. I think it should be preserved and adapted.

Andrii9090-tecnativa:
> The label with the help in blue disappear when you select by range. I think it should be preserved and adapted.

I adapted this helper
image

christian-ramos-tecnativa (approved):

pedrobaeza (approved):
/ocabot merge nobump

Please do Spanish translations in Weblate before deploying.

PR #2230, opened 2026-03-04
[IMP] product_cost_security: hide standard_price column in product tree views

emtz10:
Adds the 'product_cost_security.group_product_cost' group restriction to the 'standard_price' field in both product.product and product.template tree views.

This ensures that users without the explicit cost permission cannot visualize the cost column when triggering the "Search More" popup from sale orders or when accessing the default list views, strictly enforcing the visibility criteria.

### DIFF
--- a/product_cost_security/views/product_views.xml
+++ b/product_cost_security/views/product_views.xml
@@ -13,4 +13,28 @@
             </xpath>
         </field>
     </record>
+
+    <record id="product_template_tree_view" model="ir.ui.view">
+        <field name="model">product.template</field>
+        <field name="inherit_id" ref="product.product_template_tree_view" />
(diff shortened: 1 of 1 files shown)


hugho-ad (approved):

yajo:
Wouldn't it be better to just add the group in the model field?

emtz10:
> Wouldn't it be better to just add the group in the model field?

The group has already been added to the field.

https://github.com/OCA/product-attribute/blob/18.0/product_cost_security/models/product_product.py#L11

https://github.com/OCA/product-attribute/blob/18.0/product_cost_security/models/product_template.py#L11

yajo (approved):
Ah OK sorry, I thought that was implicit.

/ocabot merge minor

Issue #2115, opened 2025-10-24
[18.0][product_cost_security] Unable to validate incoming stock pickings without product cost read rights


uvid-sinisa:
## Module
product_cost_security

## Describe the bug
The module prevents validation of stock pickings (for example, Purchase Receipts) when installed for users without read access to product cost fields.
Inventory Users should be able to validate stock pickings even if they are not allowed to see product costs.

This issue appears to  be caused by changes in the Stock Valuation Layer (SVL) logic introduced in Odoo 18. During validation, SVL lines are generated, and the standard_price field is accessed in stock_account/models/stock_move.py, method _get_in_svl_vals:

```
unit_cost = {self.env['stock.lot']: move.product_id.standard_price}
```

Since standard_price requires read access on the product cost, this raises an AccessError for users without the required rights when product_cost_security is installed.
It seems that this call should likely use sudo() or an adjusted approach to bypass cost read restrictions during SVL valuation. However, i am not sure what changes should be applied within the product_cost_security module to prevent this error.

## To Reproduce
**Affected versions**: 18.0.1.0.0

Dependencies: stock_account installed

1. Use user without read or edit access to product costs.
2. Create a new Receipt (incoming stock picking) with any storable product.
3. Attempt to validate the picking.

Result:
Validation fails with an access error when reading product_id.standard_price.

**Expected behavior**
Stock picking validation (e.g. incoming stock picking) should succeed for users without product cost access rights. The module should not block standard warehouse operations.

<img width="1286" height="872" alt="Image" src="https://github.com/user-attachments/assets/254069cb-526f-4f02-ad56-b7b24cfe1f31" />

devravexyz:
Same issue in Sales — User can not confirm invoice

Joaco1980:
Same issue, users without read or write permissions on the cost price field cannot perform any warehouse movements.

rolandojduartem:
any advance about this?

rolandojduartem:
@pedrobaeza

pedrobaeza:
@carlos-lopez-tecnativa can you check?

carlos-lopez-tecnativa:
Sorry, but I cannot reproduce the issue with my settings.

Product category
<img width="1295" height="754" alt="Image" src="https://github.com/user-attachments/assets/cd7f71cd-c157-4bea-8fbd-2c7c0d21b5ba" />

Product
[...]

PR #2232, opened 2026-03-04
[18.0][ADD] product_cost_security_stock_account: New glue module

CarlosRoca13:
This glue module aims to avoid errors during the svl creation that needs to check the standard_price field value.

cc @Tecnativa

ping @carlos-lopez-tecnativa @carlosdauden


CarlosRoca13:
Module added as specified in Issue https://github.com/OCA/product-attribute/issues/2115.

CarlosRoca13:
I will add a test to check it

product_cost_security_stock_account/models/stock_move.py
```diff
@@ -0,0 +1,14 @@
+# Copyright 2026 Tecnativa - Carlos Roca
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+
+from odoo import models
+
+
+class StockMove(models.Model):
+    _inherit = ["stock.move"]
```
carlos-lopez-tecnativa:
```suggestion
    _inherit = "stock.move"
```

carlos-lopez-tecnativa (changes requested):
Tested and working when the user validates the picking.
@uvid-sinisa, @Joaco1980 @devravexyz @rolandojduartem @yajo could you please review this?

Another place where I think we should cache the value with sudo is when the user tries to change the product category. Odoo tries to generate a stock.valuation.layer, but an error is displayed. I think we should intercept this function [_svl_empty_stock](https://github.com/odoo/odoo/blob/d62445650c23353cb6964b7ce54e2d44572e1576/addons/stock_account/models/product.py#L678). What do you think?

Image

CarlosRoca13:
The only way I managed to avoid the error was by caching the value in the `_prepare_out_svl_vals` and `_prepare_in_svl_vals` methods.

product_cost_security_stock_account/tests/test_product_cost_security_stock_account.py
```diff
@@ -0,0 +1,109 @@
...
+from odoo.tests import new_test_user, users
+
+from odoo.addons.base.tests.common import BaseCommon
+
+
+class TestProductCostSecurityStockAccount(BaseCommon):
+    @classmethod
+    def setUpClass(cls):
+        super().setUpClass()
+        cls.user = new_test_user(
```
yajo: issue: IIUC this test [is flaky](https://github.com/OCA/odoo-community.org/blob/master/website/Contribution/CONTRIBUTING.rst#test-with-the-lowest-permissions-possible)


```suggestion
        cls.uid = new_test_user(
```

product_cost_security_stock_account/readme/DESCRIPTION.md
```diff
@@ -0,0 +1,5 @@
+Glue auto installable module between 'Product Cost Security' and 'Stock
+account' modules.
+
+It avoids errors during the svl creation that needs to check the
+standard_price field value.
```
yajo: suggestion: IMHO this description is too technical.


```suggestion
Make 'Product Cost Security' and 'Stock account' modules compatible.

Avoids permission errors during the stock valuation layer creation, which
would happen when stock users (with no access to product costs) tried to
perform valid stock-related operations, like creating stock valuation layers.
```

yajo (changes requested):
Thanks, good job! Little remarks.

CarlosRoca13:
Changes done. Thanks @yajo

yajo (approved):

Joaco1980 (approved):

pedrobaeza:
Merging as @carlos-lopez-tecnativa's requests are satisfied and he's not here this week.

/ocabot merge nobump

PR #2243, opened 2026-03-20
[18.0][IMP] product_pricelist_supplierinfo: Return from _price_compute real price value

Andrii9090-tecnativa:
Instead of returning a dummy value for `_price_compute`, we calculate the correct supplier price and return it.
This improves compatibility with other modules

@pedrobaeza @victoralmau can you review please?

@Tecnativa

### DIFF
--- a/product_pricelist_supplierinfo/models/product_pricelist_item.py
+++ b/product_pricelist_supplierinfo/models/product_pricelist_item.py
@@ -34,14 +34,12 @@ class ProductPricelistItem(models.Model):
     )
 
     def _compute_price(self, product, quantity, uom, date, currency=None):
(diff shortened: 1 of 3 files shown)


pedrobaeza (commented):
AFAIK, this is not applying possible surcharges or margins according standard. Or is it done on other place?

Andrii9090-tecnativa:
In this case applied because product.pricelist.item._compute_base_price retrieves the correct supplier price via _price_compute on product.template

image
Test in runbot

pedrobaeza (approved):
OK then

victoralmau (approved):
The changes seem consistent.

PR #2022, opened 2025-07-10
[16.0][ADD] product_attribute_value_dependent_mixin

Kev-Roche:
This technical module was written to help developers depend on product attribute values on any model by inheriting from the mixin.

cc @chafique-delli , @sebastienbeau

PR #1423, opened 2023-09-26
[14.0][ADD] product_attribute_value_dependant_mixin

chafique-delli:

PR #2250, opened 2026-03-27
[18.0][MIG] product_attribute_value_dependent_mixin: Migration to 18.0

metaminux:
Migration of `product_attribute_value_dependent_mixin` module from 16.0 to 18.0.

**Source PRs (not merged in 16.0):**
- https://github.com/OCA/product-attribute/pull/2022

cc @Kev-Roche @mourad-ehm for review


metaminux:
Hello @bealdav and @florian-dacosta 

Can you review this migration PR ?

product_attribute_value_dependent_mixin/models/product_attribute_value_dependent_mixin.py
```diff
@@ -0,0 +1,87 @@
...
+
+    available_attribute_value_domain = fields.Binary(
+        compute="_compute_available_attribute_value_domain",
+    )
+
+    @api.depends("product_tmpl_id")
+    def _compute_available_product_ids(self):
+        for rec in self:
+            rec.available_product_ids = self.env["product.product"].search(
+                [("product_tmpl_id", "=", self.product_tmpl_id._origin.id)]
```
florian-dacosta: It feels really wrong,  rec should be used in search instead of self right ?
Also, it could be refactored to make one search before the loop and then construct a dict with product_tmpl_id as a key and the product_ids as value. Would be more efficient in case of recompute on multi record.

product_attribute_value_dependent_mixin/models/product_attribute_value_dependent_mixin.py
```diff
@@ -0,0 +1,87 @@
...
+    @api.depends(
+        "product_tmpl_id",
+        "product_tmpl_id.attribute_line_ids.value_ids",
+        "available_attribute_value_domain",
+    )
+    def _compute_available_attribute_value_ids(self):
+        for rec in self:
+            if rec.product_tmpl_id:
+                rec.available_attribute_value_ids = rec.product_tmpl_id.mapped(
+                    "attribute_line_ids.value_ids"
```
florian-dacosta: mapped is not necessary here, you could write `rec.product_tmpl_id.attribute_line_ids.value_ids`
Well, I guess both are correct.

florian-dacosta (changes requested):
Technical review, not test (not sure in which submodule it is used ?)

metaminux:
Last commit fixes several bugs in the original implementation and significantly improves the documentation and test coverage.

**Typo fix in model name**

`attribute.value.dependant.mixin` (French spelling) has been renamed to `attribute.value.dependent.mixin` (correct English spelling). Since this is an `AbstractModel` with no database table, no migration script is needed — consumers just need to update their `_inherit`.

**`available_product_ids` replaced by `available_product_domain`**

The original `available_product_ids` Many2many compute field had several bugs:

- `self.product_tmpl_id` was used instead of `rec.product_tmpl_id` inside the `for rec in self` loop, causing incorrect behavior on multi-record recordsets.
- `_origin` was used unnecessarily and caused the domain `[("product_tmpl_id", "=", False)]` to be applied when `product_tmpl_id` was not set, matching all products without a template.
- The missing `if rec.product_tmpl_id` guard meant the field was never properly assigned in the `False` case.
- Performing a `search()` inside a compute is expensive and unnecessary since `product_tmpl_id.product_variant_ids` already holds the variants as a stored field.

The field has been replaced by `available_product_domain` (a `Binary` field following the same pattern already used for `available_attribute_value_domain`), and the `domain` attribute has been removed from `product_id` in the model — it belongs in the view, where the consuming module can reference `available_product_domain` explicitly. The `@api.depends` chain has also been simplified: `"product_tmpl_id.product_variant_ids"` alone is sufficient, as Odoo registers triggers for each level of a dotted dependency chain.

**`_compute_available_attribute_value_ids` removed**

This method was computing `available_attribute_value_ids`, a field that was never declared in the mixin. It was dead code and has been removed. The `@api.depends` for `available_attribute_value_domain` has also been simplified for the same reason as above, and the missing `else` branch (assigning `[]` when `product_tmpl_id` is False) has been added.

**`is_matching_product` logic fixed**

The original implementation had a structural flaw: the first check `if self.product_tmpl_id != product.product_tmpl_id` always returned `False` when `product_tmpl_id` was not set (since `False != anything` is always `True`), making it impossible to match on `product_id` or `attribute_value_ids` alone without a template.

The method has been rewritten so that each criterion is an independent filter applied only when set:

- `product_id` is checked first as the most restrictive criterion — if set, it short-circuits all other checks.
- `product_tmpl_id` is checked next — if set, the product must belong to that template.
- `attribute_value_ids` is checked last — values within the same attribute are combined with OR, values across different attributes are combined with AND. Since an attribute value can exist on multiple templates, `product_tmpl_id` and `attribute_value_ids` should be used together to avoid false positives across templates.
- If no criterion is set, the method returns `True` for any product (no constraint).

**Documentation**

The `DESCRIPTION.md` has been updated to document all fields and fully describe the matching logic, including the OR/AND semantics of `attribute_value_ids`. A new `USAGE.md` has been added explaining how to inherit the mixin and how to use the domain fields in a form view.

**Tests**

The single shallow test has been replaced by a comprehensive test suite covering all combinations of criteria. Following the pattern required by recent Odoo changes (see OCA/server-ux#1242), the `FakeModelLoader` is now loaded in `setUp`/`tearDown` instead of `setUpClass`/`tearDownClass`. The test class has been split into `TestProductAttributeValueDependentMixinCommon` (reusable by consuming modules) and `TestProductAttributeValueDependentMixin` (the actual test methods).

To get the full context of this [MIG] pull request, see also comments on previous unmerged PRs :
- #2022
- #1423

product_attribute_value_dependent_mixin/models/product_attribute_value_dependent_mixin.py
metaminux: Fixed in `_compute_available_product_domain`

product_attribute_value_dependent_mixin/models/product_attribute_value_dependent_mixin.py
metaminux: Fixed ✅

metaminux:
Hello @bealdav @florian-dacosta and @Kev-Roche 

Can I have a review on this PR please ? Some other PRs are waiting for it...

Regards,

florian-dacosta (approved):
Seems good now, thanks for the work
(technical review)

metaminux:
Hello @bealdav and @Kev-Roche 

Can you review this PR please ? @florian-dacosta already approved it...

Thanks

bealdav (approved):

Kev-Roche (approved):
LGTM

metaminux:
Hello @OCA/product-maintainers 

Can someone having write access merge this approved PR ?

Thanls,

rousseldenis:
> Hello @OCA/product-maintainers
> 
> Can someone having write access merge this approved PR ?
> 
> Thanls,

@metaminux I can, but what to do with 16.0 version ?

metaminux:
> > Hello @OCA/product-maintainers
> > Can someone having write access merge this approved PR ?
> > Thanls,
> 
> @metaminux I can, but what to do with 16.0 version ?

Hello @rousseldenis 

This PR was approved three times after the refactor, so I suppose we can merge this one and I'll update the #2022 with a new commit "backporting" the changes made here.

Does that suits you or do I have to update version 16.0 and play the migration again to 18.0 ? In any case, I'll start by updating 16.0...

rousseldenis (approved):

rousseldenis:
> > > Hello @OCA/product-maintainers
> > > Can someone having write access merge this approved PR ?
> > > Thanls,
> > 
> > 
> > @metaminux I can, but what to do with 16.0 version ?
> 
> Hello @rousseldenis
> 
> This PR was approved three times after the refactor, so I suppose we can merge this one and I'll update the #2022 with a new commit "backporting" the changes made here.
> 
> Does that suits you or do I have to update version 16.0 and play the migration again to 18.0 ? In any case, I'll start by updating 16.0...

My question was just to be sure that one should not be closed.

You can of course backport the changes

PR #2251, opened 2026-04-01
[18.0][MIG] product_readonly_security: Migration to 18.0

victoralmau:
Migration to 18.0

Please @pedrobaeza and @carlos-lopez-tecnativa can you review it?

@Tecnativa TT61748


david-banon-tecnativa (approved):

product_readonly_security/models/product_readonly_security_mixin.py
```diff
@@ -0,0 +1,40 @@
...
+from odoo.exceptions import AccessError
+from odoo.tools import config
+
+
+class ProductReadonlySecurityMixin(models.AbstractModel):
+    _name = "product.readonly.security.mixin"
+    _description = "Mixin to use Product Readonly Security"
+
+    @api.model
+    def check_access_rights(self, operation, raise_exception=True):
```
carlos-lopez-tecnativa: This function is deprecated, please use  `check_access`
https://github.com/odoo/odoo/blob/ffba9eb97876c2507f3bc137298388bf96db6b16/odoo/models.py#L4499

product_readonly_security/readme/DESCRIPTION.md
```diff
@@ -0,0 +1,4 @@
+This module creates a specific permission to allow modification of
+products and variants.
+
+Note: The "Product Creation" permission allows you to create products without needing the specific permission added by this module.
```
carlos-lopez-tecnativa, line 4: The Odoo logic is that administrator users from applications such as Purchase, Sales, Inventory, etc., can manage products.

Image

So, if I am a user with some of these groups (e.g. `Purchase Administrator`) but I do not have the `Product edition` group, I can bypass the access rules and edit, delete, or create products. Is this correct?.
Image

Image

carlos-lopez-tecnativa (changes requested):

product_readonly_security/readme/DESCRIPTION.md
victoralmau: Currently, what you're describing is not how the module has functioned since it was created. In other words, regardless of permissions in other sections (such as Purchase > Administrator), you can only make changes if you have the specific permission added by this module—this behavior is intentional.

carlos-lopez-tecnativa: Sorry, but I still don’t understand the purpose of this module.
Thinking about this scenario:
A user with only `Purchase user` permissions and no `Product creation` permissions.

With standard Odoo (without this module), the user cannot create, edit, or delete products.

So, what is the purpose of this module if, with the standard code, I get the same result?

Image

Image

Image

carlos-lopez-tecnativa (changes requested):

product_readonly_security/readme/DESCRIPTION.md
pedrobaeza: @carlos-lopez-tecnativa we want to make all the users unable to modify products, no matter the rest of the permissions (user or administrator), and be able to do it only when explicitly having this new permission.

carlos-lopez-tecnativa (approved):
> @carlos-lopez-tecnativa we want to make all the users unable to modify products, no matter the rest of the permissions (user or administrator), and be able to do it only when explicitly having this new permission.

OK, when the user does not have the new group but manages products, an exception is raised when saving the record. I expected the options to create, delete, and edit not to be displayed if the user does not have access, to avoid giving false expectations that they can create or edit records and only see the error when saving; I could have spent time creating or editing something and only seen the error message at the end. I, as a user, can feel frustrated. 
But OK, LGTM although it would be cool not to display the options if the user does not have access; at least, can this be added to the roadmap?

image

pedrobaeza:
@victoralmau has it been this way in all the versions to only stop it when trying to save?

BhaveshHeliconia (approved):
Functional review LGTM!
i agree with carlos point.

PR #2254, opened 2026-04-07
[18.0][IMP] product_cost_security: Make tests more resilient

sergio-teruel:
@Tecnativa

### DIFF
--- a/product_cost_security/models/product_cost_security_mixin.py
+++ b/product_cost_security/models/product_cost_security_mixin.py
@@ -4,6 +4,7 @@ from contextlib import suppress
 
 from odoo import api, fields, models
 from odoo.exceptions import AccessError
+from odoo.tools import config
 
 
 class ProductCostSecurityMixin(models.AbstractModel):
@@ -60,7 +61,25 @@ class ProductCostSecurityMixin(models.AbstractModel):
         extra protection to prevent only editing if the user is not in the
         `product_cost_security.group_product_edit_cost` group.
         """
-        valid_fields = super().check_field_access_rights(operation, fields)
+        valid_fields = False
+        if (
+            config["test_enable"]
+            and not self.env.context.get("testing_product_cost_security")
+            and "standard_price" in fields
+        ):
+            # In unrelated tests, users usually do not have this security group.
+            # Apply this bypass only when explicitly testing this module.
+            try:
+                valid_fields = super().check_field_access_rights(operation, fields)
+            except AccessError as e:
+                if (
+                    "You do not have enough rights to access the "
+                    'fields "standard_price" on' in str(e)
+                ):
+                    return fields
+        valid_fields = valid_fields or super().check_field_access_rights(
+            operation, fields
+        )
         if self.env.su:
             return valid_fields
         product_cost_fields = self._product_cost_security_fields().intersection(
--- a/product_cost_security/tests/test_product_cost_security.py
+++ b/product_cost_security/tests/test_product_cost_security.py
@@ -17,7 +17,13 @@ class TestProductCostSecurity(BaseCommon):
         # and attempting to change standard_price
         # Without this flag in the context,
         # the system tries to access stock.move, which the user may not have permission.
-        cls.env = cls.env(context=dict(cls.env.context, disable_auto_svl=True))
+        cls.env = cls.env(
+            context=dict(
+                cls.env.context,
+                disable_auto_svl=True,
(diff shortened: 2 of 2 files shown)


CarlosRoca13 (approved):
LGTM

product_cost_security/models/product_cost_security_mixin.py
```diff
@@ -60,6 +61,14 @@ def check_field_access_rights(self, operation, fields):
...
         `product_cost_security.group_product_edit_cost` group.
         """
+        if (
+            config["test_enable"]
+            and not self.env.context.get("testing_product_cost_security")
+            and "standard_price" in fields
+        ):
+            # By default in other tests users have not this security group, so do this
+            # only when we are testing this module.
+            return fields
```
carlos-lopez-tecnativa, line 71: The logic of this module is not exclusive to the `standard_price` field; it applies to every field that has the group `product_cost_security.group_product_cost` defined in its field definition. So, I think it is better to use the function `_product_cost_security_fields` to check this.

Additionally, I returned the `super()` call.
What do you think about this approach?

```suggestion
        product_cost_fields = self._product_cost_security_fields().intersection(
            valid_fields
        )
        if (
            config["test_enable"]
            and not self.env.context.get("testing_product_cost_security")
            and product_cost_fields
        ):
            # By default in other tests users have not this security group, so do this
            # only when we are testing this module.
            return valid_fields
```

carlos-lopez-tecnativa (changes requested):

product_cost_security/tests/test_product_cost_security.py
yajo, line 20: suggestion: alter context in test only once, here:

```suggestion
        cls.env = cls.env(context=dict(cls.env.context, disable_auto_svl=True, testing_product_cost_security=True))
```

sergio-teruel:
@yajo @carlos-lopez-tecnativa Changes done!

carlos-lopez-tecnativa (approved):

carlosdauden (approved):

carlos-lopez-tecnativa (approved):

product_cost_security/models/product_cost_security_mixin.py
```diff
@@ -60,7 +61,25 @@ def check_field_access_rights(self, operation, fields):
...
+            and "standard_price" in fields
+        ):
+            # In unrelated tests, users usually do not have this security group.
+            # Apply this bypass only when explicitly testing this module.
+            try:
+                valid_fields = super().check_field_access_rights(operation, fields)
+            except AccessError as e:
+                if (
+                    "You do not have enough rights to access the "
+                    'fields "standard_price" on' in str(e)
```
pedrobaeza, line 77: The string comparison for this is very fragile and difficult to debug if upstream the message is changed (or maybe translated). Isn't there other way to get this? Or only look for `"standard_price"` term?

product_cost_security/models/product_cost_security_mixin.py
eduezerouali-tecnativa, line 87:
```suggestion
        cost_fields = self._product_cost_security_fields()
        if (
            config["test_enable"]
            and not self.env.context.get("testing_product_cost_security")
            and cost_fields.intersection(fields)
        ):
            # In unrelated tests, users usually do not have this security group.
            # Apply this bypass only when explicitly testing this module.
            try:
                return super().check_field_access_rights(operation, fields)
            except AccessError:
                other_fields = [f for f in fields if f not in cost_fields]
                    if other_fields:
                        super().check_field_access_rights(operation, other_fields)
                    return fields
         product_cost_fields = cost_fields.intersection(valid_fields)
         if self.env.su:
             return valid_fields
```
What about checking if fields without cost security fields have any restriction?
That way we can ensure that the fields causing the problem are the cost security ones and then we do de bypass @pedrobaeza @sergio-teruel

product_cost_security/models/product_cost_security_mixin.py
```diff
@@ -60,7 +61,25 @@ def check_field_access_rights(self, operation, fields):
         extra protection to prevent only editing if the user is not in the
         `product_cost_security.group_product_edit_cost` group.
         """
-        valid_fields = super().check_field_access_rights(operation, fields)
+        valid_fields = False
+        if (
+            config["test_enable"]
+            and not self.env.context.get("testing_product_cost_security")
```
yajo, line 67: IMHO it's better to add the context check in the rule itself, so it inhibits itself when needed.

PR #2258, opened 2026-04-13
[16.0][ADD] product_restricted_tracking

AungKoKoLin1997:
This module adds a 'restricted_tracking' field to the product category model.
With this optional selection field, you can enforce that all products in a category must use a specific tracking method (Serial Number, Lot, or None).

@qrtl QT6111


product_restricted_tracking/models/product_category.py
```diff
@@ -0,0 +1,45 @@
...
+    def _check_restricted_tracking(self):
+        for categ in self:
+            if not categ.restricted_tracking:
+                continue
+            conflicting_product = self.env["product.template"].search(
+                [
+                    ("categ_id", "=", categ.id),
+                    ("tracking", "!=", categ.restricted_tracking),
+                ],
+                limit=1,
```
yostashiro: Can we remove `limit=1` and show perhaps up to 10 products, for example, so the message shows more information?

product_restricted_tracking/models/product_category.py
```diff
@@ -0,0 +1,45 @@
...
+
+
+class ProductCategory(models.Model):
+    _inherit = "product.category"
+
+    restricted_tracking = fields.Selection(
+        selection=lambda self: self.env["product.template"]
+        ._fields["tracking"]
+        .selection,
+    )
```
yostashiro, line 15: I was wondering if we should support multiple selections (e.g. 'lot' and 'serial').

@nobuQuartile Do you have a PoV? We can go with the current design if there is no concrete need.

product_restricted_tracking/tests/__init__.py
```diff
@@ -0,0 +1,3 @@
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
```
yostashiro:
```suggestion
```

product_restricted_tracking/tests/test_product_restricted_tracking.py
```diff
@@ -0,0 +1,104 @@
...
+                    "categ_id": self.category_serial.id,
+                }
+            )
+        product = self.env["product.template"].create(
+            {
+                "name": "Serial Product",
+                "tracking": "serial",
+                "categ_id": self.category_serial.id,
+            }
+        )
```
yostashiro: I suggest we move product creation to the setup.

product_restricted_tracking/tests/test_product_restricted_tracking.py
```diff
@@ -0,0 +1,104 @@
...
+        with self.assertRaises(ValidationError):
+            product.tracking = "lot"
```
yostashiro: Please add the case categ_id is changed.

product_restricted_tracking/models/product_category.py
AungKoKoLin1997: Done!

product_restricted_tracking/tests/__init__.py
AungKoKoLin1997: Done!

product_restricted_tracking/tests/test_product_restricted_tracking.py
AungKoKoLin1997: Done!

product_restricted_tracking/tests/test_product_restricted_tracking.py
AungKoKoLin1997: Done!

yostashiro (approved):
Code review. LGTM.

SanamiK:
I followed these steps, but the category I set Lot or S/N did not appear.

1. Configuration > Product Categories > NEW > Pick By Lots/By SN/No Tracking in Restricted Tracking field
2. Products > select Product Category

I guess I didn't make a mistake when setting
@AungKoKoLin1997 could you please check?

 
set tracking
product category
pick product category
pick product category2

SanamiK:
@AungKoKoLin1997 Solved the problem. 1 step of setting was missing.

SanamiK (approved):
Products Category appeared  based on the setting "Traceability > Tracking".

LGTM

Image
Image

product_restricted_tracking/models/product_template.py
```diff
@@ -0,0 +1,46 @@
...
+    )
+
+    @api.depends("tracking")
+    def _compute_categ_id_domain(self):
+        for record in self:
+            record.categ_id_domain = [
+                "|",
+                ("restricted_tracking", "=", False),
+                ("restricted_tracking", "=", record.tracking),
+            ]
```
yostashiro, line 24: Upon testing functionality, I think we should actually remove this field as it causes an usability issue for product creation. You can't select a restricted category before setting tracking.
```suggestion
```

yostashiro (changes requested):

product_restricted_tracking/models/product_template.py
AungKoKoLin1997: Yes, I agree. `constrains` will block the mistmatch cases.

yostashiro (approved):
LGTM.

SanamiK:
@AungKoKoLin1997 Could you please modify the product_category.py file so that translations can be applied to the dropdown menu?

BhaveshHeliconia (approved):

PR #2259, opened 2026-04-13
[16.0][ADD] new simple module for bulk product

quentinDupont:
In a context of organic groceries selling lot of bulk products, we need to quickly identify them so this module adds this simple computed (and inverse) field

image


MDgrap (approved):
Tested with runboat, works as intended

product_is_bulk/models/product_product.py
```diff
@@ -0,0 +1,20 @@
...
+    is_bulk = fields.Boolean()
+
+    # Onchange Section
+    @api.onchange("uom_id")
+    def onchange_uom_id(self):
+        self.is_bulk = (
+            True
+            if self.uom_id.category_id.measure_type in ["weight", "volume"]
+            else False
+        )
```
legalsylvain:
```suggestion
    is_bulk = fields.Boolean(compute="_compute_is_bulk", readonly=False)

    @api.depends("uom_id.category_id.measure_type")
    def _compute_is_bulk(self):
        for product in self:
            product.is_bulk = product.uom_id.category_id.measure_type in ["weight", "volume"]
```

Better to replace by a computed function (readonly = False). As a result, the code will be raised in more moment. for exemple, 
- if we import product massively, computation will be done. (it is not the case I think if it's onchange).
- when installing the module, the computation will be done for all existing product.

WDYT ? 

Note : maybe it is requiring an SQL script to initialize more quickly the database...)

product_is_bulk/models/product_template.py
```diff
@@ -0,0 +1,29 @@
...
+        res = super()._get_related_fields_variant_template()
+        res += [
+            "is_bulk",
+        ]
+        return res
+
+    # Onchange Section
+    @api.onchange("uom_id")
+    def onchange_uom_id(self):
+        ProductProduct.onchange_uom_id(self)
```
legalsylvain:
```suggestion

File: product_is_bulk/tests/test_product_is_bulk.py [hunk 3]
```diff
@@ -0,0 +1,23 @@
...
+        cls.uom_unit = cls.env.ref("uom.product_uom_unit")
+        cls.uom_kg = cls.env.ref("uom.product_uom_kgm")
+        cls.product1 = cls.product_obj.create(
+            {"name": "Test Product 1", "uom_id": cls.uom_unit.id}
+        )
+
+    def test_01_change_uom(self):
+        self.assertEqual(self.product1.is_bulk, False)
+        self.product1.uom_id = self.uom_kg
+        self.product1.onchange_uom_id()
```
legalsylvain:
```suggestion
```

product_is_bulk/models/product_product.py
```diff
@@ -0,0 +1,20 @@
...
+# @author: Quentin DUPONT (quentin.dupont@grap.coop)
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+
+from odoo import api, fields, models
+
+
+class ProductProduct(models.Model):
+    _inherit = "product.product"
+
+    is_bulk = fields.Boolean()
```
legalsylvain: I think that an helper could be great for that field. somethink like
 
help="Check this box if this product is sold without any manufactured packaging. It is usually the case for product  whose unit of measure is a unit of weight or volume, but it can be also the case for products priced per unit but sold in bulk (such as cookies, bread, etc.)"

legalsylvain (changes requested):
hi. thanks for your contribution ! 
some remarks inline. a blocking one regarding onchange / computed.

Otherwise, LGTM.
thanks !

product_is_bulk/models/product_product.py
quentinDupont: I chose onchange by simplicity because we also want to handle few cases where product is bulk but is not marked for sale by weight : e.g cookies sold in bulk in a jar, the price being by unit.

But indeed, maybe we can go "compute" and add an inverse function

quentinDupont: Weird, my PR description talks about compute and inverse function, maybe that was my first idea ^^

product_is_bulk/models/product_product.py
quentinDupont: ok you had the same example :-p

product_is_bulk/models/product_product.py
```diff
@@ -0,0 +1,31 @@
...
+    @api.depends("uom_id.category_id.measure_type")
+    def _compute_is_bulk(self):
+        for product in self:
+            product.is_bulk = product.uom_id.category_id.measure_type in [
+                "weight",
+                "volume",
+            ]
+
+    def _inverse_is_bulk(self):
+        pass
```
legalsylvain:
```suggestion
```

product_is_bulk/models/product_product.py
```diff
@@ -0,0 +1,31 @@
...
+
+from odoo import api, fields, models
+
+
+class ProductProduct(models.Model):
+    _inherit = "product.product"
+
+    is_bulk = fields.Boolean(
+        compute="_compute_is_bulk",
+        inverse="_inverse_is_bulk",
```
legalsylvain:
```suggestion
        readonly=False,
```

legalsylvain (commented):
I think it will work the same, and be more synthetic.

legalsylvain (approved):
thanks !

PR #2261, opened 2026-04-16
[14.0][IMP] product_logistics_uom: Length UoM to Settings View

raumschmiede-joshuaL:
In later versions the field already exists in the settings view. I added it to the 14.0:

image

Code had one intendation too much. I removed it so that the pre-commit changes look better. pre-commit does not re-indent the XML tags, only the attributes

### DIFF
--- a/product_logistics_uom/views/res_config_settings.xml
+++ b/product_logistics_uom/views/res_config_settings.xml
@@ -1,22 +1,42 @@
 <?xml version="1.0" encoding="utf-8" ?>
 <odoo>
-        <record id="res_config_settings_view_form" model="ir.ui.view">
-            <field name="name">res.config.settings.view.form.inherit.product</field>
-            <field name="model">res.config.settings</field>
-            <field name="inherit_id" ref="product.res_config_settings_view_form" />
-            <field name="arch" type="xml">
-                <field name="product_weight_in_lbs" position="attributes">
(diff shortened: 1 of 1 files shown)


raumschmiede-joshuaL:
@mt-software-de

raumschmiede-joshuaL:
Will create a PR soon to fix the pre-commit problem.

**EDIT**
Created PR https://github.com/OCA/product-attribute/pull/2262

product_logistics_uom/views/res_config_settings.xml
mt-software-de, line 1: You should do a cherry-pick. https://github.com/OCA/product-attribute/commit/b112ecea75415b3c116375226961ebd8dc9acfc4#diff-ee9090ad2f93bb9e5f5005222971731fe1dae5ca5eb20f9698bb752562a5b048
including only the new field added to the view. 
And then an separate commit for formatting the file.

mt-software-de (changes requested):
Thx. Please split the commit.

product_logistics_uom/views/res_config_settings.xml
raumschmiede-joshuaL: Eventhough the commit's name is not related to the change as it is a whole module migration? 
After that I would need to add a fix commit to make the change work with version 14.0.
And as last commit the one to format the file

PR #2201, opened 2026-02-05
[18.0][IMP] product_set: improve view product set

Saran440:
- Delete code is not used in version18
- Improve view product set

Before
selection_040

After
selection_041

PR #2098, opened 2025-10-13
[19.0][MIG] product_set: Migration to 19.0 

mmrondon:

PR #2263, opened 2026-04-16
[19.0][MIG] product_set: Migration to 19.0

CRogos:
supersede: https://github.com/OCA/product-attribute/pull/2098

includes: https://github.com/OCA/product-attribute/pull/2201

- [x] Remove `odoo_test_helper` dependency — `tests/models.py` deleted
- [x] Convert `product.set.wizard` from `AbstractModel` to `TransientModel`
- [x] Fix `test_product_set.py` — replace all demo data `env.ref()` with programmatic record creation (including `base.res_partner_1`)
- [x] Fix `test_product_set_line.py` — UoM refs retained (module data, always available)
- [x] Fix `test_product_set_wizard.py` — fully programmatic, no demo refs

@Saran440, @rousseldenis. @mmrondon Could you please review?
There were a few (breaking) changes necessary. Could you have a close look on the changes and test.


BhaveshHeliconia (approved):
Code review LGTM!

MohamedOsman7 (approved):

CRogos:
@pedrobaeza could you have a look and merge?

product_set/models/product_set_line.py
```diff
@@ -40,54 +38,27 @@ class ProductSetLine(models.Model):
         "res.company", related="product_set_id.company_id", store=True, readonly=True
     )
     name = fields.Char()
-    product_packaging_id = fields.Many2one(
```
pedrobaeza: Why removing the package part?

CRogos: Because there is no product.packaging anymore.

https://github.com/OCA/OpenUpgrade/blob/c7c7393fc92c11fefbe644b6032e3c967dd3f40d/openupgrade_scripts/scripts/product/19.0.1.2/pre-migration.py#L71

(but take your time if you want to have a closer look into this change)

pedrobaeza: OK, then you would need to add migration scripts to get `uom_id` filled according OpenUpgrade changes. And what about testing this new field?

CRogos: Could you have a look at the migration script. I currently have no setup how to test it.

product_set/models/product_set_line.py
```diff
@@ -40,54 +38,27 @@ class ProductSetLine(models.Model):
...
-        "product.packaging", domain="[('product_id', '=', product_id)]"
-    )
-    product_packaging_qty = fields.Float(
-        compute="_compute_product_packaging_qty",
-        inverse="_inverse_product_packaging_qty",
-        digits="Product Unit of Measure",
+    uom_id = fields.Many2one(
+        "uom.uom",
```
pedrobaeza: Now looking better at the migration script, this should be `product.uom` not `uom.uom`.

product_set/migrations/19.0.1.0.0/post_migrate.py
```diff
@@ -0,0 +1,25 @@
...
+    """Populate uom_id from the product for lines that previously had
+    product_packaging_id.
+    """
+
+    _logger.info("product_set: migrating product_packaging_id -> uom_id")
+
+    cr.execute(
+        """
+        UPDATE product_set_line psl
+        SET uom_id = pt.uom_id
```
pedrobaeza: As the `product_uom` is the same as the `product_packaging` one, you only need to copy its value.

pedrobaeza (changes requested):

product_set/models/product_set_line.py
CRogos: should we rename the field from uom_id  to product_uom_id?

pedrobaeza: If it's the usual nomenclature for this relation (and it seems it is), yes.

CRogos:
@MohamedOsman7 @BhaveshHeliconia could you retest? Due to pedros valid feedback, there has been some changes in the implementation.

MohamedOsman7:
@CRogos I am encountering now the same issue as here: https://github.com/OCA/sale-workflow/pull/4283#issuecomment-4352428939

image

MohamedOsman7 (approved):
i am approving this again because the quantity issue has been solved here https://github.com/OCA/sale-workflow/pull/4283

MohamedOsman7:
@pedrobaeza can we merge here?

pedrobaeza (approved):
/ocabot merge nobump

PR #2266, opened 2026-04-21
[18.0][ADD] lot_catalog: New module

victoralmau:
New module

ejemplo

@Tecnativa TT61752


victoralmau:
Please @pedrobaeza and @CarlosRoca13 can you review it?

pedrobaeza:
The name should always start with the main area, so it should be `stock_lot_catalog`, and this is something more for OCA/stock-logistics-warehouse I think.

victoralmau:
> The name should always start with the main area, so it should be `stock_lot_catalog`, and this is something more for OCA/stock-logistics-warehouse I think.

Great! I'm going to rename everything, including all the extra modules that depend on this one, move it to another repository, and also adjust any customizations that have been made using these modules... It would be really helpful to know the “correct name” (and the repository) **before** doing anything, to avoid all this extra work.

pedrobaeza:
Sorry for the inconveniences, but the rule is not new. We have always put `web`, `sale`, `stock`, `mrp`, etc as the initial prefix for every module (unless it's a whole new area, like `rma`, where the prefix is that one). About the module location, if the area is `stock` (and the core model `stock.lot` is defined there), there's not too much sense to have it here, don't you think?

victoralmau:
Moved to https://github.com/OCA/stock-logistics-warehouse/pull/2563

PR #2272, opened 2026-04-23
[16.0][ADD] product_state_shortage: Introduce shortage state management

nicolas-delbovier-acsone:
**Business Need**

This module addresses the need to manage product states effectively in scenarios where stock shortages occur. It ensures that products marked as in shortage automatically revert to their default state when new stock is received.


rousseldenis (approved):
LGTM

nicolas-delbovier-acsone:
After discussion with our Business Analyst, it seems that basing the auto revoke of the shortage status on incoming move is not sufficient to handle all the cases. 

The good idea is to revoke the status based on the available qty. Now, this computation taking quite some time, we need to do it in a cron. 

I'll do this in another PR and ping here when it is done.

nicolas-delbovier-acsone:
> After discussion with our Business Analyst, it seems that basing the auto revoke of the shortage status on incoming move is not sufficient to handle all the cases.
> 
> The good idea is to revoke the status based on the available qty. Now, this computation taking quite some time, we need to do it in a cron.
> 
> I'll do this in another PR and ping here when it is done.

See https://github.com/OCA/product-attribute/pull/2276

nicolas-delbovier-acsone:
> After discussion with our Business Analyst, it seems that basing the auto revoke of the shortage status on incoming move is not sufficient to handle all the cases.
> 
> The good idea is to revoke the status based on the available qty. Now, this computation taking quite some time, we need to do it in a cron.
> 
> I'll do this in another PR and ping here when it is done.

Well, well, well... It seems that the idea **here** is preferred by our client in the end...

product_state_shortage/models/stock_move.py
```diff
@@ -0,0 +1,22 @@
...
+    _inherit = "stock.move"
+
+    def _action_done(self, cancel_backorder=False):
+        res = super()._action_done(cancel_backorder=cancel_backorder)
+
+        incoming_moves = self.filtered(lambda m: m.picking_code == "incoming")
+        if incoming_moves:
+            shortage_products_to_reset = incoming_moves.product_id.filtered(
+                lambda p: p.product_state_id.is_shortage
+            )
```
lmignon: On each action done of incoming move you'll read all the product's inrfomations for prefetched products to get the product_state_id and then read the product_state table to read the is_shortage.....  it could cost a lot......

nicolas-delbovier-acsone: I've refactored it to avoid calling relational fields inside "filtered". Do you prefer now?

lmignon (approved):

BhaveshHeliconia (approved):
LGTM!

PR #2273, opened 2026-04-23
[19.0][ADD] product_class: classes to group product attributes

Ricardoalso:
This module introduces Product Classes to group product attributes and standardize product setup.


product_class/views/product_template_views.xml
```diff
@@ -0,0 +1,24 @@
+<?xml version="1.0" encoding="utf-8" ?>
+<odoo>
+    <record id="product_template_form_view" model="ir.ui.view">
+        <field name="model">product.template</field>
+        <field name="inherit_id" ref="product.product_template_form_view" />
+        <field name="arch" type="xml">
+            <xpath expr="//group[contains(@name, 'general')]" position="inside">
+                <field name="class_id" />
```
ivantodorovich: Can we move the field to the "Attributes & Variants" page, right before the attribute list?
I think it makes more sense in terms of UX, so users can immediately set a class and define attributes on the same place.

Image

Besides, it would also automatically hide if the group "product.group_product_variant" is missing -- which also makes sense

product_class/views/product_template_views.xml
```diff
@@ -0,0 +1,24 @@
...
+    </record>
+
+    <record id="product_template_form_attribute_class" model="ir.ui.view">
+        <field name="model">product.template</field>
+        <field name="inherit_id" ref="product.product_template_form_view" />
+        <field name="arch" type="xml">
+            <xpath expr="//field[@name='attribute_line_ids']" position="attributes">
+                <attribute
+                    name="domain"
+                >[('attribute_id', 'in', class_id.attribute_ids.ids if class_id else [])]</attribute>
```
ivantodorovich: This domain doesn't seem to be working. See video below:

https://github.com/user-attachments/assets/1d02b4b3-8799-46c4-9746-fa9255a8a54d

The constraint catches the issue and prevents saving the record, but the UI still allows to select wrong attributes.

AFAIR you can't have expressions like this in the domain attribute, so that may be why it's not working.

The way to work around this typical case is to have an intermediate `allowed_attribute_ids` non-stored computed field. At least in previous versions, maybe in 19.0 there's a better way that I'm not aware of

product_class/__manifest__.py
```diff
@@ -0,0 +1,16 @@
...
+    "website": "https://github.com/OCA/product-attribute",
+    "license": "AGPL-3",
+    "category": "Product",
+    "depends": ["product", "stock", "sale"],
+    "data": [
+        "security/ir.model.access.csv",
+        "views/product_class_views.xml",
+        "views/product_template_views.xml",
+    ],
+    "installable": True,
```
ivantodorovich, line 16: could you add yourself and me as maintainers ? 🙏🏻

product_class/tests/test_product_class.py
```diff
@@ -0,0 +1,181 @@
...
+                            {
+                                "attribute_id": self.color_attr.id,
+                                "value_ids": [Command.set([self.color_value.id])],
+                            }
+                        )
+                    ]
+                }
+            )
+
+        self.assertIn("do not belong to the selected class", str(ctx.exception))
```
ivantodorovich: prefer assertRaisesRegex everywhere, instead of this

ivantodorovich (changes requested):

product_class/models/product_class.py
```diff
@@ -0,0 +1,22 @@
...
+    _description = "Product Class"
+    _order = "name"
+
+    _name_uniq = models.Constraint(
+        "unique(name)",
+        "A product class with this name already exists.",
+    )
+
+    name = fields.Char()
+    attribute_ids = fields.Many2many(
```
ivantodorovich: It could be useful to have the inverse `class_ids` field in the `product.attribute` model and view

product_class/models/product_template.py
```diff
@@ -0,0 +1,46 @@
...
+        for product in self:
+            if not product.class_id:
+                continue
+
+            class_attribute_ids = product.class_id.attribute_ids.ids
+            invalid_lines = [
+                line
+                for line in product.attribute_line_ids
+                if line.attribute_id.id not in class_attribute_ids
+            ]
```
ivantodorovich:
```suggestion
            class_attributes = product.class_id.attribute_ids
            invalid_attributes = product.attribute_line_ids.attribute_id - class_attributes
```

product_class/views/product_template_views.xml
Ricardoalso: Hm, tricky situation here 🤔

The domain doesn’t work even with the allowed_attribute_ids approach, and the only thing preventing the user from setting an incorrect attribute is the `_check_class_attributes` constraint. I could override the `web_name_search` method, but that feels like overkill and adds complexity for limited UX benefit. 

I’m open to suggestions

Thanks :pray:

Ricardoalso: @ivantodorovich  handled it with the approach we discussed about [ce5ea6a](https://github.com/OCA/product-attribute/pull/2273/commits/ce5ea6a925b3a6244009ff3e17a8385a385580b1) 

Thanks again :rocket: 

EDIT (context):

Using `parent.class_id.attribute_ids` directly does not work and triggers : 

`
Invalid composed field class_id.attribute_ids in domain of <field name="attribute_id"> ([('id', 'in', parent.class_id.attribute_ids)] if parent.class_id else [])
`


Relating to the comment :
> AFAIR you can't have expressions like this in the domain attribute, so that may be why it's not working.

normalize_domain handles it 
```
>>> from odoo.fields import Command, Domain
>>> normalize_domain([('id', 'in', [1,2])] if True else [])
[('id', 'in', [1, 2])]
>>> normalize_domain([('id', 'in', [1,2])] if False else [])
[(1, '=', 1)]
```

product_class/models/product_template.py
```diff
@@ -0,0 +1,57 @@
...
+    def _compute_allowed_attribute_ids(self):
+        for product in self:
+            if product.class_id:
+                product.allowed_attribute_ids = product.class_id.attribute_ids
+            else:
+                # We don't want to do a search for all attributes
+                # when no class is selected and face performance issues,
+                # so we set an empty recordset instead of all attributes.
+                # We handle the case of no class selected in the domain of the view.
+                product.allowed_attribute_ids = self.env["product.attribute"]
```
ivantodorovich: Given the implementation, it seems we could simplify this like this:

```suggestion
    class_attribute_ids = fields.Many2many(
        related="class_id.attribute_ids",
        string="Class Attributes",
        help="Attributes allowed by the selected product class",
    )
```

That'll also improve the test coverage, cause currently this `_compute_allowed_attribute_ids` method is the only one not being tested

product_class/views/product_attribute_views.xml
```diff
@@ -0,0 +1,17 @@
+<?xml version="1.0" encoding="utf-8" ?>
+<odoo>
+    <record id="product_attribute_view_form" model="ir.ui.view">
+        <field name="model">product.attribute</field>
+        <field name="inherit_id" ref="product.product_attribute_view_form" />
+        <field name="arch" type="xml">
+            <xpath
+                expr="//page[@name='attribute_values']/field[@name='value_ids']"
+                position="after"
```
ivantodorovich: Can you put this in the form header.
It's a bit weird to have it after the value_ids, I'm afraid it'd go unnoticed

https://github.com/user-attachments/assets/2883fc45-605e-47ab-a9df-e8cc97be4c84

product_class/models/product_attribute.py
```diff
@@ -0,0 +1,14 @@
...
+    _inherit = "product.attribute"
+
+    class_ids = fields.Many2many(
+        comodel_name="product.class",
+        relation="product_class_attribute_rel",
+        column1="attribute_id",
+        column2="product_class_id",
+        string="Product Classes",
+        help="Product classes that include this attribute",
+    )
```
ivantodorovich, line 19: Do we need a constraint here, too? Honestly not sure.

Maybe you could write this as a unit test case:

Imagine you have a product class with some attributes, and now you created a few products using this class.
You then go to the attribute and remove the class for some of them that you already used in products.
You then end up with products that have inconsistent attributes, given the class that now has fewer attributes.
This should be prevented. We should not have been able to remove the attribute from the class in the first place because it's already used in some products.

ivantodorovich (approved):
LG.

Small comments, pre-approving

product_class/views/product_attribute_views.xml
ivantodorovich: Alternatively it could be a smart button --> maybe that's better given the same attribute can potentially be used in hundreds of different classes.

Ricardoalso: I agree :wink:  Replacing it with a smart button
image

product_class/models/product_attribute.py
Ricardoalso: When a class is removed from a product.attribute, and that class was used in a product.template, it is also removed from the product.template. However, the product.attribute records set on the product.template remain.

This behavior seems correct to me, since class_id is not a required field on product.template 🤔. It also allows users to configure product.attribute values on the product.template before defining product.class records.

ivantodorovich: I mean when an attribute is removed from a class, having both class and attribute being used in a product

Ricardoalso: Ah ! 

We get the standard error messages

`unlink` triggers `_unlink_except_used_on_product`

> You cannot delete the attribute color because it is used on the following products:


`action_archive`
> You cannot archive this attribute as there are still products linked to it

So we are safe in these cases

ivantodorovich: I don't mean actually deleting the attribute, but just unsetting its class

Ricardoalso: Okay, now I get it 😄
I will add an `@api.constrains` on attribute_ids following the same pattern as the one used for product.template.

product_class/models/product_class.py
```diff
@@ -0,0 +1,50 @@
...
+    def _check_attribute_ids_used_by_products(self):
+        for product_class in self:
+            products = self.env["product.template"].search(
+                [("class_id", "=", product_class.id)]
+            )
+            allowed_attributes = product_class.attribute_ids
+            invalid_lines = products.attribute_line_ids.filtered(
+                lambda line, allowed_attributes=allowed_attributes: line.attribute_id
+                not in allowed_attributes
+            )
```
ivantodorovich:
```suggestion
            invalid_lines = self.env["product.template.attribute.line"].search(
                [
                    ("product_tmpl_id.class_id", "=", product_class.id),
                    ("attribute_id", "not in", product_class.attribute_ids.ids),
                ]
            )
```

perf: avoid fetching mapped fields for a potentially huge list of products

ivantodorovich (approved):

cyrilmanuel (approved):

product_class/tests/test_product_class.py
```diff
@@ -0,0 +1,227 @@
...
+            "value_ids": [Command.set(values.ids)],
+        }
+    )
+
+
+class TestProductClass(TransactionCase):
+    @classmethod
+    def setUpClass(cls):
+        super().setUpClass()
+        cls.size_attr = cls.env["product.attribute"].create({"name": "Size"})
```
simahawk, line 16: disable tracking pls

simahawk (commented):
minor remark and a question: have you evaluated using attribute sets from OCA/pim?

ivantodorovich:
> minor remark and a question: have you evaluated using attribute sets from OCA/pim?

We haven't, but my understanding is that it for a different use case

Ricardoalso:
> > minor remark and a question: have you evaluated using attribute sets from OCA/pim?
> 
> We haven't, but my understanding is that it for a different use case

After taking a look at the module's description, it seems to be a far more generic module where we could potentially achieve the same things (if we set aside the view overrides done in this product_class), but with a lot more configuration tuning. :thinking:

This module is designed to be "plug and play" after installation and has a deterministic scope. But I appreciate you sharing and pointing out this module :pray:

Ricardoalso:
@ivantodorovich as discussed, some new improvements pushed [50ff6fa](https://github.com/OCA/product-attribute/pull/2273/commits/50ff6fa5e0122dd2f629b46bd9f65454783cdf0d)

product_class/models/product_template.py
```diff
@@ -26,7 +53,8 @@ def _check_class_attributes(self):
             if not product.class_id:
                 continue
 
-            class_attributes = product.class_id.attribute_ids
+            class_lines = product.class_id.attribute_line_ids
+            class_attributes = class_lines.attribute_id
```
ivantodorovich: Or simply read `product.class_attribute_ids`

product_class/models/product_template.py
```diff
@@ -43,3 +71,22 @@ def _check_class_attributes(self):
                         attrs=invalid_names,
                     )
                 )
+
+            required_attributes = class_lines.filtered("required").attribute_id
```
ivantodorovich: Or simply read `product.class_required_attribute_ids`

ivantodorovich (approved):
Thanks!!

product_class/models/product_template.py
Ricardoalso: Done :heavy_check_mark: 
Thanks :pray:

product_class/models/product_attribute.py
```diff
@@ -0,0 +1,30 @@
+from odoo import api, fields, models
```
vvrossem, line 3:
```suggestion
# Copyright 2026 Camptocamp SA
# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
from odoo import api, fields, models
```

product_class/models/product_attribute.py
```diff
@@ -0,0 +1,30 @@
...
+
+    classes_count = fields.Integer(
+        string="Product Classes Count",
+        compute="_compute_classes_count",
+    )
+
+    class_attribute_line_ids = fields.One2many(
+        comodel_name="product.class.attribute.line",
+        inverse_name="attribute_id",
+        string="Class Attribute Lines",
```
vvrossem, line 17: You can omit the `string` attribute
```suggestion
```

product_class/models/product_class.py
```diff
@@ -0,0 +1,48 @@
+from odoo import api, fields, models
```
vvrossem, line 3:
```suggestion
# Copyright 2026 Camptocamp SA
# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
from odoo import api, fields, models
```

product_class/models/product_class_attribute_line.py
```diff
@@ -0,0 +1,31 @@
+from odoo import fields, models
```
vvrossem, line 3:
```suggestion
# Copyright 2026 Camptocamp SA
# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
from odoo import fields, models
```

product_class/models/product_class_attribute_line.py
```diff
@@ -0,0 +1,31 @@
...
+        ondelete="cascade",
+        index=True,
+        string="Product Class",
+    )
+    attribute_id = fields.Many2one(
+        comodel_name="product.attribute",
+        required=True,
+        ondelete="restrict",
+        index=True,
+        string="Attribute",
```
vvrossem, line 28:
```suggestion
```

product_class/models/product_template.py
```diff
@@ -0,0 +1,91 @@
+from odoo import api, fields, models
```
vvrossem, line 3:
```suggestion
# Copyright 2026 Camptocamp SA
# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
from odoo import api, fields, models
```

product_class/models/product_template.py
```diff
@@ -0,0 +1,91 @@
...
+    class_id = fields.Many2one(
+        comodel_name="product.class",
+        string="Product Class",
+        help="Product class that constrains which attributes can be used",
+    )
+
+    class_attribute_ids = fields.Many2many(
+        comodel_name="product.attribute",
+        compute="_compute_class_attribute_ids",
+        string="Class Attributes",
```
vvrossem, line 19:
```suggestion
```

product_class/models/product_template.py
```diff
@@ -0,0 +1,91 @@
...
+        help="Attributes allowed by the selected product class",
+    )
+
+    class_required_attribute_ids = fields.Many2many(
+        comodel_name="product.attribute",
+        compute="_compute_class_attribute_ids",
+        string="Class Required Attributes",
```
vvrossem, line 26:
```suggestion
```

product_class/tests/test_product_class.py
```diff
@@ -0,0 +1,285 @@
+from psycopg2 import IntegrityError
```
vvrossem, line 3:
```suggestion
# Copyright 2026 Camptocamp SA
# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
from psycopg2 import IntegrityError
```

product_class/views/product_attribute_views.xml
```diff
@@ -0,0 +1,23 @@
+<?xml version="1.0" encoding="utf-8" ?>
+<odoo>
```
vvrossem, line 4:
```suggestion
<!-- Copyright 2026 Camptocamp SA
     License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl). -->
<odoo>
```

product_class/views/product_class_views.xml
```diff
@@ -0,0 +1,65 @@
+<?xml version="1.0" encoding="utf-8" ?>
+<odoo>
```
vvrossem, line 4:
```suggestion
<!-- Copyright 2026 Camptocamp SA
     License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl). -->
<odoo>
```

product_class/views/product_template_views.xml
```diff
@@ -0,0 +1,25 @@
+<?xml version="1.0" encoding="utf-8" ?>
+<odoo>
```
vvrossem, line 4:
```suggestion
<!-- Copyright 2026 Camptocamp SA
     License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl). -->
<odoo>
```

vvrossem (approved):
Thank you for the detailed DESCRIPTION.md and USAGE.md 👏

product_class/views/product_class_views.xml
```diff
@@ -0,0 +1,67 @@
...
+    <!-- Menu entry under Inventory > Products > Configuration -->
+    <menuitem
+        id="menu_product_class_stock_config"
+        name="Product Classes"
+        parent="stock.menu_product_in_config_stock"
+        action="product_class.product_class_action"
+        sequence="25"
+    />
+    <!-- Menu entry under Sales > Products > Configuration -->
+    <menuitem
```
ivantodorovich, line 73: Please add the group `product.group_product_variant`, so that it's only displayed if that feature is enabled

product_class/views/product_class_views.xml
```diff
@@ -0,0 +1,67 @@
...
+
+    <record id="product_class_action" model="ir.actions.act_window">
+        <field name="name">Product Classes</field>
+        <field name="res_model">product.class</field>
+        <field name="view_mode">list,form</field>
+        <field name="context">{}</field>
+    </record>
+
+    <!-- Menu entry under Inventory > Products > Configuration -->
+    <menuitem
```
ivantodorovich, line 64: Please add the group `product.group_product_variant`, so that it's only displayed if that feature is enabled

ivantodorovich (changes requested):

product_class/models/product_class.py
```diff
@@ -0,0 +1,50 @@
...
+    _name = "product.class"
+    _description = "Product Class"
+    _order = "name"
+
+    _name_uniq = models.Constraint(
+        "unique(name)",
+        "A product class with this name already exists.",
+    )
+
+    name = fields.Char()
```
ivantodorovich: Add `translate=True` please 🙏🏻 
(not sure the name_unique constraint will work afterwards, but we do need translations)

ivantodorovich: Add also an `active` field and the ribbon in the form view + filter in the list view so users can archive them

ivantodorovich (changes requested):

yankinmax:
Hello @OCA/product-maintainers can you pls take a look and possibly trigger merge?

ivantodorovich (approved):

BhaveshHeliconia (approved):
LGTM!

simahawk (commented):
This module has somehow something overlapping w/ the PIM suite, in particular the validation.. However, the attribute_set* modules add new attributes (namely fields) to the product and do not mess up w/ product attrs.

PR #2276, opened 2026-04-29
[16.0][ADD] product_state_shortage: Introduce shortage state management

nicolas-delbovier-acsone:
**Business Need**

This module addresses the need to manage product states effectively in scenarios where stock shortages occur. It ensures that products marked as in shortage automatically revert to their default state when qty available > 0.


rousseldenis (approved):
LGTM

nicolas-delbovier-acsone:
The cron data is missing from the manifest, I'll fix this

nicolas-delbovier-acsone:
A simpler idea seems to work, see: https://github.com/OCA/product-attribute/pull/2272#issuecomment-4610516915

PR #2275, opened 2026-04-29
[19.0][MIG] product_state_sale: Migration to 19.0

MohamedOsman7:
supersede: https://github.com/OCA/product-attribute/pull/2203
Depends on product_state: https://github.com/OCA/product-attribute/pull/2217 (Merged !)

PR #2278, opened 2026-04-30
[19.0][IMP] product_state: fix domain filter for state products action

MohamedOsman7:
solves: https://github.com/OCA/product-attribute/pull/2275#pullrequestreview-4202946987

### DIFF
--- a/product_state/README.rst
+++ b/product_state/README.rst
@@ -35,10 +35,10 @@ Product State
 This module introduces the state field on product template and allows
 simple product life cycle:
 
-- draft: In Development
-- sellable: Normal
-- end: End of Lifecycle
-- obsolete: Obsolete
+-  draft: In Development
+-  sellable: Normal
+-  end: End of Lifecycle
+-  obsolete: Obsolete
 
 **Table of contents**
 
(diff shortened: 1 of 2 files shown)


CRogos (approved):
LGTM

product_state/__manifest__.py
```diff
@@ -7,7 +7,7 @@
     "author": "ACSONE SA/NV, Odoo Community Association (OCA)",
     "website": "https://github.com/OCA/product-attribute",
     "category": "Product",
-    "version": "19.0.1.0.0",
+    "version": "19.0.1.0.1",
```
BhaveshHeliconia: I think the version does not need to be updated manually, since it should be handled automatically when the merge command is executed.

Other than that, LGTM!

MohamedOsman7: Done.

BhaveshHeliconia (approved):

MohamedOsman7:
@pedrobaeza Can you please merge in order to continue on https://github.com/OCA/product-attribute/pull/2275

product_state/README.rst
```diff
@@ -35,10 +35,10 @@ Product State
 This module introduces the state field on product template and allows
```
pedrobaeza, line 35: Please don't do this change.

product_state/views/product_template_views.xml
```diff
@@ -54,6 +54,6 @@
         <field name="res_model">product.template</field>
         <field name="view_mode">kanban,form,list</field>
         <field name="context">{}</field>
-        <field name="domain">[('product_state_id', '=', id)]</field>
+        <field name="domain">[('product_state_id', '=', active_id)]</field>
```
pedrobaeza, line 57: I'm not sure this `active_id` will always contain the correct data.

MohamedOsman7: It works with active_id when I trigger it from the state form, but yeah, not sure if that’s always safe.

Do you have a best practice for this?

MohamedOsman7: @pedrobaeza just wanted to follow up on this, do you think there is a more reliable approach than using active_id here? I tried record.id, did not work.

pedrobaeza: Just `id` isn't working?

MohamedOsman7: No, see here https://github.com/OCA/product-attribute/pull/2275#pullrequestreview-4202946987

pedrobaeza: OK, let's go with this

/ocabot merge patch

MohamedOsman7: I think the merge command didn't trigger because it was posted inside the review thread, it may nee dto be posted as a regular comment, right?

BhaveshHeliconia: ping @pedrobaeza

PR #2279, opened 2026-04-30
[18.0][IMP] product_sequence: default_code shouldn't be mandatory

CarlosRoca13:
Since there is already a module that makes the field mandatory, such as product_code_mandatory, adding this requirement at the database level, in my opinion, goes beyond the scope of this module.

cc @Tecnativa

ping @pedrobaeza @sergio-teruel

### DIFF
--- a/product_sequence/models/product_product.py
+++ b/product_sequence/models/product_product.py
@@ -11,7 +11,6 @@ class ProductProduct(models.Model):
     _inherit = "product.product"
 
     default_code = fields.Char(
-        required=True,
         default="/",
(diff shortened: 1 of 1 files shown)


pedrobaeza (approved):

BhaveshHeliconia (approved):
LGTM!

eduezerouali-tecnativa (approved):
ping @pedrobaeza

eduezerouali-tecnativa:
ping @pedrobaeza merge was stuck

PR #2283, opened 2026-05-05
[MIG] product_stock_state : Migration to 19.0

uncannycs:
[MIG] product_stock_state : Migration to 19.0


BhaveshHeliconia (changes requested):
@uncannycs it looks like the commit history is not aligned with the guidelines. I can only see a single commit, so the expected migration history is missing. Could you please follow the steps here: https://github.com/OCA/maintainer-tools/wiki/Migration-to-version-19.0#how-to

PR #2285, opened 2026-05-06
[19.0][MIG] product_catalog_stock: Migration to 19.0

BhaveshHeliconia:


victoralmau:
Please, cherry-pick https://github.com/OCA/product-attribute/pull/2400 to commit history before migration commit.

BhaveshHeliconia:
> Please, cherry-pick #2400 to commit history before migration commit.

Done. Thanks!

product_catalog_stock/models/stock_picking.py
```diff
@@ -24,12 +24,7 @@ def _default_order_line_values(self, child_field=False):
         return {**default_data, **new_default_data}
 
     def _get_product_catalog_domain(self):
-        return expression.AND(
-            [
-                super()._get_product_catalog_domain(),
-                [("type", "in", ["consu", "product"])],
-            ]
-        )
+        return super()._get_product_catalog_domain() & Domain("type", "!=", "service")
```
victoralmau: In my opinion, the right thing to do would be to set `type!=consu`, because otherwise we'll be including `combo`, and there's currently no support for that option.

victoralmau: Also, considering what `super()` does (in `product`) https://github.com/odoo/odoo/blob/3b7765bdec16ce98b7a5e9fac2ae3114f6a93ecd/addons/product/models/product_catalog_mixin.py#L48, I don't think it would be necessary to add any extra domains.

victoralmau (changes requested):

product_catalog_stock/models/stock_picking.py
BhaveshHeliconia: Thanks for the review!

For the first point, super() doesn't exclude service products, so we still need that filter here since services can't be used for stock moves.

And yes, I agree that using type = 'consu' is cleaner here and matches how it's handled in Odoo 19 core.

victoralmau (approved):

PR #2290, opened 2026-05-07
[16.0][IMP] product_is_bulk: correct onchange on product.template model.

legalsylvain:
rational: we use the _compute_template_field_from_variant_field / _set_product_variant_field / _get_related_fields_variant_template mechanism to compute correctly is_bulk.
However, for the time being, when user select uom_id on product template form the is_bulk is not computed on the fly, and user should wait for saving. That is not very intuitive. (it occures only on product.template form view, it is OK with product.product form view)


quentinDupont (approved):
Code review ✔️

BhaveshHeliconia (approved):
LGTM!

sebalix (approved):

PR #1668, opened 2024-06-21
[16.0][ADD] product_template_default_weight

mathieudelva:

PR #2299, opened 2026-05-20
[16.0][ADD] product_template_default_weight

Kev-Roche:
Supersedes https://github.com/OCA/product-attribute/pull/1668

This module  falling back to the template weight when the variant
weight is 0, ensuring correct weight calculation on:
- Stock moves 
- Delivery package wizard

cc @mathieudelva


bealdav (approved):
Thanks a lot, this module is definitively required.
Without it product_id.weight can answer 0 while weight on template is upper than 0.

Thanks

BhaveshHeliconia (approved):
LGTM!

product_template_default_weight/models/product_product.py
```diff
@@ -0,0 +1,17 @@
...
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+from odoo import fields, models
+
+
+class ProductProduct(models.Model):
+    _inherit = "product.product"
+
+    product_tmpl_weight = fields.Float(
+        compute="_compute_product_tmpl_weight", string="Template Weight"
```
sebalix: Why `compute` instead of `related` here?
If the goal is to override the compute method in a submodule, a comment is then welcome so we do not change it by mistake during a migration.

product_template_default_weight/models/product_product.py
```diff
@@ -0,0 +1,17 @@
...
+    )
+
+    def _compute_product_tmpl_weight(self):
```
sebalix: If the compute is kept:

```suggestion
    @api.depends("product_tmpl_id.weight")
    def _compute_product_tmpl_weight(self):
```

product_template_default_weight/models/stock_move.py
```diff
@@ -0,0 +1,20 @@
...
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+from odoo import api, models
+
+
+class StockMove(models.Model):
+    _inherit = "stock.move"
+
+    @api.depends("product_id", "product_uom_qty", "product_uom")
+    def _cal_move_weight(self):
```
sebalix, line 12: A comment telling we fully overload the standard `_cal_move_weight` method is welcome IMO

product_template_default_weight/readme/DESCRIPTION.rst
```diff
@@ -0,0 +1,11 @@
+This module ensures accurate weight calculation on stock pickings and delivery packages
```
sebalix: Not blocking, but it could be better for non-techies to use UI terms (operations or transfers)
```suggestion
This module ensures accurate weight calculation on stock operations and delivery packages
```

sebalix (commented):
LG, small comments

Kev-Roche:
Thanks @sebalix , changes done.

sebalix (approved):

product_template_default_weight/__manifest__.py
```diff
@@ -0,0 +1,20 @@
...
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+{
+    "name": "Product Template Default Weight",
+    "version": "16.0.1.0.0",
+    "category": "Product",
+    "license": "AGPL-3",
+    "summary": "Add Product Template Default Weight logic",
+    "author": "Akretion,Odoo Community Association (OCA)",
+    "maintainers": ["mathieudelva"],
```
sebalix: Should we keep the same maintainer or you also want to add yourself?

PR #2303, opened 2026-05-21
[19.0][MIG] product_sale_team: Migration to 19.0

BhaveshHeliconia:


BhaveshHeliconia:
@CristianoMafraJunior @WesleyOliveira98 @MohamedOsman7 @cgarcia-solvos Could you please review this? Thanks!

cgarcia-solvos (commented):
Code and functional review. LGTM!
Only minor comment:

product_sale_team/tests/test_product_sale_team.py
```diff
@@ -0,0 +1,15 @@
+# Copyright (C) 2025 Cetmix OÜ
+# License LGPL-3.0 or later (https://www.gnu.org/licenses/lgpl).
+
+from odoo.tests import TransactionCase, tagged
+
+
+@tagged("post_install", "-at_install")
+class TestProductSaleTeam(TransactionCase):
```
cgarcia-solvos: tracking_disable should be applied here (or inherit from BaseCommon) to align with what the guide says:

https://github.com/OCA/maintainer-tools/wiki/Migration-to-version-19.0#:~:text=Disable,way%29%2E


```suggestion
from odoo.tests import tagged
from odoo.addons.base.tests.common import BaseCommon


@tagged("post_install", "-at_install")
class TestProductionSaleTeam(BaseCommon):
```

BhaveshHeliconia: This test only inspects field metadata via _fields and doesn't create or modify any records, so tracking_disable from BaseCommon wouldn't have any effect here. TransactionCase is sufficient for this use case. Happy to change it if there's a different reason you'd prefer BaseCommon.

WesleyOliveira98 (approved):
LGTM

product_sale_team/tests/test_product_sale_team.py
cgarcia-solvos: You're right, however, in my opinion, the migration guide is a way to standardize the code for the future. If the addon grows, it's better to follow the conventions now and not wait until it's absolutely necessary. What do you think?

BhaveshHeliconia: That's a fair point — following the convention now for consistency and future-proofing makes sense. Applied the suggestion. Thanks!

cgarcia-solvos (approved):

CristianoMafraJunior (approved):
LGTM

BhaveshHeliconia:
@rousseldenis @legalsylvain Could you please review and merge this? Thanks!

rousseldenis (approved):

PR #2309, opened 2026-05-22
[17.0][FIX] product_state: add read access for internal users

miquelrosell99:
Normal users with access to products get an access error on `product.state` because the model only grants read access to sales managers, stock managers, and public users. Internal users (`base.group_user`) need read access to view products that reference a state.

**Steps to reproduce:**
1. Install product_state
2. Create a user with only `base.group_user` (no sale/stock manager)
3. Open a product form or report showing `product_state_id`

**Result:** AccessError on product.state

**Fix:** add read-only ACL for `base.group_user`.


BhaveshHeliconia (approved):
LGTM!

PR #2310, opened 2026-05-22
[18.0][IMP] product_secondary_unit: Add helper method to secondary units to get secondary qty computed

CarlosRoca13:
cc @Tecnativa TT62678

FW-PORT of: https://github.com/OCA/product-attribute/pull/1469

### DIFF
--- a/product_secondary_unit/models/product_second_unit.py
+++ b/product_secondary_unit/models/product_second_unit.py
@@ -1,6 +1,7 @@
 # Copyright 2018 Tecnativa - Sergio Teruel
 # License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
 from odoo import api, fields, models
+from odoo.tools.float_utils import float_round
 
 
 class ProductSecondaryUnit(models.Model):
@@ -47,3 +48,16 @@ class ProductSecondaryUnit(models.Model):
     def _compute_display_name(self):
         for unit in self:
             unit.display_name = f"{unit.name}-{unit.factor}"
+
+    def _get_secondary_qty(self, qty, uom, product_uom_field="uom_id"):
+        """Helper method to obtain the secondary uom quantity from a given main unit"""
+        self.ensure_one()
+        uom_product = (
+            self.product_id[product_uom_field]
+            or self.product_tmpl_id[product_uom_field]
+        )
+        if uom != uom_product:
+            qty = uom._compute_quantity(qty, uom_product)
+        return float_round(
+            qty / (self.factor or 1.0), precision_rounding=self.uom_id.rounding
+        )


sergio-teruel (approved):

product_secondary_unit/models/product_second_unit.py
```diff
@@ -47,3 +48,13 @@ class ProductSecondaryUnit(models.Model):
     def _compute_display_name(self):
         for unit in self:
             unit.display_name = f"{unit.name}-{unit.factor}"
+
+    def _get_secondary_qty(self, qty, uom, product_uom_field="uom_id"):
+        """Helper method to obtain the secondary uom quantity from a given main unit"""
+        self.ensure_one()
+        factor = self.factor * (
+            uom.factor if self.product_id[product_uom_field] != uom else 1.0
+        )
```
yostashiro: This logic doesn't consider the factor of product UoM, so the conversion will be incorrect if the product UoM is not the main one in the category (factor != 1). For example, a product whose main UoM is Dozen with a secondary unit of factor=2: _get_secondary_qty(24, unit_uom) returns 12 instead of the correct 1.

We took care of this issue in https://github.com/OCA/product-attribute/pull/2182 and you can follow the same approach as _convert_qty_to_secondary_uom in the mixin.

yostashiro (commented):
@sergio-teruel It's your turn to comment on https://github.com/OCA/product-attribute/pull/2211#issuecomment-4586766998 ;)

product_secondary_unit/models/product_second_unit.py
CarlosRoca13: Changes done! Thanks @yostashiro

product_secondary_unit/models/product_second_unit.py
```diff
@@ -47,3 +48,13 @@ class ProductSecondaryUnit(models.Model):
     def _compute_display_name(self):
         for unit in self:
             unit.display_name = f"{unit.name}-{unit.factor}"
+
+    def _get_secondary_qty(self, qty, uom, product_uom_field="uom_id"):
+        """Helper method to obtain the secondary uom quantity from a given main unit"""
+        self.ensure_one()
+        uom_product = self.product_id[product_uom_field]
```
yostashiro: I thought you'd probably like to consider the case where self.product_id is false as well.

```suggestion
        uom_product = self.product_id[product_uom_field] or self.product_tmpl_id[product_uom_field]
```

yostashiro (approved):
Code review. LGTM. 👍

pedrobaeza (approved):
/ocabot merge patch

PR #2311, opened 2026-05-22
[18.0][IMP] product_readonly_security: Disable Create/Edit/Delete in the UX if the user does not have the appropriate permissions

victoralmau:
Disable Create/Edit/Delete (always) if the user does not have the appropriate group

Related to https://github.com/OCA/product-attribute/pull/2251#pullrequestreview-4083246060

The `_check_access()` method is overridden because this is the appropriate place for Odoo to indirectly prevent the create/edit/delete actions from being displayed.

Please @pedrobaeza and @carlos-lopez-tecnativa can you review it?

@Tecnativa TT61748

### DIFF
--- a/product_readonly_security/models/product_readonly_security_mixin.py
+++ b/product_readonly_security/models/product_readonly_security_mixin.py
@@ -1,8 +1,9 @@
-# Copyright 2024 Tecnativa - Víctor Martínez
+# Copyright 2024-2026 Tecnativa - Víctor Martínez
 # License AGPL-3 - See http://www.gnu.org/licenses/agpl-3.0.html
 
-from odoo import _, api, models
-from odoo.exceptions import AccessError
+import functools
+
+from odoo import models
 from odoo.tools import config
 
 
@@ -10,11 +11,13 @@ class ProductReadonlySecurityMixin(models.AbstractModel):
     _name = "product.readonly.security.mixin"
     _description = "Mixin to use Product Readonly Security"
 
-    @api.model
-    def check_access(self, operation):
-        # Override security returning False/AccessError if not belonging to
-        # the new security group. This makest that the create, edit and delete
-        # buttons are not displayed.
+    def _check_access(self, operation):
+        # We override this method directly because we want to display an error
+        # (similar to what happens if a user don't have access -ACL- to perform
+        # an action on a model) if the user does not have the "Product edition" group.
+        # This method is very useful because it helps ensure that the Create/Edit/Delete
+        # buttons are not displayed across all sections of Odoo, including in many2one
+        # fields (example: the product field in sales order lines).
         user = self.env.user
         group = "product_readonly_security.group_product_edition"
         test_condition = not config["test_enable"] or (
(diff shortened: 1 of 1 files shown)


pedrobaeza (approved):
Good option for not touching JS.

product_readonly_security/models/product_readonly_security_mixin.py
```diff
@@ -10,6 +12,30 @@ class ProductReadonlySecurityMixin(models.AbstractModel):
     _name = "product.readonly.security.mixin"
     _description = "Mixin to use Product Readonly Security"
 
+    def get_view(self, view_id=None, view_type="form", **options):
```
carlos-lopez-tecnativa: This works when you go to any product menu, but it does not work in a many2one field: the Create and Create & Edit options are still displayed.

Could you review whether inheriting from check_access or has_access would cover this case too?
https://github.com/odoo/odoo/blob/38a34a3f84f650c71e214d35048beab8a66089ae/odoo/models.py#L4429-L4446
Image

carlos-lopez-tecnativa (changes requested):
Please don’t forget the Product Creation group. If users in this group can create products, then the options must be available even if the user does not have the Product Edition group.

Image

Image

victoralmau:
> Please don’t forget the Product Creation group. If users in this group can create products, then the options must be available even if the user does not have the Product Edition group.
> Image Image

This is intentional, similar to what happens in v17 https://github.com/OCA/product-attribute/blob/681deb75293bcdf822e2a7fbf84269729b7ca6f3/product_readonly_security/models/product_readonly_security_mixin. py#L14 If you do **NOT** have the Product Edition permission, you cannot create or edit even if you have the Product Creation permission.

victoralmau:
Another simpler approach (using the `_check_access()` method) that works in all cases (https://github.com/OCA/product-attribute/pull/2311#discussion_r3298075277).

pedrobaeza:
Why changing the approach? I think it's correct that this group forbids everything, even if the other group is granted.

victoralmau:
I have **NOT** changed the approach.

This https://github.com/OCA/product-attribute/pull/2311#issuecomment-4612580894 is for informational purposes only.

The change in this PR resolves the issue described in https://github.com/OCA/product-attribute/pull/2251#pullrequestreview-4083246060 (it works in all cases, including https://github.com/OCA/product-attribute/pull/2311#discussion_r3298075277)

pedrobaeza:
Well, I don't understand why using `partial` and so on. Can you explain a bit in the commit message the why and how it works?

product_readonly_security/models/product_readonly_security_mixin.py
```diff
@@ -1,20 +1,24 @@
...
     _name = "product.readonly.security.mixin"
     _description = "Mixin to use Product Readonly Security"
 
-    @api.model
-    def check_access(self, operation):
-        # Override security returning False/AccessError if not belonging to
-        # the new security group. This makest that the create, edit and delete
-        # buttons are not displayed.
+    def _check_access(self, operation):
+        """We override this method directly because we want to display an error
```
pedrobaeza: Thanks for the extra info. As this is an override, it should be put as comment, not as docstring, as due to the Python nature of this, the original docstring will be replaced by this one in docs generators.

product_readonly_security/models/product_readonly_security_mixin.py
```diff
@@ -27,10 +31,8 @@ def check_access(self, operation):
             and not self.env.su
             and not user.has_group(group)
         ):
-            raise AccessError(
-                _(
-                    "Sorry, you are not allowed to create/edit products. Please "
-                    "contact your administrator for further information."
-                )
+            Access = self.env["ir.model.access"]
+            return self, functools.partial(
```
pedrobaeza, line 35: This specific call is for me obscure. Why this return?

victoralmau: Comment added to the code.

pedrobaeza (approved):

cgarcia-solvos (approved):
Code Review. LGTM!

carlos-lopez-tecnativa (approved):
LGTM

PR #2315, opened 2026-05-26
[FIX] product_expiry_configurable incorrect translation gl.po

legalsylvain:
fix 

product_expiry_configurable/i18n/gl.po:214 Translation string couldn't be parsed correctly using str.format KeyError('dáche_name') - [po-python-parse-format]

ref https://github.com/OCA/product-attribute/actions/runs/26475598996/job/77960215800?pr=2314#step:7:25


legalsylvain:
Fix https://github.com/OCA/product-attribute/pull/2314

/ocabot merge patch

PR #2316, opened 2026-05-26
[FIX] product_category_code_unique: writing on many categories is broken.

legalsylvain:
rational : with the current implementation, if we write on many categories, on categories that don't have code, it will write the same code on many categories

Fixes CI of https://github.com/OCA/product-attribute/pull/2314


luisg123v (changes requested):
Code LGTM, but commit message needs some adjustments:
- Missing blank line between title and description.
- Missing trailing dot in description, but title has an extra dot.
- Lines are too long, please wrap to 72 characters.
- rational → rationale

rvalyi (approved):

BhaveshHeliconia (approved):
LGTM!

PR #2317, opened 2026-05-27
[18.0][FIX] product_sequence: Add uninstall hook to remove the NOT NULL constraint

celm1990:
When this module is installed, the default_code field is marked as required (NOT NULL at the database level). However, if the module is uninstalled, the field remains as NOT NULL, causing an error when a new product is created without a default_code.

After this commit, an uninstall hook is added to remove the NOT NULL constraint.

@pedrobaeza @CarlosRoca13 could you please review this?


pedrobaeza:
Why making the default code mandatory in this module? I think that's a mix of features.

celm1990:
> Why making the default code mandatory in this module? I think that's a mix of features.

I don't know, but if the field should no longer be required, then this PR already removes the required:
https://github.com/OCA/product-attribute/pull/2279

However, the NOT NULL constraint is not removed from the database. I think it is necessary to add my code to that PR as well.

pedrobaeza:
I would go for removing the constraint at all from the code, and do this removal in a migration script.

PR #2320, opened 2026-06-10
[19.0] [ADD] product_packaging

ivantodorovich:
This module brings back a clear **Packagings** section on product forms.

With it, you can:

- Define the additional units used to sell or handle a product (for example box, case, or pallet).
- Set a **barcode for each packaging on each product variant** (useful for scanning and logistics).
- Manage everything from the product form, in one place, with a layout closer to what users knew in Odoo 18.

Packaging units are defined once per product template and shared by all its variants. Barcodes remain specific to each variant.

## Technical note

The module reimplements the UX to configure the packagings on the product forms, but it's built on top of Odoo's new packaging mechanics using the Unit of Measure model.
The `product.packaging` model is used as a proxy to configure the lower-level `product.template.uom_ids` and its `product.uom` records for the packaging barcodes.
It also implements additional fields on the packaging level to compute and store the package weight, volume, etc..


product_packaging/models/product_packaging.py
```diff
@@ -0,0 +1,86 @@
+# Copyright 2026 Camptocamp SA (https://www.camptocamp.com).
```
legalsylvain, line 1: is it right ? product packaging was an Odoo core model, right ? 
or maybe you redefined from scratch ?

ivantodorovich: I started from scratch with a different approach.

We're not bringing back the previous feature. Here we're using the model as a proxy to configure the underlying uom_ids (Odoo's new packaging), solely for UX and informational purposes -- but the real mechanics are those of standard Odoo 19

product_packaging/hooks.py
```diff
@@ -0,0 +1,16 @@
...
+from odoo.tools import split_every
+
+
+def post_init_hook(env):
+    """Backfill packagings for products that already have ``uom_ids`` set.
+
+    Processed in batches so the ORM cache does not grow unbounded on large
+    catalogs.
+    """
+    products = env["product.product"].search([("uom_ids", "!=", False)])
```
Ricardoalso, line 13: This can cause a memory error in itself.
I would rather use : 
```
query = env["product.product"]._search([("uom_ids", "!=", False)])
all_ids = query.get_result_ids()
```

And browse them inside `split_every(1000, all_ids)`

Ricardoalso (commented):
LGTM in the overall

BhaveshHeliconia (commented):
Since Odoo 19 natively merged packaging functionality into the UoM model, could you explain the specific use case for this new module? What logistics or warehouse limitations are we trying to overcome?

product_packaging/hooks.py
ivantodorovich: AFAIK, no, `search` won't fetch any field -- just their IDs.
So no need to go through the lower-level query

ivantodorovich:
> Since Odoo 19 natively merged packaging functionality into the UoM model, could you explain the specific use case for this new module? What logistics or warehouse limitations are we trying to overcome?

1) Usability: with the new way it's a nightmare to configure product packagings barcodes. Users have to open the UoM and add the variant barcodes (`product.uom`) to the UoM record

2) Extensibility: we need to implement additional fields on the product variant packaging level (Weight, Volume, Dimensions, other custom ones..). In standard, there's no such model anymore.

ivantodorovich:
I switch to draft for there's a bug detected by our BAs. I will mark as ready again once it's resolved

product_packaging/hooks.py
Ricardoalso: My bad, I was thinking in the context of standard prefetching and cache accumulation, but using split_every combined with browse(batch_ids) here completely avoids that issue. 

This works perfectly! 😉

ivantodorovich:
All's good. Ready to review ✅

product_packaging/models/product_template.py
```diff
@@ -0,0 +1,85 @@
...
+
+        Going through the variant's real One2many applies per-row field updates
+        (weight, volume, ...) correctly, which the computed field's recompute
+        would otherwise discard.
+        """
+        self.ensure_one()
+        if len(self.product_variant_ids) == 1:
+            self.product_variant_ids.write({"packaging_ids": commands})
+
+    @api.depends("product_variant_ids.packaging_ids.uom_id")
```
BhaveshHeliconia: it Should be:
@api.depends(
    "product_variant_ids",
    "product_variant_ids.packaging_ids",
)

product_packaging/models/product_packaging.py
```diff
@@ -0,0 +1,109 @@
...
+        # OVERRIDE to add the uom to the field label
+        res = super().fields_get(allfields, attributes)
+        if self.env.context.get("uom_inline_field_labels"):
+            if "weight" in res and "string" in res["weight"]:
+                weight_uom_name = self.env[
+                    "product.template"
+                ]._get_weight_uom_name_from_ir_config_parameter()
+                res["weight"]["string"] += f" ({weight_uom_name})"
+            if "volume" in res and "string" in res["volume"]:
+                volume_uom_name = self.env[
```
BhaveshHeliconia: it should be :

ProductTemplate = self.env["product.template"]

weight_uom_name = (
 ProductTemplate._get_weight_uom_name_from_ir_config_parameter()
)
volume_uom_name = (
    ProductTemplate._get_volume_uom_name_from_ir_config_parameter()
)

product_packaging/models/product_packaging.py
```diff
@@ -0,0 +1,109 @@
...
+    )
+
+    @api.depends("uom_id", "product_id.uom_id")
+    def _compute_qty(self):
+        for packaging in self:
+            packaging.qty = packaging.uom_id._compute_quantity(
+                1.0, packaging.product_id.uom_id, round=False, raise_if_failure=False
+            )
+
+    @api.depends("product_id.weight", "uom_id.factor")
```
BhaveshHeliconia:
```suggestion
    @api.depends("product_id.weight", "uom_id.factor", "qty")
```

product_packaging/models/product_packaging.py
```diff
@@ -0,0 +1,109 @@
...
+    def _compute_weight(self):
+        for packaging in self:
+            packaging.weight = packaging.product_id.weight * packaging.qty
+
+    @api.depends("product_id.volume", "uom_id.factor")
```
BhaveshHeliconia:
```suggestion
    @api.depends("product_id.volume", "uom_id.factor", "qty")
```

product_packaging/models/product_template.py
ivantodorovich: Why? 🤔

product_packaging/models/product_packaging.py
ivantodorovich: `qty` is non-stored, so no.

`qty` ultimately/indirectly depends on the `uom_id` factor so IMO it's ok like this.

product_packaging/models/product_packaging.py
ivantodorovich: same as previous comment

product_packaging/models/product_packaging.py
ivantodorovich: it's the same thing

product_packaging/models/product_template.py
BhaveshHeliconia: When you define @api.depends('product_variant_ids.packaging_ids.uom_id'), you are telling the Odoo framework:

"Only recalculate this field if the uom_id column on an existing packaging record is updated."

Look at what happens in the database cache during different user actions:

- Scenario A: Changing an existing value (Works) A user changes a packaging line from Box ($UoM_1$) to Pallet ($UoM_2$). Odoo detects the change on uom_id and successfully triggers the compute.

- Scenario B: Adding a new packaging line (Fails) A user clicks "Add a line" and inserts a brand new packaging record. At the exact moment they create it, the framework is watching uom_id. Because this is a new record creation rather than a modification of an existing uom_id field pointer, the dependency chain can drop the event. The template field remains un-updated.

- Scenario C: Deleting a packaging line (Fails) A user deletes a packaging line. The uom_id field didn't change its value—the entire record just vanished. Because the specific field wrapper wasn't modified, the parent template continues to cache the deleted UoM, causing "Ghost" data to appear in the UI.

product_packaging/models/product_packaging.py
yankinmax:
```suggestion
    @api.depends("product_id.weight", "product_id.uom_id", "uom_id.factor")
```
If `product_id.uom_id` changes after packaging creation, `qty` changes but the stored field may stay stale

product_packaging/models/product_packaging.py
yankinmax:
```suggestion
    @api.depends("product_id.volume", "product_id.uom_id", "uom_id.factor")
```

yankinmax (approved):

product_packaging/models/product_template.py
ivantodorovich: AFAIK `product_variant_ids.packaging_ids.uom_id` implies `product_variant_ids.packaging_ids` which implies `product_variant_ids`. But I'll add the change as it won't harm

ivantodorovich:
I updated the PR to address your suggestions. Thanks!

product_packaging/models/product_packaging.py
```diff
@@ -0,0 +1,115 @@
+# Copyright 2026 Camptocamp SA (https://www.camptocamp.com).
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+
+from odoo import api, fields, models
+
+
+class ProductPackaging(models.Model):
+    _name = "product.packaging"
```
divad1196, line 8: Shouldn't it `_inherits` from Odoo's model instead since it's a proxy ?

product_packaging/models/product_product.py
```diff
@@ -0,0 +1,61 @@
...
+    packaging_ids = fields.One2many(
+        comodel_name="product.packaging",
+        inverse_name="product_id",
+        string="Packaging",
+    )
+
+    @api.model_create_multi
+    def create(self, vals_list):
+        products = super().create(vals_list)
+        # New variants inherit their template's packaging UoMs (create-only, so
```
divad1196: Phrasing of the comment is misleading IMO.
It looks like we are talking about something done by Odoo that we want to fix.

If I understood correctly, then it should be `New variants must/should inherit`

product_packaging/models/product_product.py
```diff
@@ -0,0 +1,61 @@
...
+        # variants created directly with ``packaging_ids`` keep them).
+        products._create_missing_packagings()
+        # An explicitly provided ``uom_ids`` gets wiped by its own recompute
+        # while the variant is created (no packaging exists yet); re-apply it
+        # now to materialize the packagings.
+        for product, vals in zip(products, vals_list, strict=True):
+            if vals.get("uom_ids"):
```
divad1196: Would avoid 2 lookups
```python
uom_ids = vals.get("uom_ids")
if uom_ids:
    product.uom_ids = uom_ids
```

or, since python3.8, we can use the [walrus operator](https://docs.python.org/3/whatsnew/3.8.html#assignment-expressions)...
```python
if uom_ids := vals.get("uom_ids"):
    product.uom_ids = uom_ids
```
but half of the python community hates it.

product_packaging/models/product_product.py
```diff
@@ -0,0 +1,61 @@
...
+        string="Packaging",
+    )
+
+    @api.model_create_multi
+    def create(self, vals_list):
+        products = super().create(vals_list)
+        # New variants inherit their template's packaging UoMs (create-only, so
+        # variants created directly with ``packaging_ids`` keep them).
+        products._create_missing_packagings()
+        # An explicitly provided ``uom_ids`` gets wiped by its own recompute
```
divad1196, line 22: Which recompute wipes the `uom_ids` ?

divad1196 (commented):
I am unsure about what the PR brings.
I was not aware that Odoo 19.0 merged UoM and packaging, so I checked:

There is `product.uom` which is the link between `product.product` and `uom.uom`, and it has barcodes.
https://github.com/odoo/odoo/blob/19.0/addons/product/models/product_product.py

AFAIU, point 1& 2 from the description are still there in V19 ?

product_packaging/models/product_packaging.py
ivantodorovich: Which model?

product_packaging/models/product_product.py
ivantodorovich: The `uom_ids` is a `product.template` field (odoo standard).

Whenever a variant is created, we are creating their corresponding `product.packaging` records.
Packagings is essentially a product of variants and uom_ids

product_packaging/models/product_product.py
ivantodorovich: It's a matter of taste I'd say, no real benefit except the negligible micro-optimization

product_packaging/models/product_product.py
ivantodorovich: `uom_ids` is computed from the variants' `packaging_ids`.

However, when creating a product with attributes (and so, variants), Odoo would create the product.template first and populate the variants afterwards. For some reason I could not understand, the uom_ids gets recomputed at an intermediate step when the variants are not yet created (and so, no packagings), and gets its value cleared.

The issue is covered by the unit test: `test_create_template_with_attributes_and_uom_ids`

product_packaging/models/product_packaging.py
divad1196: Based on this
> The product.packaging model is used as a proxy to configure the lower-level product.template.uom_ids and its product.uom records for the packaging barcodes.

It seems like this is a proxy on `uom.uom`

product_packaging/models/product_product.py
divad1196: Yes, I am aware of uom_ids,
this is why the comment is misleading.

The current comment looks like "New variants inherit" is a default/native behavior that we are trying to fix. A better phrasing would be "New variants **must** inherit ...". Or the function `_create_missing_packagings` is explicit enough

product_packaging/models/product_product.py
divad1196: @ivantodorovich optimization is not even on the table, completely neglagtable indeed.

I am not keen on the walrus operator, but the first proposal enters the [Parse, don’t validate](https://lexi-lambda.github.io/blog/2019/11/05/parse-don-t-validate/). 
In short: data retrieval and checks should be done once, in one place, then operate on the trusted result.

This avoid bugs caused by branching and simplifies the reading.
Little impact here, but a good pattern to follow.in general.

I won't push it, it was just an advice.

santostelmo (commented):
Another model that no longer exists is `product.packaging.type`. Could this model be integrated as well? Or any reason to not include it?

ivantodorovich:
> Another model that no longer exists is `product.packaging.type`. Could this model be integrated as well? Or any reason to not include it?

Is that the `product.packaging.level` implemented in the `product_package_level` module?

product_packaging/models/product_packaging.py
ivantodorovich: No, in fact there's no other model that represents what the `product.packaging` does: it's the packaging information at the variant level. It represents the cartesian product of `product_variants_ids * uom_ids`

I've considered using a model with the `_table` equal to the `uom_ids.relation` table. But it isn't possible because that leads to packagings defined at the template level.

Instead, the packagings are computed from the variants and template.uom_ids; and inversely they recompute the template's uom_ids when created/deleted/modified. That's why I say it's a "proxy"; because it ends up modifying (and reacting to changes) on the product.template `uom_ids` -- and leveraging this for business flows (stock, sale, purchase) instead of trying to implement a packaging_id there (like it was in previous versions)

santostelmo:
> > Another model that no longer exists is `product.packaging.type`. Could this model be integrated as well? Or any reason to not include it?
> 
> Is that the `product.packaging.level` implemented in the `product_package_level` module?

No it is another one. I am talking about [this one ](https://github.com/odoo/odoo/blob/18.0/addons/stock/models/stock_package_type.py)

`product_package_level` I saw it was implemented in [another PR](product_package_level)

ivantodorovich:
Oh, that's still there but used for Packages (stock) rather than Packagings (product).
It's the [uom.uom.package_type_id](https://github.com/OCA/odoo/blob/806fffb3f461abfd36b5f3fc1a983fd506f1fc81/addons/stock/models/product.py)

santostelmo:
> > > Another model that no longer exists is `product.packaging.type`. Could this model be integrated as well? Or any reason to not include it?
> > 
> > 
> > Is that the `product.packaging.level` implemented in the `product_package_level` module?
> 
> No it is another one. I am talking about [this one ](https://github.com/odoo/odoo/blob/18.0/addons/stock/models/stock_package_type.py)
> 
> `product_package_level` I saw it was implemented in [another PR](product_package_level)

My bad actually `product.packaging.type`is still present.

product_packaging/models/product_packaging.py
```diff
@@ -0,0 +1,118 @@
...
+
+    @api.depends("product_id.weight", "product_id.uom_id", "uom_id.factor")
+    def _compute_weight(self):
+        for packaging in self:
+            packaging.weight = packaging.product_id.weight * packaging.qty
+
+    @api.depends("product_id.volume", "product_id.uom_id", "uom_id.factor")
+    def _compute_volume(self):
+        for packaging in self:
+            packaging.volume = packaging.product_id.volume * packaging.qty
```
jbaudoux: Modifying the weight/volume of the product should not impact the packaging. For the weight, there is the tare weight that comes on top of the net weight. For the volume, it highly depends how it is packaged and it should come with length, width, height.
Those fields should be left to packaging dimension/weight modules from stock-logistics-tracking like before.
The only case where it makes sense, it is when the uom_id is the product uom_id.
But anyway, I would leave this outside this module.

product_packaging/models/product_product.py
```diff
@@ -0,0 +1,61 @@
...
+
+    def _create_missing_packagings(self):
+        """Create the packagings for the template UoMs not yet present.
+
+        Create-only: never removes packagings, so a variant created directly
+        with ``packaging_ids`` keeps them even when ``uom_ids`` is not set yet.
+        """
+        to_create = []
+        for product in self:
+            target_uoms = product.product_tmpl_id.uom_ids
```
jbaudoux, line 38: Here I'm sure we will soon need to be able to restrict a uom to a variant. Either in this module or in another.
I would call here a method to collect the packagings (uoms) of the product.

jbaudoux (commented):
I would have proxied also the product uom_id
That would provide a single place to manage attributes like dimension, weight, sale, purchase...
But maybe complex for the barcode that is char for product uom and 2many for product uoms

product_packaging/models/product_packaging.py
ivantodorovich: I dropped the depends on the product field here: https://github.com/OCA/product-attribute/compare/1473e6de41fd4210dcd44df5c053affe4403a80e..254bb056f085aaef9ed57a418c698dd502364f7b

product_packaging/models/product_product.py
ivantodorovich: Implemented here: https://github.com/OCA/product-attribute/compare/1473e6de41fd4210dcd44df5c053affe4403a80e..254bb056f085aaef9ed57a418c698dd502364f7b

product_packaging/models/product_product.py
ivantodorovich: changed in https://github.com/OCA/product-attribute/compare/1473e6de41fd4210dcd44df5c053affe4403a80e..254bb056f085aaef9ed57a418c698dd502364f7b

product_packaging/models/product_packaging.py
```diff
@@ -0,0 +1,118 @@
...
+    )
+
+    @api.depends("uom_id", "product_id.uom_id")
+    def _compute_qty(self):
+        for packaging in self:
+            packaging.qty = packaging.uom_id._compute_quantity(
+                1.0, packaging.product_id.uom_id, round=False, raise_if_failure=False
+            )
+
+    @api.depends("product_id.uom_id", "uom_id.factor")
```
jbaudoux, line 86: What about ?
```suggestion
    @api.depends("product_id.uom_id.factor", "uom_id.factor")
```

I have a small concern regarding changing the product uom. I have customers where it's a common practice to refine the product and change to another uom but with the same factor.
For example by default uom=unit and then you change to bottle or bag.
Would you add a test that if the depends are changed but not the factors, then the values are not lost?

jbaudoux: With the net weight module, it could also help but I see the weight is not recomputed using the existing net weight but the net weight gets overriddy

jbaudoux: Back on this as I have a bigger concern. Before the value was not computed and that allowed to know if the data was known and filled in or missing. How do you solve that?

jbaudoux: See migration of total_weight_from_packaging

> Behaviour is unchanged: packagings without a weight are still ignored, so their
quantity falls back to the smaller packagings and finally to the product weight.

jbaudoux (commented):
I understand you try to provide the feature of packaging as it was before. Nevertheless, I would still split it in several modules and put the concept in their respective module.
Weight and net weight in product_packaging_weight
Width, length, height, volume in peoduct_packaging_dimension

ivantodorovich:
> I understand you try to provide the feature of packaging as it was before. Nevertheless, I would still split it in several modules and put the concept in their respective module. Weight and net weight in product_packaging_weight Width, length, height, volume in peoduct_packaging_dimension

Hello!

My reasoning is that `weight` and `volume` are standard fields in the Product model, so they should be standard fields in this base packaging module too.  The fields have `optional="show"` so they can easily be hidden by users that don't care about them.

The product `net_weight` is added by `product_net_weight`, and so it requires a glue module for the packaging (https://github.com/OCA/product-attribute/pull/2322).

The product dimensions are also added by `product_dimension`, so they require a separate glue module for the packaging (https://github.com/OCA/product-attribute/pull/2327)

PR #2321, opened 2026-06-11
[18.0] [ADD] product_state_control_sale_purchase

FyasserF:
This module extends the functionality of the product state management by adding controls  on Sales and Purchase orders confirmation based on the product's lifecycle state.


product_state_control_sale_purchase/models/purchase_order.py
```diff
@@ -0,0 +1,47 @@
...
+                lambda line: (
+                    line.product_id
+                    and line.product_id.product_state_id
+                    and not line.product_id.product_state_id.authorized_to_be_bought
+                )
+            )
+
+            if forbidden_products:
+                raise ValidationError(
+                    _(
```
BhaveshHeliconia, line 42: The validation messages should be refactored to identify which specific products are blocking order confirmation (rather than a generic error message), which is crucial for usability in orders with many lines.

BhaveshHeliconia:
@FyasserF Could you add test cases?

PR #2323, opened 2026-06-12
[18.0][IMP] product_assortment: Decide if we want to use the domain fields

CarlosRoca13:
cc @Tecnativa TT63090

### DIFF
--- a/product_assortment/models/ir_filters.py
+++ b/product_assortment/models/ir_filters.py
@@ -32,6 +32,7 @@ class IrFilters(models.Model):
     record_count = fields.Integer(compute="_compute_record_count")
 
     is_assortment = fields.Boolean(default=lambda x: x._get_default_is_assortment())
+    use_partner_domain = fields.Boolean()
     partner_domain = fields.Text(default="[]", required=True)
     all_partner_ids = fields.Many2many(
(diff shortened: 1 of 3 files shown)


product_assortment/models/ir_filters.py
```diff
@@ -87,13 +91,13 @@ def get_partner_domain_fields(self):
...
 
-    @api.depends("partner_ids", "partner_domain")
+    @api.depends("use_partner_domain", "partner_ids", "partner_domain")
     def _compute_all_partner_ids(self):
         """Summarize selected partners and partners from partner domain field"""
         for ir_filter in self:
             if not ir_filter.is_assortment:
                 ir_filter.all_partner_ids = False
-            elif ir_filter.partner_domain != "[]":
+            elif self.use_partner_domain:
```
eduezerouali-tecnativa:
```suggestion
            elif ir_filter.use_partner_domain:
```
I think this is why CI is red

eduezerouali-tecnativa:
Test should be amend by adding `"use_partner_domain": True`  where need it, as no we decide if we use the domain fields.

eduezerouali-tecnativa (approved):
LGTM

PR #2113, opened 2025-10-24
[19.0][MIG] product_logistics_uom: Migration from 18.0 to 19.0

eugenios73:
product_logistics_uom migration to 19.0

Depends on https://github.com/OCA/product-attribute/pull/2089

PR #2327, opened 2026-06-16
[19.0][MIG] product_packaging_dimension: Migration to 19.0

yankinmax:
- Added new dependency on `product_packaging` along with `product_logistics_uom` to align with the 19.0 module split.

- Kept `product_packaging_dimension` focused on dimensions only: it now adds `packaging_length`, `width`, `height`, and the length UoM label.

- Removed duplicated ownership of fields now provided by `product_packaging`: `weight`, `volume`, `weight_uom_name`, and `volume_uom_name`.

- Updated `_compute_volume()` as an override with `super()`, so dimensional volume calculation extends the base packaging behavior.

- Added a pre-migration script to convert legacy per-record `length_uom_id` and `weight_uom_id` values into the configured global logistics UoMs before the old columns disappear.

- Updated views and tests to use the new base packaging view structure and global UoM configuration, without changing the module’s business purpose.

Depends on:
- [ ] https://github.com/OCA/product-attribute/pull/2320
- [ ] https://github.com/OCA/product-attribute/pull/2113


product_packaging_dimension/models/product_packaging.py
```diff
@@ -5,107 +5,128 @@
...
-        "uom.uom",
-        string="Volume Units of Measure",
-        domain=lambda self: [
-            ("category_id", "=", self.env.ref("uom.product_uom_categ_vol").id)
-        ],
-        help="Packaging volume unit of measure",
-        default=lambda self: self.env[
-            "product.template"
-        ]._get_volume_uom_id_from_ir_config_parameter(),
+    _positive_weight = models.Constraint(
```
ivantodorovich: Can you move this to `product_packaging`, because that's where the weight is defined?

product_packaging_dimension/models/product_packaging.py
```diff
@@ -5,107 +5,128 @@
...
-            "product.template"
-        ]._get_length_uom_id_from_ir_config_parameter(),
-    )
     length_uom_name = fields.Char(
         string="Length unit of measure label",
-        related="length_uom_id.name",
+        compute="_compute_length_uom_name",
     )
+    # OVERRIDE: to preserve the precision historically provided from v18.0.
+    volume = fields.Float(digits=(8, 4))
```
ivantodorovich: Maybe set digits=0 to disable rounding altogether. It makes sense for the volume

product_packaging_dimension/__manifest__.py
```diff
@@ -0,0 +1,19 @@
...
+{
+    "name": "Product Packaging Dimension",
+    "summary": "Manage packaging dimensions and weight",
+    "version": "19.0.1.0.0",
+    "category": "Product",
+    "author": "Camptocamp, Akretion, Odoo Community Association (OCA)",
+    "license": "AGPL-3",
+    "depends": [
+        "product_packaging",
+        "product_logistics_uom",
```
ivantodorovich:
```suggestion
```

Is this dependency truly needed?
It doesn't seem so, maybe I'm missing something though

yankinmax: It was historically in this module and UoMs configuration is defined there.
I would leave it here not to move the features.

product_packaging_dimension/models/product_packaging.py
yankinmax: Will do it

product_packaging_dimension/__manifest__.py
ivantodorovich: Since we're reading the length uom from the config parameter now, it doesn't seem required anymore.

We certainly don't need it so I prefer it's not enforced.
If there's no hard technical dependency, can you remove it?

yankinmax: `product_packaging_dimension` can technically call `_get_length_uom_id_from_ir_config_parameter` without `product_logistics_uom`, because Odoo core already defines that helper in https://github.com/odoo/odoo/blob/65937b5cdc6c638a9e9ff7085cbab1567ca24142/addons/product/models/product_template.py#L367.
But core only supports the legacy binary params like `product.volume_in_cubic_feet` / `product.weight_in_lbs`.
The new arbitrary UoM config params we migrated to, like `product_default_length_uom_id`, `product_default_volume_uom_id`, and `product_default_weight_uom_id`, are implemented by `product_logistics_uom` in https://github.com/OCA/product-attribute/blob/fbd0bca6b8ce9fc30db45d813635b60cdf7baf93/product_logistics_uom/models/product_template.py#L69 and exposed in settings.

yankinmax: ok for me, we can remove the dependency, there is no "hard" dependency need finally

product_packaging_dimension/views/product_packaging.xml
```diff
@@ -0,0 +1,34 @@
+<?xml version="1.0" encoding="UTF-8" ?>
+<odoo>
```
ivantodorovich, line 2: Can you add also the fields to the `list` view with `optional="show"` ?
Add also the corresponding `fields_get` override so the column labels include the uom name, like this: https://github.com/camptocamp/product-attribute/blob/4ab7af4b1e39efbad5dab588de4cebb9f917661f/product_packaging/models/product_packaging.py#L93-L109

yankinmax: Yes of course, thank you for the review

ivantodorovich (approved):
Thanks!!

imlopes (approved):
LGTM

BhaveshHeliconia (approved):
LGTM!

jbaudoux (approved):

PR #2332, opened 2026-06-22
[18.0][FIX] product_pricelist_product_price_history: fix history atta…

astirpe:
…ched to wrong variant

Description of the issue:
When a user updates a pricelist rule for a specific product variant (e.g., the 30ml bottle), the price history log incorrectly attaches itself to the first variant of that product template (e.g., the 8ml bottle). This happens because the system's logic checks for the template's default variant before checking for the specific variant that was actually modified.

Proposed solution:
Reversed the fallback logic in the write method of the pricelist item. By changing the evaluation order to prioritize the specific variant (item.product_id.id or item.product_tmpl_id.product_variant_id.id), the history log now correctly attaches to the exact variant being updated, and only falls back to the template's variant when a global template rule is modified.

### DIFF
--- a/product_pricelist_product_price_history/models/product_pricelist_item.py
+++ b/product_pricelist_product_price_history/models/product_pricelist_item.py
@@ -16,8 +16,9 @@ class ProductPricelistItem(models.Model):
                 ):
                     self.env["product.pricelist.item.history"].create(
                         {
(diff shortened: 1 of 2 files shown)


peluko00 (approved):
LGTM!

BhaveshHeliconia (approved):
LGTM!

PR #2334, opened 2026-06-23
[18.0][IMP] product_pricelist_direct_print*: batch price computation

CarlosRoca13:
The price column recomputed the pricelist price one product at a time (``_get_product_price`` per report row), which made the native ``_get_applicable_rules`` rule search run once per product. On large pricelists this dominated the report render time.

Precompute every price in a single batched ``_get_products_price`` call and inject the result in the context as ``product_prices`` so ``_compute_product_price`` reads it instead of recomputing. This collapses the per-product rule searches of the printed pricelist into a single one, both for the PDF report and the xlsx export, without changing the price values nor the public method signatures.

cc @Tecnativa TT63381

ping @pedrobaeza @carlosdauden

### DIFF
--- a/product_pricelist_direct_print/reports/report_product_pricelist.xml
+++ b/product_pricelist_direct_print/reports/report_product_pricelist.xml
@@ -5,6 +5,10 @@
     <template id="report_product_pricelist_document">
         <t t-call="web.html_container">
             <t t-set="o" t-value="o.with_context(lang=o.lang)" />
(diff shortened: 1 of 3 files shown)


pedrobaeza (approved):

carlosdauden (approved):

BhaveshHeliconia (approved):
LGTM!

PR #2335, opened 2026-06-23
[FIX] product_cost_security: Allow fieldless write() to succeed.

bobslee:
check_field_access_rights raises an AccessError whenever a non cost-editor touches a cost field. However, when fields is empty (e.g. an empty vals dict reaches write), the base ORM expands it to *every* accessible field, which includes the protected cost fields, and the mixin then wrongly denies the write.

This happens, for instance, when l10n_eu_product_adr pops every value out of the write vals (is_dangerous/adr_goods_id are delegated to the to the variants) and delegates an empty dict to super().write({}).

A field-less, non-superuser write/create targets no cost field, so we bypass the cost-edit restriction and rely on the standard ORM checks. The return value is irrelevant here: write ignores it.

### DIFF
--- a/product_cost_security/models/product_cost_security_mixin.py
+++ b/product_cost_security/models/product_cost_security_mixin.py
@@ -61,7 +61,12 @@ class ProductCostSecurityMixin(models.AbstractModel):
         `product_cost_security.group_product_edit_cost` group.
         """
         valid_fields = super().check_field_access_rights(operation, fields)
-        if self.env.su:
+        # When called without explicit fields (e.g. an empty ``write({})``),
+        # ``super()`` expands ``valid_fields`` to *all* accessible fields. Such
+        # an operation does not actually touch the cost fields, so it must not
+        # be blocked: only enforce the extra protection on the fields the
+        # caller explicitly requested.
+        if self.env.su or not fields:
             return valid_fields
         product_cost_fields = self._product_cost_security_fields().intersection(
             valid_fields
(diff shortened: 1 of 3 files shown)


sergio-teruel:
Please review https://github.com/OCA/product-attribute/pull/2254

yajo:
Why would you call `write({})` with an empty dict?

bobslee:
> Why would you call `write({})` with an empty dict?

This happens, for instance, when `⁠ l10n_eu_product_adr ⁠` pops every value out of the write `⁠ vals ⁠` (`⁠ is_dangerous ⁠⁠ / ⁠⁠ adr_goods_id ⁠` are delegated to the variants) and delegates an empty dict to `⁠ super().write({}) ⁠`.

yajo:
However, isn't this opening a problem? You're telling upstream they can write some fields, when they can't.

pedrobaeza:
Shouldn't it be ` l10n_eu_product_adr ⁠` which don't call super if `vals` is empty?

chienandalu (commented):
I think the issue is real.  In our case, the issue isn't coming from ` l10n_eu_product_adr ⁠` but from `sale_order_product_recommendation`, which precisely can trigger an empty write on the sale lines (see https://github.com/OCA/sale-workflow/blob/ba966f0e3fabaafc941b7079035fafc582f0d639/sale_order_product_recommendation/wizards/sale_order_recommendation.py#L320)

So IMO (aside from fixing those tricky bugs) we should prevent any potential future ones of the same kind impacting this weak point.

In any case, even being useless, nothing should prevent a user triggering an empty values writing on any model. The current error is really misleading in that regard.

chienandalu (approved):

PR #2342, opened 2026-07-09
[19.0] [MIG] nutritional_info: Migration to 19.0

Borruso:
https://github.com/OCA/product-attribute/issues/2075


kalpeshgajera-uncannycs:
**Suggestion:** During testing, I noticed a validation inconsistency. Duplicate nutritional types are correctly restricted on the Product Template (product.template); however, the same validation is not implemented on the Product Variant (product.product). As a result, users can add duplicate nutritional types from the variant form, and those duplicates are then reflected on the product template. It would be beneficial to apply the same validation on product.product to ensure consistent data integrity across both models.

BhaveshHeliconia (approved):
Tested locally code and functional review LGTM!

PR #2344, opened 2026-07-13
[18.0][IMP] product_secondary_unit: add decimal precision to factor

les-adhoc:
## Description

The secondary unit `factor` field is declared as a plain `Float` without `digits`, so the web client renders it with the default 2 decimals while the value is stored and used in the conversions with full precision.

When configuring a fractional conversion (e.g. `1/2.25 = 0.444444…`), the user sees a truncated value like `0.44` and cannot tell that the effective factor carries more precision. This leads to apparent conversion mismatches (quantities off by a small amount) that are actually just a display problem.

## Change

- Add a configurable decimal precision **`Secondary Unit Factor`** (default `6`).
- Use it on `product.secondary.unit.factor` via `digits="Secondary Unit Factor"`.

This mirrors how Odoo handles precision-sensitive fields elsewhere and fixes the display consistently across all views (template form, variant form, list, export), while letting admins tune the precision. Storage keeps working the same for any value within the configured precision.

## How to test

1. Enable *Units of Measure* (`uom.group_uom`).
2. On a product, add a secondary unit and set a factor with more than 2 decimals (e.g. `0.444444`).
3. The factor is displayed with 6 decimals, matching the value used in the conversions.


mav-adhoc (approved):

mav-adhoc:
LGTM

les-adhoc:
@sergio-teruel please review

sergio-teruel:
ping @carlosdauden

PR #2205, opened 2026-02-06
[16.0][OU-FIX] product_packaging_level: Fixed migration script for Migration 16.0

baguenth:
This fixes the migration script. The script currently fails when migrating to 16.0 

```
Traceback (most recent call last):
  File "/path/to/odoo/odoo/service/server.py", line 1326, in preload_registries
    registry = Registry.new(dbname, update_module=update_module)
               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "<decorator-gen-16>", line 2, in new
  File "/path/to/odoo/odoo/tools/func.py", line 87, in locked
    return func(inst, *args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/path/to/odoo/odoo/modules/registry.py", line 87, in new
    odoo.modules.load_modules(registry, force_demo, status, update_module)
  File "/path/to/odoo/odoo/modules/loading.py", line 489, in load_modules
    processed_modules += load_marked_modules(cr, graph,
                         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/path/to/odoo/odoo/modules/loading.py", line 374, in load_marked_modules
    loaded, processed = load_module_graph(
                        ^^^^^^^^^^^^^^^^^^
  File "/path/to/odoo/odoo/modules/loading.py", line 238, in load_module_graph
    migrations.migrate_module(package, 'post')
  File "/path/to/openupgrade/openupgrade_framework/odoo_patch/odoo/modules/migration.py", line 21, in migrate_module
    MigrationManager.migrate_module._original_method(self, pkg, stage)
  File "/path/to/odoo/odoo/modules/migration.py", line 170, in migrate_module
    exec_script(self.cr, installed_version, pyfile, pkg.name, stage, stageformat[stage] % version)
  File "/path/to/odoo/odoo/modules/migration.py", line 188, in exec_script
    migrate(cr, installed_version)
  File "/path/to/venv/lib/python3.12/site-packages/openupgradelib/openupgrade.py", line 2406, in wrapped_function
    func(
  File "/path/to/venv/lib/python3.12/site-packages/odoo/addons/product_packaging_level/migrations/16.0.1.0.0/post-migration.py", line 9, in migrate
    openupgrade.load_data(
  File "/path/to/venv/lib/python3.12/site-packages/openupgradelib/openupgrade.py", line 389, in load_data
[...]

PR #1963, opened 2025-04-30
[16.0][FIX] product_packaging_level: add true migration scripts

MiquelRForgeFlow:
See:
-  https://github.com/OCA/OpenUpgrade/pull/4964.

PR #2347, opened 2026-07-16
[16.0][FIX]product_packaging_level: fix migration path

hparfr:
I had the following error with during an OpenUpgrade migration:

```
  File "/odoo/links/product_packaging_level/migrations/16.0.1.0.0/post-migration.py", line 9, in migrate
    openupgrade.load_data(
  File "/odoo/lib/python3.11/site-packages/openupgradelib/openupgrade.py", line 399, in load_data
    fp = _file_open(module_name, filename)
         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/odoo/lib/python3.11/site-packages/openupgradelib/openupgrade.py", line 321, in _file_open
    raise OSError(
OSError: Couldn't find file 16.0.1.0.0/noupdate_changes.xml in the upgrade paths (/odoo/external-src/openupgrade/openupgrade_scripts/scripts)
``` 

This script was introduced in https://github.com/OCA/product-attribute/pull/1963, then corrected in https://github.com/OCA/product-attribute/pull/2205 

See other working examples:
- https://github.com/oca/reporting-engine/blob/16.0/sql_export_mail/migrations/16.0.1.0.0/post-migration.py#L7
- https://github.com/oca/helpdesk/blob/16.0/helpdesk_mgmt_rating/migrations/16.0.1.0.0/post-migration.py#L10

FYI @baguenth @MiquelRForgeFlow

### DIFF
--- a/product_packaging_level/migrations/16.0.1.0.0/post-migration.py
+++ b/product_packaging_level/migrations/16.0.1.0.0/post-migration.py
@@ -7,7 +7,7 @@ from openupgradelib import openupgrade
 @openupgrade.migrate()
 def migrate(env, version):
     openupgrade.load_data(
-        env.cr, "product_packaging_level", "16.0.1.0.0/noupdate_changes.xml"
+        env.cr, "product_packaging_level", "migrations/16.0.1.0.0/noupdate_changes.xml"
     )
     openupgrade.delete_record_translations(
         env.cr,


baguenth (approved):
LGTM.

It's strange that I never encountered this error in any of our environments. In any case, the fix is straightforward and aligns with the approach we use in all of our other migration scripts.

BhaveshHeliconia (approved):

sebalix (approved):

PR #2351, opened 2026-07-28
[18.0][FIX] product_category_code: fixed issue when duplicating a product category

astirpe:
Removed @api.returns("self", lambda value: value.id) from the copy() method because it must return a recordset, not an ID. Returning an ID causes an error when duplicating records, since the ORM expects a record, not a primitive value.


rousseldenis (approved):
Trivial. Thanks !

PR #2308, opened 2026-05-21
[19.0][MIG] product_pricelist_item_uom

JordiBForgeFlow:
Standard migration from 16.0

PR #2364, opened 2026-08-04
[19.0][REF] product_pricelist_item_uom: backport from odoo 19.3

Ricardoalso:
Add a backport of Odoo 19.3's native pricelist-rule-by-UoM feature (odoo/odoo@d2648b1d9839), restricting pricelist item rules to a specific product packaging.

OCA/product-attribute already ships product_pricelist_item_uom, but it defines uom_id on product.pricelist.item as a computed field always defaulted to the product's base UoM, and it never filters rule selection. The two modules are declared mutually exclusive via the manifest excludes key.

- [x] Will be applied above https://github.com/OCA/product-attribute/pull/2308 using a refactoring commit ([REF] product_pricelist_item_uom: backport from odoo 19.3). Check https://github.com/OCA/product-attribute/pull/2308#pullrequestreview-4868407755

### DIFF
--- a/product_pricelist_item_uom/__manifest__.py
+++ b/product_pricelist_item_uom/__manifest__.py
@@ -7,10 +7,7 @@
     "version": "19.0.1.0.0",
     "category": "Sales/Sales",
     "website": "https://github.com/OCA/product-attribute",
-    "author": "Aion Tech, Odoo Community Association (OCA)",
-    "maintainers": [
-        "SirAionTech",
-    ],
+    "author": "Aion Tech, Camptocamp SA, Odoo Community Association (OCA)",
     "license": "AGPL-3",
     "depends": [
         "product",
@@ -19,5 +16,7 @@
     "data": [
         "views/product_pricelist_item_views.xml",
         "views/product_pricelist_views.xml",
+        "views/product_views.xml",
+        "report/product_pricelist_report_templates.xml",
     ],
 }
--- a/product_pricelist_item_uom/models/product_pricelist.py
+++ b/product_pricelist_item_uom/models/product_pricelist.py
@@ -1,0 +1,36 @@
+# Copyright 2026 Camptocamp SA (https://www.camptocamp.com).
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+
+from odoo import fields, models
+from odoo.fields import Domain
+
+
+class ProductPricelist(models.Model):
+    _inherit = "product.pricelist"
+
+    def _compute_price_rule(self, products, quantity, *, uom=None, **kwargs):
+        self_with_uom = self.with_context(pricelist_item_uom_id=uom.id) if uom else self
+        return super(ProductPricelist, self_with_uom)._compute_price_rule(
+            products, quantity, uom=uom, **kwargs
+        )
+
+    def _get_applicable_rules_domain(self, products, date, **kwargs):
+        domain = Domain(super()._get_applicable_rules_domain(products, date, **kwargs))
+        uom_id = self.env.context.get("pricelist_item_uom_id")
+        if uom_id:
+            domain &= Domain("uom_id", "=", False) | Domain("uom_id", "=", uom_id)
+        return domain
+
+    def _get_related_uoms(self, product):
+        """Return the packagings having their own rule in this pricelist.
+
+        :param product: product template or product variant record.
+        :return: UoMs defined on the matching pricelist items.
+        :rtype: uom.uom
+        """
+        domain = self._get_applicable_rules_domain(
+            product, fields.Datetime.now()
+        ) & Domain("uom_id", "!=", False)
+        return (
+            self.env["product.pricelist.item"].search_fetch(domain, ["uom_id"]).uom_id
+        )
--- a/product_pricelist_item_uom/models/product_pricelist_item.py
+++ b/product_pricelist_item_uom/models/product_pricelist_item.py
@@ -2,100 +2,46 @@
 #  License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
 
 from odoo import api, fields, models
-from odoo.tools import float_compare
 
 
 class ProductPricelistItem(models.Model):
     _inherit = "product.pricelist.item"
 
     uom_id = fields.Many2one(
-        string="UoM",
+        string="Packaging",
         comodel_name="uom.uom",
-        compute="_compute_uom_id",
-        store=True,
-        readonly=False,
+        help="Restrict this rule to a specific packaging of the product. "
+        "When empty, the rule applies to the product base unit of measure.",
     )
-    uom_min_quantity = fields.Float(
-        string="Min. Quantity in UoM",
-        digits="Product Unit",
-        compute="_compute_uom_min_quantity",
-        store=True,
-        readonly=False,
-    )
+    allowed_uom_ids = fields.Many2many("uom.uom", compute="_compute_allowed_uom_ids")
 
-    @api.depends(
-        "product_id",
-        "product_tmpl_id",
-    )
-    def _compute_uom_id(self):
+    @api.depends("product_tmpl_id", "product_tmpl_id.uom_id", "product_tmpl_id.uom_ids")
+    def _compute_allowed_uom_ids(self):
         for item in self:
-            product = item.product_id or item.product_tmpl_id
-            product_uom = product.uom_id
-            if item.uom_id != product_uom:
-                item.uom_id = product_uom
+            item.allowed_uom_ids = (
+                item.product_tmpl_id.uom_id | item.product_tmpl_id.uom_ids
+            )
 
-    def _sync_uom_min_quantity(
-        self, source_quantity, source_uom, target_uom, target_qty_field
-    ):
-        """Convert `source_quantity` (in `source_uom`) to `target_uom`.
-        Assign the result to field `self.target_qty_field`, if different."""
+    @api.model_create_multi
+    def create(self, vals_list):
+        items = super().create(vals_list)
+        items.filtered(
+            lambda item: item.applied_on == "2_product_category"
+        ).uom_id = False
+        return items
+
+    def write(self, vals):
+        if vals.get("applied_on") == "2_product_category":
+            vals["uom_id"] = False
+        return super().write(vals)
+
+    def _is_applicable_for(self, product, qty_in_product_uom):
         self.ensure_one()
-        if target_uom != source_uom:
-            expected_target_quantity = source_uom._compute_quantity(
-                source_quantity,
-                target_uom,
-                raise_if_failure=False,
-                round=False,
-            )
-        else:
-            expected_target_quantity = source_quantity
-
-        # Assign only if different,
-        # according to the precision of target field
-        all_digits, precision_digits = self.fields_get(
-            allfields=[
-                target_qty_field,
-            ],
-            attributes=[
-                "digits",
-            ],
-        )[target_qty_field]["digits"]
-        if float_compare(
-            self[target_qty_field],
-            expected_target_quantity,
-            precision_digits=precision_digits,
-        ):
-            self[target_qty_field] = expected_target_quantity
-
-    @api.onchange(
-        "product_id",
-        "product_tmpl_id",
-        "uom_id",
-        "uom_min_quantity",
-    )
-    def onchange_uom_min_quantity(self):
-        for item in self:
-            product = item.product_id or item.product_tmpl_id
-            product_uom = product.uom_id
-            item._sync_uom_min_quantity(
-                item.uom_min_quantity,
-                item.uom_id,
-                product_uom,
-                "min_quantity",
+        product.ensure_one()
+        qty_to_consider = qty_in_product_uom
+        if self.uom_id and self.uom_id != product.uom_id:
+            qty_to_consider = product.uom_id._compute_quantity(
+                qty_in_product_uom, self.uom_id
             )
 
-    @api.depends(
-        "product_id",
-        "product_tmpl_id",
-        "min_quantity",
-    )
-    def _compute_uom_min_quantity(self):
-        for item in self:
-            product = item.product_id or item.product_tmpl_id
-            product_uom = product.uom_id
-            item._sync_uom_min_quantity(
-                item.min_quantity,
-                product_uom,
-                item.uom_id,
-                "uom_min_quantity",
-            )
+        return super()._is_applicable_for(product, qty_to_consider)
--- a/product_pricelist_item_uom/readme/USAGE.md
+++ b/product_pricelist_item_uom/readme/USAGE.md
@@ -1,3 +1,21 @@
+Configuration
+-------------
+
+1. Enable *Sales → Configuration → Settings → Units of Measure & Packagings*.
+2. Open a product and add the packagings it is sold in
+   (*Sales* tab, *Packagings*).
+
+Defining a rule per packaging
+-----------------------------
+
+1. Go to *Sales → Products → Pricelists* and open a pricelist.
+2. Add a price rule applied on a single product.
+3. Set the *Packaging* field to one of the product's packagings, and the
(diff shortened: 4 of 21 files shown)


ivantodorovich (approved):
Thanks!

I'm not sure we want to keep both `product_pricelist_item_uom` and this one.
I'd be more inclined to have this one replace the existing `product_pricelist_item_uom`, consuming its name, given it's essentially the same feature -- but this one is aligned to what's implemented in standard in 19.3 (and future 20.0)

ping @JordiBForgeFlow @rousseldenis @monen17

domenicostragapede (approved):

monen17:
> I'm not sure we want to keep both `product_pricelist_item_uom` and this one.

I agree it's better to avoid `exclude`ing modules.

> I'd be more inclined to have this one replace the existing `product_pricelist_item_uom`, consuming its name, given it's essentially the same feature -- but this one is aligned to what's implemented in standard in 19.3 (and future 20.0)

I'd rather go with the `REF` commit you suggested in https://github.com/OCA/product-attribute/pull/2308#issuecomment-5191216500 so the commit history of `product_pricelist_item_uom` is preserved and we can see how its logic changes/gets migrated.

Ricardoalso:
> Thanks!
> 
> I'm not sure we want to keep both `product_pricelist_item_uom` and this one. I'd be more inclined to have this one replace the existing `product_pricelist_item_uom`, consuming its name, given it's essentially the same feature -- but this one is aligned to what's implemented in standard in 19.3 (and future 20.0)
> 
> ping @JordiBForgeFlow @rousseldenis @monen17



> > I'm not sure we want to keep both `product_pricelist_item_uom` and this one.
> 
> I agree it's better to avoid `exclude`ing modules.
> 
> > I'd be more inclined to have this one replace the existing `product_pricelist_item_uom`, consuming its name, given it's essentially the same feature -- but this one is aligned to what's implemented in standard in 19.3 (and future 20.0)
> 
> I'd rather go with the `REF` commit you suggested in [#2308 (comment)](https://github.com/OCA/product-attribute/pull/2308#issuecomment-5191216500) so the commit history of `product_pricelist_item_uom` is preserved and we can see how its logic changes/gets migrated.


Thanks for the follow-up on this! I like the option of pushing a REF commit (_[REF] product_pricelist_item_uom: backport from odoo 19.3_)  on top of `product_pricelist_item_uom`, as it would indeed preserve the commit history.

I will rebase my work on `ForgeFlow:19.0-mig-product_pricelist_item_uom` and rework this PR if there is consensus among the contributors.

BhaveshHeliconia (approved):
LGTM!

PaoloYam (approved):

product_pricelist_item_uom/readme/USAGE.md
```diff
@@ -1,14 +1,35 @@
-You sell Sugar by the kg, so its UoM is kg and the product price is
-1000€: you are selling it at 1000€/kg (a bit expensive as sugar goes,
```
monen17: My beloved 1000€/kg sugar example 🥲
Is there some way you could adapt the README to use it?
I think it's very intuitive and useful to have a real (as expensive as it was 😆) example to follow through

product_pricelist_item_uom/models/product_pricelist_item.py
```diff
@@ -1,101 +1,53 @@
...
+        ).uom_id = False
+        return items
+
+    def write(self, vals):
+        if vals.get("applied_on") == "2_product_category":
+            vals["uom_id"] = False
+        return super().write(vals)
+
+    def _is_applicable_for(
+        self, product, quantity, *, uom=None, qty_in_product_uom=False, **kwargs
```
monen17: Why all these arguments?
Please keep `super`'s signature (https://github.com/odoo/odoo/blob/64fbcd7cee2709aa2542a1a21611ad66e5ebc1c2/addons/product/models/product_pricelist_item.py#L526):
```py
def _is_applicable_for(self, product, qty_in_product_uom):
```

product_pricelist_item_uom/models/product_pricelist.py
```diff
@@ -0,0 +1,78 @@
...
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+
+from odoo import fields, models
+from odoo.fields import Domain
+
+
+class ProductPricelist(models.Model):
+    _inherit = "product.pricelist"
+
+    def _compute_price_rule(
```
monen17: This is overwriting `super`'s method https://github.com/odoo/odoo/blob/64fbcd7cee2709aa2542a1a21611ad66e5ebc1c2/addons/product/models/product_pricelist.py#L169 without calling `super`, that might break inheritance for other modules

product_pricelist_item_uom/models/product_pricelist.py
```diff
@@ -0,0 +1,78 @@
...
+                    currency=currency,
+                    **kwargs,
+                )
+            else:
+                price = 0.0
+            results[product.id] = (price, suitable_rule.id)
+        return results
+
+    def _get_applicable_rules_domain(
+        self, products, date, *, quantity=None, uom=None, **kwargs
```
monen17: Please keep `super`'s signature (https://github.com/odoo/odoo/blob/64fbcd7cee2709aa2542a1a21611ad66e5ebc1c2/addons/product/models/product_pricelist.py#L248):
```py
def _get_applicable_rules_domain(self, products, date, **kwargs):
```

product_pricelist_item_uom/report/product_pricelist_report.py
```diff
@@ -0,0 +1,56 @@
+# Copyright 2026 Camptocamp SA (https://www.camptocamp.com).
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+
+from odoo import models
+
+
+class ReportProductReport_Pricelist(models.AbstractModel):
+    _inherit = "report.product.report_pricelist"
+
+    def _get_product_data(self, is_product_tmpl, product, pricelist, quantities):
```
monen17, line 10: This is overwriting `super`'s method https://github.com/odoo/odoo/blob/64fbcd7cee2709aa2542a1a21611ad66e5ebc1c2/addons/product/report/product_pricelist_report.py#L48 without calling `super`, that might break inheritance for other modules

product_pricelist_item_uom/tests/common.py
```diff
@@ -0,0 +1,53 @@
+# Copyright 2026 Camptocamp SA (https://www.camptocamp.com).
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+
+from odoo.api import SUPERUSER_ID
+from odoo.fields import Command
+from odoo.tests.common import TransactionCase
+
+
+class TestPricelistItemUomCommon(TransactionCase):
```
monen17: Please keep using `BaseCommon` as it was in the original module

product_pricelist_item_uom/views/product_pricelist_item_views.xml
```diff
@@ -1,29 +1,54 @@
...
-    <record id="product_pricelist_item_form_view" model="ir.ui.view">
-        <field
-            name="name"
-        >Add Pricelist Rule UoM fields to Pricelist Rule form view</field>
+    <!--
+        Priority is raised because ``sale`` replaces the whole
+        ``pricelist_rule_limits`` group of this form; this inheritance has to be
+        applied after it so that the packaging field is not dropped.
+    -->
+    <record id="product_pricelist_item_view_form" model="ir.ui.view">
```
monen17: I'm not asking to migrate the whole module because the logic has changed and it might be complex, but changing the `id` of this view means that during a migration the old view is deleted and the new one is created, creating issues with inherited views.
Since it's little effort, can you keep the old `id`?

product_pricelist_item_uom/views/product_pricelist_views.xml
```diff
@@ -1,27 +1,22 @@
...
-  ~ Copyright 2023 Simone Rubino - Aion Tech
-  ~ License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
-  -->
+    Copyright 2026 Camptocamp SA (https://www.camptocamp.com).
+    License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+-->
 <odoo>
-    <record id="product_pricelist_view" model="ir.ui.view">
-        <field name="name">Add Pricelist Rule UoM fields to Pricelist form view</field>
+    <record id="product_pricelist_view_form" model="ir.ui.view">
```
monen17: Please keep the old `id` or add a migration for the XMLID of the view

product_pricelist_item_uom/__manifest__.py
```diff
@@ -7,7 +7,7 @@
     "version": "19.0.1.0.0",
     "category": "Sales/Sales",
     "website": "https://github.com/OCA/product-attribute",
-    "author": "Aion Tech, Odoo Community Association (OCA)",
+    "author": "Aion Tech, Camptocamp SA, Odoo Community Association (OCA)",
     "maintainers": [
         "SirAionTech",
```
monen17: Please remove the inactive user from here

monen17 (changes requested):
Thanks for the PR!
I reviewed the code, what worries me the most is overwriting `super`'s method without calling `super`, is there some way you can sneak a `super` call in?

I also tried it but I'm getting strange multiplication factors when using the product in a quotation: 

https://github.com/user-attachments/assets/c7195976-2473-46b3-8a49-a151b175fed2

Did I misconfigure something? It's a brand new demo DB with only this module and `sale_management` installed.
The pricelist is Image and the Packagings are Image

product_pricelist_item_uom/models/product_pricelist_item.py
Ricardoalso: Oh! 🤯 I ended up being confused by the changes made to the signature[ in this commit](https://github.com/odoo/odoo/commit/d2648b1d9839), but of course, we need to stick with the code in the 19.0 branch.

product_pricelist_item_uom/models/product_pricelist.py
Ricardoalso: same as https://github.com/OCA/product-attribute/pull/2364#discussion_r3839431520

Ricardoalso:
> My beloved 1000€/kg sugar example 🥲 Is there some way you could adapt the README to use it? I think it's very intuitive and useful to have a real (as expensive as it was 😆) example to follow through

Thanks for your review :pray: 

On the multiplication factor, it is not a misconfiguration. It's how `_compute_price` works in core : 
https://github.com/odoo/odoo/blob/64fbcd7cee2709aa2542a1a21611ad66e5ebc1c2/addons/product/models/product_pricelist_item.py#L593
`fixed_price` (and `price_surcharge`/margins on `formula` rules) is always interpreted per unit of the product's *base* UoM, then multiplied by the conversion factor.

Screenshot_2026-08-23_22-53-22
Screenshot_2026-08-23_22-54-05




Since the product's base UoM is g and kg = 1000 g, `_compute_price `multiplies the fixed price by that 1000× factor whenever the line is quoted in kg.

I adapted the /product_pricelist_item_uom/readme/USAGE.md file accordingly

monen17:
Thanks for the changes!
Could you please re-request the review as shown in https://docs.github.com/en/pull-requests/how-tos/create-pull-requests/requesting-a-pull-request-review#requesting-reviews-from-collaborators-and-organization-members when it's ready? So it's listed in the to-do list in Github, otherwise I'll probably forget about it :sweat_smile:

ivantodorovich:
@Ricardoalso you can rebase now

ivantodorovich:
rebased, this should be ready now

product_pricelist_item_uom/models/product_pricelist_item.py
```diff
@@ -1,101 +1,47 @@
 #  Copyright 2023 Simone Rubino - Aion Tech
-#  License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
```
monen17: Please keep both lines aligned

product_pricelist_item_uom/models/product_pricelist_item.py
```diff
@@ -1,101 +1,47 @@
...
-    uom_min_quantity = fields.Float(
-        string="Min. Quantity in UoM",
-        digits="Product Unit",
-        compute="_compute_uom_min_quantity",
-        store=True,
-        readonly=False,
+        help="Restrict this rule to a specific packaging of the product. "
+        "When empty, the rule applies to the product base unit of measure.",
     )
+    allowed_uom_ids = fields.Many2many("uom.uom", compute="_compute_allowed_uom_ids")
```
monen17, line 16: It would be better to have an explicit `relation`, otherwise any m2m between `product.pricelist.item` and `uom.uom` would share this table.

product_pricelist_item_uom/models/product_pricelist_item.py
monen17: :+1:

product_pricelist_item_uom/readme/USAGE.md
monen17: Thanks for keeping it :pray:

product_pricelist_item_uom/models/product_pricelist.py
monen17: Now it's calling `super` :pray:

product_pricelist_item_uom/models/product_pricelist.py
monen17: The signature now matches :pray:
I'm not sure returning a `Domain` is the same as returning a `list` (like `super` does) but they should be compatible enough.

product_pricelist_item_uom/report/product_pricelist_report.py
monen17: Now it's calling `super` :pray:

product_pricelist_item_uom/tests/common.py
monen17: :+1:

product_pricelist_item_uom/views/product_pricelist_item_views.xml
monen17: :+1:

product_pricelist_item_uom/views/product_pricelist_views.xml
monen17: :+1:

product_pricelist_item_uom/__manifest__.py
monen17: :+1:

product_pricelist_item_uom/__manifest__.py
```diff
@@ -7,10 +7,7 @@
     "version": "19.0.1.0.0",
     "category": "Sales/Sales",
     "website": "https://github.com/OCA/product-attribute",
-    "author": "Aion Tech, Odoo Community Association (OCA)",
-    "maintainers": [
-        "SirAionTech",
-    ],
+    "author": "Aion Tech, Camptocamp SA, Odoo Community Association (OCA)",
```
monen17, line 10: Usually you don't add yourself to the authors but this commit changes so much that it's ok 👍

monen17 (approved):
Thanks for the changes and for the clarification in https://github.com/OCA/product-attribute/pull/2364#issuecomment-5388446887, I reviewed the code and sold some sugar 😁
I left a few notes but nothing blocking

product_pricelist_item_uom/models/product_pricelist_item.py
Ricardoalso: Stayed as close as possible to the standard :eyes: https://github.com/odoo/odoo/commit/d2648b1d983927b5df7260a16d6d1#diff-5a563f36ffd028b0d8510c6d1339b097a5067a2247cc35c0f9bdd1fd0b5a0b05R92

ivantodorovich:
Can we merge this @OCA/product-maintainers ? 🙏🏻

PR #2368, opened 2026-08-06
[19.0][MIG] product_total_weight_from_packaging

JasminSForgeFlow:
## Dependency change

`product_packaging_dimension` → `product_packaging`.

Odoo 19 removed the `product.packaging` model, and in 19.0 the OCA split puts the
packaging `weight` on the new `product_packaging` module, while
`product_packaging_dimension` keeps dimensions only. This module only needs the
weight, so it now depends on `product_packaging` directly.

Depends on:

- [ ] https://github.com/OCA/product-attribute/pull/2320

## Changes

- `get_total_weight_from_packaging()` reworked for the 19.0 packaging model:
  `product_packaging_calculator` now yields packagings identified by their
  `uom.uom`, so packaging weights are looked up on `product.packaging` keyed by
  `uom_id` instead of browsing `product.packaging` by the calculator id.
- Weight UoM conversion removed. In 19.0 a packaging has no `weight_uom_id`
  anymore (`product.packaging.weight_uom_name` is related to the product's), so
  packaging weights and the product weight are already expressed in the same UoM
  and summing them directly is correct. The dependency on `product_logistics_uom`
  (through `product_packaging_dimension`) is therefore no longer needed, and the
  README warning about the missing conversions is obsolete.
- Behaviour is unchanged: packagings without a weight are still ignored, so their
  quantity falls back to the smaller packagings and finally to the product weight.

@ForgeFlow


AaronHForgeFlow (approved):
functional + code. LGTM

PR #2369, opened 2026-08-06
[19.0][MIG] product_sequence: Migration to 19.0

bosd:
Supersedes #2127. 

Migration of `product_sequence` to 19.0, redone from scratch with the standard
OCA migration procedure. The 19.0 migration commit of @fd-oerp is cherry-picked
(with `-x`) so their authorship and work are preserved, and they have been added
to `readme/CONTRIBUTORS.md`.

### Why redo it

* #2127 squashed the administrative commits **into the functional commits**, so
  several `[MIG]`/`[IMP]` commits ended up carrying the bodies of unrelated
  `.pot`/README/Weblate commits, blurring the attribution of the real changes.
* It was branched off an older `18.0`, so it did not carry
  `[IMP] product_sequence: default_code shouldn't be mandatory` nor the Turkish,
  Croatian, Catalan and Slovenian translations that landed on `18.0` since.

### Migration changes

* `test_product_category_sequence`: since 19.0, `categ_id` is no longer required
  on `product.template`, so `assertTrue(product_4.categ_id)` no longer holds.
  Rather than dropping the case (as #2127 did), it now asserts what the module
  actually guarantees: a product without category falls back on the default
  product sequence.

### On top of the plain migration

* **New OCA icon** (`[IMP] product_sequence: use the new OCA icon`), the one
  already used by other addons of this repository.
* **Demo data** (`[ADD] product_sequence: demo data`), illustrating the three
  ways a reference is assigned:

  | Product | Category | Reference |
  |---|---|---|
  | Floor Lamp | Lamps (prefix `LAMP`) | `LAMP00001` |
  | Desk Lamp | Desk Lamps (no prefix, child of Lamps) | `LAMP00002` |
  | Office Chair | Furniture (prefix `FURN`) | `FURN00001` |
  | Extension Cord | *none* | `PR/00001` |

  The parent-category fallback is enabled on the main company so the *Desk Lamp*
  case is actually demonstrated; `test_product_parent_category_sequence`
  therefore now disables that setting explicitly instead of relying on the
  ambient company configuration.

  Three tests assert the demo references. They skip when the database has no
  demo data, which is the case on the CI since Odoo 19.0 does not load demo data
  by default.

@fd-oerp thanks for the original migration.
 the history rework could not be done as a review suggestion. @NICO-SOLUTIONS the
`copy()` discussion from #2127 is deliberately left out of this PR to keep the
migration reviewable; happy to open a follow-up for it.


bosd:
Pushed a follow-up to get Codecov green. 

Two things came out of it:

**1. The demo tests no longer skip.** Since 19.0 Odoo does not load demo data by default, so the CI databases have none and the tests were skipped there. They now load the demo files themselves via `convert_file(..., mode="init", noupdate=True)` when the records are missing, so the demo data is actually validated on the CI. Verified both ways: with and without `--with-demo`.

**2. `copy()` was broken and is fixed** (`[FIX] product_sequence: keep the reference of copied products`). This is why the two copy tests had to be disabled in 63a0f5b0 "[FIX] product_sequence: disable test", and it is the thread @NICO-SOLUTIONS opened on #2127.

Since odoo/odoo#38303 (14.0), copying a `product.product` does not copy the variant: the *template* is copied and its first variant returned. That variant is built by `_create_variant_ids` with no reference, so it gets a fresh sequence number and the `default_code` the override put into `default` was silently discarded:

* copying `PROD02` produced `PR/00007` instead of `PROD02-copy`;
* `product.copy({"default_code": "REF"})` ignored `REF` entirely.

The override now restores the requested reference on the new variant after the copy, and both tests are re-enabled. Added tests for the no-source-reference case (copy just gets a new sequence number, no `-copy` suffix) and for the multi-variant case (the reference stays on the variant and is not pushed up to the template).

@NICO-SOLUTIONS if you would rather have a copy always take a *fresh* sequence number instead of the `-copy` suffix, that is a one-line change from here  let me know and i will change it.

Reyes4711-S73 (approved):
LGTM

NICO-SOLUTIONS:
> @NICO-SOLUTIONS if you would rather have a copy always take a _fresh_ sequence number instead of the `-copy` suffix, that is a one-line change from here let me know and i will change it.

@bosd 
Thanks for tackling this and for the explanation of the Odoo core change behind it.

From my point of view, a copied product should receive a fresh sequence number instead of keeping the `-copy` suffix. `default_code` should identify the new product and not describe how the record was created.

I can see the benefit of the `-copy` suffix for temporary drafts or work copies, as it makes the origin of a duplicate immediately visible. On the other hand, if a user wants a new sequence-generated reference, this can already be achieved by setting `default_code` to `/`, but this requires an additional manual step.

However, I would not expect the `-copy` suffix to remain as the final product reference used in sales, purchasing, inventory, or integrations.

This is just my perspective though; I won´t make this decision alone. I am happy to follow the consensus of the maintainers/contributors here. If there are other advantages of keeping the `-copy` behavior, I would be interested to understand them.

If this is handled in a separate follow-up PR, that is also fine for me. However, I think we should define a clear approach for this scenario.

I'll review the code once I find some time.

NICO-SOLUTIONS:
@bosd 
I also tested this locally and found one additional detail worth mentioning.

When calling `product.product.copy()` (e.g. via the shell), the new variant first receives a fresh sequence-generated reference, which is then overwritten by the `-copy` reference. In my test, the sequence advanced from `12` to `13`, but `LAMP00012` was not assigned to any product afterwards; the final reference was `LAMP00002-copy`. So the sequence value `12` is consumed and effectively lost without ever being used as the final reference.

Together with the fact that copying the template from the UI results directly in a fresh sequence-generated reference, this means that the two copy paths currently behave differently.

I think we should clarify the intended behavior here. Since we are already touching this area during the migration, it might also be a good opportunity to clean up this legacy behavior rather than carry the ambiguity forward. Otherwise, we risk making the migration history increasingly difficult to follow as more changes are added on top of the existing behavior.

At the same time, I don't want this to become a long-running discussion for the migration, and I won't block the PR on this point. I'm happy to follow the consensus of the maintainers/contributors here.

bosd:
@NICO-SOLUTIONS thanks for testing it locally — both points are right, and the second one is a genuine flaw in my patch.

**The burned sequence number: confirmed.** The new variant is created by `_create_variant_ids` *before* my code runs, so `create()` has already called `next_by_id()`. Overwriting the reference afterwards leaves that number consumed and never assigned to anything. So my patch fixes the *lost* `default_code` but pays a sequence number for it, and it does leave `product.product.copy()` and duplicating the template from the UI behaving differently. I am not going to defend that as an end state.

**On the intended behaviour: I agree with you.** A fresh sequence number is the better contract for a module whose whole job is clean sequential references. `-copy` describes how the record came to be, not what it is, and `/` already gives the user an explicit way to ask for a new one.

**Why the fix is in this PR at all:** the two copy tests were disabled (renamed to `_tests_*`) in 63a0f5b0, so `copy()` was dead-but-uncovered code, and Codecov patch was red on a module that is new on this branch. Re-enabling those tests required the method to actually work. It was the smallest change that got the branch green — not a considered decision about what copy *should* do. Had I looked harder at the sequence consumption I would have raised the question myself instead of letting you find it.

**Proposal:** let's not hold the migration for this. Whichever consensus lands first, the follow-up is small:

* **fresh sequence number** (your preference, and mine): drop the `copy()` override entirely. Both copy paths then behave identically, nothing is consumed for nothing, and there is no code left to cover.
* **keep `-copy`**: keep the override but pass a context key so `create()` skips the sequence for the variant of a copied template, so no number is burned.

I am equally happy to settle it here instead of in a follow-up if you or the other reviewers prefer not to carry the ambiguity across the migration — it is a small diff either way, just say which one and I will push it.

Would you be willing to approve the migration itself so it is not blocked on this? The rest of the PR is the history rework, the icon and the demo data, none of which touches this question.

NICO-SOLUTIONS:
@bosd 
Thanks, that all makes sense. And thanks for your work!

Since we both agree that a fresh sequence number should be the intended behavior, I would prefer to resolve this directly in this PR rather than carry the current behavior into a follow-up. I think this might otherwise get lost once the migration is merged.

In any case, the consumed sequence number should be fixed. Since we need to touch this part of the code anyway, I think it makes sense to settle the copy behavior at the same time rather than fixing the sequence consumption now and leaving the `-copy` behavior for later.

As nobody has suggested a different approach so far, I think we have enough consensus to settle this here. Dropping the `copy()` override would give us consistent behavior for both copy paths and avoid consuming a sequence number that is never used. Also, when copying a product, Odoo already adds `(copy)` to the product name, which provides a visible indication that it is a copy and can be useful for temporary work copies.

So from my side, let's settle this here.

bosd:
@NICO-SOLUTIONS done — settled here as you proposed. The `copy()` override is dropped, so both copy paths now behave identically and no sequence number is consumed for nothing.

The commit is `[IMP] product_sequence: copied products get their own reference`. Tests:

* `test_product_copy` — the copy gets a fresh reference from its category sequence, different from the source
* `test_product_copy_without_default_code` — a source without reference copies to a new one too
* `test_product_copy_does_not_waste_a_sequence_number` — asserts `number_next_actual` advances by exactly one per copy, so the regression you found cannot come back unnoticed
* `test_product_multi_variant_reference` — the reference stays on the variant, it is not pushed up to the template

Also added a line to `readme/USAGE.md` describing the behaviour.

**One consequence you should weigh in on**, because it is not obvious and I do not want it to slip in silently. I dropped `test_product_copy_with_default_values`, which asserted that `product.copy({"default_code": "REF"})` honours `REF`. Without the override it does not: it yields a fresh sequence number instead.

That is not a regression from 18.0 — it never worked there either, which is why that test was disabled too — but it *is* a divergence from plain Odoo. Core restores an explicitly passed `default_code` at the end of `ProductTemplate.create`:

```python
for field_name in self._get_related_fields_variant_template():   # includes default_code
    if vals.get(field_name) and not template[field_name]:
        related_vals[field_name] = vals[field_name]
```

The guard is `not template[field_name]`. With this module the variant made by `_create_variant_ids` has already been given a sequence number, so `template.default_code` is truthy and the caller's value is skipped. So on any database with `product_sequence` installed, `copy(default={"default_code": ...})` from code, data files or imports is silently ignored.

I left it as is rather than adding another override, since that is exactly the kind of thing we just removed. Three ways forward, happy with any:

1. leave it, and document the limitation in the readme;
2. pin it with a test that states the behaviour explicitly, so at least it is visible;
3. fix it, by having `create()` leave the reference alone when the template already carries an explicit one.

My preference is 3 if we care about the API, 1 otherwise — but this is a fresh question, so a follow-up seems more honest than stretching this PR further.

CI and Codecov were green on the previous head; I will report back once they have run on this one. Locally: 12 tests, 0 failures, 100% line coverage on all production files, with and without `--with-demo`.

NICO-SOLUTIONS:
@bosd 
I think we can resolve the remaining `default_code` issue directly by adding a small `product.template` extension.

The idea is to pass an explicitly provided `default_code` through to `_prepare_variant_values()`. This keeps the normal sequence behavior for `copy()` and `/`, while making `copy({"default_code": "REF"})` behave like standard Odoo.

I would suggest adding:

```
from odoo import models


    _inherit = "product.template"

    def copy(self, default=None):
        default = default or {}

        if "default_code" in default and default["default_code"] != "/":
            self = self.with_context(
                product_sequence_copy_default_code=default["default_code"]
            )

        return super().copy(default=default)

    def _prepare_variant_values(self, combination):
        variant_dict = super()._prepare_variant_values(combination)

        copy_default_code = self.env.context.get(
            "product_sequence_copy_default_code"
        )
        if copy_default_code is not None:
            variant_dict["default_code"] = copy_default_code

        return variant_dict
```

I tested the three relevant cases locally:

* `copy({"default_code": "REF"})` → `REF`, sequence unchanged
* `copy()` → fresh sequence number
* `copy({"default_code": "/"})` → fresh sequence number

So this should preserve the existing `product_sequence` semantics while also keeping the explicit `default_code` API behavior compatible with Odoo.

Then we can define the corresponding tests on `product.template` as well.

bosd:
@NICO-SOLUTIONS implemented, thanks — that is a much better hook than anything I had in mind. Pushed as `[FIX] product_sequence: honour an explicitly copied internal reference`, with you as co-author since the approach and most of the code are yours.

I made one change to your snippet. `_prepare_variant_values()` is called once per combination, so on a product with attributes the context value would be stamped on *every* variant, giving them all the same reference. Odoo itself only propagates a template reference to its variant when there is exactly one (`_set_product_variant_field`), so I guarded it the same way — an empty `combination` means no attributes, hence a single variant:

```python
def _prepare_variant_values(self, combination):
    vals = super()._prepare_variant_values(combination)
    code = self.env.context.get("product_sequence_default_code")
    if code and not combination:
        vals["default_code"] = code
    return vals
```

I also used `default.get("default_code")` rather than `"default_code" in default`, so an explicit `False` falls through to the sequence instead of blanking the reference.

Your three cases are now pinned as tests, plus the multi variant one:

* `test_product_copy_with_default_code` — `copy({"default_code": "REF"})` gives `REF`, and asserts `number_next_actual` is **unchanged**, so no number is drawn for a reference nothing would use
* `test_template_copy_with_default_code` — same when the template itself is copied
* `test_product_copy_with_slash_default_code` — `"/"` still means "give me a new one"
* `test_multi_variant_copy_with_default_code` — the two variants keep distinct references and neither takes the given one

16 tests, 0 failures, 100% line coverage on every production file including the new `product_template.py`. Verified with and without `--with-demo`.

That closes the last open point from my side. If it looks right to you, an approval would be welcome so this can go in.

NICO-SOLUTIONS (approved):
Code & functional review: LGTM!

dreispt (approved):

PR #2298, opened 2026-05-19
[19.0][MIG]: product_assortment

Alexgars73:
Superseed #2124

PR #2370, opened 2026-08-07
[19.0] [MIG] product_assortment: Migration to 19.0

rven:


BhaveshHeliconia:
@rven There is already a PR opened before yours: https://github.com/OCA/product-attribute/pull/2298. Could you please review that one instead of continuing this one?

rven:
> @rven There is already a PR opened before yours: #2298. Could you please review that one instead of continuing this one?

Why isn't this one mentioned in https://github.com/OCA/product-attribute/issues/2075?
Please follow the OCA guidelines so work will not done twice!

BhaveshHeliconia:
> > @rven There is already a PR opened before yours: #2298. Could you please review that one instead of continuing this one?
> 
> Why isn't this one mentioned in #2075? Please follow the OCA guidelines so work will not done twice!

I did not create PR #2298.

That said, I do not think checking only the migration issue is enough to avoid duplicate work. Before starting a migration, I usually follow this process:

1. Check the migration issue for the target version.
2. Search the list of open PRs to see if someone has already started working on the module.
3. Check previously closed PRs for the same version to understand why they were closed and whether there is any work that can be continued.

Following this process helps avoid duplicate work and also provides useful context before starting a migration.

Perhaps it would be worth documenting this workflow in the migration guidelines so contributors know to check both the migration issue and existing PRs before starting.

BhaveshHeliconia:
ping @pedrobaeza

pedrobaeza:
@rven unless a maintainer/PSC marks the PR with the bot command, or the contributor writes explicitly in the issue, there's no trace of the migration, but you should check the opened PRs before.

rven:
> @rven unless a maintainer/PSC marks the PR with the bot command, or the contributor writes explicitly in the issue, there's no trace of the migration, but you should check the opened PRs before.

If everybody should follow these guidelines it wouldn't be an issue at all. I always look on the migration issue, not between the dozens of open PR's
image

PR #2377, opened 2026-08-14
[19.0][ADD] product_document_type

Ricardoalso:
Adds a configurable, translatable document type to product documents,
that can be managed and translated per company.


product_document_type/views/product_document_type_views.xml
```diff
@@ -0,0 +1,43 @@
...
+<odoo>
+    <record id="product_document_type_view_form" model="ir.ui.view">
+        <field name="model">product.document.type</field>
+        <field name="arch" type="xml">
+            <form>
+                <sheet>
+                    <group>
+                        <field name="name" />
+                        <field name="sequence" />
+                        <field name="active" />
```
ivantodorovich: This form view is ugly. If we were to keep it, I'd suggest removing the `sequence` here, and the `active` field should use the ribbon widget instead.

Maybe we could remove it entirely and keep only the list view with editable="bottom", like it's done in core for this kind of simple records (e.g: "Industries")

product_document_type/views/product_document_type_views.xml
```diff
@@ -0,0 +1,43 @@
...
+            </form>
+        </field>
+    </record>
+
+    <record id="product_document_type_view_list" model="ir.ui.view">
+        <field name="model">product.document.type</field>
+        <field name="arch" type="xml">
+            <list>
+                <field name="name" />
+                <field name="sequence" />
```
ivantodorovich: Should use the "handle" widget for this

product_document_type/models/product_document_type.py
```diff
@@ -0,0 +1,14 @@
...
+
+class ProductDocumentType(models.Model):
+    _name = "product.document.type"
+    _description = "Product Document Type"
+    _order = "sequence, name"
+
+    name = fields.Char(required=True, translate=True)
+    sequence = fields.Integer(default=10)
+    active = fields.Boolean(default=True)
+    company_id = fields.Many2one("res.company", default=lambda self: self.env.company)
```
ivantodorovich: The `company_id` field is not shown on any view, so it's impossible to set it.
Please add it with the `base.group_multi_company`

I would drop the default on this model, tbh, like it's done for the products --> by default it's shared


```suggestion
    company_id = fields.Many2one("res.company")
```

ivantodorovich (approved):
Thanks!

BhaveshHeliconia (approved):
LGTM!

ivantodorovich:
Hello @OCA/product-maintainers can we merge this? 🙏🏻

product_document_type/models/product_document.py
```diff
@@ -0,0 +1,9 @@
+# Copyright 2026 Camptocamp SA (https://www.camptocamp.com).
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+from odoo import fields, models
+
+
+class ProductDocument(models.Model):
+    _inherit = "product.document"
+
+    document_type_id = fields.Many2one("product.document.type", index=True)
```
sebalix: Nitpick:
```suggestion
    document_type_id = fields.Many2one("product.document.type", index="btree_not_null")
```

sebalix (approved):
Small remark, LG

product_document_type/models/product_document.py
ivantodorovich: @Ricardoalso

sebalix (approved):

PR #2379, opened 2026-08-19
[19.0][MIG] sale_product_matrix_secondary_unit: Migration to 18.0

CarlosRoca13:
cc @Tecnativa TT64135

ping @carlos-lopez-tecnativa @pedrobaeza


pedrobaeza:
/ocabot migration sale_product_matrix_secondary_unit

Check runboat

CarlosRoca13:
Ready 4 review

pedrobaeza (approved):
Tested on runboat.

Image

Image

Not sure if entering the matrix the first time, the secondary unit should be selected by default instead the main one. How it was in the previous version?

sale_product_matrix_secondary_unit/models/product_template.py
```diff
@@ -0,0 +1,21 @@
...
+class ProductTemplate(models.Model):
+    _inherit = "product.template"
+
+    def _get_template_matrix(self, **kwargs):
+        matrix = super()._get_template_matrix(**kwargs)
+        # The default secondary unit
+        if self.sale_secondary_uom_id:
+            matrix["secondary_unit_id"] = self.sale_secondary_uom_id.id
+        # Optional secondary units
+        if self.secondary_uom_ids:
```
carlos-lopez-tecnativa: LGTM, just a minor change. I think we should filter to only active records because the field uses `active_test=False`.
https://github.com/OCA/product-attribute/blob/eaa42bddc45a212f7ab72309c52b64b6d95c3de3/product_secondary_unit/models/product_template.py#L14

Image

Image

carlos-lopez-tecnativa (changes requested):

carlos-lopez-tecnativa:
> Not sure if entering the matrix the first time, the secondary unit should be selected by default instead the main one. How it was in the previous version?


Only if you set Default Secondary Unit for Sales in the product form
image

sale_product_matrix_secondary_unit/models/product_template.py
CarlosRoca13: Change done!

carlos-lopez-tecnativa (approved):
LGTM

CarlosRoca13:
I think this is ready to be merged

PR #2402, opened 2026-09-11
[19.0][IMP] product_cost_security: Enhance product cost security mixin with module test mode checks

carlos-lopez-tecnativa:
Due to constant incompatibilities with other modules that create products or read the standard_price field in test mode, this commit adds an option to enable the checks in test mode only for certain modules. You can inherit the method to add more modules or add the module directly here.

This PR is an alternative to https://github.com/OCA/product-attribute/pull/2254 while this other PR is being merged. However, this approach is more extensible on a per-module basis.

@Tecnativa @pedrobaeza @sergio-teruel @carlosdauden @yajo, could you please review this?

### DIFF
--- a/product_cost_security/models/product_cost_security_mixin.py
+++ b/product_cost_security/models/product_cost_security_mixin.py
@@ -2,7 +2,7 @@
 # License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl-3.0)
 from contextlib import suppress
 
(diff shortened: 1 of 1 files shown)


sergio-teruel (approved):
LGTM!!

carlosdauden (approved):

pedrobaeza (approved):
/ocabot merge patch

Issue #1326, opened 2023-04-24
product_lot_sequence: Incrementation issues


grindtildeath:
## Module
product_lot_sequence

## Describe the bug
When opening the detailed operations of an assigned picking for a product 
tracked by serial numbers, Odoo systematically calls `_get_next_serial` even
if there is not any serial number to generate.
Moreover, the widget allowing to generate the serial numbers will not call 
the sequence but only increment the number according to the next serial, 
potentially leading to a sequence that is not in sync anymore with the created
serial numbers.

## To Reproduce
**Affected versions**: 15.0, 16.0

1. Either use a global sequence or a product specific sequence to generate serial numbers
2. On a reception from the supplier for such a product, opening the detailed operation tab will call `_get_next_serial` and increment the sequence, leading to multiple issues:
    * Calling assign serial numbers will create new numbers that will create `stock.production.lot` records without incrementing the sequence accordingly
    * The same numbers could be generated on different receptions if these are not validated as soon as the numbers are generated

**Expected behavior**
Sequence is only incremented for generated lot/serial numbers.

**Additional context**
cf https://github.com/OCA/product-attribute/pull/1323#issuecomment-1511681322

grindtildeath:
Can we please keep this issue open? :pray:

gurneyalex:
What is the status of this issue? Still there in recent versions of Odoo ?

grindtildeath:
Looks like my PR from 15.0 was ported to 16.0 and the commit exists in branches 17.0 and 18.0, so we can close this

PR #2405, opened 2026-09-16
[18.0][IMP] product_lot_sequence: mass create sequence names and resync

henrybackman:
Mass creation moved the sequence next number only by one for the whole batch and later would hand out used serial numbers. Now resyncing the sequences at creation by pattern matching the names to identify sequences and current numbers

The generate dialog had no default and empty produced "0", "1", "2" instead of sequence names. action_generate_lot_line_vals now fills it from the sequence.

Added two system parameters, both defaulting to the current behaviour:

- product_lot_sequence.consume_on_create: propose a number without taking it, so a canceled form leaves no gaps. The number is consumed only in create.
- product_lot_sequence.duplicate_check: refuse a duplicate sequence number. Values are "no", "check" and "admin".

Only names matching sequence prefix and suffix are considered, so a vendor's own serial number doesn't move the sequence and is never refused.


gurneyalex (approved):

gurneyalex:
I think you can update the known issues / roadmap section of the readme as this addition is supposed to fix #1326 (cc @grindtildeath )

product_lot_sequence/models/stock_lot.py
```diff
@@ -17,52 +18,170 @@ def _get_sequence_policy(self):
...
 
+    @api.model
+    def _consume_on_create(self):
+        """Whether a proposed name only takes its number once the lot exists"""
+        return (
+            self.env["ir.config_parameter"]
+            .sudo()
+            .get_param("product_lot_sequence.consume_on_create")
+            == "True"
+        )
```
grindtildeath: ⚠️ Use str2bool for this ⚠️

product_lot_sequence/models/stock_move.py
```diff
@@ -1,24 +1,31 @@
...
+        # propose the next serial of the sequence
+        self.ensure_one()
+        if (
+            self.display_assign_serial
+            and not self.next_serial
+            and self.product_id.tracking == "serial"
+            and self.state == "assigned"
+        ):
+            self.next_serial = self.env["stock.lot"]._propose_next_name(self.product_id)
         return super().action_show_details()
```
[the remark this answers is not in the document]
grindtildeath: Same concern

product_lot_sequence/models/stock_lot.py
```diff
@@ -17,52 +18,170 @@ def _get_sequence_policy(self):
...
+        for sequence, lots in self._lots_by_sequence().items():
+            sequence._resync_from_lot_names(lots.mapped("name"))
+
+    def _check_lot_name_available(self):
+        """Refuse a sequence name that already exists"""
+        policy = self._get_duplicate_check_policy()
+        if policy == "no":
+            return
+        if policy == "admin" and self.env.user.has_group("base.group_system"):
+            return
```
grindtildeath: I guess you want to ensure the policy is set to `check` before doing what comes after. Otherwise, having an existing parameter with whatever value would do the same than check which is probably not wanted/expected.
IMO the default should probably be `no`

product_lot_sequence/models/stock_lot.py
```diff
@@ -17,52 +18,170 @@ def _get_sequence_policy(self):
...
+        if seq_policy == "global":
+            sequence = self._get_name_sequence()
+            return {sequence: self} if sequence else {}
+        if seq_policy == "product":
+            grouped = self.grouped(lambda lot: lot.product_id.product_tmpl_id)
+            return {
+                template.lot_sequence_id: lots
+                for template, lots in grouped.items()
+                if template.lot_sequence_id
+            }
```
grindtildeath: 🔴  If the same sequence is used on different templates, only the lots from the last template will be kept.

product_lot_sequence/readme/CONFIGURE.md
```diff
@@ -15,6 +15,30 @@ If any other value is used for this System Parameter, then you will get
...
+
+When "False", the proposed serial number is taken from the sequence as soon
+as it is shown. Canceling the form or replacing the value leaves a gap in
+the numbering.
+
+When "True", proposing a number reads it, and the sequence is moved
+forward only when the lot is created. Mass creation moves the sequence
+past the whole batch rather than by a single number.
+
+## Refusing a serial number that already exists
```
grindtildeath:
```suggestion
## Refusing a serial number that already exists on another product
```

product_lot_sequence/readme/CONFIGURE.md
```diff
@@ -15,6 +15,30 @@ If any other value is used for this System Parameter, then you will get
...
+
+System Parameter \`product_lot_sequence.duplicate_check\`, default "no".
+
+Odoo only enforces that a serial number is unique per product, which is
+not enough when one sequence numbers every product.
+
+- "no": not checked, default behavior.
+- "check": check uniqueness for everyone.
+- "admin": same as check, except for admins same as "no"
```
grindtildeath: IMO there shouldn't be an exception for admin.

However, I feel like this is a new feature that should probably be done in a separated module since it's an extra restriction that is not necessarily needed for those is using this module, but could be of use for those not using this module and are using the default Odoo sequence.

product_lot_sequence/models/stock_lot.py
```diff
@@ -17,52 +18,170 @@ def _get_sequence_policy(self):
...
+        for lot_vals in vals_list:
+            if lot_vals.get("name"):
+                continue
+            product = self.env["product.product"].browse(lot_vals.get("product_id"))
+            sequence = self._get_name_sequence(product)
+            if sequence:
+                lot_vals["name"] = sequence._next()
+        lots = super().create(vals_list)
+        lots._check_lot_name_available()
+        lots._resync_sequence()
```
grindtildeath, line 137: Shouldn't this be done only for consume on create?

product_lot_sequence/models/stock_move.py
```diff
@@ -1,24 +1,31 @@
...
-                )
+        # propose the next serial of the sequence
+        self.ensure_one()
+        if (
+            self.display_assign_serial
+            and not self.next_serial
+            and self.product_id.tracking == "serial"
+            and self.state == "assigned"
+        ):
+            self.next_serial = self.env["stock.lot"]._propose_next_name(self.product_id)
```
grindtildeath: AFAIU if 2 users are concurrently editing different moves using the same sequence, this will now assign the same `next_serial` value on both stock moves, which will result in an error for the second user once the first one will have created the lots, without the ability to fix the value of next_serial. 😕

grindtildeath (changes requested):
That's a big one, with nice additions and some more concerning ones. IMO it would have been great to keep concepts separated in smaller commits to ease the review process.

The issue I do have with this implementation is that it doesn't rely on the DB sequence anymore to avoid concurrency issues, and makes me feel it solves some by introducing others. Moreover, I feel like the code is now way more complex and it would probably make more sense to have a dedicated module for the sequence resync to keep the different concepts/features separated.

product_lot_sequence/models/stock_move.py
henrybackman: Yes, this was a valid finding and I've implemented a fix for this 🙏 The next name is peeked every time if the name wasn't consumed yet, otherwise the previous name will be kept

product_lot_sequence/models/stock_lot.py
henrybackman: The resync fixes a bug in this module where creating multiple serials at once only increases the next value of the sequence by one, resulting in duplicate serials in consequent creations. But I'll think about what can be extracted to a separate module, however this I would keep in this module to fix that obvious bug

product_lot_sequence/models/stock_lot.py
henrybackman: Great catch, didn't think of this! I've reworked this to take into account cases when multiple templates share the same sequence and added a test for it too

product_lot_sequence/models/stock_lot.py
henrybackman: Yes, added that 👍

product_lot_sequence/models/ir_sequence.py
```diff
@@ -0,0 +1,103 @@
...
+            # %(year)s -> width 4 -> "20200001" -> "2020"
+            parts.append(rf"\d{{{PLACEHOLDER_WIDTHS[key]}}}")
+            position = match.end()
+        # add the remaining literal after the placeholders
+        parts.append(re.escape(text[position:]))
+        return "".join(parts)
+
+    def _lot_name_regex(self):
+        """Match the names this sequence produces, capturing their number"""
+        self.ensure_one()
```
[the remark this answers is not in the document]
henrybackman: Two concurrent forms are allowed the read the same proposed names, the duplicates will be caught during create. Only one create will succeed, others will fail. This is acceptable because we don't want to consume the sequence on preview, only on create. The failed forms can be refreshed with new values

product_lot_sequence/models/stock_lot.py
```diff
@@ -18,51 +18,54 @@ def _get_sequence_policy(self):
...
+    def _get_name_sequence(self, product=None):
+        """Return the correct sequence for the lot, empty when none applies"""
         seq_policy = self._get_sequence_policy()
-        if seq_policy != "product":
-            return self.env["ir.sequence"].next_by_code("stock.lot.serial")
-        return ""
+        if seq_policy == "product":
+            if product:
+                return product.product_tmpl_id.lot_sequence_id
+        elif seq_policy == "global":
```
grindtildeath, line 38: Nitpicking, but this looks a bit like a breaking change, as in case no sequence policy is defined, `_default_name` would return an empty string instead of calling next on `stock.lot.serial` as it used to. Whereas it used to return an empty string only in case the policy was defined as "product". Therefore I would change this condition to an `else`.

product_lot_sequence/models/ir_sequence.py
```diff
@@ -80,3 +80,28 @@ def _lot_name_for(self, number):
...
+
+        Only names this sequence could have produced are considered, and the
+        counter is never moved backwards.
+        """
+        self.ensure_one()
+        regex = self._lot_name_regex()
+        numbers = [int(match.group(1)) for match in map(regex.match, names) if match]
+        if not numbers:
+            return
+        self._lock()
```
grindtildeath, line 102: Could you use `try_lock_for_update` here instead of having to define your own function?

product_lot_sequence/models/stock_lot.py
```diff
@@ -55,6 +55,31 @@ def generate_lot_names(self, first_lot, count):
             for index in range(count)
         ]
 
+    def _lots_by_sequence(self):
+        """Group the lots by the sequence their names should follow"""
+        seq_policy = self._get_sequence_policy()
+        if seq_policy == "global":
```
grindtildeath, line 95: Again I would try to keep what's needed for global as default behaviour in case the seq_policy is not product.

Other option would be to change `_get_sequence_policy` and return "global" whenever the value of system parameter is not product

product_lot_sequence/models/stock_lot.py
```diff
@@ -55,6 +55,31 @@ def generate_lot_names(self, first_lot, count):
...
+            grouped = self.grouped(lambda lot: lot.product_id.product_tmpl_id)
+            for template, lots in grouped.items():
+                sequence = template.lot_sequence_id
+                if not sequence:
+                    continue
+                # several templates may point at one sequence, so accumulate
+                by_sequence[sequence] = by_sequence.get(sequence, self.browse()) | lots
+            # callers lock these in order: sort so two batches holding the same
+            # sequences cannot take them in opposite order and deadlock
+            return dict(sorted(by_sequence.items(), key=lambda item: item[0].id))
```
grindtildeath, line 109: That looks a bit over engineered ... or dare I say AI generated ❓ 😆 

If you need a list of lots grouped by sequence and ordered by id you could just do as you did before but with a defaultdict, which makes it much more easy to read.

```
if seq_policy == "product":
    lot_ids_by_sequence = defaultdict(list)
    for lot in self:
        seq = lot.product_id.product_tmpl_id.lot_sequence
        if not seq:
            continue
        lot_ids_by_sequence[seq].append(lot.id)
    return {seq: self.browse(sorted(lot_ids)) for seq, lot_ids in lot_ids_by_sequence.items()}
```
However, I'm not sure why it needs to be sorted...

product_lot_sequence/models/stock_move.py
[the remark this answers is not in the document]
grindtildeath: I don't see how this would be solved... 🫣

grindtildeath (commented):
Looks quite better, but can you please submit the companion module in another PR? 🙏

henrybackman:
> Looks quite better, but can you please submit the companion module in another PR? 🙏

I moved the stock_lot_unique_name here:
- https://github.com/OCA/product-attribute/pull/2413

For the other items, let's go through them together today to avoid further review rounds

product_lot_sequence/models/ir_sequence.py
henrybackman: This is available in version 19.0, added TODO for future migration

product_lot_sequence/models/stock_lot.py
henrybackman: Changed the _get_sequence_policy to always return either product or global policy

product_lot_sequence/models/stock_lot.py
henrybackman: After looking into this, the concern regarding unsorted keys is not real and we simplified the code. It was AI  :sweat_smile:

product_lot_sequence/models/stock_lot.py
```diff
@@ -17,52 +18,170 @@ def _get_sequence_policy(self):
...
+
+    @api.model
+    def generate_lot_names(self, first_lot, count):
+        # number the batch from the sequence, not from the name's digits: core
+        # increments the last run of digits it finds, which merges with prefix
+        # (ABC2026) and suffix (3M) digits into one number, "ABC20260000013M"
+        sequence = self._get_name_sequence()
+        match = sequence and sequence._lot_name_regex().match(first_lot or "")
+        if not match:
+            return super().generate_lot_names(first_lot, count)
```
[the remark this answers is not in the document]
henrybackman: The `_get_name_sequence()` now always returns either "product" or "global"

product_lot_sequence/models/stock_move.py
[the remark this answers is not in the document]
henrybackman: After investigating, this override is found unnecessary and is now removed. The `next_serial` field is not used anymore but was still referenced by outdated tests (also removed now)

product_lot_sequence/models/stock_lot.py
henrybackman: The `_get_sequence_policy` now returns always either "product" or "global". Invalid values will default to "global"

product_lot_sequence/readme/CONFIGURE.md
henrybackman: This feature is moved to `stock_lot_unique_name` module:
- https://github.com/OCA/product-attribute/pull/2413

product_lot_sequence/readme/CONFIGURE.md
henrybackman: This feature is moved to `stock_lot_unique_name` module and as discussed will be implemented using groups instead of configuration parameters (TBD)
- https://github.com/OCA/product-attribute/pull/2413

product_lot_sequence/models/stock_lot.py
henrybackman: The `_get_sequence_policy` now returns always either "product" or "global"

henrybackman:
We went through the remaining issues with @grindtildeath and I've added the fixup commits on the changes. If this looks good now, I can rebase and prepare for the merge

